PR #2, opened 2014-12-19
Switch to CS2.0 core library

southwood:
This is the core library for the android project, it compiles in Java1.6 and can send CS2.0 telemetry. To use it a platform specific context must be implemented (implement ITelemetryContext) and CoreTelemetryClient must be extended to instantiate the context.


littleaj:

PR #126, opened 2015-03-03
Performance Counter adding container

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Sets the timeout to wait before the first reporting.
+     *
+     * The number must be a positive number
+     *
+     * Note that the method will be effective if called before the first call to the 'register' method.
+     * @param startCollectingIntervalInMillis Timeout to wait before the first collection of performance counters in milliseconds.
+     */
+    public void setStartCollectingIntervalInMillis(long startCollectingIntervalInMillis) {
+        if (startCollectingIntervalInMillis <= 0) {
```
Yonisha: why 0 is not valid? can't i start collecting on right on app startup?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -0,0 +1,223 @@
...
+    /**
+     * Sets the timeout to wait between collection of Performance Counters.
+     *
+     * The number must be a positive number
+     *
+     * Note that the method will be effective if called before the first call to the 'register' method.
+     * @param collectingIntervalInMillis The timeout to wait between collection of Performance Counters.
+     */
+    public void setCollectingIntervalInMillis(long collectingIntervalInMillis) {
+        if (collectingIntervalInMillis <= 0) {
```
Yonisha: I think that we need to enforce a minimal value of at least few seconds, otherwise if a user define 0.0001 we can get like an infinite loop here.

Yonisha: In the future, to be more robust, this minimal value should be the sum of counters' timeouts.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -0,0 +1,223 @@
...
+    private void scheduleWork() {
+        threads.scheduleAtFixedRate(
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        for (PerformanceCounter performanceCounter : performanceCounters.values()) {
+                            try {
+                                performanceCounter.report(telemetryClient);
+                            } catch (Throwable e) {
+                                InternalLogger.INSTANCE.error("Exception while reporting performance counter '%s': '%s'", performanceCounter.getId(), e.getMessage());
```
Yonisha: Worth logging the counter type (i.e. simple name) since the id won't help you when looking at the logs.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -0,0 +1,223 @@
...
+     * A private method that is called only when the container needs to start
+     * collecting performance counters data. The method will schedule a callback
+     * to be called, it will initialize a {@link com.microsoft.applicationinsights.TelemetryClient} that the Performance Counters
+     * will use to report their data, and it will also register itself a the {@link com.microsoft.applicationinsights.internal.shutdown.SDKShutdownActivity}
+     */
+    private void initialize() {
+        if (!initialized) {
+            synchronized (INSTANCE) {
+                if (!initialized) {
+                    createThreadToCollect();
```
Yonisha: Threads

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -0,0 +1,223 @@
...
+                },
+                startCollectingIntervalInMillis,
+                collectingIntervalInMillis,
+                TimeUnit.MILLISECONDS);
+
+        // Register the instance so the container is stopped when the application exists.
+        SDKShutdownActivity.INSTANCE.register(INSTANCE);
+    }
+
+    private void createThreadToCollect() {
```
Yonisha: Threads

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -46,11 +47,16 @@
         private boolean stopped = false;
 
         private final ArrayList<ChannelFetcher> fetchers = new ArrayList<ChannelFetcher>();
+        private final ArrayList<Stoppable> stoppables = new ArrayList<Stoppable>();
 
         public synchronized void register(ChannelFetcher fetcher) {
```
Yonisha: Why not change the channels + logger to implement Stoppable and reuse the code?

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -92,6 +99,19 @@ private void stopChannels() {
...
                     }
                 } catch (Throwable t) {
+                    InternalLogger.INSTANCE.error("Failed to stop channel: '%s'", t.getMessage());
+                }
+            }
+        }
+        /**
+         * Make sure no exception is thrown!
+         */
+        private void stopStoppables() {
```
Yonisha: Test?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounter.java
```diff
@@ -0,0 +1,37 @@
+/*
```
Yonisha: I think it is better to define it as an abstract class so you'll be able to enforce restrictions like timeouts and such.

Yonisha:
signed off

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
gupele: basically not, when the application starts, I think we need to let it initialize for two reasons: 1. we don't want to take too much time, 2. the performance counters in this stage are not telling too much

Yonisha: Good point. So maybe it is better to set the minimal value to a few seconds.

gupele: I will, currently those setters are only for testing, I will change the code to allow that to be the only case for that

PR #134, opened 2015-03-12
Added PerformanceModule annotation

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/annotation/PerformanceModule.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import java.lang.annotation.Target;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Target({ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PerformanceModule {
```
Yonisha: I wonder what we'll do with the other annotations for telemetry modules/initializers - we already have TelemetryModule / TelemetryInitializer classes... WDTH?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -128,6 +133,51 @@ private void setInternalLogger(ConfigFileParser parser, TelemetryConfiguration c
...
+            public Class<? extends Annotation>[] annotations() {
+                return new Class[]{PerformanceModule.class};
+            }
+
+            @Override
+            public void reportTypeAnnotation(Class<? extends Annotation> annotation, String className) {
+                performanceModuleNames.add(className);
+            }
+        };
+        final AnnotationDetector cf = new AnnotationDetector(reporter);
```
Yonisha: what is cf?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -128,6 +133,51 @@ private void setInternalLogger(ConfigFileParser parser, TelemetryConfiguration c
...
+            @SuppressWarnings("unchecked")
+            public Class<? extends Annotation>[] annotations() {
+                return new Class[]{PerformanceModule.class};
+            }
+
+            @Override
+            public void reportTypeAnnotation(Class<? extends Annotation> annotation, String className) {
+                performanceModuleNames.add(className);
+            }
+        };
```
Yonisha: please move out do a dedicated class

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -128,6 +133,51 @@ private void setInternalLogger(ConfigFileParser parser, TelemetryConfiguration c
...
+        final AnnotationDetector cf = new AnnotationDetector(reporter);
+        try {
+            cf.detect("com.microsoft.applicationinsights");
```
Yonisha: I think it is better to encapsulate this entire logic of the TypeReporter & AnnotationDetector.
A class with a simple 'detect' API for a given annotation type. The code will be cleaner.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -128,6 +133,51 @@ private void setInternalLogger(ConfigFileParser parser, TelemetryConfiguration c
...
+            for (String performanceModuleName : performanceModuleNames) {
+                TelemetryModule module = null;
+                try {
+                    module = createInstance(performanceModuleName, TelemetryModule.class);
+                    if (module != null) {
+                        modules.add(module);
+                    } else {
+                        InternalLogger.INSTANCE.error("Failed to create performance module: '%s'", performanceModuleName);
+                    }
+                } catch (Exception e) {
```
Yonisha: This is redundant - createInstance can't throw any exception.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -128,6 +133,51 @@ private void setInternalLogger(ConfigFileParser parser, TelemetryConfiguration c
         InternalLogger.INSTANCE.initialize(loggerOutput, loggerData.items);
     }
 
+    private List<TelemetryModule> setPerformanceCounters(ConfigFileParser parser) {
```
Yonisha: getPerformanceCounters (the method doesn't set anything :))

Yonisha:
signed off

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: right, I will update the name thanks

core/src/main/java/com/microsoft/applicationinsights/internal/annotation/PerformanceModule.java
gupele: It is a good question, I thought that we should go with self explanatory names, like PerformanceModule, its location under package 'annotation' should help distinguish, I think we should try to make sure that it is clear to our users

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I think you are right, I will that on the next pull request

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I agree

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I don't know :) will change

PR #152, opened 2015-03-29
Loading default PCs as fallback for JBoss web servers

Yonisha:
In case of a failure to scan for PC modules in a JBoss web server, I simply load the modules by providing hard-coded module names.

This is only a TEMPORARY solution until we'll investigate and understand why the annotation scanner fail to run. I don't think we should invest more on it right now since the ROI is low (we get the same result with hard-coded module names but with the price of ugly workaround).

Guy, note that I saw only 'private byte' and 'process CPU' PCs have been registered. Does it make sense?


gupele:
I am merging this pull-request.
We need to find the reason for the issue, plus we need to find out why only two are registered

PR #172, opened 2015-04-28
Features/dll ready

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/util/PropertyHelper.java
```diff
@@ -28,8 +30,9 @@
 /**
  * Helper class for reading data from a properties file found on the class path.
  */
-public class PropertyHelper
-{
+public final class PropertyHelper {
+    public final static String SDK_VERSION_FILE_NAME = "sdk-version.properties";
```
amnonshmsft: This is a generic helper class while "sdk-version.properties" is a file designated for only for specific purpose (getting sdk-version numbers) so I wouldn't mix these two. A better design would be to define another class for extracting these properties and then it can be used both by the SdkVersionContextInitializer and your code for placing the native artifacts

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
```diff
@@ -32,18 +32,16 @@
...
-    final String sdkVersionFileName = "sdk-version.properties";
     final String sdkPrefix = "Java";
 
     @Override
     public void initialize(TelemetryContext context) {
         String sdkVersion = sdkPrefix;
-        try {
-            Properties sdkVersionProps = PropertyHelper.getProperties(sdkVersionFileName);
+
+        Properties sdkVersionProps = PropertyHelper.getSdkVersionProperties();
```
amnonshmsft: See my comment in PropertyHelper about moving "sdk-version.properties" into it

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -33,10 +43,12 @@
  * Created by gupele on 3/30/2015.
  */
 public final class JniPCConnector {
+    public static final String AI_BASE_FOLDER = "AISDK";
+    public static final String AI_NATIVE_FOLDER = "native";
     public static final String PROCESS_SELF_INSTANCE_NAME = "__SELF__";
 
     private static final String BITS_MODEL_64 = "64";
-    private static final String NATIVE_LIBRARY_64 = "applicationinsights-core-native-win64";
+    private static final String NATIVE_LIBRARY_64 = "applicationinsights-core-native-win64.dll";
```
amnonshmsft: For completeness please also add NATIVE_LIBRARY_32 (even we won't use it in this release)

gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/util/PropertyHelper.java
gupele: I'm not sure I agree. PropertyHelper is a helper in the core context. Why have another class for that? what if we need to have another propertis file that is loaded from various places?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
         return true;
     }
 
-    private static boolean loadLibrary() {
+    /**
+     * The method will try to extract the dll for the Windows performance counters to a local
+     * folder and then will try to load it. The method will do all that by doing the following things:
+     * 1. Find the OS type (64/32) currently supports only 64 bit.
+     * 2. Will find the path to extract to, which is %temp%/sdk_version_number
```
amnonshmsft: The path should include at least another level for grouping applicationinsights data, i.e. %temp%/<applicationinsights>/<sdk_version_number>, or even better %temp%/<applicationinsights>/<artifact_name>/<sdk_version_number>

This will also be a good preparation if we will decide to change the location under which we persist throttled telemetry.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
+     * 3. Find out whether or not the file is already exists in that directory
```
amnonshmsft: Find out whether or not the file already exists (remove the word "is")

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
+     * 4. If the dll is not there, the method will extract it from the jar to that directory
+     * 5. The method will call System.load to load the dll and by doing so we are ready to use it
+     * @return true on success, otherwise false
+     * @throws IOException If there are errors in opening/writing/reading/closing etc.
+     *         Note that the method might throw RuntimeExceptions due to critical issues
+     */
+    private static boolean loadLibrary() throws IOException {
         String model = System.getProperty("sun.arch.data.model");
         String libraryToLoad = NATIVE_LIBRARY_64;
         if (!BITS_MODEL_64.equals(model)) {
```
amnonshmsft: Don't you want to log the failure reason, e.g. unsupported platform in this case?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
gupele: The code will extract the dll into %temp/AISDK/version_number/dll, I should update the comment to reflect that

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
gupele: good point, thanks

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
gupele: thanks

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
amnonshmsft: Also please consider adding the artifact name, i.e. "%temp/AISDK/artifact_name/version_number"

core/src/main/java/com/microsoft/applicationinsights/internal/util/PropertyHelper.java
amnonshmsft: The way I see it, PropertyHelper is just a general helper class to help other classed loading custom property files, and should contain specific files inside it. Otherwise, what makes the sdk-version file different then the other properties files it gets special treatment and its own method while the other files do not? 
Also I wouldn't want to extend this file for every properties file I add to the project.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
             return false;
         }
 
-        System.loadLibrary(libraryToLoad);
+        File dllPath = buildDllLocalPath();
+
+        File dllOnDisk = new File(dllPath + File.separator + libraryToLoad);
```
amnonshmsft: Please do not try to construct paths manually - it is highly error prone and moreover, the File class has a specific constructor to help you with that:

File(File parent, String child) - Creates a new File instance from a parent abstract pathname and a child pathname string.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
+        if (properties == null) {
+            throw new RuntimeException("Failed to find SDK Version Properties file.");
+        }
+
+        String version = properties.getProperty("version");
+        if (version == null) {
+            throw new RuntimeException("Failed to find SDK version.");
+        }
+
+        String dllPathName = System.getProperty("java.io.tmpdir") + File.separator + AI_BASE_FOLDER + File.separator + AI_NATIVE_FOLDER + File.separator + version;
```
amnonshmsft: See my above comment about composing paths - please use File(File parent, String child) instead.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
+        File dllPath = new File(dllPathName);
+
+        if (!dllPath.exists()) {
+            dllPath.mkdirs();
+        }
+
+        if (!dllPath.exists() || !dllPath.canRead() || !dllPath.canWrite()) {
```
amnonshmsft: Consider moving the canRead() and carWrite() checks be done on the temp folder - if you can write something there then you can probably write something under any path you create there. Also, if you can't write to the temp folder then mkdirs() is going to fail.

You should, however, keep the check that dllPath exists here.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -126,15 +140,94 @@ private static boolean initNativeCode() {
...
+        }
+
+        System.load(dllOnDisk.toString());
+
         InternalLogger.INSTANCE.trace("Successfully loaded library '%s'", libraryToLoad);
+
         return true;
     }
+
+    private static void extractToLocalFolder(File dllOnDisk, String libraryToLoad) throws IOException {
```
amnonshmsft: I think that it might be useful having a helper class/method for extracting a resource into a given folder instead of having it embedded in this class. WDYT?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
gupele: I think that in the future we will have a utility class that will produce the base folder, but currently I'm not sure I understand what is the difference between doing it manually and writing multiple new File statements

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -63,6 +75,8 @@ public static boolean initialize() {
 
             if (loadLibrary()) {
                 return initNativeCode();
+            } else {
+                InternalLogger.INSTANCE.error("Failed to load native dll, Windows performance counters will not be used.");
```
amnonshmsft: Please also log the exception message here.

PR #175, opened 2015-05-01
Integration test framework using AppInsights export to azure storage

moralt:
Integration tests framework currently sends http requests and compares the results from Application Insights - Export with the expected results.


web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
+package com.microsoft.applicationinsights.azuretest;
```
amnonshmsft: Please add a license statement to this file (see other files for reference)

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        System.out.println("Before clearing: Queue contains " + queue.getApproximateMessageCount() + " items");
+        System.out.println("Clearing queue ...");
+        queue.clear();
+    }
+
+    private ArrayList<TelemetryItem> sentHttpGetRequests() throws Exception {
+
+        String serverAddress = config.getTestServerAddress();
+        int tomcat7Port = config.getTomcat7Port();
+        String applicationFollder = config.getApplicationFolder();
```
amnonshmsft: Typo in Follder

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        String requestId1 = UUID.randomUUID().toString();
+        String requestId2 = UUID.randomUUID().toString();
+        String requestId3 = UUID.randomUUID().toString();
+
+        URI uri1 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/books?id=Thriller&runId=" + runId + "&requestId=" + requestId1);
```
amnonshmsft: Should be just port - the test should not care what web server it works with

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        queue.downloadAttributes();
+        System.out.println("Before clearing: Queue contains " + queue.getApproximateMessageCount() + " items");
+        System.out.println("Clearing queue ...");
+        queue.clear();
+    }
+
+    private ArrayList<TelemetryItem> sentHttpGetRequests() throws Exception {
+
+        String serverAddress = config.getTestServerAddress();
+        int tomcat7Port = config.getTomcat7Port();
```
amnonshmsft: tomcat7Port - should be just port - the test should not care what web server it works with

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        String applicationFollder = config.getApplicationFolder();
+
+        String requestId1 = UUID.randomUUID().toString();
+        String requestId2 = UUID.randomUUID().toString();
+        String requestId3 = UUID.randomUUID().toString();
+
+        URI uri1 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/books?id=Thriller&runId=" + runId + "&requestId=" + requestId1);
+        URI uri2 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/loan?title=Gone%20Girl&id=030758836x&subject=Thriller&runId=" + runId + "&requestId=" + requestId2);
+        URI uri3 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/notExistingWebFage?runId=" + runId + "&requestId=" + requestId3);
+
```
amnonshmsft: You're have the same URL construction logic in the above 3 lines.
Why not refactoring it into a method, i.e. 
URI constructUrl(server, port, app, path) {
    return new URI("http://" + server + ":" + port + "/" + app + path);
}

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: Also in configuration file? Will we use a different config for each test?

amnonshmsft: The environment against which the test is executed should be set from outside.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
amnonshmsft: The package name should reflect the what's tested (spring app?) and not how this test is done (using azure).
Please review the names in the UT to find a more appropriate package name (note that changing the package name probably also implies moving the files to a different folder)

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Hashtable;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Created by moralt on 30/4/2015.
+ */
```
amnonshmsft: For consistency with the rest of the files please change the date format to the american style, i.e. 4/30/2015.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+public class CompareTelemetryWithAzureExportTests {
```
amnonshmsft: CompareTelemetryWithAzureExportTests is not a class name but a function name...
A better name would be something like "TelemetryComparer"

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+    String runId = UUID.randomUUID().toString();
+    String configEnvironmentVariable = "INTEGRATION_TEST_CONFIGURATION";
+    IntegrationTestConfiguration config;
+    private static final int millisecondsInSecond = 1000;
+
+    /**
+     * Sends GET requests to server and expects to get similar result from app insights
+     * @throws Exception
```
amnonshmsft: The result of GET is Http status code + body while from app insights export you're getting the logged telemetry, so I'm not sure that I understand what you mean by "expects to get similar result from app insights"...

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        initConfiguration();
+        clearQueue();
+
+        ArrayList<TelemetryItem> expectedTelemetries = sentHttpGetRequests();
+        ArrayList<TelemetryItem> realTelemetries = getTelemetryFromAzure(DocumentType.Requests);
+
+        ArrayList<TelemetryItem> missingTelemetry = getMissingExpectedTelemetry(realTelemetries, expectedTelemetries);
+        if (missingTelemetry.size() > 0) {
+            for (TelemetryItem item : missingTelemetry) {
```
amnonshmsft: These tests are going to be executed unattended so this output may go nowhere.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        }
+        else {
+            System.out.println("Test passed successfully");
+        }
+    }
+
+    private void initConfiguration() throws JAXBException {
+        String location = System.getenv(configEnvironmentVariable);
+        Map<String, String> env =  System.getenv();
+        System.out.println("Configuration file located in " + location);
```
amnonshmsft: `env` looks like an unused parameter

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        try {
```
amnonshmsft: I'm not sure that you need this try/catch just for writing a message to System.out below. The exception thrown would probably include the failure reason.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+            System.out.println("Failed to load configuration file " + location);
+            throw e;
+        }
+    }
+
+    /**
+     * Clears the Azure queue from all existing messages
+     * @throws Exception
+     */
+    private void clearQueue() throws Exception {
```
amnonshmsft: method name does not reflect which Queue is to be clear. Please rename to clearAzureQueue.

Also, this function assumes that `config` had been initialized. However, it would be better to make this function more generic by having it accept the relevant parameters, i.e. `account`, `queueClient` and `queueName`.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        CloudQueueClient queueClient = account.createCloudQueueClient();
+        CloudQueue queue = queueClient.getQueueReference(config.getQueueName());
+
+        queue.downloadAttributes();
+        System.out.println("Before clearing: Queue contains " + queue.getApproximateMessageCount() + " items");
+        System.out.println("Clearing queue ...");
+        queue.clear();
+    }
+
+    private ArrayList<TelemetryItem> sentHttpGetRequests() throws Exception {
```
amnonshmsft: send and not sent...

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        String serverAddress = config.getTestServerAddress();
+        int tomcat7Port = config.getTomcat7Port();
+        String applicationFollder = config.getApplicationFolder();
+
+        String requestId1 = UUID.randomUUID().toString();
```
amnonshmsft: Why not refactoring `UUID.randomUUID().toString();` into `String getRandomUUIDString() {}`?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        URI uri1 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/books?id=Thriller&runId=" + runId + "&requestId=" + requestId1);
+        URI uri2 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/loan?title=Gone%20Girl&id=030758836x&subject=Thriller&runId=" + runId + "&requestId=" + requestId2);
+        URI uri3 = new URI("http://" + serverAddress + ":" + tomcat7Port + "/" + applicationFollder + "/notExistingWebFage?runId=" + runId + "&requestId=" + requestId3);
+
+        TelemetryItem expectedTelemetry1 = sendGet(uri1, requestId1);
+        TelemetryItem expectedTelemetry2 = sendGet(uri2, requestId2);
+        TelemetryItem expectedTelemetry3 = sendGet(uri3, requestId3);
+
+        ArrayList<TelemetryItem> expectedTelemetries = new ArrayList<TelemetryItem>();
```
amnonshmsft: Generally it would be much more elegant to use loops and arrays/collections here rather then having each action appear 3 times with 3 variables.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        return expectedTelemetries;
+    }
+
+    /**
+     * Sends HTTP GET request and returns the expected telemetry from AppInsights
+     * @param uri The URI for the request
+     * @return The expected telemetry items from AppInsights
+     * @throws Exception
+     */
+    private TelemetryItem sendGet(URI uri, String requestId) throws Exception {
```
amnonshmsft: `sendHttpGetRequest`

amnonshmsft: Also, please refactor by separating the logic for composing and sending the GET request and calculating the expected telemetry from the URL.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        connection.setRequestMethod("GET");
+
+        //add request header
+        connection.setRequestProperty("User-Agent", config.getUserAgent());
+
+        System.out.println("Sending 'GET' request to URL: " + uri.toString());
+        Integer responseCode = connection.getResponseCode();
+        System.out.println("Response Code : " + responseCode);
+
+        // Create expected result
```
amnonshmsft: Please refactor by separating the logic for composing and sending the GET request and calculating the expected telemetry from the URL.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+        try {
+            String[] params = uri.getQuery().split("&");
+            for (String param : params) {
+                String name = param.split("=")[0];
+                String value = param.split("=")[1];
+                expectedResult.setProperty("queryParameter." + name, value);
+            }
+        }
+        catch (Exception e) {
```
amnonshmsft: In what scenarios are we expecting this to fail and should such failures be ignored?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        return expectedResult;
+    }
+
+    /**
+     * Connects to Azure export and retrieves the telemetry items
+     * @param docType The document type to retrieve from Azure
+     * @return The retrieved telemetry items
+     * @throws Exception
+     */
+    private ArrayList<TelemetryItem> getTelemetryFromAzure(DocumentType docType) throws Exception {
```
amnonshmsft: What we're actually doing here is getting the logged telemetry. The fact that it is done from Azure is less important.
Therefore, I'd suggest calling this function `getLoggedTelemetry` or something similar

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        System.out.println("Creating Azure storage account connection");
```
amnonshmsft: Also, is it possible to get all the telemetry, regardless of the specific docType?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        CloudStorageAccount account = CloudStorageAccount.parse(config.getStorageConnectionString());
+        CloudBlobClient blobClient = account.createCloudBlobClient();
+
+        System.out.println("Creating Azure queue connection");
+        CloudQueueClient queueClient = account.createCloudQueueClient();
+        CloudQueue queue = queueClient.getQueueReference(config.getQueueName());
+
+        ArrayList<JSONObject> telemetryAsJson = new ArrayList<JSONObject>();
+        StopWatch stopWatch = new StopWatch();
```
amnonshmsft: Do we really need stopWatch here?
Why not simply taking the startTime and comparing it with the current time?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+     * Converts a string to multiple JSON objects
+     * @param jString The string to convert
+     * @return ArrayList of JSON objects
+     */
+    private ArrayList<JSONObject> getJsonsFromString(String jString) throws JSONException {
+        System.out.println("Extracting JSON objects from string");
+        String[] jsonStrings = jString.split("\n");
+        ArrayList<JSONObject> jsonObjects = new ArrayList<JSONObject>();
+
+        for(String s : jsonStrings) {
```
amnonshmsft: `for (` (i.e. add space)

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+     */
+    private ArrayList<String> getBlobUrisFromQueueMessages(DocumentType docType, ArrayList<CloudQueueMessage> messages) throws Exception {
+        ArrayList<String> blobUris = new ArrayList<String>();
+
+        System.out.println("Extracting blob URIs of document type " + docType.toString() + " from " + messages.size() + " messages");
+        for (CloudQueueMessage message : messages) {
+            JSONObject messageContentAsJson = new JSONObject(message.getMessageContentAsString());
+            String msgDocType = messageContentAsJson.getString("DocumentType");
+
+            if(msgDocType.equals(docType.toString())) {
```
amnonshmsft: `if (` (add space)

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/TelemetryItem.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * Tests if the properties of the this item equals to the properties of another telemetry item
+     * @param obj The other object
+     * @return True if equals, otherwise false.
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if(this == obj) {
```
amnonshmsft: `if (` (add space - also in the other occurrences below)

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        System.out.println("Got " + blobUris.size() + " blob URIs with document type " + docType.toString());
+        return blobUris;
+    }
+
+    /**
+     * Converts a string to multiple JSON objects
+     * @param jString The string to convert
+     * @return ArrayList of JSON objects
+     */
+    private ArrayList<JSONObject> getJsonsFromString(String jString) throws JSONException {
```
amnonshmsft: How about renaming to `convertToJson`?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+
+    /**
+     * Converts JSON object to TelemetryItem by it's document type
+     * @param docType The document type of the telemetry
+     * @param json The JSON object
+     * @return A TelemetryItem
+     * @throws Exception
+     */
+    private TelemetryItem getTelemetryItemFromJson(DocumentType docType, JSONObject json) throws Exception {
+        if (docType == DocumentType.Requests) {
```
amnonshmsft: Why not using a switch statement?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+            return getRequestTelemetryItemFromJson(json);
+        }
+
+        if (docType == DocumentType.PerformanceCounters) {
+            return getPerformanceCounterTelemetryItemFromJson(json);
+        }
+
+        throw new Exception("Can't find document type: " + docType.toString());
```
amnonshmsft: You don't need to look for the docType as you get it as a parameter. It would be better to use "Unsupported document type"

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    /**
+     * Converts JSON object to Request TelemetryItem
+     * @param json The JSON object
+     * @return A TelemetryItem
+     */
+    private TelemetryItem getRequestTelemetryItemFromJson(JSONObject json) throws URISyntaxException, JSONException {
```
amnonshmsft: Why not using OO and letting concrete classes inheriting from TelemetryItem to do this job?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        System.out.println("Converting JSON object to telemetry item with document type " + DocumentType.Requests.toString());
+        String address       = json.getJSONArray("request").getJSONObject(0).getString("url");
```
amnonshmsft: Please store `json.getJSONArray("request").getJSONObject(0)` in a local variable.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        Integer port         = json.getJSONArray("request").getJSONObject(0).getJSONObject("urlData").getInt("port");
+        Integer responseCode = json.getJSONArray("request").getJSONObject(0).getInt("responseCode");
+        String userAgent     = json.getJSONObject("context").getJSONObject("device").getString("userAgent");
+
+        JSONArray parameters = json.getJSONArray("request").getJSONObject(0).getJSONObject("urlData").getJSONArray("queryParameters");
+        Hashtable<String, String> queryParameters = new Hashtable<String, String>();
+        for (int i = 0; i < parameters.length(); ++i) {
+            String name  = parameters.getJSONObject(i).getString("parameter");
```
amnonshmsft: Please store `parameters.getJSONObject(i)` in a local variable and give it a meaningful name

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    /**
+     * Converts JSON object to PerformanceCounter TelemetryItem
+     * @param json The JSON object
+     * @return A TelemetryItem
+     */
+    private TelemetryItem getPerformanceCounterTelemetryItemFromJson(JSONObject json) throws JSONException {
+        System.out.println("Converting JSON object to telemetry item with document type " + DocumentType.PerformanceCounters.toString());
+        String category = json.getJSONArray("performanceCounter").getJSONObject(0).getString("categoryName");
```
amnonshmsft: Please use a local variable for `json.getJSONArray("performanceCounter").getJSONObject(0)`

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        String instance = json.getJSONArray("performanceCounter").getJSONObject(0).getString("instanceName");
+
+        TelemetryItem telemetryResult = new TelemetryItem(DocumentType.PerformanceCounters, ""); // TODO when implementing the usage of performance counters test, think of logic for id.
```
amnonshmsft: Please place the comment above this line and not at the end of it

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+        return telemetryResult;
+    }
+
+    /**
+     * Tests if the expected telemetry exists in the real telemetry from AppInsights
+     * @param realTelemetryItems The real telemetry
+     * @param expectedTelemetryItems The expected telemetry
+     * @return A collection of expected telemetry items that were not found in the real telemetry. If all items exists, returns empty collection.
+     */
+    private ArrayList<TelemetryItem> getMissingExpectedTelemetry(ArrayList<TelemetryItem> realTelemetryItems,
```
amnonshmsft: We should be able to check bi-directional inclusion. Therefore please change this method name and local variables to be symmetric and not assume that one of the collections is the real and the other is the expected.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
```diff
@@ -0,0 +1,342 @@
...
+                                                                 ArrayList<TelemetryItem> expectedTelemetryItems) {
```
amnonshmsft: Also, have you considered using Set instead of ArrayList?

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/Helpers.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.applicationinsights.azuretest;
+
+
+/**
+ * Created by moralt on 30/4/2015.
+ */
+public class Helpers {
```
amnonshmsft: I really don't think that you need a Helper class just for sleep, especially when you use it only in one place.

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/IntegrationTestConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Unmarshaller;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+import java.io.File;
+
+/**
+ * Created by moralt on 30/4/2015.
+ */
+@XmlRootElement(name = "IntegrationTestConfiguration")
```
amnonshmsft: Could have been much simpler but just using Java built in **[Properties](http://docs.oracle.com/javase/7/docs/api/java/util/Properties.html)** object

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/TelemetryItem.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Hashtable;
+
+/**
+ * Created by moralt on 30/4/2015.
+ */
+public class TelemetryItem {
+    private static Hashtable<DocumentType, ArrayList<String>> defaultPropertiesToCompare = new Hashtable<DocumentType, ArrayList<String>>();
+    private DocumentType docType;
+    private String id;
+    private Hashtable<String, String> properties = new Hashtable<String, String>();
```
amnonshmsft: You have a built in object called [Properties](http://docs.oracle.com/javase/7/docs/api/java/util/Properties.html) which is actually implemented as Hashtable, so no need to define one your own

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: Fixed

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: Right

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: Right

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: Possible but redundant

web/src/integTest/java/com/microsoft/applicationinsights/azuretest/CompareTelemetryWithAzureExportTests.java
moralt: I think it's more elegant this way than using "new Date().getTime() - startTime.getTime() < timeToPoll ...."

littleaj:

PR #177, opened 2015-05-05
Writer implementation

Yonisha:
Writer configuration schema (as part of CollectD configuration file):
Performance Counters - CollectD Integration - Adding an Application Insights writer:

<Plugin ApplicationInsightsWriter>
    InstrumetationKey 123-123-123-123-123
    SDKLogger true
    Exclude Plugin1:DataSource1,DataSource2 Plugin2 Plugin3:DataSource1


amnonshmsft:
General: please change the commit message to something more meaningful with a context, i.e. "Performance Counters - CollectD Integration - Adding an Application Insights writer"

collectd/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+apply from: "$buildScriptsDir/common-java.gradle"
+apply from: "$buildScriptsDir/publishing.gradle"
+
+def getCollectDLibDir() {
+    return System.env.'COLLECTD'
```
amnonshmsft: Looks like redundant code - why not simply setting `collectdLibDir` directly?

`def collectdLibDir =  System.env.'COLLECTD'`

amnonshmsft: Please rename COLLECTD to COLLECTD_HOME to be more aligned with the java environment variable class names, e.g. JAVA_HOME, GROOVY_HOME, M2_HOME (maven), etc.

(BTW, according to the above JAVA_JRE_6 should be renamed to JAVA_JRE6_HOME and I might do it in a couple of days)

collectd/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+}
+
+def collectdLibDir = getCollectDLibDir()
+def archivePath = new File(collectdLibDir, "lib/collectd-api.jar")
```
amnonshmsft: Please add a check that the relevant file exist and if not throw a `ProjectConfigurationException` with an appropriate description (you can see an example in `common-java.gradle`)

amnonshmsft: `archivePath` is a too general name, especially assuming that the chances that we'll have other dependent archives from another components (i.e. not CollectD) added are rather low.
Also I think that you can skip using this variable and simply replace the dependency below with `fileTree(dir: collectLibDir, include: 'lib/*.jar')`.

If you decide to leave it then please rename it to `archivesPath` (`archives` in plural)

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.HashMap;
+import java.util.List;
+
+import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import com.microsoft.applicationinsights.collectd.internal.ApplicationInsightsWriterLogger;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.util.LocalStringsUtils;
+import com.microsoft.applicationinsights.telemetry.MetricTelemetry;
+import org.collectd.api.*;
```
amnonshmsft: Please import the specific classes rather than all classes.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * Created by yonisha on 4/26/2015.
+ */
+public class ApplicationInsightsWriter implements CollectdConfigInterface,
+        CollectdInitInterface,
```
amnonshmsft: Please move CollectdConfigInterface one line down so all implemented interfaces will be aligned

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+    private TelemetryConfiguration telemetryConfiguration;
+    private TelemetryClient telemetryClient;
+    private ApplicationInsightsWriterLogger logger = null;
+
+    // endregion Members
+
+    // region Ctors
+
+    public ApplicationInsightsWriter() {
+        Collectd.registerConfig(PLUGIN_NAME, this);
```
amnonshmsft: Passing `this` to another method withing a constructors is considered as a bad practice as the object is not yet fully initialized and you do not know how the called method is going to use it. 
Please check whether you can move this registration code to somewhere else or make sure in CollectD documentation that it is safe to use it this way.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+        Collectd.registerInit(PLUGIN_NAME, this);
+        Collectd.registerWrite(PLUGIN_NAME, this);
+        Collectd.registerShutdown(PLUGIN_NAME, this);
+
+        this.logger = new ApplicationInsightsWriterLogger();
+        this.telemetryConfiguration = TelemetryConfiguration.getActive();
+    }
+
+    protected ApplicationInsightsWriter(TelemetryClient telemetryClient, ApplicationInsightsWriterLogger logger) {
```
amnonshmsft: Who is going to use this constructor except of the test?
It this is it's only purpose then it would be cleaner to create a test class inheriting from this class and overriding the constructor.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+    // region Public
+
+    /**
+     * Reads the configuration provided for the writer plugin.
+     * @param configuration The configuration.
+     * @return Zero on success, false otherwise.
+     */
+    public int config(OConfigItem configuration) {
+        logger.logInfo("Loading configuration.");
+        logger.logDebug("Configuration found: " + configuration);
```
amnonshmsft: The configuration is not found here - you get it as a parameter :)
Also - you're assuming that configuration.getString() will print something meaningful here, e.g. a name, class, etc.
I just want to make sure that this is the case

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+        List<OConfigItem> children = configuration.getChildren();
+        if (children.size() < 1) {
+            logger.logError("Configuration error: application instrumentation key must be set.");
+
+            return 1;
+        }
+
+        for (int i = 0; i < children.size(); i++) {
```
amnonshmsft: Can we use a foreach loop here, i.e.
for (OConfigItem child : children) { ...}

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+     */
+    public int config(OConfigItem configuration) {
+        logger.logInfo("Loading configuration.");
+        logger.logDebug("Configuration found: " + configuration);
+
+        List<OConfigItem> children = configuration.getChildren();
+        if (children.size() < 1) {
+            logger.logError("Configuration error: application instrumentation key must be set.");
+
+            return 1;
```
amnonshmsft: What's the meaning of 1?
Please use constants with meaningful names (also relevant for the return values below)

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+                    logger.logError(key + " configuration option needs exactly 1 argument [true/false].");
+                    return 2;
+                }
+
+                boolean enableSDKLogger = Boolean.parseBoolean(values.get(0).toString());
+                if (enableSDKLogger) {
+                    InternalLogger.INSTANCE.initialize("CONSOLE", new HashMap<String, String>());
+                }
+            } else {
+                logger.logWarning("Unknown configuration option '" + key + "'.");
```
amnonshmsft: Is it possible that this configuration key is intended to another writer or to CollectD core?
I'm asking because in the case above this shouldn't be a warning.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+
+        logger.logInfo("Configuration loaded successfully.");
```
amnonshmsft: What about the case that you went through all the keys but could not find the instrumentation key?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+                this.telemetryConfiguration.setInstrumentationKey(instrumentationKey);
+            } else if (key.equalsIgnoreCase(SDK_LOGGER_CONFIGURATION_KEY)) {
+                List<OConfigValue> values = child.getValues();
+                if (values.size() != 1) {
+                    logger.logError(key + " configuration option needs exactly 1 argument [true/false].");
+                    return 2;
+                }
+
+                boolean enableSDKLogger = Boolean.parseBoolean(values.get(0).toString());
```
amnonshmsft: Is it probably possible that parsing of the boolean value will fail in case of type so you should handle such case.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+                String instrumentationKey = values.get(0).toString();
+                if (LocalStringsUtils.isNullOrEmpty(instrumentationKey)) {
+                    logger.logError (INSTRUMENTATION_KEY_CONFIGURATION_KEY + "' configuration option is mandatory, plugin will be disabled");
+
+                    return 1;
+                }
+
+                this.telemetryConfiguration.setInstrumentationKey(instrumentationKey);
+            } else if (key.equalsIgnoreCase(SDK_LOGGER_CONFIGURATION_KEY)) {
```
amnonshmsft: There's an inconsistency in the CONFIGURATION_KEY constant names:

* `INSTRUMENTATION_KEY_CONFIGURATION_KEY` - the actual value for the instrumentation key
* `SDK_LOGGER_CONFIGURATION_KEY` - whether to enable or disable the internal logger

So it would be better to rename the later to **`ENABLE_`**`SDK_LOGGER_CONFIGURATION_KEY`

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+        return 0;
+    }
+
+    /**
+     * Shutting down the writer plugin.
+     * @return Zero on success, non-zero otherwise.
+     */
+    public int shutdown() {
+        return 0;
```
amnonshmsft: We don't want to flush our data here?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+        logger.logInfo("Initializing...");
+
+        try {
+            this.telemetryClient = new TelemetryClient(this.telemetryConfiguration);
+
+            logger.logInfo("Initialization completed.");
+        } catch (Throwable e ) {
+            logger.logError("Initialization failed, plugin will be disabled:\n" + e.toString());
+
+            return 1;
```
amnonshmsft: What's the meaning of 1?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Initializes the writer plugin.
+     * @return Zero on success, non-zero otherwise.
+     */
+    public int init() {
+        logger.logInfo("Initializing...");
+
+        try {
+            this.telemetryClient = new TelemetryClient(this.telemetryConfiguration);
+
```
amnonshmsft: Please remove the empty line

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+        Number value = valueList.getValues().get(index);
+        if (value != null) {
+            telemetry.setValue(value.doubleValue());
+        }
+
+        return telemetry;
+    }
+
+    protected static String generateMetricName(ValueList valueList, DataSource dataSource) {
```
amnonshmsft: Is it possible to somehow cache these values?
I'm asking since in each write we're constructing probably the same metric names over and over, so this sounds as an inefficient process.

collectd/build.gradle
Yonisha: cannot drop this variable if I want to throw exception if not found.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: This is how collectd development wiki guided

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: Yes, this is the case...

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+        DataSource dataSource = getDataSource(valueList.getDataSet(), index);
+
+        MetricTelemetry telemetry = new MetricTelemetry();
+
+        if (dataSource != null) {
+            telemetry.setMin(dataSource.getMin());
+            telemetry.setMax(dataSource.getMax());
+        }
+
+        String telemetryName = generateMetricName(valueList, dataSource);
```
amnonshmsft: Would calling this method work if dataSource == null?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: not possible.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,227 @@
...
+
+        return 0;
+    }
+
+    // endregion Public
+
+    // region Private
+
+    private static MetricTelemetry createMetricTelemetry(ValueList valueList, int index) {
+        DataSource dataSource = getDataSource(valueList.getDataSet(), index);
```
amnonshmsft: I would find this method to be more readable if the mandatory stuff would appear before the optional stuff, i.e.:

    private static MetricTelemetry createMetricTelemetry(ValueList valueList, int index) {
        MetricTelemetry telemetry = new MetricTelemetry();

        DataSource dataSource = getDataSource(valueList.getDataSet(), index);
        String telemetryName = generateMetricName(valueList, dataSource);
        telemetry.setName(telemetryName);

        Number value = valueList.getValues().get(index);
        if (value != null) {
            telemetry.setValue(value.doubleValue());
        }

        if (dataSource != null) {
            telemetry.setMin(dataSource.getMin());
            telemetry.setMax(dataSource.getMax());
        }

        return telemetry;
    }

collectd/src/main/java/com/microsoft/applicationinsights/collectd/internal/ApplicationInsightsWriterLogger.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+package com.microsoft.applicationinsights.collectd.internal;
+
+import org.collectd.api.Collectd;
+
+/**
+ * Created by yonisha on 5/4/2015.
+ *
+ * This is class is mainly for testability of the writer.
```
amnonshmsft: Please strive to keep isolation between test-specific code and product code.
A better approach would be to define an interface for logging and having the product code calling it (in case a logger was provided). Then the test code would provide the product code the implementation for this logger.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: cannot fail - it's simply return false if the input is null or not 'true'

collectd/build.gradle
amnonshmsft: Why not? :)

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: nope. {x}_CONFIGURATION_KEY --> the {x} matches the string that will appear in the configuration. Since in our AI config xml the logger element called SDKLogger, I aligned it here.

collectd/build.gradle
Yonisha: Never mind, i thought you meant to put the fileTree inside the dependency declaration :)
Anyway, keeping it as it (changed the var name and added a validation)

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: why is that?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: I don't like wrapping an object that is being under test, it is somehow misses the point :\
It is commonly used to add a ctor for testability purposes (note that it is protected).

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: true, I thought we didn't implement the flush yet :)

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: yes

collectd/src/main/java/com/microsoft/applicationinsights/collectd/internal/ApplicationInsightsWriterLogger.java
Yonisha: I did it for simplicity. Even if I would define an interface and use it in the product code I would still have to create a concrete class for production code.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: true, added a validation for that

amnonshmsft:
Please add a context to your commit messages (e.g. PC/CollectD) so it would be easy to understand to which feature these belong to

collectd/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+plugins {
+    id 'java'
```
amnonshmsft: No need to apply the java plug-in - it is already applied inside `common-java.gradle`

collectd/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+    compile project(':core')
+    compile files(collectDLibPath)
+    testCompile group: 'org.mockito', name: 'mockito-all', version: '1.8.0'
+    testCompile group: 'junit', name: 'junit', version: '4.11'
+}
+
+shadowJar {
+    // We need to exclude CollectD API library since it is already provided by CollectD.
+    // Nevertheless exclude doesn't work, probably due to its dependency type (file), so we simply relocating it.
+    relocate('org.collectd.api', 'deprecated.org.collectd.api')
```
amnonshmsft: Assuming that this is a function call it would be better to use Groovy convention of calling methods without using parenthesis

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+    private static final String PLUGIN_NAME = "ApplicationInsightsWriter";
+    private static final String DEFAULT_AI_LOGGER_OUTPUT = "CONSOLE";
+
+    // endregion Consts
+
+    // region Members
+
+    private TelemetryConfiguration telemetryConfiguration;
+    private TelemetryClient telemetryClient;
+    private Dictionary<String, PluginExclusion> excludedPluginsDictionary;
```
amnonshmsft: [Dictionary](http://docs.oracle.com/javase/6/docs/api/java/util/Dictionary.html) class is obsolete so unless it is required by CollectD framework you should use the Map interface instead.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+     */
+    public ApplicationInsightsWriter() {
+        Collectd.registerConfig(PLUGIN_NAME, this);
+        Collectd.registerInit(PLUGIN_NAME, this);
+        Collectd.registerWrite(PLUGIN_NAME, this);
+        Collectd.registerShutdown(PLUGIN_NAME, this);
+
+        this.logger = new ApplicationInsightsWriterLogger();
+        this.telemetryConfiguration = TelemetryConfiguration.getActive();
+        this.excludedPluginsDictionary = new Hashtable<String, PluginExclusion>();
```
amnonshmsft: Is synchronization an issue? If not then HashMap is considered faster than Hashtable.
You can read about the differences between the two in [here](http://stackoverflow.com/questions/267312/difference-between-a-hashmap-and-a-dictionary-adt) and [here](http://javahungry.blogspot.com/2014/03/hashmap-vs-hashtable-difference-with-example-java-interview-questions.html).

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+    /**
+     * Writes the given values to Application Insights.
+     * @param valueList The values to be sent.
+     * @return Zero on success, non-zero otherwise.
+     */
+    public int write(ValueList valueList) {
+        // We're avoiding writing traces in this method to not spam the logs, as this method is being called very frequently.
+        logger.logDebug("Writing values (" + valueList.getValues().size() + "):\n" + valueList.toString());
+
+        String pluginName = valueList.getPlugin();
```
amnonshmsft: If this method is being called very frequently won't it be best to store the data source to be included in a cache instead of iterating over all plugins and calling `isDataSourceExcluded` on them?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+
+    // region Private
+
+    private static MetricTelemetry createMetricTelemetry(ValueList valueList, int index) {
+        MetricTelemetry telemetry = new MetricTelemetry();
+
+        DataSource dataSource = getDataSource(valueList.getDataSet(), index);
+        String telemetryName = generateMetricName(valueList, dataSource);
+        telemetry.setName(telemetryName);
+        telemetry.getContext().getUser().setUserAgent(COLLECTD_USER_AGENT);
```
amnonshmsft: I might be wrong here but as far as I understand the user agent field was meant to allow AI customers to understand what browsers are used by their customers. Therefore I'm not sure that setting this value for CollectD is the right thing to do as it might affect the UX (we should consult our Redmond partners about that).

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+            telemetry.setMin(dataSource.getMin());
+            telemetry.setMax(dataSource.getMax());
+        }
+
+        return telemetry;
+    }
+
+    private static void setHostMachineProperty(MetricTelemetry telemetry, String host) {
+        if (LocalStringsUtils.isNullOrEmpty(host)) {
+            host = "N/A";
```
amnonshmsft: Consider defining `N/A` as a constant and moving it to the constant sections

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
```diff
@@ -0,0 +1,283 @@
...
+            source.append('/').append(typeInstance);
+        }
+
+        return source.toString();
+    }
+
+    protected static String generateMetricName(ValueList valueList, DataSource dataSource) {
+        String metricName = generateSourceName(valueList);
+
+        if (dataSource != null) {
```
amnonshmsft: Can dataSource be empty?

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ConfigurationException.java
```diff
@@ -0,0 +1,31 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.collectd;
+
+/**
+ * Created by yonisha on 5/6/2015.
+ */
+public class ConfigurationException extends Exception {
```
amnonshmsft: `javax.naming` package already defines [ConfigurationException](http://docs.oracle.com/javase/7/docs/api/javax/naming/ConfigurationException.html) so please check whether we can use it instead of this class.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/internal/ApplicationInsightsWriterLogger.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public ApplicationInsightsWriterLogger() {
+        this.enabled = true;
+    }
+
+    /**
+     * Constructs new instance of @ApplicationInsightsWriterLogger
+     * @param enabled True to enable the plugin, false otherwise.
+     */
+    public ApplicationInsightsWriterLogger(boolean enabled) {
```
amnonshmsft: If you're not providing a setter for enabling/disabling the logger so what's the point in providing a constructor that can create a disabled object?
Also I couldn't find where you're using this constructor except in the UT.

collectd/src/main/java/com/microsoft/applicationinsights/collectd/internal/WriterConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+    public static WriterConfiguration buildConfiguration(OConfigItem configuration) throws ConfigurationException {
+        logger.logDebug("Parsing configuration: " + configuration);
+
+        WriterConfiguration writerConfiguration = new WriterConfiguration();
+        List<OConfigItem> children = configuration.getChildren();
+
+        for (OConfigItem child : children) {
+            String key = child.getKey();
+
+            if (key.equalsIgnoreCase(INSTRUMENTATION_KEY_CONFIGURATION_KEY)) {
```
amnonshmsft: For improved readability and cleaner code consider handling each key in its own function, i.e.

    if (key.equalsIgnoreCase(INSTRUMENTATION_KEY_CONFIGURATION_KEY)) {
        writerConfiguration.instrumentationKey = parseInstrumentationKey(child);
    } else if (key.equalsIgnoreCase(SDK_LOGGER_CONFIGURATION_KEY)) {
        writerConfiguration.isLoggerEnabled = parseIsLoggerEnabled(child);
    } else if ...

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: Not an issue. thanks

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: I always add new line before a trace and before a return statement, it is more readable imo

collectd/src/main/java/com/microsoft/applicationinsights/collectd/ApplicationInsightsWriter.java
Yonisha: Not sure I understand how you can avoid this check

PR #178, opened 2015-05-07
Added the ability to set the size of the buffer and the timeout to wa…

gupele:
…it supporting configuration and online changes

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
@@ -61,11 +61,17 @@ import com.google.common.base.Preconditions;
  */
 public final class InProcessTelemetryChannel implements TelemetryChannel {
     private final static int DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 500;
+    private final static int MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 1;
+    private final static int MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 500;
 
     private final static int TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS = 10;
+    private final static int MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS = 1;
+    private final static int MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS = 30;
 
     private final static String DEVELOPER_MODE = "DeveloperMode";
-    private final static String EndpointAddress = "EndpointAddress";
+    private final static String ENDPOINT_ADDRESS = "EndpointAddress";
+    private final static String MAX_TELEMETRY_ITEMS_IN_QUEUE = "MaxTelemetryItemsInQueue";
+    private final static String SEND_INTERVAL_IN_SECONDS = "SendIntervalInSeconds";
 
     private boolean developerMode = false;
     private static TransmitterFactory s_transmitterFactory;
@@ -86,7 +92,20 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
      * @param developerMode True will behave in a 'non-production' mode to ease the debugging
      */
     public InProcessTelemetryChannel(String endpointAddress, boolean developerMode) {
-        initialize(endpointAddress, developerMode);
+        initialize(endpointAddress, developerMode, DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER, TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS);
+    }
+
+    /**
+     * Ctor
+     * @param endpointAddress Must be empty string or a valid uri, else an exception will be thrown
+     * @param developerMode True will behave in a 'non-production' mode to ease the debugging
+     * @param maxQueueItemCount Max number of Telemetries we keep in the buffer, when reached we will send the buffer
+     *                          Note, value should be between TRANSMIT_BUFFER_MIN_TIMEOUT_IN_MILLIS and TRANSMIT_BUFFER_MAX_TIMEOUT_IN_MILLIS inclusive
+     * @param sendIntervalInMillis The maximum number of milliseconds to wait before we send the buffer
+     *                          Note, value should be between MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER and MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER inclusive
+     */
+    public InProcessTelemetryChannel(String endpointAddress, boolean developerMode, int maxQueueItemCount, int sendIntervalInMillis) {
+        initialize(endpointAddress, developerMode, maxQueueItemCount, sendIntervalInMillis);
     }
 
     /**
@@ -97,13 +116,23 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
     public InProcessTelemetryChannel(Map<String, String> nameAndValues) {
         boolean developerMode = false;
         String endpointAddress = null;
+        int maxQueueItemCount = DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER;
+        int sendIntervalInMillis = TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS;
 
         if (nameAndValues != null) {
             developerMode = Boolean.valueOf(nameAndValues.get(DEVELOPER_MODE));
-            endpointAddress = nameAndValues.get(EndpointAddress);
+            endpointAddress = nameAndValues.get(ENDPOINT_ADDRESS);
+            maxQueueItemCount = translateNumber(nameAndValues.get(MAX_TELEMETRY_ITEMS_IN_QUEUE),
+                    MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER,
+                    MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER,
+                    DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER);
+            sendIntervalInMillis = translateNumber(nameAndValues.get(SEND_INTERVAL_IN_SECONDS),
+                    MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS,
+                    MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS,
+                    TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS);
         }
 
-        initialize(endpointAddress, developerMode);
+        initialize(endpointAddress, developerMode, maxQueueItemCount, sendIntervalInMillis);
     }
 
     /**
@@ -124,7 +153,7 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
             this.developerMode = developerMode;
             int maxTelemetriesInBatch = this.developerMode ? 1 : DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER;
 
-            telemetryBuffer.setMaxTelemetriesInBatch(maxTelemetriesInBatch);
+            setMaxTelemetriesInBatch(maxTelemetriesInBatch);
         }
     }
 
@@ -181,11 +210,42 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
         telemetryBuffer.flush();
     }
 
+    /**
+     * Sets the buffer size
+     * @param maxTelemetriesInBatch should be between MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER
+     *                              and MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER inclusive
+     */
+    public void setMaxTelemetriesInBatch(int maxTelemetriesInBatch) {
+        if (maxTelemetriesInBatch >= MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER &&
+                maxTelemetriesInBatch <= MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER) {
+            telemetryBuffer.setMaxTelemetriesInBatch(maxTelemetriesInBatch);
+        } else {
+            InternalLogger.INSTANCE.trace("Value %s is ignored, value should be between %s and %s, inclusive",
+                    maxTelemetriesInBatch, MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER, MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER);
+        }
+
+    }
+
+    /**
+     * Sets the time tow wait before flushing the internal buffer
+     * @param transmitBufferTimeoutInSeconds should be between MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS
+     *                              and MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS inclusive
+     */
+    public void setTransmitBufferTimeoutInSeconds(int transmitBufferTimeoutInSeconds) {
+        if (transmitBufferTimeoutInSeconds >= MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS &&
+                transmitBufferTimeoutInSeconds <= MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS) {
+            telemetryBuffer.setTransmitBufferTimeoutInSeconds(transmitBufferTimeoutInSeconds);
+        } else {
+            InternalLogger.INSTANCE.trace("Value %s is ignored, value should be between %s and %s seconds, inclusive",
+                    transmitBufferTimeoutInSeconds, MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS, MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS);
+        }
+    }
+
     private void writeTelemetryToDebugOutput(Telemetry telemetry) {
         InternalLogger.INSTANCE.trace("InProcessTelemetryChannel sending telemetry");
     }
 
-    private synchronized void initialize(String endpointAddress, boolean developerMode) {
+    private synchronized void initialize(String endpointAddress, boolean developerMode, int maxQueueItemCount, int sendIntervalInSeconds) {
         makeSureEndpointAddressIsValid(endpointAddress);
 
         if (s_transmitterFactory == null) {
@@ -193,7 +253,7 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
         }
 
         telemetriesTransmitter = s_transmitterFactory.create(endpointAddress);
-        telemetryBuffer = new TelemetryBuffer(telemetriesTransmitter, DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER, TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS);
+        telemetryBuffer = new TelemetryBuffer(telemetriesTransmitter, maxQueueItemCount, sendIntervalInSeconds);
         setDeveloperMode(developerMode);
     }
 
@@ -215,4 +275,16 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
         }
     }
 
+    private int translateNumber(String valueString, int minimum, int maximum, int defaultValue) {
+        try {
+            int value = Integer.parseInt(valueString);
+            if (value < minimum || value > maximum) {
+                return defaultValue;
+            }
+
+            return value;
+        } catch (NumberFormatException e) {
+            return defaultValue;
+        }
+    }
 }
--- a/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
@@ -279,74 +279,26 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
         }
     }
 
-    /**
-     * The method will translate the numbers and will make sure it is within the limits
-     * If the value is not an int the default value will be returned
-     * Else, will return the outcome of 'makeSureNumberIsWithinLimits'
-     * @param propertyName The value to translate
-     * @param minimum The minimum value allowed, the value must be >= to that value
-     * @param maximum The maximum value allowed, the value must be <= to that value
-     * @param defaultValue Return this value in case the valueString is not an int
-     * @param propertyName The name of the property to display if needed in log messages
-     * @return The value within limits as stated in the param declarations above
-     */
-    private int translateNumber(Map<String, String> values, int minimum, int maximum, int defaultValue, String propertyName) {
-        String valueString = null;
-        try {
-            valueString = values.get(propertyName);
-            if (Strings.isNullOrEmpty(valueString)) {
-                return defaultValue;
-            }
-
-            int value = Integer.parseInt(valueString);
-            return makeSureNumberIsWithinLimits(value, minimum, maximum, propertyName);
-        } catch (NumberFormatException e) {
-            InternalLogger.INSTANCE.trace("'%s': bad format for value '%s', therefore using '%d'", propertyName, valueString, defaultValue);
-
-            return defaultValue;
-        }
-    }
-
-    /**
-     * Will return the value if (value >= minimum and value <= maximum) otherwise the limit that is closer
-     * @param value The value to check
-     * @param minimum The value should be >= from that value
-     * @param maximum The Value should be <= from that value
-     * @param propertyName The property name to display in log messages
-     * @return Will return the value if (value >= minimum and value <= maximum) otherwise the limit that is closer
-     */
-    private int makeSureNumberIsWithinLimits(int value, int minimum, int maximum, String propertyName) {
-        if (value < minimum) {
-            InternalLogger.INSTANCE.trace("'%s': value '%d' is lower than the limit '%d', therefore the limit will be used", propertyName, value, minimum);
-
-            return minimum;
-        } else if (value > maximum) {
-            InternalLogger.INSTANCE.trace("'%s': value '%d' is higher than the limit '%d', therefore the limit will be used", propertyName, value, maximum);
-
-            return maximum;
-        }
-
-        return value;
-    }
-
     private LimitsEnforcer createDefaultMaxTelemetryBufferCapacityEnforcer(Integer currentValue) {
         LimitsEnforcer maxItemsInBatchEnforcer =
-                LimitsEnforcer.createWithClosestLimitOnError(MAX_MAX_TELEMETRY_BUFFER_CAPACITY_NAME,
+                LimitsEnforcer.createWithClosestLimitOnError(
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -61,11 +61,17 @@
  */
 public final class InProcessTelemetryChannel implements TelemetryChannel {
     private final static int DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 500;
+    private final static int MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 1;
+    private final static int MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 500;
 
     private final static int TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS = 10;
+    private final static int MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS = 1;
+    private final static int MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS = 30;
```
amnonshmsft: Regarding the max values if the user decides to override these then I'm not sure that we should limit him, and even if we do I think that these limits should be about 10 times higher, e.g. ~5000 telemetries per container and ~300 seconds per transmition

gupele: Fine with me, I will speak to Harel about that to get the numbers

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -61,11 +61,17 @@
...
 
     private final static String DEVELOPER_MODE = "DeveloperMode";
-    private final static String EndpointAddress = "EndpointAddress";
+    private final static String ENDPOINT_ADDRESS = "EndpointAddress";
+    private final static String MAX_TELEMETRY_ITEMS_IN_QUEUE = "MaxTelemetryItemsInQueue";
```
amnonshmsft: These new two configuration elements should be aligned with the .Net SDK.

Also queue is an internal implementation details so it might be better to use MaxTelemetryItemsBatchSize or something similar

gupele: The names were approved by Harel

gupele: Better said, were given by Harel

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -97,13 +116,23 @@ public InProcessTelemetryChannel(String endpointAddress, boolean developerMode)
     public InProcessTelemetryChannel(Map<String, String> nameAndValues) {
         boolean developerMode = false;
         String endpointAddress = null;
+        int maxQueueItemCount = DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER;
+        int sendIntervalInMillis = TRANSMIT_BUFFER_DEFAULT_TIMEOUT_IN_SECONDS;
```
amnonshmsft: 1. Please use the full word "Milliseconds"
2. This variable is in milliseconds while the constant is in seconds so I guess you need to multiply by 1000

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
amnonshmsft: It is not just being aligned with Harel but we should strive to be as aligned with the .Net as possible, so we should do a double check here.
Anyway - this shouldn't delay your submission - we can update it later.

amnonshmsft:
Please add a context to the beginning of your commit, e.g. "Core - Channel Configurability - Added the ability to ..."

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -97,13 +116,23 @@ public InProcessTelemetryChannel(String endpointAddress, boolean developerMode)
...
 
         if (nameAndValues != null) {
             developerMode = Boolean.valueOf(nameAndValues.get(DEVELOPER_MODE));
-            endpointAddress = nameAndValues.get(EndpointAddress);
+            endpointAddress = nameAndValues.get(ENDPOINT_ADDRESS);
+            maxQueueItemCount = translateNumber(nameAndValues.get(MAX_TELEMETRY_ITEMS_IN_QUEUE),
```
amnonshmsft: Please align the parameters either by moving the first parameter one line below or moving the rest of the parameters to be aligned with the first one (the same goes to the next call as well)

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -181,11 +210,42 @@ public void flush() {
...
+     * Sets the buffer size
+     * @param maxTelemetriesInBatch should be between MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER
+     *                              and MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER inclusive
+     */
+    public void setMaxTelemetriesInBatch(int maxTelemetriesInBatch) {
+        if (maxTelemetriesInBatch >= MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER &&
+                maxTelemetriesInBatch <= MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER) {
+            telemetryBuffer.setMaxTelemetriesInBatch(maxTelemetriesInBatch);
+        } else {
+            InternalLogger.INSTANCE.trace("Value %s is ignored, value should be between %s and %s, inclusive",
```
amnonshmsft: Shouldn't it be a warning?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
gupele: Thanks, it is a typo, should be sendIntevalInSeconds

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -181,11 +210,42 @@ public void flush() {
         telemetryBuffer.flush();
     }
 
+    /**
+     * Sets the buffer size
+     * @param maxTelemetriesInBatch should be between MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER
+     *                              and MAX_NUMBER_OF_TELEMETRIES_PER_CONTAINER inclusive
+     */
+    public void setMaxTelemetriesInBatch(int maxTelemetriesInBatch) {
+        if (maxTelemetriesInBatch >= MIN_NUMBER_OF_TELEMETRIES_PER_CONTAINER &&
```
amnonshmsft: Please consult with Harel regarding the desired behavior here as we might want that if the provided buffer is bigger than the maximum we'll set the value to the maximum value and if it is lower than the minimum then we'll set it to a value lower than the minimum.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
gupele: We currently have 'error' or 'trace' we can change to 'error'

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -181,11 +210,42 @@ public void flush() {
...
+
+    }
+
+    /**
+     * Sets the time tow wait before flushing the internal buffer
+     * @param transmitBufferTimeoutInSeconds should be between MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS
+     *                              and MAX_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS inclusive
+     */
+    public void setTransmitBufferTimeoutInSeconds(int transmitBufferTimeoutInSeconds) {
+        if (transmitBufferTimeoutInSeconds >= MIN_TRANSMIT_BUFFER_TIMEOUT_IN_SECONDS &&
```
amnonshmsft: Same comment as above regarding the behavior in case the provided value is higher/lower than the limits.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
amnonshmsft: Here you're using **Batch** while in other place you're using "Queue". Please be consistent (I personally prefer "batch").

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +275,16 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
         }
     }
 
+    private int translateNumber(String valueString, int minimum, int maximum, int defaultValue) {
```
amnonshmsft: This might be the place to apply the over-max/under-min value behavior

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -116,17 +116,21 @@ public TelemetryBuffer(TelemetriesTransmitter sender, int maxTelemetriesInBatch,
...
      * @param maxTelemetriesInBatch The max amount of Telemetries that are allowed in a batch.
      */
-    public void setMaxTelemetriesInBatch(int maxTelemetriesInBatch)
-    {
-        this.maxTelemetriesInBatch = maxTelemetriesInBatch;
+    public void setMaxTelemetriesInBatch(int maxTelemetriesInBatch) {
+        synchronized (lock) {
+            if (telemetries != null && maxTelemetriesInBatch < telemetries.size()) {
+                // Request for smaller buffers, we flush if our buffer contains more elements
+                flush();
```
amnonshmsft: Is flush() public? If not we might want it to be so we can use it in other places as well.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ChannelXmlElement.java
```diff
@@ -66,6 +68,24 @@ public void setDeveloperMode(boolean developerMode) {
         this.developerMode = developerMode;
     }
 
+    public String getMaxTelemetryItemsInQueue() {
+        return maxTelemetryItemsInQueue;
+    }
+
+    @XmlElement(name="MaxTelemetryItemsInQueue")
+    public void setMaxTelemetryItemsInQueue(String maxTelemetryItemsInQueue) {
```
amnonshmsft: Queue or batch? :)

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBufferTest.java
```diff
@@ -347,6 +357,89 @@ public void testFlushWithSevenInTheBuffer() throws Exception {
...
+        mockSender.waitForFinish(1L);
+    }
+
+    @Test
+    public void testSetMaxTelemetriesInBatchWithSmallerSize() {
+        MockSender mockSender = new MockSender()
+                .setExpectedNumberOfScheduleSendCalls(0)
+                .setExpectedNumberOfSendNowCalls(1)
+                .setExpectedTelemetriesNumberInScheduleSend(0)
+                .setExpectedTelemetriesNumberInSendNow(2);
```
amnonshmsft: Please use a constant for the telemetries in SendNow instead of just 2 (here, and below)

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBufferTest.java
```diff
@@ -347,6 +357,89 @@ public void testFlushWithSevenInTheBuffer() throws Exception {
...
+            testedBuffer.add("mockTelemetry");
+        }
+        testedBuffer.setMaxTelemetriesInBatch(1);
+
+        mockSender.waitForFinish(1L);
+    }
+
+    @Test
+    public void testSetMaxTelemetriesInBatchWithSmallerSizeButLargerThanWhatInBuffer() {
+        MockSender mockSender = new MockSender()
```
amnonshmsft: Please use constants for numbers

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
amnonshmsft: It is not an error as it is not a critical issue and we can continue and on the other hand I don't want this message to get masked by other tracing messages.
So maybe we can just add warning to our internal logger?

core/src/main/java/com/microsoft/applicationinsights/internal/config/ChannelXmlElement.java
gupele: I also prefer batch, Harel will try to align us with the rest

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
gupele: flush is already public

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
gupele: will do

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -60,12 +61,19 @@
  * Created by gupele on 12/17/2014.
  */
 public final class InProcessTelemetryChannel implements TelemetryChannel {
-    private final static int DEFAULT_NUMBER_OF_TELEMETRIES_PER_CONTAINER = 500;
+    private final static int DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY = 500;
+    private final static int MIN_MAX_TELEMETRY_BUFFER_CAPACITY = 1;
+    private final static int MAX_MAX_TELEMETRY_BUFFER_CAPACITY = 1000;
+    private final static String MAX_MAX_TELEMETRY_BUFFER_CAPACITY_NAME = "MaxTelemetryBufferCapacity";
```
amnonshmsft: Is the parsing of these new options has UT coverage?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -86,24 +94,53 @@ public InProcessTelemetryChannel() {
...
     }
 
     /**
      * This Ctor will query the 'namesAndValues' map for data to initialize itself
      * It will ignore data that is not of its interest, this Ctor is useful for building an instance from configuration
-     * @param nameAndValues - The data passed as name and value pairs
+     * @param namesAndValues - The data passed as name and value pairs
      */
-    public InProcessTelemetryChannel(Map<String, String> nameAndValues) {
+    public InProcessTelemetryChannel(Map<String, String> namesAndValues) {
```
amnonshmsft: Why not just using java `Properties` object instead of a map?
Also, `namesAndValues` is a too generic name to be used in the static interface - what are these namesAndValues used for? Where someone can find a documentation about these?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+        return value;
+    }
+
+    private LimitsEnforcer createDefaultMaxTelemetryBufferCapacityEnforcer(Integer currentValue) {
+        LimitsEnforcer maxItemsInBatchEnforcer =
+                LimitsEnforcer.createWithClosestLimitOnError(MAX_MAX_TELEMETRY_BUFFER_CAPACITY_NAME,
+                        MIN_MAX_TELEMETRY_BUFFER_CAPACITY,
+                        MAX_MAX_TELEMETRY_BUFFER_CAPACITY,
+                        DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY,
+                                                             currentValue == null ? DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY : currentValue);
```
amnonshmsft: Alignment...

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+        return maxItemsInBatchEnforcer;
+    }
+
+    private LimitsEnforcer createDefaultSendIntervalInSecondsEnforcer(Integer currentValue) {
+        LimitsEnforcer sendIntervalInSecondsEnforcer =
+                LimitsEnforcer.createWithClosestLimitOnError(FLUSH_BUFFER_TIMEOUT_IN_SECONDS_NAME,
+                        MIN_FLUSH_BUFFER_TIMEOUT_IN_SECONDS,
+                        MAX_FLUSH_BUFFER_TIMEOUT_IN_SECONDS,
+                        DEFAULT_FLUSH_BUFFER_TIMEOUT_IN_SECONDS,
+                                                             currentValue == null ? DEFAULT_FLUSH_BUFFER_TIMEOUT_IN_SECONDS : currentValue);
```
amnonshmsft: Alignment

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+        } else if (value > maximum) {
+            InternalLogger.INSTANCE.trace("'%s': value '%d' is higher than the limit '%d', therefore the limit will be used", propertyName, value, maximum);
+
+            return maximum;
+        }
+
+        return value;
+    }
+
+    private LimitsEnforcer createDefaultMaxTelemetryBufferCapacityEnforcer(Integer currentValue) {
```
amnonshmsft: I think you can skip the word `Default` in `createDefaultMaxTelemetryBufferCapacityEnforcer`.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+                LimitsEnforcer.createWithClosestLimitOnError(MAX_MAX_TELEMETRY_BUFFER_CAPACITY_NAME,
+                        MIN_MAX_TELEMETRY_BUFFER_CAPACITY,
+                        MAX_MAX_TELEMETRY_BUFFER_CAPACITY,
+                        DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY,
+                                                             currentValue == null ? DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY : currentValue);
+
+        return maxItemsInBatchEnforcer;
+    }
+
+    private LimitsEnforcer createDefaultSendIntervalInSecondsEnforcer(Integer currentValue) {
```
amnonshmsft: I think you can skip the word `Default` in `createDefaultSendIntervalInSecondsEnforcer`

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+     * If the value is not an int the default value will be returned
+     * Else, will return the outcome of 'makeSureNumberIsWithinLimits'
+     * @param propertyName The value to translate
+     * @param minimum The minimum value allowed, the value must be >= to that value
+     * @param maximum The maximum value allowed, the value must be <= to that value
+     * @param defaultValue Return this value in case the valueString is not an int
+     * @param propertyName The name of the property to display if needed in log messages
+     * @return The value within limits as stated in the param declarations above
+     */
+    private int translateNumber(Map<String, String> values, int minimum, int maximum, int defaultValue, String propertyName) {
```
amnonshmsft: Who uses this method?
Also, wouldn't it belong to the LimitsEnforces?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -215,4 +279,75 @@ private void makeSureEndpointAddressIsValid(String endpointAddress) {
...
+
+    /**
+     * Will return the value if (value >= minimum and value <= maximum) otherwise the limit that is closer
+     * @param value The value to check
+     * @param minimum The value should be >= from that value
+     * @param maximum The Value should be <= from that value
+     * @param propertyName The property name to display in log messages
+     * @return Will return the value if (value >= minimum and value <= maximum) otherwise the limit that is closer
+     */
+    private int makeSureNumberIsWithinLimits(int value, int minimum, int maximum, String propertyName) {
```
amnonshmsft: If `translateNumber` is not used then `makeSureNumberIsWithinLimits` is probably redundant as well

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -92,12 +98,17 @@
...
+        }
 
-    public TransmissionFileSystemOutput(String folderPath) {
-        Preconditions.checkNotNull(folderPath, "folderPath must be a non-null value");
+        capacityEnforcer = LimitsEnforcer.createWithClosestLimitOnError(MIN_CAPACITY_MEGABYTES,
+                                                                        MAX_CAPACITY_MEGABYTES,
+                                                                        DEFAULT_CAPACITY_MEGABYTES,
+                                                                        MAX_TRANSMISSION_STORAGE_CAPACITY_NAME,
+                                                                        maxTransmissionStorageCapacity);
+        capacityInKB = capacityEnforcer.getCurrentValue() * 1024;
```
amnonshmsft: Since you're using 1024 in multiple places please define it as a constant (KILO?)

core/src/main/java/com/microsoft/applicationinsights/internal/config/ChannelXmlElement.java
```diff
@@ -66,6 +69,33 @@ public void setDeveloperMode(boolean developerMode) {
...
+    public String isMaxTransmissionStorageFilesCapacityInMB() {
+        return maxTransmissionStorageFilesCapacityInMB;
+    }
+
+    @XmlElement(name="MaxTransmissionStorageFilesCapacityInMB")
+    public void setMaxTransmissionStorageFilesCapacityInMB(String maxTransmissionStorageFilesCapacityInMB) {
+        this.maxTransmissionStorageFilesCapacityInMB = maxTransmissionStorageFilesCapacityInMB;
+    }
+
     public Map<String, String> getData() {
```
amnonshmsft: See my above comment about using `Properties` instead of `Map<String, String>`

core/src/main/java/com/microsoft/applicationinsights/internal/config/ChannelXmlElement.java
```diff
@@ -36,7 +36,10 @@
 public class ChannelXmlElement {
```
amnonshmsft: As far as I understand this class does not hold an XML element but the channel configuration, so I think it should be named accordingly

core/src/main/java/com/microsoft/applicationinsights/internal/util/LimitsEnforcer.java
```diff
@@ -0,0 +1,128 @@
...
+        this.minimum = minimum;
+        this.defaultValue = defaultValue;
+        this.currentValue = normalizeValue(currentValue);
+    }
+
+    public static LimitsEnforcer createWithDefaultOnError(String propertyName, int minimum, int maximum, int defaultValue, Integer currentValue) {
+        return new LimitsEnforcer(Type.DEFAULT_ON_ERROR, minimum, maximum, defaultValue, currentValue, propertyName);
+    }
+
+    public static LimitsEnforcer createWithDefaultOnError(int minimum, int maximum, int defaultValue, String propertyName, String currentValue) {
```
amnonshmsft: In the first and third overloads of this function the `propertyName` parameter is the first overload and in the second and forth versions it is the forth.
Please be consistent with the parameters order to prevent unnecessary confusion for the user.

core/src/main/java/com/microsoft/applicationinsights/internal/util/LimitsEnforcer.java
```diff
@@ -0,0 +1,128 @@
...
+        return new LimitsEnforcer(Type.DEFAULT_ON_ERROR, minimum, maximum, defaultValue, translate(propertyName, currentValue), propertyName);
+    }
+
+    public static LimitsEnforcer createWithClosestLimitOnError(String propertyName, int minimum, int maximum, int defaultValue, Integer currentValue) {
+        return new LimitsEnforcer(Type.CLOSEST_LIMIT_ON_ERROR, minimum, maximum, defaultValue, currentValue, propertyName);
+    }
+
+    public static LimitsEnforcer createWithClosestLimitOnError(int minimum, int maximum, int defaultValue, String propertyName, String currentValue) {
```
amnonshmsft: Same as above regarding the consistency of method overloads

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBufferTest.java
amnonshmsft: Reiterating my ask to use constants...

PR #184, opened 2015-05-12
Core - Added search folders when locating conf file and log printings

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -77,6 +79,14 @@ private InternalLogger() {
     }
 
     /**
+     * This method lets configuration publish messages before the logger is initialized
+     * @param configMode
+     */
+    public synchronized void configMode(boolean configMode) {
+        this.configMode = configMode;
+    }
```
Yonisha: I think this is not necessary, since we set it to true.

core/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -197,6 +207,21 @@ public void trace(String message, Object... args) {
...
+     * Either, before reading the file, or when there was error reading that configuration file
+     * @param requestLevel - The level of the message
+     * @param message - The message to print
+     * @param args - The arguments that are part of the message
+     */
+    public void logConfig(LoggingLevel requestLevel, String message, Object... args) {
+        if (configMode) {
+            if (requestLevel.getValue() >= LoggingLevel.INFO.getValue()) {
+                new ConsoleLoggerOutput().log(createMessage(requestLevel.toString(), message, args));
+            }
```
Yonisha: Isn't is more simple that 'config' will be a logging level such as that comes before info, warning, error?

core/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -266,8 +291,14 @@ private void onInitializationError(String errorMessage) {
...
-            loggerOutput.log(createMessage(requestLevel.toString(), message, args));
+        if (initialized) {
+            if (requestLevel.getValue() >= loggingLevel.getValue()) {
+                loggerOutput.log(createMessage(requestLevel.toString(), message, args));
+            }
+        } else {
+            if (requestLevel.getValue() >= LoggingLevel.WARN.getValue()) {
+                new ConsoleLoggerOutput().log(createMessage(requestLevel.toString(), message, args));
+            }
         }
```
Yonisha: Why creating a console logger if not initialized?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -78,9 +79,18 @@
...
+            String configurationFileName = getConfigurationFileName();
+            if (Strings.isNullOrEmpty(configurationFileName)) {
+                return;
+            }
+
+            ApplicationInsightsXmlConfiguration applicationInsights = builder.build(configurationFileName);
             if (applicationInsights == null) {
+                InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.ERROR, "Failed to read configuration file");
                 configuration.setChannel(new InProcessTelemetryChannel());
+                return;
```
Yonisha: +1. I intended to fix that as part of the findbug fixes :+1:

core/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
Yonisha: Another thing is that in the current implementation, we can simply use current methods instead of logConfig + passing the logging level

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,87 @@ private String getConfigurationFileName() {
...
-                    InternalLogger.INSTANCE.trace("Found configuration file: '%s'", configurationPath);
-                    return configurationPath;
-                }
-            } catch (URISyntaxException e) {
-                InternalLogger.INSTANCE.error("Failed to find configuration file: '%s'", e.getMessage());
+            // If not found as a resource, trying to load from the executing jar directory
+            String configurationFile = getConfFromSourceLocation();
+            if (configurationFile != null) {
+                return configurationFile;
             }
```
Yonisha: can you change to if-else please? more readable (imo)

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,87 @@ private String getConfigurationFileName() {
...
         } else {
             String configurationFile = resource.getFile();
-            InternalLogger.INSTANCE.trace("Found configuration file: '%s'", configurationFile);
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Found configuration file: '%s'", configurationFile);
             return configurationFile;
         }
+    }
+
+    private String getConfFromClassPath(ClassLoader classLoader) {
+        if ((classLoader instanceof URLClassLoader)) {
```
Yonisha: Can we use "System.getProperty("java.class.path")? this will return all locations and we can get all the directories from it

Yonisha: In addition, in what cases this class loader will be of type URLClassLoader?

core/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
amnonshmsft: I agree with Yoni that using `configMode` + `logConfig` looks kind of redundant. 
The way I see it is that we could generally simplify this by logging configuration issue through special channel (e.g. the `logConfig` method), and thus we won't need the `configMode` anymore
Also, I'd like to suggest to rename `logConfig` to something more generic (e.g. `adminLog`, `consoleLog`, `directLog`) so we won't be limited for using this channel only for configuration error reporting.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -78,9 +79,18 @@
      */
     public final void initialize(TelemetryConfiguration configuration) {
         try {
-            ApplicationInsightsXmlConfiguration applicationInsights = builder.build(getConfigurationFileName());
+            InternalLogger.INSTANCE.configMode(true);
+
+            String configurationFileName = getConfigurationFileName();
```
amnonshmsft: Actually you're looking for the file path and not for the file name which is know (ApplicationInsights.xml)

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -78,9 +79,18 @@
...
+            if (Strings.isNullOrEmpty(configurationFileName)) {
+                return;
+            }
+
+            ApplicationInsightsXmlConfiguration applicationInsights = builder.build(configurationFileName);
```
amnonshmsft: Please rename the variable name to reflect that it is a configuration, i.e. `applicationInsightsConfig`.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -97,7 +107,9 @@ public final void initialize(TelemetryConfiguration configuration) {
 
             initializeComponents(configuration);
         } catch (Exception e) {
-            InternalLogger.INSTANCE.error("Failed to initialize configuration, exception: %s", e.getMessage());
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.ERROR, "Failed to initialize configuration, exception: %s", e.getMessage());
+        } finally {
+            InternalLogger.INSTANCE.configMode(false);
```
amnonshmsft: See Yoni's and my comments about using `configMode`. Also, if you decide to stick with it please use java convention, i.e. `setConfigMode`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
         ClassLoader classLoader = TelemetryConfigurationFactory.class.getClassLoader();
```
amnonshmsft: See my above comment about this method name

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
         URL resource = classLoader.getResource(fileToParse);
 
-        // If not found as a resource, trying to load from the executing jar directory
         if (resource == null) {
-            try {
-                String jarFullPath = TelemetryConfigurationFactory.class.getProtectionDomain().getCodeSource().getLocation().toURI().getPath();
-                File jarFile = new File(jarFullPath);
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Did not find resource configuration file");
```
amnonshmsft: "Could not find resource named ApplicationInsights.xml"

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
-                    String jarDirectory = jarFile.getParent();
-                    String configurationPath = jarDirectory + File.separator + fileToParse;
-
-                    InternalLogger.INSTANCE.trace("Found configuration file: '%s'", configurationPath);
-                    return configurationPath;
-                }
-            } catch (URISyntaxException e) {
-                InternalLogger.INSTANCE.error("Failed to find configuration file: '%s'", e.getMessage());
+            // If not found as a resource, trying to load from the executing jar directory
+            String configurationFile = getConfFromSourceLocation();
```
amnonshmsft: `configurationFile` --> `configurationFilePath`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
         ClassLoader classLoader = TelemetryConfigurationFactory.class.getClassLoader();
         URL resource = classLoader.getResource(fileToParse);
 
-        // If not found as a resource, trying to load from the executing jar directory
         if (resource == null) {
```
amnonshmsft: Your code would be more readable if you'll start with handling the case where `resource != null` as this will eliminate the need for an `else`:

    URL resource = classLoader.getResource(fileToParse); 
    if (resource != null) { 
        String configurationFile = resource.getFile(); 
        InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Found configuration file: '%s'", configurationFile); 
        return configurationFile; 
    } 

    InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Did not find resource configuration file"); 
    // If not found as a resource, trying to load from the executing jar directory 
    String configurationFile = getConfFromSourceLocation(); 
    if (configurationFile != null) { 
        return configurationFile; 
    } 
 
    // If not found as a resource, trying to load from the class path 
    return getConfFromClassPath(classLoader);

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
amnonshmsft: It may be a religious thing, but generally I prefer having as little `else`s as possible. The reason for that is that once we complete with the `if` part we either returned or continuing in the main flow of the function and not in a sub-flow.
This also makes the code less nested, especially in cases where there are several nested conditions.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
         } else {
             String configurationFile = resource.getFile();
-            InternalLogger.INSTANCE.trace("Found configuration file: '%s'", configurationFile);
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Found configuration file: '%s'", configurationFile);
             return configurationFile;
         }
+    }
+
+    private String getConfFromClassPath(ClassLoader classLoader) {
+        if (!(classLoader instanceof URLClassLoader)) {
```
amnonshmsft: Doesn't this mean a bug?
In this case it would be more appropriate throwing an invalid argument exception

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
+            return null;
+        }
+
+        HashSet<String> checkedUrls = new HashSet<String>();
+
+        URL[] urls = ((URLClassLoader)classLoader).getURLs();
+        for (URL url : urls) {
+            String configurationPath = url.getPath();
+            if (configurationPath.endsWith(".jar")) {
```
amnonshmsft: Wouldn't it be better letting standard classes such as FilenameUtils do this work for us?
Also, which I can understand this code it is still a bit criptic - if you decide to leave it please wrap it in a method, e.g. `extractUrlFolder` or something similar

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
+                int index = configurationPath.lastIndexOf('/');
+                if (index != -1) {
+                    configurationPath = configurationPath.substring(0, index + 1);
+                } else {
+                    continue;
+                }
+            } else {
+                if (!configurationPath.endsWith("/")) {
```
amnonshmsft: NormalizePath?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
+                    configurationPath += '/';
+                }
+            }
+
+            if (checkedUrls.contains(configurationPath)) {
+                continue;
+            }
+
+            String configurationFile = configurationFileExists(configurationPath);
```
amnonshmsft: I would expect a method named `configurationFileExists` to return a Boolean and not a string. Please rename this function so it would better reflect what it does and returns.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
+                InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.ERROR, "Can not access folder '%s'", jarFullPath);
+            }
+        } catch (URISyntaxException e) {
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.ERROR, "Failed to find configuration file, exception: '%s'", e.getMessage());
+        }
+
+        return null;
+    }
+
+    private String configurationFileExists(String path) {
```
amnonshmsft: See my above comment about the name of this method

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -327,27 +339,94 @@ private String getConfigurationFileName() {
...
+        File configFile = new File(path, fileToParse);
+
+        if (configFile.exists()) {
+            InternalLogger.INSTANCE.logConfig(InternalLogger.LoggingLevel.INFO, "Found configuration file in: '%s'", path);
+            return configFile.getAbsolutePath();
+        } else {
```
amnonshmsft: As you're returning in the `if` part there's no need for an else - just place your code after the `if` scope

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: Not sure, but I will do that on a separate commit

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I agree

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: No, that is not a bug although usually the condition will be true and if so we will try to find where the urls are

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I think that this should do

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

PR #185, opened 2015-05-13
Adding 32Bit support

Yonisha:


core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -157,11 +157,7 @@ private static boolean initNativeCode() {
      */
     private static boolean loadLibrary() throws IOException {
         String model = System.getProperty("sun.arch.data.model");
```
amnonshmsft: Regarding `String model = System.getProperty("sun.arch.data.model");` - would this be right property to check also for Java OpenJDK (Zulu) distribution?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -68,22 +67,23 @@
...
                 InternalLogger.INSTANCE.error("Jni connector is only used on Windows OS.");
                 return false;
             }
 
-            if (loadLibrary()) {
-                return initNativeCode();
-            } else {
-                InternalLogger.INSTANCE.error("Failed to load native dll, Windows performance counters will not be used.");
-            }
+            loadLibrary();
```
amnonshmsft: `loadLibrary` is a too generic name - please rename to `loadNativeLibrary`

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -68,22 +67,23 @@
...
+            initializationResult = initNativeCode();
         } catch (Throwable e) {
-            InternalLogger.INSTANCE.error("Failed to load native dll, Windows performance counters will not be used: '%s'.", e.getMessage());
+            InternalLogger.INSTANCE.error(
+                    "Failed to load native dll, Windows performance counters will not be used. " +
```
amnonshmsft: Alignment - first line (`"Failed to load...') should be 4 chars after `InternalLogger...`.
Second line `"Please make sure...` should be aligned with the first line.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -68,22 +67,23 @@
...
             }
 
-            if (loadLibrary()) {
-                return initNativeCode();
-            } else {
-                InternalLogger.INSTANCE.error("Failed to load native dll, Windows performance counters will not be used.");
-            }
+            loadLibrary();
+            initializationResult = initNativeCode();
         } catch (Throwable e) {
```
amnonshmsft: Why not initializing the native code as part of `loadNativeLibrary`?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
Yonisha: Good point.
I've verified it on zulu 1.6 64 bit JVM and it works as expected

PR #191, opened 2015-05-18
Removing set/get value functions and fixing duration in json

Yonisha:
PR contains the following:
1. Removing set/get value method from dependency telemetry.
2. Used the duration value in the json sent to the BE.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/Duration.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/Duration.java
@@ -122,6 +122,17 @@ public final class Duration {
         return milliseconds;
     }
 
+    /**
+     * Gets the total milliseconds of the duration.
+     * @return The total milliseconds of the duration.
+     */
+    public long getTotalMilliseconds() {
+        return (days * SECONDS_IN_ONE_DAY * 1000) +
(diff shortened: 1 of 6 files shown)


samples/src/main/java/com/microsoft/applicationinsights/sample/AiCore.java
```diff
@@ -25,9 +25,12 @@
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.jar.JarFile;
```
Yonisha: will remove it

core/src/main/java/com/microsoft/applicationinsights/telemetry/Duration.java
```diff
@@ -122,6 +122,17 @@ public int getMilliseconds() {
         return milliseconds;
     }
 
+    /**
+     * Gets the total milliseconds of the duration.
+     * @return The total milliseconds of the duration.
+     */
+    public long getTotalMilliseconds() {
+        return (days * SECONDS_IN_ONE_DAY * 1000) +
+                (hours * SECONDS_IN_ONE_HOUR * 1000) +
```
amnonshmsft: alignment

PR #192, opened 2015-05-18
Disabling user/session cookies generation

Yonisha:
Disabling user/session cookies generation by defaulting the modules to NOT generate cookies. Only if a cookie contains in the request, the module will use it information to update the telemetry sent to AI BE.

### DIFF
--- a/web/src/test/java/com/microsoft/applicationinsights/web/extensibility/modules/WebSessionTrackingTelemetryModuleTests.java
+++ b/web/src/test/java/com/microsoft/applicationinsights/web/extensibility/modules/WebSessionTrackingTelemetryModuleTests.java
@@ -89,7 +89,6 @@ public class WebSessionTrackingTelemetryModuleTests {
 
     // region Tests
 
-    @Ignore
     @Test
     public void testNewSessionIsCreatedWhenCookieNotExist() throws Exception {
         CookiesContainer cookiesContainer = HttpHelper.sendRequestAndGetResponseCookie();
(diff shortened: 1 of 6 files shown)


web/src/test/java/com/microsoft/applicationinsights/web/extensibility/modules/WebSessionTrackingTelemetryModuleTests.java
```diff
@@ -89,6 +89,7 @@ public static void classCleanup() throws Exception {
 
     // region Tests
 
+    @Ignore
     @Test
```
amnonshmsft: Several things:
 1. Please be consistent with the order of @Ignore and @Test. 
 2. Is it possible to keep these tests alive and to change the generateNewSessions/generateNewUsers programatically? The reason for that is that if one day we'll decide to roll back this value for these members we won't need to recall to un-ignore these tests as well.
 3. Also, if on the other hand we'll decide to drop this code altogether then it would be better to locate all the ignored tests next to each other

web/src/test/java/com/microsoft/applicationinsights/web/extensibility/modules/WebUserTrackingTelemetryModuleTests.java
```diff
@@ -62,11 +66,17 @@
     @BeforeClass
     public static void classInitialize() throws Exception {
         server.start();
+
+        // Set mock channel
+        channel = MockTelemetryChannel.INSTANCE;
```
amnonshmsft: I'm not sure that this change is related to the user/session cookies (i.e. try not to mix unrelated changes in a single commit)

PR #199, opened 2015-05-24
Core: fixing configuration file resolution algorithm

amnonshmsft:


core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -333,75 +331,57 @@ private boolean setChannel(ChannelXmlElement channelXmlElement, TelemetryConfigu
...
-            }
+        return configurationFile;
+    }
 
-            String configurationFile = getConfigurationAbsolutePath(configurationPath);
-            if (configurationFile != null) {
-                return configurationFile;
-            } else {
-                checkedUrls.add(configurationPath);
+    private String getConfigurationFromDefaultClassLoader() {
```
gupele: should be 'getConfigurationFromCurrentClassLoader'

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -333,75 +331,57 @@ private boolean setChannel(ChannelXmlElement channelXmlElement, TelemetryConfigu
...
-                } else {
-                    continue;
-                }
-            } else {
-                if (!configurationPath.endsWith("/")) {
-                    configurationPath += '/';
-                }
-            }
+        if (configurationFile != null) {
+            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Configuration file found in: '%s'", configurationFile);
```
gupele: "Configuration file has been successfully found in:......." this is what is written in the Blob

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -333,75 +331,57 @@ private boolean setChannel(ChannelXmlElement channelXmlElement, TelemetryConfigu
...
 
-        // If not found as a resource, trying to load from the class path
-        return getConfFromClassPath(classLoader);
-    }
-
-    private String getConfFromClassPath(ClassLoader classLoader) {
-        if (!(classLoader instanceof URLClassLoader)) {
-            return null;
+        // If still not found try to get it from the class path
+        if (configurationFile == null) {
```
gupele: This should be in first check of the 'if (configurationFile == null)' this is activated only if the second one produces null

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -85,7 +83,7 @@ public final void initialize(TelemetryConfiguration configuration) {
                 return;
             }
 
-            ApplicationInsightsXmlConfiguration  applicationInsightsConfig = builder.build(configurationFile);
+            ApplicationInsightsXmlConfiguration applicationInsightsConfig = builder.build(configurationFile);
```
gupele: I think all this logic should be placed in a dedicated class, which is easier to test

amnonshmsft:
Code submitted by Guy in another PR.

PR #200, opened 2015-05-25
Core - Fixed and added tests to the configuration file search

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
```diff
@@ -0,0 +1,172 @@
...
+
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+
+/**
+ * Created by gupele on 5/25/2015.
+ */
+public final class ConfigurationFileLocator {
```
amnonshmsft: Why not making this a static class?

PR #201, opened 2015-05-26
[BugFix][Cookies] Updating user/session modules to be aligned with JS…

Yonisha:
… cookies format

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
@@ -60,7 +60,7 @@ public class DateTimeUtils {
      * @throws java.text.ParseException Thrown when failed to parse the given string.
      */
     public static Date parseRoundTripDateString(String roundTripString) throws ParseException {
-        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSSS'Z'");
+        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH");
         Date date = roundTripDateFormat.parse(roundTripString);
 
         return date;
@@ -72,7 +72,7 @@ public class DateTimeUtils {
      * @return Round-trip date string
      */
     public static String formatAsRoundTripDate(Date date) {
-        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSSS'Z'");
+        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH");
         return roundTripDateFormat.format(date);
     }
 }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
@@ -32,6 +32,8 @@ import java.util.concurrent.TimeUnit;
  */
 public class DateTimeUtils {
 
+    private static final String ROUND_TRIP_DATE_FORMAT = "yyyy-MM-dd'T'HH";
+
     private DateTimeUtils() {
     }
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
```diff
@@ -60,7 +60,7 @@ public static long getDateDiff(Date laterDate, Date date, TimeUnit timeUnit) {
      * @throws java.text.ParseException Thrown when failed to parse the given string.
      */
     public static Date parseRoundTripDateString(String roundTripString) throws ParseException {
```
amnonshmsft: What's the difference between `parseRoundTripDateString` and `formatAsRoundTripDate`?
Do we really need both these methods?
If we do please at least refactor

core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
```diff
@@ -60,7 +60,7 @@ public static long getDateDiff(Date laterDate, Date date, TimeUnit timeUnit) {
...
-        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSSSSS'Z'");
+        SimpleDateFormat roundTripDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH");
```
amnonshmsft: Please move `"yyyy-MM-dd'T'HH"` into a constant

core/src/test/java/com/microsoft/applicationinsights/internal/util/DateTimeUtilsTests.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Created by yonisha on 5/26/2015.
+ */
+public class DateTimeUtilsTests {
+
+    @Test
+    public void testParseRoundTripDateString() throws ParseException {
+        final String dateStr = "2015-05-26T07:00:14+01:00";
+        String reversedDateStr = parseRoundTripDateString(dateStr);
```
amnonshmsft: `parseRoundTripDateString` returns a Date rather than a string so it would be better to explicitly call toString() here.

core/src/test/java/com/microsoft/applicationinsights/internal/util/DateTimeUtilsTests.java
```diff
@@ -0,0 +1,60 @@
+package com.microsoft.applicationinsights.internal.util;
```
amnonshmsft: Would it be valuable to add negative tests as well here?

core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
```diff
@@ -60,7 +60,7 @@ public static long getDateDiff(Date laterDate, Date date, TimeUnit timeUnit) {
      * @throws java.text.ParseException Thrown when failed to parse the given string.
```
amnonshmsft: General - does this change also impacts the server-side cookies?

web/src/test/java/com/microsoft/applicationinsights/web/internal/cookies/UserCookieTests.java
```diff
@@ -74,8 +74,9 @@ public static void initialize() throws Exception {
     public void testCookieParsedSuccessfully() throws Exception {
         UserCookie userCookie = new UserCookie(defaultCookie);
 
+        Date expectedAcquisitionTime = DateTimeUtils.parseRoundTripDateString(DateTimeUtils.formatAsRoundTripDate(acquisitionTime));
```
amnonshmsft: What is the meaning of `expected` here?

core/src/main/java/com/microsoft/applicationinsights/internal/util/DateTimeUtils.java
Yonisha: no

core/src/test/java/com/microsoft/applicationinsights/internal/util/DateTimeUtilsTests.java
Yonisha: Don't think so..

core/src/test/java/com/microsoft/applicationinsights/internal/util/DateTimeUtilsTests.java
Yonisha: nope, this is not the DateTimeUtils method, but a private one.

PR #202, opened 2015-05-26
Core - Added clear log messages for jars mismatches

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -81,6 +82,11 @@ public static TransmissionNetworkOutput create() {
     }
 
     public static TransmissionNetworkOutput create(String endpoint) {
+        if (!ClassDataUtils.INSTANCE.isClassExists("org.apache.http.conn.HttpClientConnectionManager")) {
+            String errorMessage = String.format("HttpClient Jars mismatch: please make sure version 4.3.1 of HttpClient is used");
```
amnonshmsft: I think that we should be good with v4.3.
Can you please verify it? (you can simply change the version we're using in build.gradle and see that things don't get broken)

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -81,6 +82,11 @@ public static TransmissionNetworkOutput create() {
...
+            throw new IllegalStateException(errorMessage);
```
amnonshmsft: According to the description of [IllegalStateException](http://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html):

> Signals that a method has been invoked at an illegal or inappropriate time

So I'm not sure that this is the right exception to throw here.
IMHO more suitable exceptions would be ClassNotFound exception with the message above.

core/src/main/java/com/microsoft/applicationinsights/internal/config/JaxbAppInsightsConfigurationBuilder.java
```diff
@@ -44,7 +44,11 @@ public ApplicationInsightsXmlConfiguration build(String filename) {
 
             return applicationInsights;
         } catch (JAXBException e) {
-            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Failed to parse configuration file: '%s'", e.getMessage());
+                if (e.getCause() != null) {
+                InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Failed to parse configuration file '%s': '%s'", filename, e.getCause().getMessage());
```
amnonshmsft: 1. Indentation
2. Please make sure that this would print the right message in case the XML contains forbidden leading characters or just unexpected tags

gupele: made sure

core/src/main/java/com/microsoft/applicationinsights/internal/reflect/ClassDataUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    INSTANCE;
+
+    private ClassDataVerifier verifier = new ClassDataVerifierImpl();
+
+    public void initialize() {
+        verifier = new ClassDataVerifierImpl();
+    }
+
+    @Override
+    public boolean isClassExists(String className) {
```
amnonshmsft: I'm not sure that `isClassExists` (and `isMethodExists` below) is a valid English. I think a better name then would be to replace `is` with `verify` or `check`, i.e. `verifyClassExists` and `verifyMethodExists`. WDYT?

gupele: Agree

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -92,6 +97,13 @@ private void onException(Exception e) {
      */
     public void init(FilterConfig config){
         try {
+            if (!ClassDataUtils.INSTANCE.isMethodExists(javax.servlet.http.HttpServletResponse.class, "getStatus")) {
+                InternalLogger.INSTANCE.error(
```
amnonshmsft: Shouldn't this log also appear in the appserver logs, i.e. should be logged always?

gupele: correct, will be changed to 'logAlways'

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -92,6 +97,13 @@ private void onException(Exception e) {
...
+                        "Servlet Jars mismatch. The Java SDK needs the Servlet version 3.0, Web request tracking filter will be disabled.");
```
amnonshmsft: Please change to the following:
_Unsupported servlet version. The Application Insights web request tracking filter requires the provided javax.servlet library to implement servlet spec 3.0 or above. The web request tracking filter will be disabled._

gupele: will do

PR #203, opened 2015-05-27
[CollectD][Build] Build refinements and repo documentation (md file) …

Yonisha:
…update

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,8 @@ Prerequisites
 
 Getting started
 ---------------
-1.  Set JAVA_HOME environment variable to point to point to the [JDK installation directory]
-2.  Recommended IDE is IntelliJ idea by using ‘Import Project’ rather than ‘Open Project’.
-3.  You can use gradlew.bat on Windows to get gradle installed.
+1.  Set JAVA_HOME environment variable to point to point to the JDK installation directory
+2.  To build run ./gradlew build on Linux systems or gradlew.bat build on Windows systems.
 
 Using Eclipse IDE
 -----------------
@@ -26,14 +25,16 @@ Using Eclipse IDE
 
 CollectD Plugin - Optional
 --------------------------
-For contributing code to Application Insights CollectD writer plugin, please do the following:
-1.  Download CollectD Java API sources and compile them using JDK 1.6 (OpenSDK is preferable).
+To build Application Insights CollectD writer plugin, please do the following:
+
+1.  Download CollectD Java API sources and compile them using JDK 1.6.
     The output jar should be named: 'collectd-api.jar'.
     More info on compiling CollectD sources can be found here: https://collectd.org/dev-info.shtml
 2.  Create a new directory for CollectD library you just created, and set a new environment variable 'COLLECTD_HOME'
     pointing to that folder.   
 3.  Copy the new jar into %COLLECTD_HOME%/lib
-4.  Refresh Application Insights project. CollectD writer plugin sub-project should now be loaded.
+4.  Reload Application Insights project. CollectD writer plugin sub-project should now be loaded.
+    IDE restart may be required in order to identify the new environment variable.
 
 Notes
 -----
--- a/README.md
+++ b/README.md
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
 
 Prerequisites
 -------------
-1. Java SDK 1.6 or higher
+1.  Java SDK 1.6 or higher
 
 Getting started
 ---------------
-1. Set JAVA_HOME environment variable to point to point to the [JDK installation directory]
-2. Recommended IDE is IntelliJ idea by using ‘Import Project’ rather than ‘Open Project’.
-3. You can use gradlew.bat on Windows to get gradle installed.
+1.  Set JAVA_HOME environment variable to point to point to the [JDK installation directory]
+2.  Recommended IDE is IntelliJ idea by using ‘Import Project’ rather than ‘Open Project’.
+3.  You can use gradlew.bat on Windows to get gradle installed.
 
 Using Eclipse IDE
 -----------------
-1. Install gradle from http://www.gradle.org/installation
-2. Add GRADLE_HOME/bin to your PATH environment variable
-3. In build.gradle add line [apply plugin: "eclipse"]
-4. In Eclipse used File->Import Existing Project in a workspace.
-5. Use [gradle build] to build the project from the command line.
+1.  Install gradle from http://www.gradle.org/installation
+2.  Add GRADLE_HOME/bin to your PATH environment variable
+3.  In build.gradle add line [apply plugin: "eclipse"]
+4.  In Eclipse used File->Import Existing Project in a workspace.
+5.  Use [gradle build] to build the project from the command line.
+
+CollectD Plugin - Optional
+--------------------------
+For contributing code to Application Insights CollectD writer plugin, please do the following:
+1.  Download CollectD Java API sources and compile them using JDK 1.6 (OpenSDK is preferable).
+    The output jar should be named: 'collectd-api.jar'.
+    More info on compiling CollectD sources can be found here: https://collectd.org/dev-info.shtml
+2.  Create a new directory for CollectD library you just created, and set a new environment variable 'COLLECTD_HOME'
+    pointing to that folder.   
+3.  Copy the new jar into %COLLECTD_HOME%/lib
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
 -------------
-1. Java SDK 1.6 or higher
+1.  Java SDK 1.6 or higher
 
 Getting started
 ---------------
-1. Set JAVA_HOME environment variable to point to point to the [JDK installation directory]
-2. Recommended IDE is IntelliJ idea by using ‘Import Project’ rather than ‘Open Project’.
-3. You can use gradlew.bat on Windows to get gradle installed.
+1.  Set JAVA_HOME environment variable to point to point to the [JDK installation directory]
```
amnonshmsft: I think we can take "JDK installation directory" out of the square brackets

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+2.  Recommended IDE is IntelliJ idea by using ‘Import Project’ rather than ‘Open Project’.
```
amnonshmsft: I think we can remove this line as we have no reason to recommend our customers how to work.

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+3.  You can use gradlew.bat on Windows to get gradle installed.
```
amnonshmsft: Actually this talks about the standard gradle wrapper so I think it would be better to replace it with:
> To build run `./gradlew build` on Linux systems or `gradlew.bat build` on Windows systems

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
 
 Using Eclipse IDE
```
amnonshmsft: You might want to add instructions for _Using IntelliJ IDE_ here instead of lines in the Getting Started

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
 -----------------
-1. Install gradle from http://www.gradle.org/installation
-2. Add GRADLE_HOME/bin to your PATH environment variable
-3. In build.gradle add line [apply plugin: "eclipse"]
-4. In Eclipse used File->Import Existing Project in a workspace.
-5. Use [gradle build] to build the project from the command line.
+1.  Install gradle from http://www.gradle.org/installation
+2.  Add GRADLE_HOME/bin to your PATH environment variable
+3.  In build.gradle add line [apply plugin: "eclipse"]
```
amnonshmsft: Is this the recommended way to work with Eclipse?
I'm a bit surprised that there's no simpler way, more IntelliJ like

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+4.  In Eclipse used File->Import Existing Project in a workspace.
+5.  Use [gradle build] to build the project from the command line.
+
+CollectD Plugin - Optional
+--------------------------
+For contributing code to Application Insights CollectD writer plugin, please do the following:
```
amnonshmsft: For contributing code to --> To build

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+1.  Download CollectD Java API sources and compile them using JDK 1.6 (OpenSDK is preferable).
```
amnonshmsft: If CollectD builds smoothly also with Oracle SDK then I think that we can skip the recommendation for using the OpenSDK

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+    The output jar should be named: 'collectd-api.jar'.
+    More info on compiling CollectD sources can be found here: https://collectd.org/dev-info.shtml
```
amnonshmsft: Actually this item can be replaced by:
> Follow the instructions in https://collectd.org/dev-info.shtml to build CollectD project.
> The output jar should be named 'collectd-api.jar'.

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
+2.  Create a new directory for CollectD library you just created, and set a new environment variable 'COLLECTD_HOME'
+    pointing to that folder.   
+3.  Copy the new jar into %COLLECTD_HOME%/lib
+4.  Refresh Application Insights project. CollectD writer plugin sub-project should now be loaded.
```
amnonshmsft: What do you mean by "Refresh Application Insights project"?
I'm quite sure that since the environment of the IDE application does not include the `COLLECTD_HOME` environment variable just refreshing would not be enough but the user will have to re-open the IDE and then refresh.

README.md
```diff
@@ -8,21 +8,32 @@ Please refer to [AppInsights-Home](https://github.com/Microsoft/ApplicationInsig
...
-3. In build.gradle add line [apply plugin: "eclipse"]
-4. In Eclipse used File->Import Existing Project in a workspace.
-5. Use [gradle build] to build the project from the command line.
+1.  Install gradle from http://www.gradle.org/installation
+2.  Add GRADLE_HOME/bin to your PATH environment variable
+3.  In build.gradle add line [apply plugin: "eclipse"]
+4.  In Eclipse used File->Import Existing Project in a workspace.
+5.  Use [gradle build] to build the project from the command line.
+
+CollectD Plugin - Optional
```
amnonshmsft: This section is not aligned (use the View buttons to see how this file looks like in HTML).
Also as this item is optional consider placing it as a note (in a similar way to the instructions of how to build with Java 6)

collectd/build.gradle
```diff
@@ -24,7 +24,14 @@ plugins {
 }
 
 apply from: "$buildScriptsDir/common-java.gradle"
-apply from: "$buildScriptsDir/publishing.gradle"
```
amnonshmsft: Don't we want to publish this jar?

collectd/build.gradle
```diff
@@ -24,7 +24,14 @@ plugins {
...
+
+jar {
+    classifier = 'standalone'
```
amnonshmsft: As we're only distributing CollectD as a self-contained jar I'm not sure we need this classifier here.
If we decide to use it anyway then I think it would be better to investigation whether there is standard naming convention for such distributions

collectd/build.gradle
```diff
@@ -36,15 +43,11 @@ if (!collectDLibPath.exists()) {
 archivesBaseName = 'applicationinsights-collectd'
 
 dependencies {
-    compile project(':core')
+    compile group: 'com.microsoft.azure', name: 'applicationinsights-core', version: '0.9.5'
```
amnonshmsft: Why not using project dependency?
If you have no choice but to use external dependency then please do not use hard-coded fixed version.

collectd/build.gradle
```diff
@@ -36,15 +43,11 @@ if (!collectDLibPath.exists()) {
...
 
-// region Publishing properties
-
-projectPomName = project.msftAppInsights + " Java SDK CollectD Writer Plugin"
-projectPomDescription = "This is the CollectD writer plugin for " + project.msftAppInsightsJavaSdk
-
-// endregion Publishing properties
\ No newline at end of file
+// Building the shadow (fat) jar after compiling sources.
+tasks.assemble.doLast { tasks.shadowJar.execute() }
```
amnonshmsft: Can't this be achieved by using dependencies which is more "Gradle-y" than calling execute()?
i.e. 

    build.dependesOn shadowJar
    shadowJar.dependesOn assemble // not sure this is needed

README.md
Yonisha: you don't need to build the entire project, just the Java api sources. leaving as is.

collectd/build.gradle
Yonisha: Not to Maven, as discussed.

collectd/build.gradle
Yonisha: Disabled the default jar, we don't need it.

collectd/build.gradle
Yonisha: What's not gradle-y in the line above?
The two lines you wrote won't provide the what mine line provides.

PR #209, opened 2015-06-03
Agent - first version

gupele:


agent/build.gradle
```diff
@@ -0,0 +1,36 @@
+
```
amnonshmsft: License license license

agent/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+plugins {
+    id 'com.github.johnrengelman.shadow' version '1.2.1'
+}
+
+apply from: "$buildScriptsDir/common-java.gradle"
+apply from: "$buildScriptsDir/publishing.gradle"
```
amnonshmsft: We need apply `publishing.gradle` only if we'd like to publish it via Maven

agent/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    classifier = ''
+    relocate 'org.objectweb.asm', 'com.microsoft.applicationinsights.agent.dependencies.asm'
+}
+archivesBaseName = 'applicationinsights-agent'
+
+jar {
+    dependsOn shadowJar
+    enabled = false
+    manifest {
+        attributes("Agent-Class": "com.microsoft.applicationinsights.agent.internal.agent.AgentImplementation",
```
amnonshmsft: Please align mappings

agent/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+dependencies {
+    compile group: 'org.ow2.asm', name: 'asm', version: '5.0.3'
+    compile group: 'org.ow2.asm', name: 'asm-commons', version: '5.0.3'
+    testCompile group: 'commons-io', name: 'commons-io', version: '2.4'
+    testCompile group: 'junit', name: 'junit', version: '4.11'
+    testCompile group: 'org.mockito', name: 'mockito-all', version: '1.8.0'
+}
+
+projectPomName = project.msftAppInsights + " Java SDK Agent Module"
```
amnonshmsft: Only relevant if we'd like to publish it via Maven

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+import org.objectweb.asm.commons.AdviceAdapter;
+
+/**
+ * A base class that lets its derived classes to enjoy utilities
+ * like creating Temp variables managing end of methods method names etc.
```
amnonshmsft: Temp --> temporary.
Also, I'm not sure that I understand what you mean by _"managing end of methods method names"_

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * Created by gupele on 5/11/2015.
+ */
+abstract class AdvancedAdapter extends AdviceAdapter {
```
amnonshmsft: What do you mean by the word advanced?
Can you pick a more meaningful name here?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+
+    protected final boolean reportExecutionTime;
+
+    protected enum ExitStatus {
+        EXIT_WITH_EXCEPTION,
+        EXIT_WITH_RETURN_VALUE,
+        EXIT_VOID,
+        UNKNOWN
```
amnonshmsft: For naming consistency please rename UNKNOWN to EXIT_UNKNOWN

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassNamesProvider.java
```diff
@@ -0,0 +1,215 @@
...
+     * @return The {@link ClassInstrumentationData}
+     */
+    @Override
+    public ClassInstrumentationData getAndRemove(String className) {
+        return classesToInstrument.remove(className);
+    }
+
+    private void addHibernate() {
+        HashSet<String> methodNamesOnly = new HashSet<String>();
+        methodNamesOnly.add("delete");
```
amnonshmsft: I think it would be more elegant here and below defining a string array with the method names, forbidden paths, etc. and then call the relevant methods on the relevant string arrays in a loop.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassNamesProvider.java
```diff
@@ -0,0 +1,215 @@
...
+
+        sqlClasses.add("org/postgresql/core/BaseStatement");
+        sqlClasses.add("org/postgresql/jdbc2/AbstractJdbc2Statement");
+        sqlClasses.add("org/postgresql/jdbc3g/AbstractJdbc3gStatement");
+        sqlClasses.add("org/postgresql/jdbc4/AbstractJdbc4Statement");
+        sqlClasses.add("org/postgresql/jdbc4/Jdbc4Statement");
+        sqlClasses.add("org/postgresql/jdbc4/Jdbc4PreparedStatement");
+        sqlClasses.add("org/postgresql/jdbc4/Jdbc4CallableStatement");
+
+        HashSet<String> methodNamesOnly = new HashSet<String>();
```
amnonshmsft: I think that you can skip the word `Only` in `methodNamesOnly`

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassNamesProvider.java
```diff
@@ -0,0 +1,215 @@
...
+        methodNamesOnly.add("delete");
+        methodNamesOnly.add("get");
+        methodNamesOnly.add("insert");
+        methodNamesOnly.add("list");
+        methodNamesOnly.add("update");
+
+        addToClasses("org/hibernate/impl/StatelessSessionImpl", InstrumentedClassType.SQL, methodNamesOnly);
+    }
+
+    private void populateForbiddenClasses() {
```
amnonshmsft: What are forbidden classes?
Please add documentation

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassNamesProvider.java
```diff
@@ -0,0 +1,215 @@
...
+        forbiddenPaths.add("org/apache");
+        forbiddenPaths.add("com/microsoft/applicationinsights");
+        forbiddenPaths.add("sun/nio/");
+        forbiddenPaths.add("sun/rmi/");
+        forbiddenPaths.add("com/sun/jmx/");
+        forbiddenPaths.add("sun/net/www/http/KeepAlive");
+        forbiddenPaths.add("com.google");
+    }
+
+    private void populateSqlClasses() {
```
amnonshmsft: For better modularity I think it would be better to have SQL and Http in separate classes

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public static void premain(String args, Instrumentation inst) {
+
+        try {
+            agentJarLocation = getAgentJarLocation();
+            appendJarsToBootstrapClassLoader(inst);
+            loadJarsToBootstrapClassLoader(inst);
+        } catch (Throwable throwable) {
+            InternalAgentLogger.INSTANCE.error("Agent is NOT activated: failed to load to bootstrap class loader: " + throwable.getMessage());
+            System.exit(-1);
```
amnonshmsft: Should we exit the program in case the agent failed to load?

agent/build.gradle
amnonshmsft: As we decided that the agent will be published in as a beta version we should explicitly set its version (override the common setting)

agent/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+plugins {
+    id 'com.github.johnrengelman.shadow' version '1.2.1'
+}
+
+apply from: "$buildScriptsDir/common-java.gradle"
+apply from: "$buildScriptsDir/publishing.gradle"
+
+shadowJar {
+    classifier = ''
+    relocate 'org.objectweb.asm', 'com.microsoft.applicationinsights.agent.dependencies.asm'
```
amnonshmsft: Per my understanding the agent is loaded in the bootstrap context which does not contain this library and also only a single agent can be loaded, so my question is whether there's a real risk for name clash that requires this relocate

agent/build.gradle
amnonshmsft: Wouldn't "Premain-Class" suffice here?
I.e. do we need both "Agent-Class" and "Premain-Class"?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+    protected final boolean reportExecutionTime;
+
+    protected enum ExitStatus {
+        EXIT_WITH_EXCEPTION,
+        EXIT_WITH_RETURN_VALUE,
+        EXIT_VOID,
+        EXIT_UNKNOWN
+    }
+
+    protected static class TempVar {
```
amnonshmsft: Please place internal classes before/after the class member variables (but not in the middle of it)

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+            return;
+        }
+
+        if (opcode != Opcodes.ATHROW) {
+            byteCodeForMethodExit(opcode);
+        }
+    }
+
+    @Override
+    public void visitMethodInsn(int opcode,String owner,String name,String desc, boolean isMethodOwnerAnInterface) {
```
amnonshmsft: Q - if you're just delegating to your super class do you really need to override this method?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+        super.visitMethodInsn(opcode, owner, name, desc, isMethodOwnerAnInterface);
+    }
+
+    protected AdvancedAdviceAdapter(boolean reportExecutionTime, int api, MethodVisitor methodVisitor, int access, String owner, String methodName, String desc) {
```
amnonshmsft: To many parameters/too long line - please place parameters in separate lines.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
```diff
@@ -0,0 +1,169 @@
...
+
+    protected void duplicateTop(Type type) {
+        if (type.getSize() == 2) {
+            dup2();
+        } else {
+            dup();
+        }
+    }
+
+    protected void activateEnumMethod(Class<?> clazz, String methodName, String methodSignature, Object... args) {
```
amnonshmsft: It would be good if you can add some documentation about what we're doing here

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultMethodInstrumentor.java
```diff
@@ -0,0 +1,135 @@
...
+ * substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.agent;
```
amnonshmsft: Do we really need the `agent` part twice in the package name (i.e. `agent.internal.agent`)?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultMethodInstrumentorsFactory.java
```diff
@@ -0,0 +1,45 @@
...
+ * Created by gupele on 5/20/2015.
+ */
+final class DefaultMethodInstrumentorsFactory implements MethodInstrumentorsFactory {
+    private final ClassDataProvider provider;
+
+    public DefaultMethodInstrumentorsFactory(ClassDataProvider provider) {
+        this.provider = provider;
+    }
+
+    public DefaultMethodInstrumentor getMethodVisitor(MethodInstrumentationDecision decision, int access, String desc, String className, String methodName, MethodVisitor methodVisitor) {
```
amnonshmsft: Too long line/too many parameters - please place function parameters in separate lines.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpMethodInstrumentor.java
```diff
@@ -0,0 +1,70 @@
...
+package com.microsoft.applicationinsights.agent.internal.agent;
+
+import java.net.URL;
+
+import com.microsoft.applicationinsights.agent.internal.coresync.impl.ImplementationsCoordinator;
+
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Type;
+
+/**
```
amnonshmsft: What's the purpose of this class?
Please add documentation

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecision.java
```diff
@@ -0,0 +1,40 @@
...
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.agent;
+
+/**
+ * The class holds the type of actions that should be done on an instrumented method
+ * The class is the 'decision' for the method after taking into consideration init/configuration data
```
amnonshmsft: I'm not sure that I'm clear with the word "decision" here.
Why not simply using "configuration"?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.agent;
+
+/**
+ * Created by gupele on 6/2/2015.
+ */
+final class MethodInstrumentationDecisionBuilder {
```
amnonshmsft: Again, not sure that the I'd use the word decision here

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+    public boolean reportCaughtExceptions;
+    public boolean reportExecutionTime;
+
+    public MethodInstrumentationDecisionBuilder withReportCaughtExceptions(boolean reportCaughtExceptions) {
```
amnonshmsft: Do we really need `withReportCaughtExceptions` and `withReportExecutionTime`?
Why not just get these parameters in the constructor or in `create()`?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationInfo.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * The class holds the data for methods.
+ * The class should collect data for methods of one class.
+ *
+ * For each method the class holds a {@link com.microsoft.applicationinsights.agent.internal.agent.MethodInstrumentationDecision}
+ * that will be used later on if the method of the class is found in runtime.
+ *
+ * The class knows how to 'merge' the information of methods so methods without signatures
+ * mean that all methods with that name qualify, to mark all the methods, set priority between methods
+ *
```
amnonshmsft: > qualify, to

-->

> qualify. To

?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationInfo.java
```diff
@@ -0,0 +1,137 @@
...
+ * Created by gupele on 5/29/2015.
+ */
+public final class MethodInstrumentationInfo {
+    private final static String ANY_SIGNATURE_MARKER = "";
+    private MethodInstrumentationDecision allClassMethods = null;
+
+    private static class MethodInfo {
+
+        public HashMap<String, MethodInstrumentationDecision> descriptions = new HashMap<String, MethodInstrumentationDecision>();
```
amnonshmsft: Why descriptions?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationRequest.java
```diff
@@ -0,0 +1,55 @@
...
+            throw new IllegalArgumentException("methodName must be non-null non-empty value.");
+        }
+
+        this.reportCaughtExceptions = reportCaughtExceptions;
+        this.reportExecutionTime = reportExecutionTime;
+        this.methodName = methodName;
+        this.methodSignature = methodSignature;
+    }
+
+    public MethodInstrumentationRequest(String methodName, boolean reportCaughtExceptions, boolean reportExecutionTime) {
```
amnonshmsft: In this constructor overload you're just calling the other constructor with `methodName=null` which means that an exception will get thrown immediately, so I'm not sure why we need this specific overload...

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationRequestBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.agent;
+
+/**
+ * Created by gupele on 6/2/2015.
+ */
+final class MethodInstrumentationRequestBuilder {
```
amnonshmsft: In all uses of this class you're calling all methods, i.e. setting all the fields.
So why not simply passing all the values in the constructor or to `create()`?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/StringUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ * substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.agent;
```
amnonshmsft: I'm not sure that this class should live in this agent and not in a more general one (e.g. common/tools/utils?)

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/SqlStatementMethodInstrumentor.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    protected void onMethodEnter() {
+
+        super.visitFieldInsn(GETSTATIC, implementationCoordinatorInternalName, "INSTANCE", implementationCoordinatorJavaName);
+
+        mv.visitLdcInsn(getMethodName());
+        mv.visitVarInsn(ALOAD, 0);
+        if (numberOfArgs > 1) {
+            mv.visitVarInsn(ALOAD, 1);
+            mv.visitMethodInsn(INVOKEVIRTUAL, implementationCoordinatorInternalName, ON_ENTER_METHOD_NANE, ON_ENTER_METHOD_SIGNATURE, false);
```
amnonshmsft: Your making the following call both for in case the condition is satisfied or not:

    mv.visitMethodInsn(INVOKEVIRTUAL, implementationCoordinatorInternalName, ON_ENTER_METHOD_NANE, ON_ENTER_METHOD_SIGNATURE, false);

So why not taking it out of the if and else parts?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentBuiltInConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.config;
+
+/**
+ * Created by gupele on 6/5/2015.
+ */
+public class AgentBuiltInConfiguration {
+    private final boolean enabled;
```
amnonshmsft: In some similar data classes you're defining the class fields as public and do not have getters (e.g. ClassInstrumentationData, MethodInstrumentationDecision, MethodInstrumentationRequest) and in some you're defining the fields as private and do not define getters (e.g. this class).
While it doesn't make much difference I think it would be better to be consistent in all classes (I'm personally in favor of having getters).

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentBuiltInConfigurationBuilder.java
```diff
@@ -0,0 +1,56 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.config;
+
+/**
+ * Created by gupele on 6/5/2015.
+ */
+public class AgentBuiltInConfigurationBuilder {
```
amnonshmsft: See my previous comments about using this pattern.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfigurationBuilderFactory.java
```diff
@@ -0,0 +1,52 @@
...
+            }
+        } catch (Throwable t) {
+            InternalAgentLogger.INSTANCE.error("Failed to create builder: '%s'" + t.getMessage());
+        }
+
+        return null;
+    }
+
+    public AgentConfigurationBuilder createDefaultBuilder() {
+        return new XmlAgentConfigurationBuilder();
```
amnonshmsft: Can you send an example how such xml looks like?

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilderTest.java
```diff
@@ -0,0 +1,215 @@
...
+            writer.println("<HTTP enabled=\"true\"/>");
+
+            writer.println("</BuiltIn>");
+        }
+    }
+
+    private static class ForbiddenSectionTestPrinter implements Printer {
+        @Override
+        public void print(PrintWriter writer) {
+            writer.println("<ForbiddenPrefixes>");
```
amnonshmsft: Please take this XML content out into an xml file and load it as string from there (instead of building its string like this).

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilderTest.java
```diff
@@ -0,0 +1,215 @@
...
+            writer.println("<Prefix>a.b.AClass1</Prefix>");
+
+            writer.println("</ForbiddenPrefixes>");
+        }
+    }
+
+    private static class ConfigurationTestPrinter implements Printer {
+        @Override
+        public void print(PrintWriter writer) {
+            writer.println("<Class name=\"a.AClass1\">");
```
amnonshmsft: As above

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.microsoft.applicationinsights.agent.internal.agent.ClassInstrumentationData;
+
+/**
+ * Defines the interface for concrete classes that represent configuration data for the Agent
+ * Created by gupele on 5/17/2015.
+ */
+public interface AgentConfiguration {
+
+    AgentBuiltInConfiguration getBuiltInSwitches();
```
amnonshmsft: You're returning a configuration so why not sticking with this terminology?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfigurationDefaultImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public Map<String, ClassInstrumentationData> getRequestedClassesToInstrument() {
+        return classesToInstrument;
+    }
+
+    @Override
+    public AgentBuiltInConfiguration getBuiltInSwitches() {
+        return builtInConfiguration;
```
amnonshmsft: See my comment above about switches vs. configuration

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassDataProvider.java
```diff
@@ -0,0 +1,240 @@
...
+ */
+class DefaultClassDataProvider implements ClassDataProvider {
+    private final static String HIBERNATE_SESSION_IMPL_CLASS_NAME = "org/hibernate/impl/SessionImpl";
+    private final static String HIBERNATE_STATELESS_SESSION_IMPL_CLASS_NAME = "org/hibernate/impl/StatelessSessionImpl";
+
+    private final static String HTTP_CLASS_NAME = "sun/net/www/protocol/http/HttpURLConnection$HttpInputStream";
+    private final static String HTTP_METHOD_NAME = "read";
+    private final static String HTTP_METHOD_SIGNATURE = "([BII)I";
+
+    private final static String[] FORBIDDEN_CLASS_PREFIXES = new String[] {
```
amnonshmsft: What do you mean by forbidden?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+        public final String key;
+
+        public RegistrationData(ClassLoader classLoader, AgentNotificationsHandler handler, String key) {
+            this.classLoader = classLoader;
+            this.handler = handler;
+            this.key = key;
+        }
+    }
+
+    private static ConcurrentHashMap<String, RegistrationData> implementations = new ConcurrentHashMap<String, RegistrationData>();
```
amnonshmsft: Please locate all private fields either before the methods or after these, but not in the middle

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+        } catch (Throwable t) {
+        }
+    }
+
+    @Override
+    public void onMethodEnterSqlStatement(String name, Statement statement, String sqlStatement) {
+        try {
+            AgentNotificationsHandler implementation = getImplementation();
+            if (implementation != null) {
+                if (statement != null && sqlStatement != null && sqlStatement.length() == 0) {
```
amnonshmsft: Wouldn't you want to set `sqlStatement` also in case that `sqlStatement == null`?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+            AgentNotificationsHandler implementation = getImplementation();
+            if (implementation != null) {
+                implementation.onMethodFinish(name);
+            }
+        } catch (Throwable t) {
+        }
+    }
+
+    @Override
+    public String getName() {
```
amnonshmsft: Why null?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+            implementations.put(implementationName, new RegistrationData(classLoader, handler, implementationName));
+
+            return implementationName;
+        } catch (Throwable throwable) {
+            InternalAgentLogger.INSTANCE.error("Exception: '%s'", throwable.getMessage());
+            return null;
+        }
+    }
+
+    public Collection<RegistrationData> getRegistered() {
```
amnonshmsft: This method is not used

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLogger.java
```diff
@@ -0,0 +1,173 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.logger;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+import com.microsoft.applicationinsights.agent.internal.agent.StringUtils;
```
amnonshmsft: Isn't it kind of code duplication of the SDK internal logger?
If so why not making it a generic class and have the SDK use this logger as well?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+        public String[] arguments;
+        public long interval;
+        public InstrumentedClassType type;
+        public Object result;
+    }
+
+    private static class ThreadData {
+        public final LinkedList<MethodData> methods = new LinkedList<MethodData>();
+
+        public void done() {
```
amnonshmsft: It doesn't make much sense to me calling `threadData.done`.
Why not calling this method `reset` or `clean`?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+            methods.clear();
+        }
+    }
+
+    static final class ThreadDataThreadLocal extends ThreadLocal<ThreadData> {
+        private ThreadData threadData;
```
amnonshmsft: This name is a kind of cumbersome. Why not simply calling it `ThreadLocalData`?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+    private final String name;
+
+    public CoreAgentNotificationsHandler(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void onException(String className, String methodName, Throwable throwable) {
+        try {
+            if (throwable instanceof Exception) {
```
amnonshmsft: If throwable is not an instance of Exception, can we also track it somehow?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+                telemetryClient.trackException((Exception)throwable);
+            }
+        } catch (Throwable t) {
+        }
+    }
+
+    @Override
+    public void onMethodEnterURL(String name, URL url) {
+        String urlAsString = url == null ? null : url.toString();
```
amnonshmsft: To improve code redability please place the `url == null` in parenthesis.
Alternatively you can change the code to:
    
    String urlAsString = null;
    if (url != null) {
        urlAsString = url.toString();
    }

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/AgentNotificationsHandler.java
```diff
@@ -0,0 +1,48 @@
...
+public interface AgentNotificationsHandler {
+    String getName();
+
+    void onException(String className, String methodName, Throwable throwable);
+
+    void onMethodEnterURL(String name, URL url);
+
+    void onMethodEnterSqlStatement(String name, Statement statement, String sqlStatement);
+
+    void onDefaultMethodEnter(String name);
```
amnonshmsft: What does **default** method enter means?
Please add documentation

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLoggerTest.java
```diff
@@ -0,0 +1,78 @@
...
+ * substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.logger;
```
amnonshmsft: How is it turned on/off?

agent/build.gradle
gupele: Yes, theoretically asm can be used by others, there is no need to take the risk

agent/build.gradle
gupele: Thanks

agent/build.gradle
gupele: We need to decide how to publish, that is a temporary solution that lets us build and test the agent with the sdk jars

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
gupele: Sure, typo

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
gupele: Thanks, left overs from tests

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
gupele: I think so, otherwise the user might get the impression that things are ok and will be surprised that the events are not in the portal, I think the 'fail-fast' approach is better

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassDataProvider.java
gupele: Classes that should not be instrumented

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultMethodInstrumentor.java
gupele: I think it is ok

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecision.java
gupele: The class carries the information for instrumentation. The name decision is to imply that the data is the final decision of what to instrument after taking into consideration other factors as configuration, forbidden lists etc.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
gupele: Using the builder pattern

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationInfo.java
gupele: Will change

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationRequest.java
gupele: It will throw only if 'methodName' is null, which it should do, but the overload is for methods that don't care about the signature

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationRequestBuilder.java
gupele: Using the builder pattern

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassDataProvider.java
amnonshmsft: Then I think that the work "Excluded" would be more natural here

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
amnonshmsft: But what's the justification for using this pattern here?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/StringUtils.java
gupele: I agree, will be moved

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationRequestBuilder.java
amnonshmsft: But what's the justification for using this pattern here?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfiguration.java
gupele: Will change to configuration

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfigurationBuilderFactory.java
gupele: There is one in the onenote

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/AgentNotificationsHandler.java
gupele: Changed

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
gupele: Changed

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
gupele: Added documentation

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
gupele: Changed to notificationHandlersData

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
gupele: Thanks

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLogger.java
gupele: I thought about that, but this means that we are going to have another jar for that, since the core should not rely on having the Agent jar in runtime, so meanwhile we have a duplication, almost, which I think will do for now

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLoggerTest.java
gupele: It is 'off' by default and can b turned on in configuration

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: Removed

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: Done

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: Throwable is the base class for all exceptions and erros, I will change the method name

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassDataProvider.java
gupele: Changed

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
gupele: Readability & read only objects

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentBuiltInConfiguration.java
gupele: Changed to getters

agent/build.gradle
amnonshmsft: I'm not sure that I get it

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultMethodInstrumentor.java
amnonshmsft: I'm not saying it is not ok but just that it is redundant

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecision.java
amnonshmsft: Still, I'd think that configuration reflects better the content of this class, especially considering the fact that this class just holds members and does not make any decision.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/MethodInstrumentationDecisionBuilder.java
amnonshmsft: Both reasons does not call for using this pattern:
* **Readability** - you have many other places in code where you're passing a class 4+ parameters
* **Readonly** - you don't need this pattern for that - just set these fields in the constructor

Am I missing something?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/AgentConfigurationBuilderFactory.java
amnonshmsft: Our customers/external contributors do not have an access to our OneNote - it would helpful adding it somewhere, at least under the tests

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLoggerTest.java
amnonshmsft: In which configuration? Where will this configuration be defined?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+        startMethod(InstrumentedClassType.HTTP, name, urlAsString);
+    }
+
+    @Override
+    public void onMethodEnterSqlStatement(String name, Statement statement, String sqlStatement) {
```
amnonshmsft: This handler handle several types of data.
I think it would be more elegant delegating the handling of specific data types to specific handlers.
WDYT?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+        telemetry.setDependencyKind(DependencyKind.Undefined);
+
+        InternalLogger.INSTANCE.trace("Sending RDD event for '%s'", methodData.name);
+
+        telemetryClient.track(telemetry);
+    }
+
+    private void sendHTTPTelemetry(MethodData methodData, boolean isException) {
+        if (methodData.arguments != null && methodData.arguments.length == 1) {
+            String url = methodData.arguments[0];
```
amnonshmsft: Is it possible that `methodData.arguments.length > 1`?
If so, do you still want to skip the telemetry?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -0,0 +1,234 @@
...
+            Duration duration = new Duration(nanoToMilliseconds(methodData.interval));
+
+            InternalLogger.INSTANCE.trace("Sending HTTP RDD event, URL: '%s'", url);
```
amnonshmsft: 2 Qs:
1. Why not using `trackClient.trackDependency`?
2. I also want to make sure that there won't be double logging, i.e. here and by `telemetryClient.Track` (or `trackClient.trackDependency`)

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * Note, that the method will take care of situations where the filter is already declared in the 'web.xml'
+     *
+     * @param sce The Servlet Context Event that we use to extract the data we need.
+     */
+    @Override
+    public void contextInitialized(ServletContextEvent sce) {
+        try {
+            ServletContext context = sce.getServletContext();
+
```
amnonshmsft: Why adding empty line after each statement?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Created by gupele on 5/7/2015.
+ */
+//@WebListener
+public final class WebAppInitializer implements ServletContextListener {
+
+    private final static String FILTER_NAME = "ApplicationInsightsWebFilter";
+    private final static String WEB_INF_FOLDER = "WEB-INF/";
+
+    private String name;
```
amnonshmsft: `name` is defined as a member but is not used this way as you're passing it as a parameter to the other functions (TMHO the latter is better)

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override
+    public void contextInitialized(ServletContextEvent sce) {
+        try {
+            ServletContext context = sce.getServletContext();
+
+            name = getName(context);
+
+            String key = registerWebApp(name);
+
+            addFilter(context);
```
amnonshmsft: I think it would make the code more readable if the call to `addFilter` would appear just after getting the context as this way all the other commands related to the registration would appear , i.e.:

    ServletContext context = sce.getServletContext();
    addFilter(context);
    name = getName(context);
    String key = registerWebApp(name);
    WebRequestTrackingFilter.setName(key);

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
```diff
@@ -0,0 +1,165 @@
...
+
+        InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Did not find AI-Filter");
+        return null;
+    }
+
+    private String getName(ServletContext context) {
+        String name = null;
+        try {
+            String contextPath = context.getContextPath();
+            if (Strings.isNullOrEmpty(contextPath)) {
```
amnonshmsft: I'm a bit surprised that there's no more elegant name to extract the app name from this context :(

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterImpl.java
```diff
@@ -0,0 +1,204 @@
...
+
+        setKeyOnTLS();
+
+        if (isInitialized) {
+            isRequestProcessedSuccessfully = invokeSafeOnBeginRequest(req, res);
+        }
+
+        try {
+            chain.doFilter(req, res);
+        } catch (ServletException se) {
```
amnonshmsft: Instead of having 3 `catch`es here with the same code wouldn't it be nicer to have a single catch that will do the filtering, i.e.:

    catch (Exception e) {
        if (e instanceof ServletException || e instanceof IOException || e instanceof RuntimeException) {
            onException(e);
        }
        throw e;
    }

gupele: This will not compile. The signature of the method throws IOException or ServletExcedption, not Exception

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
gupele: I know. I thought that we should start with this, later on we can add support for explicit naming through configuration,

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebAppInitializer.java
gupele: I think that current code reflects better the algorithm: get the name, register, create the filter with the name found/or not

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: I don't think it matters, no way of double logging

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: No argument means there is no URL data, so I think there is no much help for the user that there was an HTTP call but we don't know what the URL was, but this is a very small thing to change, I will consult Harel on that issue

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
gupele: I thought about that, but I decided to leave things as is, in the future I might change that if we support more types that have their own way of fetching data

PR #210, opened 2015-06-04
Core - Adding support for HttpClient older versions

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.applicationinsights.internal.channel.common;
```
amnonshmsft: License license license...

amnonshmsft:
Please configure your IDE to automatically add the license statement to new files

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSenderFactory.java
```diff
@@ -0,0 +1,22 @@
...
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.reflect.ClassDataUtils;
+
+/**
+ * Created by gupele on 6/4/2015.
+ */
+final class ApacheSenderFactory {
+    public ApacheSender create() {
+        if (!ClassDataUtils.INSTANCE.verifyClassExists("org.apache.http.conn.HttpClientConnectionManager")) {
+            String errorMessage = String.format("Found an old version of HttpClient jar, for best performance consider upgrading to version 4.3+");
```
amnonshmsft: Why using String.format if you're not formatting anything?
Actually you can just pass the message (`"Found an old..."`) directly to `InternalLogger.INSTANCE.warn()`

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -75,18 +77,14 @@
     private volatile boolean stopped;
 
     // Use one instance for optimization
-    private final CloseableHttpClient httpClient;
+    private final ApacheSender httpClient;
+//    private final CloseableHttpClient httpClient;
```
amnonshmsft: Please remove commented out code

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -120,12 +114,9 @@ public synchronized void stop(long timeout, TimeUnit timeUnit) {
...
             return;
         }
-        try {
-            s_senderThreadsManager.stopAllSendersBackOffActivities();
-            httpClient.close();
-        } catch (IOException e) {
-            InternalLogger.INSTANCE.error("Failed to close http client, exception: %s", e.getMessage());
-        }
+
+        s_senderThreadsManager.stopAllSendersBackOffActivities();
```
amnonshmsft: The removed `try/catch` lines were indeed redundant?

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43Test.java
```diff
@@ -0,0 +1,16 @@
...
+import org.apache.http.client.HttpClient;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public final class ApacheSender43Test {
+    @Test
+    public void testHttpClientType() {
+        HttpClient tested = new ApacheSender42().getHttpClient();
```
amnonshmsft: According to the class name you should create an `ApacheSender43` here and not 42

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42Test.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.applicationinsights.internal.channel.common;
```
amnonshmsft: Actually I think you can merge this file and the next file (`ApacheSender43Test.java`) into a single file named `ApacheSenderTest.java` and just have different test names instead.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSenderFactoryTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.mockito.Mockito;
+
+import java.lang.reflect.Field;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.anyString;
+
+public final class ApacheSenderFactoryTest {
+    @Test
+    public void testBadJar() throws NoSuchFieldException, IllegalAccessException {
```
amnonshmsft: I'm not sure that I'd name these tests Bad and Good jars as we support both versions so both should be considered as good...
Instead I would rename these tests to `testV42Lib` or `testNewJar` (less preferred for the case one day we'll decide to support even older/newer version of this library.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutputTest.java
```diff
@@ -31,15 +31,4 @@
 import static org.mockito.Matchers.anyString;
```
amnonshmsft: If this class does not have any test then why keeping this file?

PR #214, opened 2015-06-10
Core - Internal Logger bug fix and added tests

gupele:


amnonshmsft:
Do you also want to move the logger out of the internal path?
(It would be better to do it in a separate commit)

PR #215, opened 2015-06-10
Core - Changing error severity levels for configuration search and ch…

gupele:
…anged succesful message

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
@@ -59,9 +59,9 @@ public final class ConfigurationFileLocator {
         }
 
         if (configurationFile != null) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
```diff
@@ -147,8 +147,6 @@ private String getConfigurationAbsolutePath(String path) {
         File configFile = new File(path, configurationFileName);
 
         if (configFile.exists()) {
-            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Found configuration file in: '%s'", path);
```
amnonshmsft: I think we should leave this line so in case the user got confused and placed ApplicationInsights.xml in several locations he will be able to tell easily what file was used.

gupele: main method, 'getConfigurationFile' prints on exit, the status & the full path of the file if found, which should be enough

PR #224, opened 2015-06-22
Test/integration test refactoring

Yonisha:
Sorry for the big PR - you weren't here.


web/src/integTest/resources/testSettings.properties
```diff
@@ -1,3 +0,0 @@
-maxWaitTime=180
```
amnonshmsft: These settings should be external to the test, thus should be placed in a properties file.
Also, I might want the build to generate this file, so please revert this change and use this file instead of the constants you defined.

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryQueue.java
```diff
@@ -0,0 +1,81 @@
...
+import java.security.InvalidKeyException;
+import java.util.ArrayList;
+
+/**
+ * Created by yonisha on 6/16/2015.
+ */
+public class ApplicationTelemetryQueue {
+
+    private static final int QUEUE_VISIBILITY_TIMEOUT_IN_SECONDS = 180;
+    private static final int QUEUE_MESSAGE_BATCH_SIZE = 32;
```
amnonshmsft: Please take this parameter to the properties file (see my above comment)

test/webapps/bookstore-spring/src/main/java/com/springapp/mvc/extensions/WebRequestRunIdTelemetryInitializer.java
```diff
@@ -0,0 +1,48 @@
+/*
```
amnonshmsft: I wonder if this is the right location for this telemetry initializer:
1. This looks like a generic extension that can be used by any application and not just by this app so it can be placed under a common area
2. It is more like a test code so it can be placed under the test framework code

test/webapps/bookstore-spring/src/main/java/com/springapp/mvc/extensions/WebRequestRunIdTelemetryModule.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (runId == null) {
+            return;
+        }
+
+        RequestTelemetry httpRequestTelemetry = ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry();
+        httpRequestTelemetry.getProperties().put(RUN_ID_QUERY_PARAM_NAME, runId);
+    }
+
+    private String getRunIdFromQueryString(String queryString) {
```
amnonshmsft: I wonder whether there is standard code for parsing the query string (rather than doing it manually)

test/webapps/bookstore-spring/src/main/resources/ApplicationInsights.xml
```diff
@@ -7,13 +7,22 @@
...
+
+		<!-- This initializer is used by the integration tests to verify correlation between requests & instrumented code. -->
+		<Add type="com.springapp.mvc.extensions.WebRequestRunIdTelemetryInitializer" />
 	</TelemetryInitializers>
 	<ContextInitializers>
 		<Add type="com.springapp.mvc.MyContextInitializer" />
 	</ContextInitializers>
 	<SDKLogger></SDKLogger>
+	<Channel>
+		    <DeveloperMode>true</DeveloperMode>
```
amnonshmsft: Tabs to spaces

web/build.gradle
```diff
@@ -52,6 +52,8 @@ whenPomConfigured = { p ->
         }
     }
     writePomToArtifactsDirectory(p, project.name)
+
+    System.get
```
amnonshmsft: ??

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationBlobClient.java
```diff
@@ -0,0 +1,77 @@
...
+ * substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.framework;
```
amnonshmsft: This is test framework, not the SDK framework

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryManager.java
```diff
@@ -0,0 +1,154 @@
...
+    public ApplicationTelemetryManager(String storageAccountConnectionString, String queueName) throws StorageException, InvalidKeyException, URISyntaxException {
+        this.applicationTelemetryQueue = new ApplicationTelemetryQueue(storageAccountConnectionString, queueName);
+        this.applicationBlobClient = new ApplicationBlobClient(storageAccountConnectionString);
+    }
+
+    public List<TelemetryItem> getApplicationTelemetries(String runId, int numberOfExpectedTelemetries) throws Exception {
+
+        List<TelemetryItem> telemetriesForRunId = getTelemetriesFromCacheWithRunId(runId);
+
+        if (telemetriesForRunId.size() < numberOfExpectedTelemetries) {
```
amnonshmsft: It would be good also the check the other case `telemetriesForRunId.size() > numberOfExpectedTelemetries` for the case that for some reason a telemetry is being sent twice (we had such bug in the past)

web/src/integTest/java/com/microsoft/applicationinsights/test/utils/PropertiesUtils.java
```diff
@@ -1,44 +0,0 @@
-package com.microsoft.applicationinsights.test.utils;
```
amnonshmsft: It would be better to keep this class so it can be used both for the environment property file and for the test specific property file

web/src/integTest/java/com/microsoft/applicationinsights/framework/TestEnvironment.java
```diff
@@ -47,4 +45,33 @@ public String getApplicationStorageConnectionString() {
     public String getApplicationStorageExportQueueName() {
         return envProps.getProperty(KEY_APPLICATION_STORAGE_EXPORT_QUEUE_NAME);
     }
+
+    public static Properties loadPropertiesFromFile(String filePath) throws IOException
```
amnonshmsft: This logic should reside in the PropertyUtils class that you removed so it can be reused for all property files used by the tests

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryQueue.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private CloudQueue queue;
+
+    public ApplicationTelemetryQueue(String connectionString, String queueName) throws URISyntaxException, StorageException, InvalidKeyException {
+        CloudStorageAccount account = CloudStorageAccount.parse(connectionString);
+        CloudQueueClient queueClient = account.createCloudQueueClient();
+        queue = queueClient.getQueueReference(queueName);
+
+        // This can be a problem if the queue is used by several tests concurrently.
```
amnonshmsft: So is there a way to prevent this problem?
I'm asking as there are chances that at some point we'd like to run the tests concurrently

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryQueue.java
```diff
@@ -0,0 +1,81 @@
...
+        this.clear();
+    }
+
+    public ArrayList<CloudQueueMessage> retrieveMessages() throws StorageException {
```
amnonshmsft: Maybe rename to `retrieveAllMessages`?

web/src/integTest/java/com/microsoft/applicationinsights/framework/HttpRequestClient.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+
+/**
+ * Created by yonisha on 6/16/2015.
+ */
+public class HttpRequestClient {
```
amnonshmsft: Why not just `HttpClient`?

web/src/integTest/java/com/microsoft/applicationinsights/framework/HttpRequestClient.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // URI structure: http://<server>:<port>/<app_name>/<path>
+    private static final String REQUEST_URI_TEMPLATE = "http://%s:%s/%s/%s";
+    private static final String COOKIE_HEADER_KEY = "Cookie";
+
+    private HttpRequestClient() {
+    }
+
+    public static int sendHttpRequest(URI uri, List<String> requestCookies) throws Exception {
```
amnonshmsft: As this class is named `httpRequestClient` (or simply `httpClient`), you can simply call this method `sendRequest`

web/src/integTest/java/com/microsoft/applicationinsights/web/spring/TestSettings.java
```diff
@@ -1,35 +0,0 @@
-package com.microsoft.applicationinsights.web.spring;
```
amnonshmsft: This file should be kept as well to hold the test properties

web/src/integTest/java/com/microsoft/applicationinsights/web/spring/WebTelemetryTests.java
```diff
@@ -21,263 +21,207 @@
...
-    /**
-     * Sends GET requests to server and expects that will telemetry from app insights and it will include the correct information about the request
-     * @throws Exception
-     */
-    @Test
-    public void sendHttpRequestsTest() throws Exception {
+    // region Tests run IDs
 
-        initTestConfiguration();
+    private static final String TEST_REQUEST_COMMON_PROPERTIES_RUN_ID = "1";
```
amnonshmsft: Please use generated GUIDs rather than static numbers.
Reason - if we'll want to run tests against Tomcat and Jetty (or just different app server versions) in parallel then using a static ID would cause a clash

web/src/integTest/java/com/microsoft/applicationinsights/web/spring/WebTelemetryTests.java
```diff
@@ -21,263 +21,207 @@
...
+                testEnv.getApplicationStorageConnectionString(), testEnv.getApplicationStorageExportQueueName());
 
-        for (String uriPath : uriPathsToRequest) {
-            String requestId = Helpers.getRandomUUIDString();
-            URI uri = Helpers.constructUrl(serverAddress, port, applicationName, uriPath + "&requestId=" + requestId);
-            int responseCode = sendHttpGetRequest(uri);
-            TelemetryItem expectedTelemetry = createExpectedResult(uri, requestId, responseCode);
-            expectedTelemetries.add(expectedTelemetry);
-        }
+        // It takes to telemetries to be exported. Therefore, we're sending requests required for all tests in order
```
amnonshmsft: It takes time?

test/webapps/bookstore-spring/src/main/java/com/springapp/mvc/extensions/WebRequestRunIdTelemetryInitializer.java
Yonisha: The web app is deployed before, and regardless, the integration tests. Therefore you cannot define the initializer as part of the f/w.

test/webapps/bookstore-spring/src/main/java/com/springapp/mvc/extensions/WebRequestRunIdTelemetryModule.java
Yonisha: Not familiar with such code..

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryQueue.java
Yonisha: There is, but it requires some changes in the architecture. Let's do it offline.

web/src/integTest/java/com/microsoft/applicationinsights/framework/ApplicationTelemetryManager.java
Yonisha: verified during the test

test/webapps/bookstore-spring/src/main/java/com/springapp/mvc/extensions/WebRequestRunIdTelemetryInitializer.java
amnonshmsft: Why is that?
Isn't the only dependency is in the applicationinsights.xml configuration file, i.e. there is no compile time dependency?
Even if there is you can make this **test** application dependent on the test framework.

PR #226, opened 2015-06-28
Core - Exception Telemetry will create an Exception if passed null

gupele:


littleaj:

PR #227, opened 2015-07-08
Core - Adding support for build data that is sent as context properties

gupele:


core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/BuildContextInitializer.java
```diff
@@ -0,0 +1,86 @@
...
+    final static String BUILD_DEFAULT_VALUE = "unknown";
+
+    private String gitBranch;
+    private String gitCommit;
+    private String gitUrl;
+    private boolean hasBuildData;
+
+    public BuildContextInitializer() {
+        try {
+            InputStream inputStream = getClass().getClassLoader().getResourceAsStream(BUILD_INFO_FILE_NAME);
```
amnonshmsft: Please use the **[PropertyHelper](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/internal/util/PropertyHelper.java)** class.

Also, in a separate check in please factor out the build info from the PropertyHelper class (see our **[old discussion here](https://github.com/Microsoft/ApplicationInsights-Java/pull/172)**).

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -141,6 +142,7 @@ private void setContextInitializers(ContextInitializersXmlElement contextInitial
         // To keep with prev version. A few will probably be moved to the configuration
         initializerList.add(new SdkVersionContextInitializer());
         initializerList.add(new DeviceInfoContextInitializer());
+        initializerList.add(new BuildContextInitializer());
```
amnonshmsft: This depends on whether we plan to submit this code as part of the SDK or just as an example

amnonshmsft:
Please do not merge before syncing with me

littleaj:

PR #229, opened 2015-07-12
Core - Align with prev versions of http client

gupele:


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -31,6 +31,8 @@
  * Created by gupele on 6/4/2015.
  */
 interface ApacheSender {
+    final static int DEFAULT_REQUEST_TIMEOUT_IN_MILLIS = 60000;
```
amnonshmsft: Is this value configurable?
If not then please remove the word DEFAULT from the constant name

PR #235, opened 2015-07-15
Samples/adding sample context initializer

gupele:


samples/build.gradle
```diff
@@ -21,6 +21,25 @@
 
 apply from: "$buildScriptsDir/common-java.gradle"
 
+import com.microsoft.applicationinsights.build.tasks.PropsFileGen
+
+def sourceOriginFileDir = "$project.buildDir/src/main/resources"
```
amnonshmsft: Please move the definition of `sourceOriginFileDir` inside the task and also there's no need for defining `targetDir` as you actually have it (`sourceOriginFileDir`).
Your task should therefore look like this (also added comments):

    task generateSourceOriginProperties() {
        // Place source properties in a Properties object 
        Properties props = new Properties()
        props.setProperty "git.commit", System.env.'GIT_COMMIT' ?: "unknown"
        props.setProperty "git.branch", System.env.'GIT_BRANCH' ?: "unknown"
        props.setProperty "git.repo", System.env.'GIT_URL' ?: "unknown"

        // Then save these properties into a resource file named source-origin.properties 
        def sourceOriginFileDir = "$project.buildDir/src/main/resources"    
        File targetFile = new File(sourceOriginFileDir, "source-origin.properties")
        sourceOriginFileDir.mkdirs()
        FileOutputStream out = new FileOutputStream(targetFile.getAbsolutePath())
        props.store out, "Git source information"
        out.close()
    }

PR #238, opened 2015-07-21
Core - added i-key sources JVM option, Env variable, xml file

gupele:


amnonshmsft:
Can you please add UT, at least for the environment variable part?
(i.e. set an environment variable and then expect the telemetry client to use it)

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -59,6 +61,9 @@
 
     private String performanceCountersSection = DEFAULT_PERFORMANCE_MODULES_PACKAGE;
 
+    final static String ENV_VARIABLE_I_KEY = "APPLICATION_INSIGHTS_IKEY";
+    final static String JVM_VARIABLE_I_KEY = "-D" + ENV_VARIABLE_I_KEY;
```
amnonshmsft: Are you sure that this would be the name of the variable provided by the JVM?
Per my understanding when java is executed with a flag `-Dproperty=value` then the property name is `property` rather than `-Dproperty`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -108,6 +113,11 @@ public final void initialize(TelemetryConfiguration configuration) {
         }
     }
 
+    private void initializeMinimumCapabilities(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
```
amnonshmsft: Both the channel and the instrumentation key are not capabilities.
Please rename this function to something line `setFundamentalConfiguration`or `setCoreConfiguration`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -79,20 +84,20 @@ public final void initialize(TelemetryConfiguration configuration) {
...
-                configuration.setChannel(new InProcessTelemetryChannel());
+                initializeMinimumCapabilities(null, configuration);
                 return;
             }
 
             ApplicationInsightsXmlConfiguration applicationInsightsConfig = builder.build(configurationFile);
             if (applicationInsightsConfig == null) {
                 InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Failed to read configuration file");
-                configuration.setChannel(new InProcessTelemetryChannel());
+                initializeMinimumCapabilities(applicationInsightsConfig, configuration);
```
amnonshmsft: In this case `applicationInsightsConfig == null` so why not passing to `initializeMinimumCapabilities` null explicitly?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+     * @param userConfiguration The configuration that was represents the user's configuration in ApplicationInsights.xml.
      * @param configuration The configuration class.
      * @return True if succeeded.
      */
-    private boolean setInstrumentationKey(String instrumentationKey, TelemetryConfiguration configuration) {
+    private boolean setInstrumentationKey(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
         try {
-            configuration.setInstrumentationKey(instrumentationKey);
+            // First, try to find the i-key as an Env variable 'APPLICATION_INSIGHTS_IKEY'
+            RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
```
amnonshmsft: Why not simply using 'System.getProperty(ENV_VARIABLE_I_KEY)'?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+     * Next we will try to fetch the i-key from the ApplicationInsights.xml
+     * @param userConfiguration The configuration that was represents the user's configuration in ApplicationInsights.xml.
      * @param configuration The configuration class.
      * @return True if succeeded.
      */
-    private boolean setInstrumentationKey(String instrumentationKey, TelemetryConfiguration configuration) {
+    private boolean setInstrumentationKey(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
         try {
-            configuration.setInstrumentationKey(instrumentationKey);
+            // First, try to find the i-key as an Env variable 'APPLICATION_INSIGHTS_IKEY'
```
amnonshmsft: Wrong comment - first you check whether it was set as a system property

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
     }
 
     /**
-     * Setting an instrumentation key
-     * @param instrumentationKey The instrumentation key found in the configuration.
+     * Setting an instrumentation key:
+     * First we try the JVM option '-DAPPLICATION_INSIGHTS_IKEY=i_key'
```
amnonshmsft: We shouldn't check for a JVM option but for a **[system property](https://docs.oracle.com/javase/tutorial/essential/environment/sysprop.html)**

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+
+                    String ikey = ikeyData[1];
+                    if (!Strings.isNullOrEmpty(ikey)) {
+                        configuration.setInstrumentationKey(argument);
+                        return true;
+                    }
+                }
+            }
+
+            // Second, try to find the i-key as an JVM argument '-DAPPLICATION_INSIGHTS_IKEY=i_key'
```
amnonshmsft: Wrong comment - here you're actually looking for an environment variable

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+     * Setting an instrumentation key:
+     * First we try the JVM option '-DAPPLICATION_INSIGHTS_IKEY=i_key'
+     * Next we will try the environment variable 'APPLICATION_INSIGHTS_IKEY',
+     * Next we will try to fetch the i-key from the ApplicationInsights.xml
+     * @param userConfiguration The configuration that was represents the user's configuration in ApplicationInsights.xml.
      * @param configuration The configuration class.
      * @return True if succeeded.
      */
-    private boolean setInstrumentationKey(String instrumentationKey, TelemetryConfiguration configuration) {
+    private boolean setInstrumentationKey(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
```
amnonshmsft: You're returning a `boolean` value but no one is using it...
Also, returning a value here is redundant as one can simply check whether `configuration.getInstrumentationKey() != null`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,20 +175,55 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+            // Second, try to find the i-key as an JVM argument '-DAPPLICATION_INSIGHTS_IKEY=i_key'
+            String ikey = System.getenv(ENV_VARIABLE_I_KEY);
+            if (!Strings.isNullOrEmpty(ikey)) {
+                configuration.setInstrumentationKey(ikey);
+                return true;
+            }
+
+            // Else, try to find the i-key in the ApplicationInsights.xml
+            if (userConfiguration != null) {
+                configuration.setInstrumentationKey(userConfiguration.getInstrumentationKey());
```
amnonshmsft: Can `userConfiguration.getInstrumentationKey()` return and empty iKey or null?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: no, I started with that since I wanted to use one name so the user is more likely to remember, and easier docs/blogs

But, I don't have any objections to any other name, lets consult with Harel too, and then I will change that name

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok, I will change the name 'setMinimumConfiguration'

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: right

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: ok

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I will go over the comments again

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I will change that line and will also trim the value and will only assign if the configuration is not null and not empty

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: I think it is nicer, and more readable this way

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
gupele: right, thanks

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -59,6 +61,8 @@
 
     private String performanceCountersSection = DEFAULT_PERFORMANCE_MODULES_PACKAGE;
 
+    final static String ENV_VARIABLE_I_KEY = "APPLICATION_INSIGHTS_IKEY";
```
amnonshmsft: Regarding the constant name then this is no longer only the name of the environment variable but also for the system variable.
Therefore consider renaming it to something like `EXTERNAL_PROPERTY_IKEY` or `EXTERNAL_PROPERTY_IKEY_NAME`

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,19 +174,46 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+     * Next we will try to fetch the i-key from the ApplicationInsights.xml
+     * @param userConfiguration The configuration that was represents the user's configuration in ApplicationInsights.xml.
      * @param configuration The configuration class.
-     * @return True if succeeded.
      */
-    private boolean setInstrumentationKey(String instrumentationKey, TelemetryConfiguration configuration) {
+    private void setInstrumentationKey(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
         try {
-            configuration.setInstrumentationKey(instrumentationKey);
+            // First, try to find the i-key as a system property '-DAPPLICATION_INSIGHTS_IKEY=i_key'
```
amnonshmsft: Regarding the comment it still refers to the previous implementation (i.e. `-DAPPLICATION_.INS..`).
Consider changing to 

> First, check whether an i-key was provided as a java system property (i.e. using the -D flag)

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -165,19 +174,46 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+            }
+
+            // Second, try to find the i-key as an environment variable 'APPLICATION_INSIGHTS_IKEY'
+            ikey = System.getenv(ENV_VARIABLE_I_KEY);
+            if (!Strings.isNullOrEmpty(ikey)) {
+                configuration.setInstrumentationKey(ikey);
+                return;
+            }
+
+            // Else, try to find the i-key in the ApplicationInsights.xml
```
amnonshmsft: The second `the` (the one before ApplicationInsights.xml is redundant:

    // Else, try to find the i-key in ApplicationInsights.xml

core/src/test/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactoryTest.java
```diff
@@ -33,19 +33,23 @@
...
 import static org.junit.Assert.assertTrue;
-import static org.mockito.Matchers.anyString;
+import static org.junit.Assert.assertNull;
 
 public final class TelemetryConfigurationFactoryTest {
 
     private final static String MOCK_IKEY = "c9341531-05ac-4d8c-972e-36e97601d5ff";
     private final static String MOCK_ENDPOINT = "MockEndpoint";
     private final static String NON_VALID_URL = "http:sd{@~fsd.s.d.f;fffff";
+    private final static String APP_INSIGHTS_IKEY_TEST_VALYE = "ds";
```
amnonshmsft: Typo in `VALYE`

core/src/test/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactoryTest.java
```diff
@@ -73,6 +77,32 @@ public void initialize(TelemetryConfiguration configuration) {
     }
 
     @Test
+    public void configurationWithNullIkeyTest() {
```
amnonshmsft: Where's the code for trimming?
Also, what about the case that there is a valid iKey preceded/followed by spaces?

Issue #253, opened 2015-08-11
Configuration Files on the Classpath present in a Jar File will not load


smflorentino:
I'm currently writing a plugin to enable AppInsights on my project (it's a Maven project). I created a plugin, also a Maven project, and bundled the ApplicationInsights.xml file inside, as a resource. This works fine in my IDE, since the file is not inside of a jar, but on the filesystem. However, once I package the project and use the resulting artifact as a dependency (a jar of the plugin gets placed on the classpath), AppInsights sees the configuration file but cannot load it from the jar.

I tracked down the source of the bug - if the ApplicationInsights.xml file is present in a jar on the classpath, the SDK still tries to load as a file - but it should be loaded as a resource or input stream, since its not a file on the filesystem per se.

I'm able to reproduce this as a failing unit test in the SDK, as I've pushed to my branch on the PR, where I'm hoping a fix can be discussed/implemented. 

Other people seem to be experiencing the problem too - as I found this StackOverflow question when searching for answers:
http://stackoverflow.com/questions/31789083/issue-unmarshalling-object-from-file-inside-jar/31931056#31931056

PR #254, opened 2015-08-11
Repro config file load failure when config is in jar

smflorentino:
PR for #253 
I'm currently writing a plugin to enable AppInsights on my project (it's a Maven project). I created a plugin, also a Maven project, and bundled the ApplicationInsights.xml file inside, as a resource. This works fine in my IDE, since the file is not inside of a jar, but on the filesystem. However, once I package the project and use the resulting artifact as a dependency (a jar of the plugin gets placed on the classpath), AppInsights sees the configuration file but cannot load it from the jar.

I tracked down the source of the bug - if the ApplicationInsights.xml file is present in a jar on the classpath, the SDK still tries to load as a file - but it should be loaded as a resource or input stream, since its not a file on the filesystem per se.

I added a jar to the classpath that contains a configuration file, as well as a unit test that demonstrates the bug. ConfigurationFileLocator has some code than can be commented in to demonstrate this behavior as well.  

Other people seem to be experiencing the problem too - as I found this StackOverflow question when searching for answers:
http://stackoverflow.com/questions/31789083/issue-unmarshalling-object-from-file-inside-jar/31931056#31931056


littleaj:

Issue #278, opened 2016-02-12
testSDKInfoFileIsWrittenWithInstrumentationKey fails on Fedora Linux 22


olivierdagenais:
I ran `gradlew test` on my Fedora 22 VM and one of the tests in `core` failed:

Class: com.microsoft.applicationinsights.extensibility.initializer.docker.DockerContextInitializerTests
Method: testSDKInfoFileIsWrittenWithInstrumentationKey

```
Argument(s) are different! Wanted:
fileFactory.create(
    "/usr/appinsights/docker/sdk.info",
    "InstrumentationKey=A-test-instrumentation-key"
);
-> at com.microsoft.applicationinsights.extensibility.initializer.docker.DockerContextInitializerTests.testSDKInfoFileIsWrittenWithInstrumentationKey(DockerContextInitializerTests.java:133)
Actual invocation has different arguments:
fileFactory.create(
    "/usr/appinsights/docker/sdk.info",
    "InstrumentationKey=c9341531-05ac-4d8c-972e-36e97601d5ff"
);
```

A quick spelunking in the code suggests that the test might be sensitive to ordering since it [indirectly] accesses the `TelemetryConfiguration` singleton and its configuration wasn't loaded from the `ApplicationInsights.xml` test resource.

Yonisha:
Hi Olivier,

Thanks for your feedback and suggestion.
We'll address this issue shortly.

beckylino:
@olivierdagenais are you still experiencing any issues with this test?

olivierdagenais:
@beckylino I haven't used this project in over a year.  I had encountered this test failure when I was testing a fix for another issue and figured the maintainers might be interested to know some of the tests were unreliable.  I don't currently have a Fedora Linux VM to test with, sorry.

PR #279, opened 2016-02-12
Try to use a per-user temporary directory

olivierdagenais:
# Motivation
Our team uses ApplicationInsights-Java in a few of our products and reports have started coming in from customers trying to use our products under Linux and finding themselves unable to do so when more than one user is involved.  The best, most comprehensive example is in the thread [TFS Command Line Client crashes in Linux](https://social.msdn.microsoft.com/Forums/vstudio/en-US/37561345-25d0-4bd9-8d3d-40531c7c117a/tfs-command-line-client-crashes-in-linux?forum=tee).

The cause seems to be that on GNU/Linux (among others), the temporary directory is **shared** and ApplicationInsights-Java wasn't designed to accommodate that situation.  See the third paragraph of [Separate /tmp directories for each user] (http://www.chiark.greenend.org.uk/~peterb/uxsup/project/tmp-per-user/) for a much more eloquent description of the problem.  On Windows and Mac OS X, it appears [interactive?] users already get their own temporary directory, which is why I suspect it took so long to find this issue.

I haven't dug deeply into the `TransmissionFileSystemOutput` class' use of the `transmissions` temporary folder to see if it's sufficient to have it be per-user (as opposed to per-user, per-process), but this is at least a start.

## Potential alternative implementation
I'm new to this project, so pardon my ignorance.  While investigating the issue, it occurred to me that it's possible `TransmissionFileSystemOutput` may have been written with the expectation that the contents of its `transmissions` temporary folder would stick around between runs of the program emitting telemetry and, should such a program be updated with the next release of ApplicationInsights-Java that contains this fix, any files in the `/tmp/transmissions` folder used by previous releases of this library will be abandoned.

If this is a terrible situation, we could overcome it by having the `TransmissionsFileSystemOutput` be more aware of the temporary folders and have it move the contents of the old one to the new one whenever it starts up.  A lot more work for potentially very few telemetry files, so I'm not sure if it's worth it.

## The "hacky" alternative implementation I don't really want to talk about
Lastly, and I'm afraid to bring this one up, but it's technically the least disruptive option of the bunch: the workaround discovered by our users was to increase the permissions on the temporary folders such that multiple users could run programs that emit telemetry files to the same folder.  In other words, adding something like:

```java
folder.setWritable(true, false);
```
...right after creating the folder.

The thing is, although that might appear to fix the problem, the `TransmissionsFileSystemOutput` may be thread-safe but I'm not sure if it's process-safe, such that two processes (whether from the same user or from separate users) emitting and attempting to transmit telemetry files wouldn't end up in a race condition in cleaning up the folder, with the loser of the race throwing an `IOException` because a file stopped existing in the middle of processing.

# Manual testing
I ran `gradlew test` on the following platforms:
* Oracle Java 1.8.0_71-b15 (64-bit) on Windows 10 (10.0.10586)
* Oracle Java 1.7.0_71-b14 (64-bit) on Mac OS X (Yosemite 10.10.5)
* OpenJDK 1.8.0_71-b15 (64-bit) on Fedora 22
    * This is the only one where an unrelated test is failing. I've opened Issue #278 to track that.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
@@ -47,6 +47,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import com.microsoft.applicationinsights.internal.channel.TransmissionOutput;
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 
+import com.microsoft.applicationinsights.internal.util.FileSystemUtils;
 import com.microsoft.applicationinsights.internal.util.LimitsEnforcer;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
@@ -100,7 +101,7 @@ public final class TransmissionFileSystemOutput implements TransmissionOutput {
 
     public TransmissionFileSystemOutput(String folderPath, String maxTransmissionStorageCapacity) {
         if (folderPath == null) {
-            folderPath = new File(System.getProperty("java.io.tmpdir"), TRANSMISSION_DEFAULT_FOLDER).getPath();
+            folderPath = new File(FileSystemUtils.getTempDir(), TRANSMISSION_DEFAULT_FOLDER).getPath();
         }
 
         capacityEnforcer = LimitsEnforcer.createWithClosestLimitOnError(MIN_CAPACITY_MEGABYTES,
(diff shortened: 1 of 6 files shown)


olivierdagenais:
Ping!

olivierdagenais:
I rebased my changes on top of 66c798e.

olivierdagenais:
The changes from this pull request were eventually incorporated as part of 9e87647d659978182fedde17c9893cdf139b07d4.

Issue #282, opened 2016-06-03
Method setStatus(int sc, String sm) in ApplicationInsightsHttpResponseWrapper must be override


mkosiorek:
In ApplicationInsightsHttpResponseWrapper status code is not properly mapped when method setStatus(int sc, String sm) is used.
Jersey library (org.glassfish.jersey.containers:jersey-container-servlet-core:2.22.1) is using deprecated method setStatus in class org.glassfish.jersey.servlet.internal.ResponseWriter
```
        final String reasonPhrase = responseContext.getStatusInfo().getReasonPhrase();
        if (reasonPhrase != null) {
            response.setStatus(responseContext.getStatus(), reasonPhrase);
        } else {
            response.setStatus(responseContext.getStatus());
        }
```
Please update code and override missing method:

```
    @Override
    public void setStatus(int sc, String sm) {
        super.setStatus(sc, sm);

        this.httpStatusCode = sc;
    }
```

OriYosefiMSFT:
Thank you for reporting and thank you @knapsu for contributing a fix!

PR #285, opened 2016-06-16
HTTP response code is not always set properly

knapsu:
ApplicationInsightsHttpResponseWrapper does not always set HTTP status code properly resulting in displaying code '200' in Azure portal even for failed requests.
The reason is the missing override for setStatus(int sc, String sm) method.


knapsu:
This pull request addresses issue #282

PR #288, opened 2016-07-01
Propagate log context data

sfthomas:
Details of the Logging Contexts (MDC, ThreadContext, etc) are not sent to AI - this fix adds them as Custom Properties for each logging framework.


sayhi2mukul:
Hi Yonisha,

Will this change be published as new version in maven repository.
http://mvnrepository.com/artifact/com.microsoft.azure/applicationinsights-logging-log4j2 shows 1.0.4 as latest version published in march. 

Thanks,
Mukul

OriYosefiMSFT:
Hi Mukul, 

We plan to release another version (including to Maven) soon. We do want to add a few more things first before we release the binaries though. 

Thanks,
Ori

Issue #292, opened 2016-07-26
Memory leak when logging exception telemetry


jgogstad:
The `WebRequestTrackingFilter` uses `ThreadLocal` to store state. This state is not cleaned up in a `finally`-clause, but it's cleaned up on [this line](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java#L187). That line is never reached if `chain.doFilter` throws an exception since the exception handling rethrows the exception.

By the way, wouldn't the rethrow make it impossible to correlate the exception with the request? Since no telemetry in sent with request information? How does that correlation work?

By the way2: It's good practice to have classes not marked with `final` and use `protected` for fields and methods when writing libraries. When bugs such as these occur, it is much easier for clients to deal with them. With the code as it stands now, I'm left with forking the project and releasing an in-house version with a fix, or copy all the code into my code base.

PR #293, opened 2016-07-27
#292: Fixes memory leak

jgogstad:
Fixes the memory leak described in #292


jgogstad:
that was supposed to be separate pull requests, and I see I accidentaly used our CI-user for the commits. I'll close this PR and create two separate ones

PR #294, opened 2016-07-27
Cleanup ThreadContext in finally-clause

### LINKED ISSUE #292: Memory leak when logging exception telemetry

jgogstad:
Fixes #292

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -73,31 +73,27 @@ public final class WebRequestTrackingFilter implements Filter {
      */
     public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
         ApplicationInsightsHttpResponseWrapper response = new ApplicationInsightsHttpResponseWrapper((HttpServletResponse)res);
-        boolean isRequestProcessedSuccessfully = true;
-
         setKeyOnTLS(key);
 
-        if (isInitialized) {
-            isRequestProcessedSuccessfully = invokeSafeOnBeginRequest(req, response);
-        }
+        boolean isRequestProcessedSuccessfully = invokeSafeOnBeginRequest(req, response);
 
         try {
             chain.doFilter(req, response);
+            invokeSafeOnEndRequest(req, response, isRequestProcessedSuccessfully);
         } catch (ServletException se) {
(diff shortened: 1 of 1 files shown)


OriYosefiMSFT:
Thanks for contributing the solution. I understand the issue you are reporting but I'm concerned that the proposed code may have an issue as well. 
webModulesContainer.invokeOnEndRequest would invoke the module's OnEndRequest which uses the TLS (e.g. [here](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java#L113) ).
After the proposed changed, the TLS would have already been cleaned up in the "finally" clause.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -181,10 +182,6 @@ private boolean invokeSafeOnBeginRequest(ServletRequest req, ServletResponse res
     private void invokeSafeOnEndRequest(ServletRequest req, ServletResponse res) {
         try {
             webModulesContainer.invokeOnEndRequest(req, res);
```
OriYosefiMSFT: This call invokes the modules' onEndRequest method which uses the TLS. After cleanup above, the TLS would no longer be available

jgogstad:
True, sorry for the lack of proper testing on my part. Pushed a fix that also addresses #298. I restructured the conditionals a bit in order to make the exception handling more readable. As long as you're targeting JRE 1.6, we can't collapse the catch-blocks so the code is duplicated.

gupele:
Thanks for the fix!

Issue #281, opened 2016-05-26
Response codes > 300 are treated as failed


rdpeake:
Our server returns lots of "304 not modified", however this library only flags responses between 200 and 300 as success.

Is there an intention behind this choice?

adriaanthomas:
Same problem here.

Looking at [the code](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java#L120) it suggests this is intentional, but it looks strange to me. Why not do a `telemetry.setSuccess(response.getStatus() < 400)`?

OriYosefiMSFT:
will consider response codes >=400 (with the exception of 401) as failures. Anything under is considered success,

PR #295, opened 2016-07-27
Consider 4xx and 5xx as failures

jgogstad:
Redirects shouldn't be considered failures, as reported in #281


web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
```diff
@@ -117,7 +117,7 @@ public void onEndRequest(ServletRequest req, ServletResponse res) {
 
             ApplicationInsightsHttpResponseWrapper response = ((ApplicationInsightsHttpResponseWrapper)res);
             if (response != null) {
-                telemetry.setSuccess(response.getStatus() >= 200 && response.getStatus() < 300);
+                telemetry.setSuccess(response.getStatus() < 400);
```
OriYosefiMSFT: Thanks for the contribution.
I see the  .NET SDK [also treat 401 as success](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Core/Managed/Shared/DataContracts/RequestTelemetry.cs#L233)

Please keep the same behavior so the SDKs are aligned. Thanks.

OriYosefiMSFT:
I will add another commit to identify 401 as success

PR #296, opened 2016-07-29
Features/filter

gupele:


samples/src/main/java/com/microsoft/applicationinsights/sample/DebugProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.microsoft.applicationinsights.extensibility.TelemetryProcessor;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+/**
+ * Sample filter
+ *
+ * Created by gupele on 7/26/2016.
+ */
+public class DebugProcessor implements TelemetryProcessor {
```
OriYosefiMSFT: May be worth to make this sample a little easier to read:
1. I would rename to SampleProcessor (instead of DebugProcessor)
2. Maybe rename setBad to setPass
3. Add documentation that explains the goal of the filter and that in the sample filter setPass chooses whether to pass all telemetry or to drop all telemetry.

core/src/main/java/com/microsoft/applicationinsights/extensibility/TelemetryProcessor.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.microsoft.applicationinsights.extensibility;
+
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+/**
+ * Created by gupele on 7/26/2016.
+ */
+public interface TelemetryProcessor {
+    boolean process(Telemetry telemetry);
```
OriYosefiMSFT: Maybe worth adding documentation about the meaning of the return value

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -172,6 +179,37 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
...
+                }
+                ArrayList<TelemetryProcessorXmlElement> legal = new ArrayList<TelemetryProcessorXmlElement>();
+                for (TelemetryProcessorXmlElement element : b) {
+                    String fullTypeName = builtInMap.get(element.getType());
+                    if (StringUtils.isNullOrEmpty(fullTypeName)) {
+                        InternalLogger.INSTANCE.error("Failed to find built in processor: '%s', ignored", element.getType());
+                        continue;
+                    }
+                    element.setType(fullTypeName);
+                    legal.add(element);
```
OriYosefiMSFT: Maybe validProcessors or availableProcessors would be more readable than legal

core/src/main/java/com/microsoft/applicationinsights/internal/processor/PageViewTelemetryFilter.java
```diff
@@ -0,0 +1,140 @@
...
+import com.microsoft.applicationinsights.internal.annotation.BuiltInProcessor;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.telemetry.PageViewTelemetry;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+/**
+ * The class can filter out PageViewTelemetries that
+ *   have a duration which is less than a predefined value
+ *   have URLs that has parts that are not needed, based on configuration
+ *   have unneeded page names that were predefined in configuration
```
OriYosefiMSFT: worth elaborating a little about what it means part that are not needed, etc.

core/src/main/java/com/microsoft/applicationinsights/internal/processor/PageViewTelemetryFilter.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ *  Illegal value will prevent from the filter from being used.
```
OriYosefiMSFT: I think "Invalid values would prevent the filter from being used" would be clearer

core/src/main/java/com/microsoft/applicationinsights/internal/processor/PageViewTelemetryFilter.java
```diff
@@ -0,0 +1,140 @@
...
+                    return false;
+                }
+            }
+        }
+
+        if (notNeededNames.contains(asPVT.getName())) {
+            return false;
+        }
+
+        long pvtDuration = asPVT.getDuration();
```
OriYosefiMSFT: May be worth to do the getDuration validation first. It's more lightweight than the URL/Name scan and if the duration is below the threshold, we can avoid the lengthier string.contains operations

core/src/main/java/com/microsoft/applicationinsights/internal/processor/RequestTelemetryFilter.java
```diff
@@ -0,0 +1,150 @@
...
+
+        private FromTo(int from, int to) {
+            this.from = from;
+            this.to = to;
+        }
+    }
+
+    private long minimumDurationInMS = 0;
+    private boolean hasBlocked;
+    private final Set<String> exactBadResponseCodes = new HashSet<String>();
```
OriYosefiMSFT: I think something like specificlyIgnoredResponseCodes would be clearer

core/src/main/java/com/microsoft/applicationinsights/internal/processor/RequestTelemetryFilter.java
```diff
@@ -0,0 +1,150 @@
...
+    private final List<FromTo> fromTos = new ArrayList<FromTo>();
```
OriYosefiMSFT: I think ignoredResponseCodeRange would be clearer

core/src/main/java/com/microsoft/applicationinsights/internal/processor/RequestTelemetryFilter.java
```diff
@@ -0,0 +1,150 @@
...
+import com.microsoft.applicationinsights.internal.annotation.BuiltInProcessor;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.telemetry.Duration;
+import com.microsoft.applicationinsights.telemetry.RequestTelemetry;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+/**
+ * The class can filter out RequestTelemetries that
+ *   have a duration which is less than a predefined value
+ *   have http codes that are not needed based on configuration
```
OriYosefiMSFT: Maybe "have specific HTTP responses codes or HTTP response code ranges"

core/src/main/java/com/microsoft/applicationinsights/internal/processor/RequestTelemetryFilter.java
```diff
@@ -0,0 +1,150 @@
...
+            return true;
+        }
+
+        if (!hasBlocked && minimumDurationInMS <= 0) {
+            return true;
+        }
+
+        if (telemetry instanceof RequestTelemetry) {
+            RequestTelemetry requestTelemetry = (RequestTelemetry)telemetry;
+            String requestRespohseCode = requestTelemetry.getResponseCode();
```
OriYosefiMSFT: maybe responseCode instead fo requestResponseCode? The combination of request and response together is a bit confusing (at least to me :-) )

core/src/main/java/com/microsoft/applicationinsights/internal/processor/SyntheticSourceFilter.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Set;
+
+import com.microsoft.applicationinsights.agent.internal.common.StringUtils;
+import com.microsoft.applicationinsights.extensibility.TelemetryProcessor;
+import com.microsoft.applicationinsights.internal.annotation.BuiltInProcessor;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+/**
+ * The class will filter out telemetries that come from 'unneeded' configured list of source names
```
OriYosefiMSFT: It's not really clear what this filter does. I think it may be helpful to elaborate a little more

jgogstad:
Any documentation on how to configure the processors that came in 1.0.5? For example how to configure the `Set`s in `PageViewTelemetryFilter`.

gupele:
Hi @jgogstad,

The filters go into a new XML tag "TelemetryProcessor". There you can choose and customize the built in ones and/or new custom filters

For example,
```
<ApplicationInsights>
   <TelemetryProcessors>

      <BuiltInProcessors>
           <Processor type="TraceTelemetryFilter">
                  <Add name="FromSeverityLevel" value="ERROR"/>
           </Processor>

           <Processor type="RequestTelemetryFilter">
                  <Add name="MinimumDurationInMS" value="100"/>
                  <Add name="NotNeededResponseCodes" value="200-400"/>
           </Processor>

           <Processor type="PageViewTelemetryFilter">
                  <Add name="DurationThresholdInMS" value="100"/>
           </Processor>

           <Processor type="SyntheticSourceFilter"/>

     </BuiltInProcessors>

     <CustomProcessors>
        <Processor type="com.microsoft.applicationinsights.sample.SampleFilter">
            <Add name="Pass" value="false"/>
        </Processor>
    </CustomProcessors>

 </TelemetryProcessors>
</ApplicationInsights>

```
In this example the configuration file declares built in filters and one custom one.
Basically, to instantiate a filter you need to implement the TelemetryProcessor interface

The method to implement is `'public boolean process(Telemetry telemetry)`' where a telemetry is passed to the method which returns a boolean value:
'True'  - The telemetry is 'approved'. Note that other filters can veto the telemetry which will prevent it from being sent
'False' - The telemetry is not approved. This means the telemetry will not be sent

Filters are a good way to filter out telemetries you don't care for and by doing so reduce the traffic.

The filter can get data from configuration by having a 'setXXX' method that matches its property name.
For example, if the filter has a property 'percentage', it should have a 'setPercentage(string percentage)' method and then you can pass data from configuration like this:

<Processor type="full.path.to.CustomFilter>
	<Add name "Percentage" value="50"/>
</Processor>

The framework will instantiate the class and will use reflection to set the value in the proper method (please read the 'SampleFilter).

The framework also sets its own 'common' filters to ease the work of users.

The following are the current 'built in' filters

TraceTelemetryFilter
	This filter can filter out log telemetries. In our above example, only logs with severity level of Error and above (Critical) will be sent.
	This is a good way to have your app produce logs in lower severity level but on the same time, send only the ones that will be fully helpful
	

PageViewTelemetryFilter
	This filter can filter out PageViewTelemetries.
		You can set the 'NotNeededUrls' to collection of urls to prevent them from being sent. For example: <`Add name="NotNeededUrls" value="badsite1, badsite2"/>
`		which means that Urls that contain those keywords will not be sent
		
		You can set the 'NotNeededNames' much in the same way to prevent PageView with those names
		
		You can set the 'DurationThresholdInMS' to filter out PageViews that had a duration of less than this value
		
	
RequestTelemetryFilter
	This filter can filter out RequestTelemetries
		You can set the 'NotNeededResponseCodes' to prevent from requests with those response codes to be sent. For example `<Add name="NotNeededResponseCodes" value="200-400,500"/>`
		
		You can set the 'MinimumDurationInMS' to send only 'slow' requests.
		

SyntheticSourceFilter
	This filter can filter out all telemetries that are originated in a synthetic source (you can also set the source names to block)
	

TelemetryEventFilter
	This filter can filter out EventTelemtries
		This filter can be useful in case you use the SDK for creating Event Telemetries but then in production, you would like to prevent some of them from being sent without re-compiling your code.
		


We will soon have blog post for that and please note that we are working to add more 'built in' filters and enhance the current ones.

Thanks

jgogstad:
Thanks for the clarification! I tried you sample configuration, and it resulted in a `NullPointerException`. See #305

PR #304, opened 2016-08-08
Support method names that follow java naming standard

jgogstad:
When configuring processors, it's natural for java developers to follow the naming standard for the rest of the project, that is lower case first character in method names.


littleaj:

PR #305, opened 2016-08-08
Fixed NullpointerException when no fields are set on processors.

jgogstad:
The current implementation of `TelemetryConfigurationFactory` throws NullPointerException whenever it encounters processors that are not configured with any attributes, e.g.

```
<ApplicationInsights>
   <TelemetryProcessors>
     <CustomProcessors>
        <Processor type="com.example.MyProcessor"/>
    </CustomProcessors>
 </TelemetryProcessors>
</ApplicationInsights>
```

There's also a bug where the same processor instance is added multiple times. This occurs when multiple fields are set on a processor.

This PR fixes both (and includes PR #304 ).

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
@@ -455,18 +455,26 @@ public enum TelemetryConfigurationFactory {
         }
 
         for (TelemetryProcessorXmlElement className : classNames) {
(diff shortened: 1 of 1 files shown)


gupele:
@jgogstad, thanks for your valuable input.

We will not use this PR since we have an ongoing background effort to refactor the TC Factory, I already started doing that for the TelemetryProcessor, I will use your input for the PR I am going to submit soon

Thanks!

jgogstad:
Brilliant! We'll use a fork until it's ready. Looking forward to see the result. Closing this.

PR #302, opened 2016-08-07
Features/catch runtime exceptions

gupele:

PR #306, opened 2016-08-08
Support for exception tracking of exceptions handled by Spring

jgogstad:
The current implementation of `WebRequestTrackingFilter` requires exceptions to propagate to the filter in order for them to be reported. That does not happen if the exception is handled by the underlying MVC or Webservice framework.

In the case of Spring MVC, an attribute with the exception is added to the request. Note that the attribute name in this PR is only used by Spring 4 (but there are not Spring 4 dependencies). I'll investigate how Spring 3 behaves, I can't find the same attribute there. The attribute names are configurable, I'm pretty sure that CXF and Jersey behaves the same way, so the module _should_ be usable for those frameworks as well.

I have some tests that verifies this module, they depend on `MockMvc` from Spring 4 in order to call webservices (see [this link](https://spring.io/guides/tutorials/bookmarks/) for an introduction). I have frankly no idea how to test webservices in Spring 3, so as for now the PR is without tests.

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
@@ -1,0 +1,74 @@
+package com.microsoft.applicationinsights.web.extensibility.modules;
+
+import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import com.microsoft.applicationinsights.extensibility.TelemetryModule;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.web.servlet.DispatcherServlet;
+
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+public class SpringMvcRequestTrackingTelemetryModule implements WebTelemetryModule, TelemetryModule {
+
+    private TelemetryClient telemetryClient;
+    private List<String> exceptionAttributesList = asList(DispatcherServlet.class.getName() + ".EXCEPTION");
+    private String exceptionAttributes;
+
+    @Override
+    public void initialize(TelemetryConfiguration configuration) {
+        telemetryClient = new TelemetryClient(configuration);
+    }
+
+    @Override
+    public void onBeginRequest(ServletRequest req, ServletResponse res) {
+
+    }
+
+    @Override
+    public void onEndRequest(ServletRequest req, ServletResponse res) {
+        for (String exceptionAttribute : exceptionAttributesList) {
+            Exception exception = (Exception) req.getAttribute(exceptionAttribute);
+            if (exception != null) {
+                telemetryClient.trackException(exception);
+            }
+        }
+    }
+
+    public TelemetryClient getTelemetryClient() {
+        return telemetryClient;
+    }
+
+    public void setTelemetryClient(TelemetryClient telemetryClient) {
+        this.telemetryClient = telemetryClient;
+    }
+
+    public String getExceptionAttributes() {
+        return exceptionAttributes;
+    }
+
+    public void setExceptionAttributes(String exceptionAttributes) {
+        try {
+            if (StringUtils.isBlank(exceptionAttributes)) {
+                return;
+            }
+            String[] attributes = exceptionAttributes.split(",");
+            exceptionAttributesList = new ArrayList<>(attributes.length);
+            for (String attribute : attributes) {
+                if (!StringUtils.isBlank(exceptionAttributes)) {
+                    exceptionAttributesList.add(attribute);
+                }
+            }
+            InternalLogger.INSTANCE.trace(String.format("SpringMvcRequestTrackingTelemetryModule: set ExceptionAttributes: %s", exceptionAttributes));
+        } catch (Throwable t) {
+            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, String.format("SpringMvcRequestTrackingTelemetryModule: failed to parse ExceptionAttributes: %s", exceptionAttributes));
+            throw t;
+        }
+
+    }
+}


OriYosefiMSFT:
@gupele, wouldn't [this PR](https://github.com/Microsoft/ApplicationInsights-Java/pull/302) allow catching the Spring exceptions (as well as others)?

jgogstad:
That looks interesting, but I don't really see how it's going to work. Maybe you can cast some light on it?

It is not really interesting to track *all* exceptions thrown in the JVM, many exceptions are thrown by libraries and then later handled by the same library. For instance, this library throws `ClassNotFoundException`s that are later handled in order to check if certain classes are on the classpath. This is a common pattern, throw in an IoC-framework and an ORM, and you'll have dozens all the time. For application insights, users are interesting in tracking "exceptions that shouldn't have occured".

Tracking unhandled exceptions goes without saying, but tracking exceptions that are handled, but results in "5xx internal server error" is quite difficult, how do you know which exception to track? We need to be able to differentiate between "normal exceptions" that are thrown and handled as part of regular application flow, and "irregular exceptions" that are thrown by "something" and and handled by the MVC- or webservice library. I don't see how to cover these cases without explicitly integrating with the MVC- or webservice libraries in question.

gupele:
I'm not sure I totally agree. Yes, runtime exceptions might be thrown and caught by libraries but this usually will be done on initialization phases. Usually, after the library is initialized, runtime exceptions are not the way to drive the flow. Having said that, there is a point in filtering out those exceptions, one way to do so might be by using a Filter which I am going to submit soon which will let users filter out exceptions that are not originated from certain packages, another option is basically do that by adding conf options in the configuration file, AI-Agent.xml, where the users can enable/disable this feature.

jgogstad:
Custom filters would be nice, especially if they get passed the originating `Class` that the exception originated from.

Even though exceptions are rarely used to control application flow per se, it is still many exceptions that are normal and expected. Take for instance opening an web page in a browser and then close the page. If the entire page didn't load before the user closed it, the webserver (Jetty in my case) will throw an `IOException` deep down somewhere because the TCP-socket was prematurely terminated, or terminated while the web server was writing to it. That `IOException` is of course handled long before it reaches any user code, but it's still perfectly normal and perfectly acceptable. If you attach a debugger to a running application and ask it to break on every exception, you'll see lots of these, especially if an ORM like Hibernate is involved.

I'm not saying tracking all Runtime exceptions is necessarily a bad idea, it's just not very straight forward to get value from it in big applications. I'd love to see an approach though, I just believe I'll see a lot of irrelevant exceptions until I'm convinced otherwise. Until then, I believe that modules like this one have value since they 1) have 100% "this is a relevant exception"-rate and 2) doesn't depend on an active javaagent.

gupele:
Exceptions that derive flows are more likely to be checked exceptions, like IOException. It is not likely to find runtime exceptions that do that. This is why catching runtime exceptions might be interesting. Also, I am going to add the ability to send only runtime exceptions that originated from interesting packages, and ability to suppress runtime exceptions that are not interesting at all no matter where they came from, say java.lang.NumberFormatException which might give the user the needed information for digging into the issue, I am going to add that to the PR so hopefully we will have this feature next version

Having said that, the module might be helpful for those who are not using the agent, I will review the code soon.

jgogstad:
That's a good point. It might not be as much noise as I initially pictured. Looking forward to try it out!

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public void setTelemetryClient(TelemetryClient telemetryClient) {
+        this.telemetryClient = telemetryClient;
+    }
+
+    public String getExceptionAttributes() {
+        return exceptionAttributes;
+    }
+
+    public void setExceptionAttributes(String exceptionAttributes) {
```
OriYosefiMSFT: I'm not sure I understand when this method is used. Shouldn't it be called in 'onEndRequest'?

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.servlet.ServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+public class SpringMvcRequestTrackingTelemetryModule implements WebTelemetryModule, TelemetryModule {
+
+    private TelemetryClient telemetryClient;
+    private List<String> exceptionAttributesList = asList(DispatcherServlet.class.getName() + ".EXCEPTION");
```
OriYosefiMSFT: Can this be used via the 'EXCEPTION_ATTRIBUTE' field instead of be set explicitly?

jgogstad: No. `EXCEPTION_ATTRIBUTE` is only a field in Spring MVC 4. Application insights is compiled against Spring MVC 3 so that would cause a compilation failure.

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
jgogstad: It's a configuration parameter for specifying other attributes on `HttpServletRequest` where ApplicationInsights should look for exceptions. The provided default configuration (line 20) only specify an attribute available in Spring MVC 4. Clients that use containers like CXF, Jersey or Spring 3 will have to configure this module with the attribute value that the framework use.

OriYosefiMSFT: What is the flow of events you are expecting for users of a different framework?
Is it expected for the web filter to call the module's `setExceptionAttributes` with the relevant attributes.  
Is fo, I'm not sure there is a simple way to do this.
Can you please give an example of how one would configure the module with a different attribute?

gupele: It will be quite impossible for users to use this, so I think the method should be removed, the class should be renamed to be *SpringMVC4* and have a dedicated SpringMVC3 class now, or later on, and naturally they can share code, this way the classes are more explicit and clearly state what they do, and we keep things in configuration rather then coding

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
```diff
@@ -0,0 +1,74 @@
...
+    public void onEndRequest(ServletRequest req, ServletResponse res) {
+        for (String exceptionAttribute : exceptionAttributesList) {
+            Exception exception = (Exception) req.getAttribute(exceptionAttribute);
+            if (exception != null) {
+                telemetryClient.trackException(exception);
+            }
+        }
+    }
+
+    public TelemetryClient getTelemetryClient() {
```
gupele: Why is this method needed?

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
```diff
@@ -0,0 +1,74 @@
...
+        telemetryClient = new TelemetryClient(configuration);
+    }
+
+    @Override
+    public void onBeginRequest(ServletRequest req, ServletResponse res) {
+
+    }
+
+    @Override
+    public void onEndRequest(ServletRequest req, ServletResponse res) {
```
gupele: I think this method should be under try/catch

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
```diff
@@ -0,0 +1,74 @@
+package com.microsoft.applicationinsights.web.extensibility.modules;
```
gupele: Header is missing

gupele: Can you please add a unit test class?

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/SpringMvcRequestTrackingTelemetryModule.java
OriYosefiMSFT: I believe this is invoked through `safeInvokeOnEndRequest` (or something like it) which wraps the invocation with try/catch

gupele: I think the try/catch should be added, as all other modules do, and not rely on the caller.

littleaj:

PR #316, opened 2016-08-18
Allow configuration of PC collection cadence

gupele:
Allow users to set the cadence of performance counter collection

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/PerformanceCountersXmlElement.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/PerformanceCountersXmlElement.java
@@ -21,9 +21,7 @@
 
 package com.microsoft.applicationinsights.internal.config;
 
-import javax.xml.bind.annotation.XmlElement;
-import javax.xml.bind.annotation.XmlElementWrapper;
-import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.*;
 import java.util.ArrayList;
 
 /**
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/config/PerformanceCountersXmlElement.java
```diff
@@ -32,6 +30,7 @@
 @XmlRootElement(name="PerformanceCounters")
 public class PerformanceCountersXmlElement {
     private boolean useBuiltIn = true;
+    private long cadenceInSec = 60;
```
OriYosefiMSFT: I think frequency or even collectionFrequency would be clearer than cadence

OriYosefiMSFT: You already defined a constant with the default value. It's better if you can use this constant everywhere instead of writing 60 in two different places

core/src/test/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainerTest.java
```diff
@@ -111,14 +112,14 @@ public void testSetStartCollectingIntervalInMillisZero() {
 
     @Test
     public void testSetCollectingIntervalInMillisMillisNegativeNumber() {
-        PerformanceCounterContainer.INSTANCE.setCollectingIntervalInMillis(-100);
-        assertTrue(PerformanceCounterContainer.INSTANCE.getCollectingIntervalInMillis() > 0);
+        PerformanceCounterContainer.INSTANCE.setCadenceInSec(-1);
+        assertTrue(PerformanceCounterContainer.INSTANCE.getCadenceInSec() > 0);
```
OriYosefiMSFT: You can check it's >= MIN_CADENCE_IN_SECONDS

PR #326, opened 2016-09-13
Agent standalone

gupele:


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlParserUtils.java
```diff
@@ -46,7 +46,20 @@ public static Element getFirst(NodeList nodes) {
...
+    public static String getAttribute(Element element, String attributeName) {
+        if (element == null) {
+            return null;
+        }
+
+        String attributeValue = element.getAttribute(attributeName);
+        return attributeValue;
+    }
+
+    public static boolean getEnabled(Element element, String attributeName) {
```
OriYosefiMSFT: it is a generic bool value which can be used for things other than enabled/disabled, no? maybe better to call it getBoolValue or something like this for clarity?

PR #343, opened 2016-11-08
allow indicating a metric is actually a custom performance counter

OriYosefiMSFT:


core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java
```diff
@@ -59,6 +59,13 @@ public MetricTelemetry(String name, double value) {
     }
 
     /**
+     * indicate that this metric is a custom performance counter and should be sent to the performance counters table
+     */
+    public void markAsPerfCounter(){
```
cijothomas: markAsCustomPerfCounter as function name would better explain what this does.

PR #367, opened 2017-04-11
Updated the agent project not to be transitive when resolved for core

isaac84:
So the org.objectweb package doesn't end up in the root of the core jar.

This causes us a SecurityException due to another version of the asm jar (3.2 in our case) in our deployed WAR.

I ran the gradle build locally and all the existing tests here still pass.


Dmitry-Matveev:
Thank you!

Issue #369, opened 2017-04-18
Apache HttpClient 4.3.5 vulnerable (CVE-2015-5262)


sanjeevp:
Apache HttpClient 4.3.5 has a known vulnerability and should be upgraded to the latest version: 4.5.3.

See: [CVE-2015-5262](https://nvd.nist.gov/vuln/detail/CVE-2015-5262)

Dmitry-Matveev:
Fixed in 1.0.9. Thanks for reporting!

PR #374, opened 2017-05-24
Bugfix/362

debugthings:
Based on feed back from @isaac84 and PR #367  I've updated the core JAR to remove the transitive dependencies for guava, which depends on JSR305 which has it's own dependency on javax.annotation. 

I also removed the transitive dependency from core and agent, even though it's not really required in this situation as the web JAR could have a need for a shadow jar in the future. This will save on bug hunts later.

Lastly I updated to the latest version of the httpclient based on the recommendation of [CVE-2015-5262](https://nvd.nist.gov/vuln/detail/CVE-2015-5262) however this vulnerability is classified as a DoS. In short it is a bug in the httpclient that does not honor the SSL handshake timeout.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -39,8 +39,8 @@ dependencies {
     compile ([group: 'eu.infomas', name: 'annotation-detector', version: '3.0.4'])
     compile ([group: 'commons-io', name: 'commons-io', version: '2.4' ])
     compile ([group: 'org.apache.commons', name: 'commons-lang3', version: '3.1']) 
-    compile ([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3'])
-    compile ([group: 'com.google.guava', name: 'guava', version: '12.0.1']) { transitive = false }
+    compile ([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.3.5'])
+    compile ([group: 'com.google.guava', name: 'guava', version: '12.0.1'])
     testCompile group: 'junit', name: 'junit', version: '4.11'
     testCompile group: 'org.mockito', name: 'mockito-all', version: '1.8.0'
     testCompile group: 'com.google.code.gson', name: 'gson', version: '1.7.2'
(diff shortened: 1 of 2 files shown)


sharwell:
@debugthings Is there any way you can separate the bug fix for #369 into its own PR? I'm getting requests from external users regarding the status of closing that specific CVE but I'm not knowledgeable enough about this SDK to review all of what you've included here.

core/build.gradle
```diff
@@ -36,11 +36,11 @@ archivesBaseName = 'applicationinsights-core'
 
 dependencies {
     provided (project(':agent')) { transitive = false }
-    compile group: 'eu.infomas', name: 'annotation-detector', version: '3.0.4'
-    compile group: 'commons-io', name: 'commons-io', version: '2.4'
-    compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.1'
-    compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.3.5'
-    compile group: 'com.google.guava', name: 'guava', version: '12.0.1'
+    compile ([group: 'eu.infomas', name: 'annotation-detector', version: '3.0.4'])
```
Dmitry-Matveev: Does it mean that our customer who used core and web with transitive dependencies and relied on them will need to copy them manually from now on? If so, you'd need to update a documentation in Azure Docs repo as well and PR them :)

core/build.gradle
```diff
@@ -36,11 +36,11 @@ archivesBaseName = 'applicationinsights-core'
...
+    compile ([group: 'commons-io', name: 'commons-io', version: '2.4' ])
+    compile ([group: 'org.apache.commons', name: 'commons-lang3', version: '3.1']) 
+    compile ([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3'])
```
Dmitry-Matveev: I remember there were some version specific things in our code around apache.http - did you check that code and telemetry transmission are fine with that change?

core/build.gradle
debugthings: I think I will opt to keep this transitive and relocate the javax packages out of the root of the JAR. This should satisfy that requirement. Look for an updated commit soon

core/build.gradle
debugthings: This version does work. But per @sharwell I will move this to another PR to be reviewed.

sharwell: @debugthings Thank you so much for this. I know it's more work for you but I know it means a lot to one downstream user.

Issue #347, opened 2016-11-17
Error when using `TelemetryClient.trackSessionState`


tobiasdiez:
The code
````java
        telemetryClient = new TelemetryClient();
        telemetryClient.trackSessionState(SessionState.Start);
        telemetryClient.trackEvent("TestEvent");
        telemetryClient.trackSessionState(SessionState.End);
````
results in two identical errors
> Failed to send, Bad request  : {"itemsReceived":1,"itemsAccepted":0,"errors":[{"index":0,"statusCode":400,"message":"Unknown domain type 'Microsoft.ApplicationInsights.SessionStateData'"}]}

The session is not tracked in the online portal but the custom event "TestEvent" shows up (thus the connection works in principle). How can I track session start and end?

Dmitry-Matveev:
trackSessionState() is going to be removed from the upcoming version of the API.
Currently, only HockeyApp SDK supports SessionState, Java SDK and .NET SDK cannot successfully submit this telemetry type. It was already removed from [TelemetryClient](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/TelemetryClient.cs) in .NET SDK.

To workaround this limitation one can submit a metric with the duration of the session and plot it in AI.

PR #375, opened 2017-05-24
Removing trackSessionState and depreciating supporting classes

debugthings:
Per issue #347 we need to remove this method. This is a breaking change to bring us up to a similar API level in .NET and should be noted in the release notes.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -108,7 +108,16 @@ public class TelemetryClient {
                 Strings.isNullOrEmpty(getContext().getInstrumentationKey()) || configuration.isTrackingDisabled();
     }
 
-     /**
+    /**
+     * Sends the specified state of a user session to Application Insights.
+     * @param sessionState {@link com.microsoft.applicationinsights.telemetry.SessionState}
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -108,16 +108,7 @@ public boolean isDisabled() {
                 Strings.isNullOrEmpty(getContext().getInstrumentationKey()) || configuration.isTrackingDisabled();
     }
 
-    /**
-     * Sends the specified state of a user session to Application Insights.
-     * @param sessionState {@link com.microsoft.applicationinsights.telemetry.SessionState}
-     *                     value indicating the state of a user session.
-     */
-    public void trackSessionState(SessionState sessionState) {
```
Dmitry-Matveev: Should we keep it "deprecated" for at least one version to give some time for switching over to a new API?
Also, since this is a top-level API breaking change, we might need to mark it deprecated for this major version and only remove in the next major version like Java SDK 2.x.x. What do you think?

debugthings: True. Pulling the whole API could break users. Maybe mark it @Depreciated and change the implementation to call trackEvent() instead?

PR #383, opened 2017-06-08
Apache Http 4.5.3

Dmitry-Matveev:


Dmitry-Matveev:
@debugthings , I saw you did this (update to 4.5.3) before but then you rolled back, did you have some issues with this update? It looks like we need it to solve SSL Handshake issue our customers reported in addition to #369 .

Issue #389, opened 2017-06-16
Expose all fields of dependency tables in the data model 


beckylino:

dhaval24:
@grlima I believe that you are addressing part of this in #457 like exposing parent-id etc.

dhaval24:
@grlima with the cross component telemetry correlation and updated schema this feature should be mostly complete now. Let me know your thoughts.

PR #396, opened 2017-06-30
update the schema to latest

SergeyKanzhelev:
This is the first step of migrating to the latest clean schema of Application Insights. There are lots of things to do next: add new fields, deprecate and remove the use of performance counters, clean up context tags. But this PR is sufficient by itself. Other changes will make it too big

#389


core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/ExceptionData.java
```diff
@@ -1,175 +1,196 @@
...
+        
         writer.write("exceptions", exceptions);
-        writer.write("measurements", measurements);
-
-        if (severityLevel != null) {
-            writer.write("severityLevel", severityLevel.toString());
-        }
-
+        
+        writer.write("severityLevel", severityLevel);
```
Dmitry-Matveev: Does write() check for null or severityLevel cannot be null now?

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/MessageData.java
```diff
@@ -1,131 +1,150 @@
...
         writer.write("ver", ver);
+        
         writer.write("message", message);
-
-        if (severityLevel != null) {
-            writer.write("severityLevel", severityLevel.toString());
-        }
-
+        
+        writer.write("severityLevel", severityLevel);
```
Dmitry-Matveev: Same here.

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/DataPoint.java
```diff
@@ -1,179 +1,218 @@
...
      * Serializes the beginning of this object to the passed in writer.
      * @param writer The writer to serialize this object to.
-     * @throws IOException Might be throw during serialization.
      */
-    protected void serializeContent(JsonTelemetryDataSerializer writer) throws IOException {
+    protected void serializeContent(JsonTelemetryDataSerializer writer) throws IOException
+    {
         writer.write("name", name);
-
-        if (!DataPointType.Measurement.equals(kind)) {
```
SergeyKanzhelev: I think it was an optimization to not send the value that has a default: https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/DataPoint.bond#L13 Shouldn't break anything, I'll thing how to improve code generator

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/ExceptionData.java
SergeyKanzhelev: https://github.com/Microsoft/ApplicationInsights-Java/pull/396/files#diff-dc6856d601a520155e7f485732bcbf45R90

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/MessageData.java
SergeyKanzhelev: same - write will now check it

SergeyKanzhelev:
@Dmitry-Matveev  thank you for reviewing such a long PR!

PR #411, opened 2017-08-18
Dhdoshi/spring boot name bug

### LINKED ISSUE #359: Not possible to register WebApp name/key properly when using WebRequestTrackingFilter in Spring boot
Hi,

When using the WebRequestTrackingFilter with a Spring boot application on the root context path the filter cannot register the web app name/key properly:

```
AI: ERROR 30-01-2017 12:15, 22: WebApp name is not found, unable to register WebApp
AI: INFO 30-01-2017 12:15, 22: Successfully registered the filter 'ApplicationInsightsWebFilter
```

This seem to be due to dependencies to either having a (non-root) servlet context path set for the app or that the app is found under a class-loading path containing "WEB-INF", which is not the case for this Spring boot app.

Would be good if the initialisation config could read either the spring boot app name or to have some external configuration value to set for the web app name/key.

Thanks in advance

dhaval24:
Fix #359 
springbootbug

- [ ] CHANGELOG.md updated

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -58,7 +58,7 @@ public final class WebRequestTrackingFilter implements Filter {
     private boolean agentIsUp = false;
     private final LinkedList<ThreadLocalCleaner> cleaners = new LinkedList<ThreadLocalCleaner>();
 
-    //Only for Spring Boot Application (External Name Setting)
+    //For External Name Setting
     private String applicationName;
 
     // endregion Members
@@ -210,6 +210,10 @@ public final class WebRequestTrackingFilter implements Filter {
     public WebRequestTrackingFilter() {
     }
 
+    /**
+     * Parameterized constructor, used to set Application name explicitly
+     * @param applicationName
+     */
     public WebRequestTrackingFilter(String applicationName) {
         this.applicationName = applicationName;
     }
@@ -224,14 +228,10 @@ public final class WebRequestTrackingFilter implements Filter {
                 InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Agent was not found. Skipping the agent registration");
                 return;
             }
-
             ServletContext context = filterConfig.getServletContext();
-
             String name = getName(context);
-
             String key = registerWebApp(name);
             setKey(key);
-
             InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Successfully registered the filter '%s'", FILTER_NAME);
         } catch (Throwable t) {
             InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Failed to register '%s', exception: '%s'", FILTER_NAME, t.getMessage());
@@ -266,7 +266,7 @@ public final class WebRequestTrackingFilter implements Filter {
 
     private String getName(ServletContext context) {
 
-        // If Application is of type Spring Boot
+        // If Application Name is set externally
         if (applicationName != null) {
             return applicationName;
         }
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -27,6 +27,7 @@ import com.microsoft.applicationinsights.agent.internal.coresync.impl.AgentTLS;
 import com.microsoft.applicationinsights.common.CommonUtils;
 import com.microsoft.applicationinsights.internal.agent.AgentConnector;
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.schemav2.Internal;
 import com.microsoft.applicationinsights.internal.util.ThreadLocalCleaner;
 
 import javax.servlet.Filter;
@@ -39,10 +40,12 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Date;
 import java.util.LinkedList;
+import java.util.Properties;
 
 /**
  * Created by yonisha on 2/2/2015.
@@ -229,8 +232,8 @@ public final class WebRequestTrackingFilter implements Filter {
                 return;
             }
             ServletContext context = filterConfig.getServletContext();
-            String name = getName(context);
-            String key = registerWebApp(name);
+            applicationName = getName(context);
+            String key = registerWebApp(applicationName);
             setKey(key);
             InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Successfully registered the filter '%s'", FILTER_NAME);
         } catch (Throwable t) {
@@ -266,12 +269,12 @@ public final class WebRequestTrackingFilter implements Filter {
 
     private String getName(ServletContext context) {
 
-        // If Application Name is set externally
-        if (applicationName != null) {
-            return applicationName;
+        String name = getApplicatioNameFromProperties();
+
+        if (name != null) {
+            return name;
         }
 
-        String name = null;
         try {
             String contextPath = context.getContextPath();
             if (CommonUtils.isNullOrEmpty(contextPath)) {
@@ -295,6 +298,38 @@ public final class WebRequestTrackingFilter implements Filter {
             InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Exception while fetching WebApp name: '%s'", t.getMessage());
         }
 
+        return name;
+    }
+
+    /*This method is used to parse application name from the properties file*/
+    private String getApplicatioNameFromProperties() {
+
+        Properties properties = new Properties();
+        InputStream input = null;
+        String name = null;
+        try {
+            String fileName = "application.properties";
+            input = WebRequestTrackingFilter.class.getClassLoader().getResourceAsStream(fileName);
+            if (input == null) {
+                return null;
+            }
+
+            properties.load(input);
+            name = properties.getProperty("spring.application.name");
+        }
+        catch (IOException ex) {
+            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, "Exception in loading properties file: '%s'",ex.getMessage());
+        }
+        finally {
+            if (input != null) {
+                try {
+                    input.close();
+                }
+                catch (IOException e) {
+                    InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Error while closing properties file '%s'",e.getMessage());
+                }
+            }
+        }
         return name;
     }
 
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -213,14 +213,6 @@ public final class WebRequestTrackingFilter implements Filter {
     public WebRequestTrackingFilter() {
     }
 
-    /**
-     * Parameterized constructor, used to set Application name explicitly
-     * @param applicationName
-     */
-    public WebRequestTrackingFilter(String applicationName) {
-        this.applicationName = applicationName;
-    }
-
     private synchronized void initialize(FilterConfig filterConfig) {
         try {
 
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -21,29 +21,28 @@
 
 package com.microsoft.applicationinsights.web.internal;
 
+import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import com.microsoft.applicationinsights.agent.internal.coresync.impl.AgentTLS;
+import com.microsoft.applicationinsights.common.CommonUtils;
+import com.microsoft.applicationinsights.internal.agent.AgentConnector;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.util.ThreadLocalCleaner;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Date;
 import java.util.LinkedList;
-
-import javax.servlet.Filter;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
-import javax.servlet.FilterChain;
-import javax.servlet.FilterConfig;
-import javax.servlet.ServletContext;
-import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
-import com.microsoft.applicationinsights.common.CommonUtils;
-import com.microsoft.applicationinsights.TelemetryClient;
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
See the checklist in PR description? Please follow the check boxes:

1. Mention the change in public surface - new constructor. 
2. Update `CHANGELOG.md`

Also make sure that the new public method has a Javadoc with explanation how it when to use it.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -59,20 +58,24 @@
     private boolean agentIsUp = false;
     private final LinkedList<ThreadLocalCleaner> cleaners = new LinkedList<ThreadLocalCleaner>();
 
+    //Only for Spring Boot Application (External Name Setting)
```
SergeyKanzhelev: I don't think we need to limit use of it for Spring Boot. May one decide to do it in regular app?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -257,6 +265,12 @@ private String registerWebApp(String name) {
     }
 
     private String getName(ServletContext context) {
+
+        // If Application is of type Spring Boot
```
SergeyKanzhelev: again, do not limit it to spring boot only

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -59,20 +58,24 @@
     private boolean agentIsUp = false;
     private final LinkedList<ThreadLocalCleaner> cleaners = new LinkedList<ThreadLocalCleaner>();
 
+    //Only for Spring Boot Application (External Name Setting)
+    private String applicationName;
```
SergeyKanzhelev: make it `final`

SergeyKanzhelev: how `applicaitonName` will be used? Is it send along the telemetry? If it will end up as `roleName` - worse naming it appropriately

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: I am not quite sure about it, may be I am wrong though. I think other application like Spring/Jsp/Servlet etc  they pick up configuration with xml files. However, in Spring boot configuration is done with Java classes and application name is set using application.properties file. Therefore there was a need to explicitly set the name. Correct me if I am wrong here.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: Sure can make that as final

dhaval24: For, this one I have a question too. I think even without setting the name Telemetry flows in the portal. Only it comes in AI logs. I don' know the exact implication of it, in this context

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: 1. As soon as we make new constructor public, we cannot control if it's going to be used for Spring Boot only or not.
2. There might be another framework we haven't tried yet that required application name setup outside of configuration xml files.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: @dhaval24 , please check where this name is used through the code and whether it makes it into the telemetry item at all. If not, keeping it as `applicationName` is ok.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: On a side note - is there any implication if name is set in both configuration and the code with this constructor for non-SpringBoot app? Name from the code will take precedence and any mappings created for the previous name will be ignored.

web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
```diff
@@ -217,5 +319,28 @@ private FilterAndTelemetryClientMock createInitializedFilterWithMockTelemetryCli
 
         return new FilterAndTelemetryClientMock(filter, mockTelemetryClient);
     }
+
+
+    /*For Parametrized constructor testing*/
+    private FilterAndTelemetryClientMock createInitializedParametrizedFilterWithMockTelemetryClient(boolean withTelemetryClient, boolean clientThrows) throws ServletException, NoSuchFieldException, IllegalAccessException {
+        Filter filter = createParametrizedInitializedFilter("My-App");
```
Dmitry-Matveev: There is no validation that "My-app" has actually made it though and changed the way the code works.
You may want to add a test that checks my-app assignment or tests a change in the execution flow caused by the name being pre-assigned.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: @SergeyKanzhelev I just looked into the code and it seems that we cannot make the applicationName final because if we do so it has to be initialized in the constructor itself (either through parameterized constructor or default constructor) and we do not have a way to do it with default constructor because with non spring boot application the name of the application is caught from the servlet context and in our implementation it is retrieved from getName() method which parses the context and is invoked post constructor call. Let me know.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: @Dmitry-Matveev I believe that the name from the code will take precedence as the getName() method will return after executing the if statement as soon as it sees the name is set externally.

Dmitry-Matveev: @dhaval24 , Thank for looking into this! Then the question is - will changing name right in the code affect the behavior of the SDK in some unpredictable way?
For example, if name from the code takes precedence, could it be that some configuration settings are not picked up (simply because they are mapped to the name used in the configuration but not in code)? If no such or similar case is possible - we should be OK with this change.

dhaval24: @Dmitry-Matveev honestly, I am not quite sure if this can affect the SDK in some unpredictable way. Do you have any ideas how I can check this?

dhaval24:
@SergeyKanzhelev @Dmitry-Matveev In the recent commits, I have removed the public constructor as it could have possible consequences of breaking other components like java agent, because whenever you set the name of the web-app it also gets the same name registered with the Java Agent and this only happens initially. Now if we expose public constructor, then someone might change the name of the application in the middle of the execution and in this case Java Agent might get unhooked. 

There were couple of work-arounds that I had:
1. Whenever this constructor is called, set the name in agent too (kind of hacky)
2. Expose public method instead of constructor (still public) and a check before assigning the name- would free us from our end but customers would have false impression.
3. Private method which reads from the properties file (thought would be the best)

However, one caveat - this is still a patch, a full fleged solution would be to like having another config module similar to what we have for XML, though its worth while to investigate if we need those efforts at all. 

Please provide your suggestions.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -206,24 +213,28 @@ private void setKeyOnTLS(String key) {
     public WebRequestTrackingFilter() {
     }
 
+    /**
+     * Parameterized constructor, used to set Application name explicitly
+     * @param applicationName
+     */
+    public WebRequestTrackingFilter(String applicationName) {
```
Dmitry-Matveev: It looks like public constructor is still present - maybe merge issue?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -284,6 +301,38 @@ private String getName(ServletContext context) {
...
+        String name = null;
+        try {
+            String fileName = "application.properties";
+            input = WebRequestTrackingFilter.class.getClassLoader().getResourceAsStream(fileName);
+            if (input == null) {
+                return null;
+            }
+
+            properties.load(input);
+            name = properties.getProperty("spring.application.name");
```
Dmitry-Matveev: Is it the way we're going to use for spring only or the name should be more generic?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -284,6 +301,38 @@ private String getName(ServletContext context) {
         return name;
     }
 
+    /*This method is used to parse application name from the properties file*/
+    private String getApplicatioNameFromProperties() {
```
Dmitry-Matveev: We might want to add a unit test for this new behavior.

dhaval24: @Dmitry-Matveev I agree that there should be test to test this behavior. However the WebRequestTrackingFilter does not implement any interface that we can mock using Mockito framework. The only way that we can write a testcase for this using JUnit can be actually reading from the file and doing the verification. Now that comes with a caveat. We need to store that file into our source code somewhere. If you have any other ideas on testing let me know.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: @Dmitry-Matveev yes as of now I am not aware of any other framework that reads the value from the properties file though this is through a very preliminary research. There can be places where this is the only way but not in my knowledge. For all other cases we are already reading from application context. Also just pushed the corrected version where the public constructor has been removed which was accidentally present in the previous commit.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: You can make `WebRequestTrackingFilter` implement an interface if it's required for tests since you control the code. Storing file within test project for test purposes is perfectly fine - you can consider such a file a 'test resource'.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: This property is going to be our public interface to configure application name that something can take dependency on, so changing it once released is not an easy task.
I think renaming this property before release to a more generic scope rather than "spring" will make us stay on the safe side and prevent "deprecation"/rename issues in the future.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
gavlyukovskiy: @Dmitry-Matveev interface will not help here, mocking interface makes only interface methods available for mocking. Better to make class not final and Mockito will be able to mock it.

gavlyukovskiy:
Hard coding getting property `spring.application.name` from `application.properties` from the classpath it is not the proper way to go. According to section _24.3 Application property files_ of [Spring Boot Reference Guide](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html) not only `application.properties` can be used, but also `application.yaml` and located not only in the classpath, but also in application directory or `config` subdirectory. For example we use `config` subdirectory for deployment to set deployment specific properties.

I understand that `spring.application.name` is not good candidate for such overriding, but at least case with `application.yaml` is very popular. 
I think the most reliable way is to give ability to the user to define this property using constructor, assuming that Spring Boot users will make something like that:
```
@Bean
public FilterRegistrationBean webRequestTrackingFilter(@Value("${spring.application.name}") String applicationName) {
    FilterRegistrationBean registration = new FilterRegistrationBean();
    registration.setFilter(new WebRequestTrackingFilter(applicationName));
    registration.addUrlPatterns("/*");
    registration.setName("webRequestTrackingFilter");
    registration.setOrder(1);
    return registration;
}
```

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
Dmitry-Matveev: @gavlyukovskiy, makes sense. @dhaval24, could you please take a look at this approach?

Dmitry-Matveev:
@gavlyukovskiy, I remember @dhaval24 was trying to address this with the constructor but as far as I remember the continuation on this, the things didn't play well down the road for some reason. @dhaval24 would have more details to share.

dhaval24:
@gavlyukovskiy I wast taking the exact same approach as you suggested in the first place however after some digging into the code I found that. we use the same name to bind Java Agent also. As I expressed in the previous comments on this PR  there is possible risk of de-binding the agent from SDK because the person might be able to change  the name of the field by calling the public constructor with argument. However, if I think deeply on this the constructor is only called when the user intents to get a new instance and hence might be okay.

However, the issue with this approach is that we might expose a public constructor which is not required at all. Also, when I was dealing with this I had an opinion that we might need a separate module to read the configurations from the properties file in the same way as we do for XML because I believe that there are several spring boot applications that are standalone only and doesn't have War files, these applications don't have any xml files in theory so it would be best if we can read all the config from properties file but then we had a discussion that is it worth the effort and therefore we haven't proceed on this yet. Let me know if you have any better ideas other than public constructor (may be addressing wider problems relating to spring boot)

gavlyukovskiy:
Reading configuration from properties will not help much for Spring Boot users. It is not just read application.properties because of 
1. Config location  - `file:./config/`, `file:./`, `classpath:/config/`, `classpath:/` by default and can be overridden by using parameter `--spring.config.location=classpath:/myprops.properties` when running jar/war
2. Different property types  - `.properties`, `.yaml` 
3. [Profiles](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html) - you may have `application-prod.properties` and `application-dev.properties` 
4. [Relaxed binding](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding) - property `spring.application.name` can be read from system variables `SPRING_APPLICATION_NAME`

The only way to handle this is using Spring classes that are aware of those type of configuration - `Environment.getProperty("property")` (no relaxed binding support), `@Value("${property}")` (no relaxed binding support), `@ConfigurationProperties`. 
Latter beside relaxed binding support also a type safe and has [metadata support](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html), that is used in IDE's to autocomplete properties and show their description.

If you want to support Spring Boot I think the better way is to add some property map (or tags) in your `TelemetryConfiguration` to use in your core project and a way to configure it from outside (while configuring Spring application context). Then create [spring boot starter](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-auto-configuration.html) module for Application Insights, say `application-insights-spring-boot-starter`. In documentation there are suggestion to create two different modules _auto-configuration_ and _starter_ but it makes sense only if you have more than 1 starter with different libraries setup. In this starter can configure `TelemetryConfiguration` and set all external properties you need using spring configuration.

If you want to look at starter examples you may refer to [3rd-party spring boot starters page](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-starters#starters) or to [my project](https://github.com/gavlyukovskiy/spring-boot-data-source-decorator) with few starters for data source decorating libraries.

We in Cloudyn (Azure Cost Management) are interested in integration with Application Insights as all our applications are spring boot we anyhow will create some kind of a starter for our needs. If spring boot is in your priorities we can cooperate on this, my Microsoft mail is v-arhavl@microsoft.com.

dhaval24:
@gavlyukovskiy This makes much more sense to me now. Totally agree with your point that reading from properties won't suffice. Thanks for your interest in using Application Insights and definitely SpringBoot is one of the most popular web frameworks in Java community and it makes sense to have first class support for SpringBoot. I will start an email thread with you.

dhaval24:
Closing this PR as public constructor is now introduced to support this as part of #449 PR Merge

dhaval24:
#461 is still active and work will be in progress on that.

PR #415, opened 2017-08-24
Dhaval/revert set sample rate null check

dhaval24:
Fixing RequestTelemetry sending bug with SchemaV2. Re-introduced null check in setSampleRate() and fixed the write method.

- [x] CHANGELOG.md updated

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java
@@ -139,11 +139,8 @@ public class Envelope
     /**
      * Sets the SampleRate property.
      */
-    public void setSampleRate(Double value) {
-        if (value != null) {
+    public void setSampleRate(double value) {
             this.sampleRate = value;
-        }
-
     }
     
     /**
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java
@@ -139,8 +139,11 @@ public class Envelope
     /**
      * Sets the SampleRate property.
      */
-    public void setSampleRate(double value) {
-        this.sampleRate = value;
+    public void setSampleRate(Double value) {
+        if (value != null) {
+            this.sampleRate = value;
+        }
+
     }
     
     /**
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseSampleSourceTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseSampleSourceTelemetry.java
@@ -17,6 +17,8 @@ public abstract class BaseSampleSourceTelemetry<T extends Domain> extends BaseTe
     @Override
     protected void setSampleRate(Envelope envelope) {
         Double currentSP = getSamplingPercentage();
-        envelope.setSampleRate(currentSP);
+        if (currentSP != null) {
+            envelope.setSampleRate(currentSP);
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java
```diff
@@ -139,8 +139,11 @@ public double getSampleRate() {
     /**
      * Sets the SampleRate property.
      */
-    public void setSampleRate(double value) {
-        this.sampleRate = value;
+    public void setSampleRate(Double value) {
```
SergeyKanzhelev: this file is autogenerated. Instead of changing this file - can you please fix the code that attempts to set this to `null` or resulting code that fails because of unexpected value.

Alternatively you'd need to modify code generator, but I think it's better to make this code works with `null` value

dhaval24: Sure, I can see and fix the code which tries to set this to null as you mentioned this file is autogenerated.

dhaval24: @SergeyKanzhelev Updated Envelope.java to remove null check and add it to the place where it is being called in BaseSampleSourceTelemetry.java. Hope this works

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseSampleSourceTelemetry.java
```diff
@@ -17,6 +17,8 @@ public void reset() {
     @Override
     protected void setSampleRate(Envelope envelope) {
         Double currentSP = getSamplingPercentage();
-        envelope.setSampleRate(currentSP);
+        if (currentSP != null) {
```
SergeyKanzhelev: so how the `reset` worked before? It calls this method and passes `null`. Perhaps it should be a fix in serialization and null is a `valid` value.

dhaval24: @SergeyKanzhelev getSamolingPercentage() returns null and null gets assigned to currentSP. Then the call to envelope.setSampleRate(currentSP) sets null to the sampleRate of the envelope overriding its default value of 100.00D. That was the flow before this change and therefore I added the null check to see if the currentSP has null value then don't override the default.

SergeyKanzhelev: I understand that flow wouldn't change as the only thing you did is moved the check one level up. My question is if one called `reset` on something that was set to non-default value - is it expected that reset will simply do nothing? I'd expect that it will at least set the value to 100

dhaval24: I got your question now. I checked the implementation of reset and right now it tries to pass null and null gets set. I don't think so it is setting it back to 100 though I may be wrong

public void reset() {
        this.setSamplingPercentage((Double)null);
    }

dhaval24: @SergeyKanzhelev upon closely looking at the code I found following flow:
1. setSampleRate() and setSamplePercentage() are two methods.
2. User has the permission to set sampling percentage which in-turn is internally set to sampling rate. 
3. when the telemetry.reset() is called it sets the sampling rate as null and next time when the telemtry is pushed, it sees that sampling rate is null and picks the default sampling rate as 100.0D

Therefore, I believe there should be no need of change though I might be wrong. Please let me know what are your thoughts.

PR #420, opened 2017-09-07
Delay retrying to send telemetry if an exception is hit

leantk:
I originally made changes to delay retries on specific exceptions but users are hitting this for more reasons such as working offline, working through a proxy, etc. They see >500 requests in less than a minute in these cases which have taken down their proxy. I don't know any reason why there shouldn't be backoff if there is an exception in general instead of constantly retrying. I made the change to delay no matter the exception. I tested on my personal machine and I stopped seeing the constant retries every second.


Dmitry-Matveev:
Discussed offline, approving PR but waiting for [this one to complete first](https://github.com/Microsoft/ApplicationInsights-Java/pull/405) (may require small merge commit).

Dmitry-Matveev:
This PR covers the case for the `IllegalStateException` as well, so merging this one first. We'll continue discussion with @andresol in #405 .
@leantk, could you please add changelog.md entry for this change?

leantk:
@Dmitry-Matveev I made the changelog update. If you need me to do anything else let me know.

Issue #403, opened 2017-07-27
Exceeding property length invalidates custom event


chadequate:
In the .Net implementation when I would set the value of a custom event property to a string that was longer than the max allowed, the message would simply be truncated and event would still post. In Java the event never posts. Not sure which way is correct, but the inconsistency threw me off.

dhaval24:
Closing issue. #421 addressed this.

PR #421, opened 2017-09-10
Dhdoshi/sanitize json

### LINKED ISSUE #401: value of custom property should be sanitized
Setting custom property to any value that conflicts with JSON will break serialization - invalid JSON will be generated

dhaval24:
Fix #401 

Introduced a class SanitizationUtils.java containing method to sanitize string with escape characters to correctly create JSON.

Introduced SanitizationUtilsTest class containing methods to test SanitizationUtils.java

Modified the write(...) method in JsonTelemetryDataSerializer to write the sanitized string

- [x] CHANGELOG.md updated


core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     * JSON Sanitization failure
+     * @param text
+     * @return Sanitized String suitable for JSON
+     */
+    public static String sanitizeStringForJSON(String text) {
+
+        final StringBuilder result = new StringBuilder();
+        StringCharacterIterator iterator = new StringCharacterIterator(text);
+        for (char curr = iterator.current(); curr != iterator.DONE; curr = iterator.next()) {
+            if( curr == '\"' ){
```
grlima: Since you always append the same thing ("\\") to the char, I don't think you need those if statements... perhaps it'd be cleaner to have all chars which need escaping in a list/map and have one if.

dhaval24: @grlima agree with your point. For a while I had thought the same. However, I thought that since there are finite number of if statements (within limits ), I believe creating a Map will be an overhead that can impact when the size of string is really large. Querying a hashmap means calculating hashcode and performing the lookup which can be really problematic if there are too many queries and ofcourse initializing an object type has substantial more overhead. This method is already O(n) in both space and size and for the performance perspective would be better if we can keep if-else as they are more preliminary statements. Let me know your thoughts.

core/src/test/java/com/microsoft/applicationinsights/common/SanitizationUtilsTests.java
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.applicationinsights.common;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class SanitizationUtilsTests {
+
+    @Test
+    public void testSanitizeStringForJSON() {
+        String result = SanitizationUtils.sanitizeStringForJSON("{\"entityType\":\"CONTACTNAME\",\"contactField\":\"\",\"query\":\"zuck\",\"candidates\":[{\"id\":\"2\",\"name\":\"Andrew Zuck\",\"aliases\":[],\"phoneNumbers\":[{\"type\":\"MOBILENUMBER\",\"number\":\"+1 (555) 222-2222\"}],\"emailAddresses\":[],\"streetAddresses\":[]}],\"scores\":{\"2\":2.327993297520463}}");
```
Dmitry-Matveev: This test does not seem to validate `\f`, `\b`, `\n`, `\r`, `\t`, `/`, `\` cases - please add another test case(s) to cover for this.

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
Dmitry-Matveev: Do we need to take care of the escaped single quote as well? `\'`

dhaval24: @Dmitry-Matveev added \' escape character and also added the test case.

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.text.StringCharacterIterator;
+
+/**
+ * Created by dhdoshi on 09/10/2017
+ *
+ * This class provides utility functions to sanitize strings
+ * for various formats. Currently it supports JSON sanitization
+ */
+public final class SanitizationUtils {
```
SergeyKanzhelev: why you want it to be public?

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * This method appends escape characters to input String to prevent
+     * JSON Sanitization failure
+     * @param text
+     * @return Sanitized String suitable for JSON
+     */
+    public static String sanitizeStringForJSON(String text) {
```
SergeyKanzhelev: I suggest to pass the max length as an additional argument so you do not need to sanitize excessively long string

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
```diff
@@ -0,0 +1,57 @@
...
+                result.append("\\n");
+            }
+            else if(curr == '\r'){
+                result.append("\\r");
+            }
+            else if(curr == '\t'){
+                result.append("\\t");
+            }
+            else {
+                result.append(curr);
```
SergeyKanzhelev: check out .NET imoplementation. There are plenty of UNICODE control characters you'd want to sanitize as well

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
dhaval24: @SergeyKanzhelev I addressed this in my current commits. Please have a look. This solves #403 also.

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
dhaval24: @SergeyKanzhelev I think we can't make a class private and I believed it would be best to create a class for sanitization and put all related methods there. In the future we might need to sanitize for eg URL or something else we can put this there. It also helps in better organization for testing purpose too.

SergeyKanzhelev: why you cannot make class internal? What' the idea of exposing this API publicly?

dhaval24: @SergeyKanzhelev I don't purposefully intend to make an API public. I totally agree that it can added unnecessary constraint of maintaining it as people can take dependency over it. In my previous comment I did not meant that we can't create inner classes in Java. I can well make this class as an inner class inside JsonTelemetryDataSerializer / even down just create a private method there itself. My thoughts were that it might not be the best place as this method may have utility in other parts of the SDK (though not sure) and in that case we might not have the leverage to reuse the same method. Further, putting it inside JsonTelemetryDataSerializer might make it difficult to test as its a final class and doesn't implement and interface. That means we can't mock the behavior of it with Mockito framework and hence the only option would be to create the actual object of the larger class with more external dependencies just for the mere purpose of testing. PS: this are just my thoughts from very limited experience and I can be totally wrong. Please provide your guidance on what you think would be the best for this scenario.

SergeyKanzhelev: You can make it package-private by removing the `public` keyword. This way you will have minimal code change and easy to read code without exposing these helpers to the world.

core/src/main/java/com/microsoft/applicationinsights/common/SanitizationUtils.java
dhaval24: @SergeyKanzhelev Probably I found a potential issue with this. Lets think of a situation when we get an excessively long string almost around max length which has large number of escape characters which will need to be sanitized. Now we truncate it before sanitization but post sanitization the string will again be larger so breeze will respond with 500 if we send that. Quickfix: Truncate before and after sanitization. 
Corner Case: post sanitize truncate can remove some of the sanitized characters and cause the same problem.
Concrete Solution: before sanitization count the occurnace of the characters to be handeled and truncate to that length so sanitiaztion does not exceed max length.
Please let me know your opinion!

dhaval24:
@SergeyKanzhelev @Dmitry-Matveev addressed all concerns. Can you please review so that it is ready for merging.

dhaval24:
@SergeyKanzhelev can you please review the changes?

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # CHANGELOG
 
 ## Version 1.0.10
+- Fixed Exceeded property length leads to lost event.
+- Fixed issue of sending custom event.
+- Method `sanitizeStringForJSON` takes string input and converts it to JSON friendly string.
```
SergeyKanzhelev: this method is not public. Rewrite it as of "sanitization is now working"

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
...
+- Class `com.microsoft.applicationinsights.telemetry.SanitizationUtils` has methods to sanitize string for JSON.
```
SergeyKanzhelev: not a public class. No need to mention it

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
...
+- Fixed Request Telemetry Sending bug with new schema.
```
SergeyKanzhelev: when Sending starts with upper case letter? More details would be benefitial (maybe at least issue number)

Issue #426, opened 2017-09-14
All the fields might not have been truncated resulting in possible endpoint rejection


dhaval24:
In JsonTelemetryDataSerializer the SDK probably doesn't truncate all the fields with the max size that endpoint accepts, might possibly result in the endpoint rejection of the telemetry. 

One of the upcoming PR addresses this issue for the Custom Event, however the issue might be widespread across other fields also and might be other SDK's too.

dhaval24:
#429 is related

dhaval24:
This issue has been addressed

PR #431, opened 2017-09-20
Reorganize sanitization logic (Please review carefully)

### LINKED ISSUE #429: The current Sanitization and truncation logic does not adhere to new schema
The current Sanitization logic placed in Sanitizer.java does not adhere to the new schema (has old specifications for Maximum length allowed of various fields) 

The new schema has specific lengths for various kinds of telemetry whereas the current has a common length for name. Sanitization is based on regex which is not correct, needs to be done according to json specifications. 

Solution: Move all the sanitization logic to JsonTelemetryDataSerializer class where the actual json string is created. Update the Bond Definitions to automatically generate write methods with parameters.

dhaval24:
Fix #429 #426 
.
This pull request moves all the sanitization logic to JsonTelemetryDataSerialiazer, updates bond definitions to generate write methods with limits according to our new schema. 

This pull request also obsoletes many classes in SchemaV2 package which are now no longer in use.
Obsoletes most of the methods of Sanitizer class, disables the pre-sanitization in track() method of TelemetryClient and obsoletes sanitize() method in Telemetry interface and all the classes which implement it.

This pull requests also adds getter and setter for OperationParentId. 

There are many tests also which are updated to incorporate these changes. New tests have been added to JsonTelemetryDataSerializerTest class to check serialization and length limits.

Note: There are lot of changes please review it carefully before we can merge this! I would suggest create a local build from this branch and test on your Java applications if possible. Also provide advice on if there is need for writing more test cases. 

- [x] CHANGELOG.md updated


core/src/main/java/com/microsoft/applicationinsights/internal/util/Sanitizer.java
```diff
@@ -168,7 +187,57 @@ private static String MakeKeyNonEmpty(String key) {
         return key;
     }
 
+    @Obsolete
```
Dmitry-Matveev: Do we need to mark private methods as deprecated? Is it for better internal usage and API tracking?

core/src/main/java/com/microsoft/applicationinsights/internal/util/Sanitizer.java
```diff
@@ -168,7 +187,57 @@ private static String MakeKeyNonEmpty(String key) {
...
     private static String truncate(String value, int maxLength) {
         return value.length() > maxLength ? value.substring(0, maxLength) : value;
     }
+
+    @Obsolete
```
Dmitry-Matveev: That's an agile development, we add deprecated methods :)

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
```diff
@@ -320,35 +335,76 @@ private void writeName(String name) throws IOException {
...
-                case '\t':
-                    out.write("\\t");
-                    break;
-                default:
-                    out.write(c);
+    private String sanitizeStringForJSON(String text, int maxLength) {
+
+        final StringBuilder result = new StringBuilder();
+        StringCharacterIterator iterator = new StringCharacterIterator(text);
+        for (char curr = iterator.current(); curr != iterator.DONE && result.length() < maxLength - 2; curr = iterator.next()) {
```
Dmitry-Matveev: Minor: 2 looks like a magic number here.

core/src/test/java/com/microsoft/applicationinsights/telemetry/SanitizerTest.java
```diff
@@ -21,150 +21,145 @@
...
-    }
-
-    @Test
-    public void testNonValidBadFormat2UUID() {
-        boolean valid = Sanitizer.isUUID("c934153105ac4d8c972e36e97601d5ffc934153105ac4d8c972e36e97601d5ff");
-        assertFalse(valid);
-    }
-
-    @Test
-    public void testValidUUIDWithComma() {
```
Dmitry-Matveev: Is this and all other tests are now covered? I see several things being addressed in the new tests like JSON special characters and length, but not sure if all of those test cases are still addressed in the rest of the tests. If we have code that relies on any of the validated checks, we may want to keep similar tests around.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -416,11 +416,11 @@ public void track(Telemetry telemetry) {
             throw new IllegalArgumentException("Instrumentation key cannot be undefined.");
         }
 
-        try {
-            telemetry.sanitize();
-        } catch (Throwable t) {
-            InternalLogger.INSTANCE.error("Exception while sanitizing telemetry: '%s'",t.getMessage());
-        }
+//        try {
```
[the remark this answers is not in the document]
dhaval24: @Dmitry-Matveev sure I purposefully kept the comments for now, once everything is fine I will remove off the comments . Sometimes its just more handy to comment instead of moving back to older commit using git :-p

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Application.java
```diff
@@ -32,10 +32,12 @@
 import com.microsoft.applicationinsights.telemetry.Duration;
 import com.microsoft.applicationinsights.telemetry.JsonTelemetryDataSerializer;
 import com.google.common.base.Preconditions;
+import org.apache.http.annotation.Obsolete;
 
 /**
  * Data contract class Application.
  */
+@Obsolete
```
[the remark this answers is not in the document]
dhaval24: @Dmitry-Matveev on briefly reading this documentation it seems like using @Deprecated notation makes the compiler to generate warning when someone tries to use it and also makes in javadoc itself, so wouldn't it suffice to replace with just @Deprecated tag? Would you like me to replace all the @Obsolete tags in this PR with it?

core/src/main/java/com/microsoft/applicationinsights/internal/util/Sanitizer.java
dhaval24: I have marked private methods in this class as obsolete because public methods of this class use those private methods, so kind of I cannot remove it but we are no longer using it. Makes sense?

core/src/main/java/com/microsoft/applicationinsights/internal/util/Sanitizer.java
dhaval24: The reason it is still there is because if in turn someone uses this classes public method which in theory they shouldn't this will atleast makes things right. If you want I can modify those public methods and remove the usage of this method and then remove this method totally. Let me know on that.

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
dhaval24: Yes 2 is a magic number. This is to ensure safety because you need that two additional character space if the last letter is an escape letter. Any better ways to address this?

core/src/test/java/com/microsoft/applicationinsights/telemetry/SanitizerTest.java
dhaval24: @Dmitry-Matveev I tried to cover all the scenarios which were addressed by the new PR. The rest of the commented tests in this class are related to UUID which we no longer use (Atleast that is what I know).

dhaval24:
@Dmitry-Matveev I have made the changes requested by you to update the @Obsolete tag to @Deprecated tag and also removed the comments present in the code except for sanitization test class as you mentioned we might have to take some of their test cases somewhere. Please have a look

core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/AvailabilityData.java
```diff
@@ -233,16 +233,16 @@ protected void serializeContent(JsonTelemetryDataSerializer writer) throws IOExc
         super.serializeContent(writer);
         writer.write("ver", ver);
         
-        writer.write("id", id);
+        writer.write("id", id, 64, true);
```
SergeyKanzhelev: Alternatively to `true` or `false` you can have two methods - one `write` and another `writeRequired`. This way you save one runtime check and code a bit more readable

dhaval24: @SergeyKanzhelev done. Thanks for the input!

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
Dmitry-Matveev: You can make it constant with the name explaining what it is.

core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java
```diff
@@ -28,6 +28,7 @@
 
 import com.google.common.base.Strings;
 import com.microsoft.applicationinsights.internal.util.Sanitizer;
+import org.apache.http.annotation.Obsolete;
```
Dmitry-Matveev: We do not need those imports anymore across all files - you switched everything to `@Deprecated`.

core/src/test/java/com/microsoft/applicationinsights/telemetry/SanitizerTest.java
Dmitry-Matveev: We removed non only UUID tests here but also tests that validate Empty case, Null case, All Valid case and so on and it looks like we replaced them with the single validation of the "JSON needs sanitization" case. We might need to bring some of these tests back (not physically but logically).

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -186,9 +186,6 @@ public String getInstrumentationKey() {
      * @throws IllegalArgumentException when the new value is null or empty
      */
     public void setInstrumentationKey(String key) {
-        if (!Sanitizer.isUUID(key)) {
```
Dmitry-Matveev: Why don't we need to validate that IKey is GUID anymore? Is it new approach across all SDKs or this removal was unintentional?

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
dhaval24: Yes I can make it constant. Will change it.

core/src/test/java/com/microsoft/applicationinsights/telemetry/SanitizerTest.java
dhaval24: so for the non empty cases, we have introduced different writer methods write () and writerReq(). Now if writeReq() is called without the required field we populate it with default ones.  I will think if I can add test to check default scenario. Lets discuss on Monday what all other tests can be added in this.

core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java
dhaval24: @Dmitry-Matveev by all the files do you mean across whole SDK we should remove @Obsoelete? Because there are many files that I haven't touched at all and they contain @Obsolete annotations. I will optimize all imports for the file  I touched so that there are no unused import.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
dhaval24: @Dmitry-Matveev this was not unintentional. When I talked with Sergey he told that me about this. I don't remember the exact reason but he told we use some different format too.

dhaval24: @Dmitry-Matveev I have removed unused imports from most of the places, removed obsolete and replaced with Depricated at maximum locations and added couple of test cases which logically cover maximum scenarios. Please review.

dhaval24:
@Dmitry-Matveev I have tried to addressed all possible issues for this PR. Can you please have a look at it so that if possible we can merge this too and have one more CTI pass before final release.

PR #437, opened 2017-10-04
 Adding quotes around security level write method to fix serialization

### LINKED ISSUE #436: TraceTelemetry w/Severity set is no longer appearing in Insights
I picked up Version 1.0.10 yesterday.  After doing so, all my trace lines with severities set stopped writing to application insights.

ie:
```
TraceTelemetry traceTelemetry = new TraceTelemetry("hello", SeverityLevel.Information);
 OperationContext operation = traceTelemetry.getContext().getOperation();
 operation.setId("MyTransactionId");
 telemetry.trackTrace(traceTelemetry);
```

The above stopped tracing with 1.0.10

while if I do the same thing, but set the SeverityLevel to null, it works:
```
TraceTelemetry traceTelemetry = new TraceTelemetry("hello", null);
 OperationContext operation = traceTelemetry.getContext().getOperation();
 operation.setId("MyTransactionId");
 telemetry.trackTrace(traceTelemetry);
```

I rolled back to 1.0.9 because of this defect.

dhaval24:
Fix #436 

This was a sanitization issue. Adding quotes around write(String name, com.microsoft.applicationinsights.internal.schemav2.SeverityLevel value) method fixed it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Version 1.0.11
+- Fixed issue #436 (TraceTelemetry with Severity is not shown in UI)
+
 ## Version 1.0.10
 - `track()` method of 'com.microsoft.applicationinsights.TelemetryClient' is now modified. No longer performing pre-sanitization
 - All Sanitization will now occur in `com.microsoft.applicationinsights.telemetry.JsonTelemetryDataSerializer` class. 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Version 1.0.11
-- Fixed issue #436 (TraceTelemetry with Severity is not shown in UI)
+- Fixed issue #436 (TraceTelemetry with Severity is not shown in UI). This fixes a regression issue with `TelemetryClient.trackTrace` and `TelemetryClient.trackException`.
 
 ## Version 1.0.10
 - `track()` method of 'com.microsoft.applicationinsights.TelemetryClient' is now modified. No longer performing pre-sanitization
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
@@ -95,7 +95,9 @@ public final class JsonTelemetryDataSerializer {
     public void write(String name, com.microsoft.applicationinsights.internal.schemav2.SeverityLevel value) throws IOException {
         if (value != null) {
             writeName(name);
+            out.write(JSON_COMMA);
             out.write(String.valueOf(value));
+            out.write(JSON_COMMA);
             separator = JSON_SEPARATOR;
         }
     }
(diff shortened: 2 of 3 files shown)


dhaval24:
@littleaj as discussed please add a test case which you have already created around the issue in this branch. Thanks.

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
```diff
@@ -95,7 +95,9 @@ public void write(String name, int value) throws IOException {
     public void write(String name, com.microsoft.applicationinsights.internal.schemav2.SeverityLevel value) throws IOException {
         if (value != null) {
             writeName(name);
+            out.write(JSON_COMMA);
```
Dmitry-Matveev: Do you need a changelog item for this?

dhaval24: @Dmitry-Matveev Yes we would need changelog. They are public methods so I think so yes.

littleaj:
@Dmitry-Matveev @dhaval24 I added a test for this scenario.

core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java
littleaj: @Dmitry-Matveev @dhaval24 added changelog

dhaval24:
@littleaj This looks good to me.

core/src/test/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializerTest.java
```diff
@@ -275,18 +282,27 @@ public void test() throws IOException {
...
         stubClass.getM1().put("key2", 6);
         stubClass.getM1().put("key3", 7);
+        stubClass.setSeverity(SeverityLevel.Critical);
         StringWriter stringWriter = new StringWriter();
         JsonTelemetryDataSerializer tested = new JsonTelemetryDataSerializer(stringWriter);
         stubClass.serialize(tested);
         tested.close();
         String str = stringWriter.toString();
-        StubClass bac = new Gson().fromJson(str, StubClass.class);
+        System.out.println("serailized=\n"+str);
```
dhaval24: Do we need this print statement here? I believe it will output on console which everyone (other devs) would not be aware of without looking at the code.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## Version 1.0.11
```
dhaval24: @littleaj as we changed the public facing method do we need to provide description of what has been changed in the implementation of the method? Rest looks fine.

core/src/test/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializerTest.java
littleaj: Gradle suppresses sysout by default, but I can take this out.

CHANGELOG.md
littleaj: I can modify this to reflect that this fixes a regression in trackTrace/trackException

PR #445, opened 2017-10-09
removing escape for single quote, not needed for Json sanitization

### LINKED ISSUE #443: Sanitize NULL values in strings
Certain systems can return null values in the exception message. This can cause problems when sending the telemetry and cause it not to be captured by Application Insights.

dhaval24:
Fix #443 

Updating the sanitization logic. There is no need to sanitize single quotes according to JSON Standards. 
(https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringEscapeUtils.html#escapeJson-java.lang.String-)

Further, I am not sure what all types of languages and character schema we support. However, this sanitization logic matches the dot net logic in most possible ways.


dhaval24:
@Dmitry-Matveev does this look good? Can we merge so master is fully functional?

PR #446, opened 2017-10-10
Removing the call for empty method et.setExceptionHandledAt() and ass…

dhaval24:
Removing the calls to empty method setExceptionHandledAt() and deprecating its associated enum.  Also removed associated phantom test which cleans up the build too (no warnings about tests of deprecated methods).

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionHandledAt.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionHandledAt.java
@@ -25,6 +25,7 @@ package com.microsoft.applicationinsights.telemetry;
  * This enumeration is used by ExceptionTelemetry to identify if and where exception was handled.
  *
  * This Enum is now deprecated.
+ * @deprecated As of 1.0.11
  */
 @Deprecated
 public enum ExceptionHandledAt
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionHandledAt.java
```diff
@@ -23,7 +23,10 @@
 
 /**
  * This enumeration is used by ExceptionTelemetry to identify if and where exception was handled.
+ *
+ * This Enum is now deprecated.
  */
+@Deprecated
```
littleaj: Was this replaced by something else? If so, let's add the `@deprecated` javadoc field pointing to the replacement.

It may be good practice to include something like
```
/**
 * @deprecated As of 1.0.11
*/
```

dhaval24: Hi thanks for this suggestion. Last time when I checked this the docs said that there is no longer a need to write @deprecated tag for docs. It should be automatically generated thats why I left. But nonetheless I can add.

littleaj: I see; that should be fine. Please point me to where that is documented. I'm still figuring out our javadoc story.

dhaval24: May be I am not finding the right link where I saw that anyways added that. Will let you know if I hit across that.

Issue #418, opened 2017-08-29
java.lang.VerifyError (ByteCode Instrumentation Error) while collecting SQL dependencies 


dhaval24:
When collecting SQL dependency using Java Agent with Tomcat as web-server I hit following error:

java.lang.VerifyError: Expecting a stackmap frame at branch target 19
Exception Details:
  Location:
    oracle/jdbc/driver/OracleStatement.removeChild(Loracle/jdbc/driver/OracleStatement;)V @5: if_acmpne
  Reason:
    Expected stackmap frame at this location.
  Bytecode:
    0x0000000: 2b2a b401 6ba6 000e 2a2b b401 6fb5 016b
    0x0000010: a700 202a b401 6b4d 2cb4 016f 2ba5 000b
    0x0000020: 2cb4 016f 4da7 fff3 2c2b b401 6fb5 016f
    0x0000030: 2b01 b501 6d2b 01b5 016f b1            

	oracle.jdbc.driver.T4CDriverExtension.allocateStatement(T4CDriverExtension.java:56)
	oracle.jdbc.driver.PhysicalConnection.createStatement(PhysicalConnection.java:1507)
	oracle.jdbc.driver.PhysicalConnection.createStatement(PhysicalConnection.java:1478)
	com.Servlet.doPost(Servlet.java:43)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:660)
	javax.servlet.http.HttpServlet.service(HttpServlet.java:741)
	org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:53)
	com.microsoft.applicationinsights.web.internal.WebRequestTrackingFilter.doFilter(WebRequestTrackingFilter.java:81)

The possible reason is due to the stricter rules of byte code instrumentation induced in JDK 1.7+

One work around is to use the following flag in in the tomcat startup script:
**Xverify:none**

This flag suppress the byte code validation but might violate security.

dhaval24:
Here are the links which explain this issue in more detail:
http://blog.takipi.com/oracles-latest-java-8-update-broke-your-tools-how-did-it-happen/
https://dzone.com/articles/oracles-latest-java-8-update
http://chrononsystems.com/blog/java-7-design-flaw-leads-to-huge-backward-step-for-the-jvm

SergeyKanzhelev:
So did you fix the underlying problem of injecting an invalid code?

dhaval24:
[...]

Issue #365, opened 2017-04-07
Relocate all Web Jar dependencies


Dmitry-Matveev:
As per [this comment](https://github.com/Microsoft/ApplicationInsights-Java/commit/799a839b59a41f56f2247a4f68c3c1cb86e18ad9#commitcomment-21670121) in the source code, the current shadow Jar process does not relocate `javax.annotation` nor `org.objectweb` causing Security Exception like the following:
```
java.lang.SecurityException: class "org.objectweb.asm.MethodAdapter"'s signer information does not match signer information of other classes in the same package
at java.lang.ClassLoader.checkCerts(ClassLoader.java:898)
at java.lang.ClassLoader.preDefineClass(ClassLoader.java:668)
at java.lang.ClassLoader.defineClass(ClassLoader.java:761)
at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
```

debugthings:
#367 Related

Dmitry-Matveev:
Partially addressed in 1.0.8, to review in 1.0.9.

markpjohnson:
This is a high priority for my team. Is there any chance this will get resolved soon? We are awaiting several new features, but this issue is a blocker

dhaval24:
@markpjohnson thank you for notifying us your blocking issue. As this is a security issue and of High P. We will discuss this in our triage meeting. It would be really helpful if you can provide us with all the other features that you are looking for in your monitoring needs so that we can put them in our list and have appropriate milestones for them. If you don't mind letting us know what is the current monitoring solution you are using or is this more of experimental phase for your team?

dhaval24:
@Dmitry-Matveev if you have any idea that SDK is supposed to repackage every dependency that we use into the microsoft namespace. I believe we should do it to avoid any conflict with user jars in the project.

Dmitry-Matveev:
That was the intent all along and Core Jar was simply selected as the first one to implement shadow jar. We can now think of extending this solution to the other jars we ship if there are no issues uncovered while releasing shadow Core Jar for the last couple times.

dhaval24:
@Dmitry-Matveev as I look closely into the code this is about relocating some dependencies in the code/build.gradle which are already addressed in the code. The PR for bytecode fixes also relocates the jars for web and therefore we should be good to close this issue. Let me know.

Dmitry-Matveev:
[...]

Issue #286, opened 2016-07-01
SQL Server prepared statement setObject instrumentation error


sfthomas:
```
java.lang.VerifyError: StackMapTable format error: bad type array size in method com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.setObject(ILjava/lang/Object;)V
    at com.microsoft.sqlserver.jdbc.SQLServerConnection.prepareStatement(SQLServerConnection.java:2167)
    at com.microsoft.sqlserver.jdbc.SQLServerConnection.prepareStatement(SQLServerConnection.java:1853)
    at org.apache.tomcat.dbcp.dbcp2.DelegatingConnection.prepareStatement(DelegatingConnection.java:287)
    at org.apache.tomcat.dbcp.dbcp2.DelegatingConnection.prepareStatement(DelegatingConnection.java:287)
    at org.hibernate.engine.jdbc.internal.StatementPreparerImpl$5.doPrepare(StatementPreparerImpl.java:146)
    at org.hibernate.engine.jdbc.internal.StatementPreparerImpl$StatementPreparationTemplate.prepareStatement(StatementPreparerImpl.java:172)
    at org.hibernate.engine.jdbc.internal.StatementPreparerImpl.prepareQueryStatement(StatementPreparerImpl.java:148)
    at org.hibernate.loader.Loader.prepareQueryStatement(Loader.java:1927)
    at org.hibernate.loader.Loader.executeQueryStatement(Loader.java:1896)
    at org.hibernate.loader.Loader.executeQueryStatement(Loader.java:1874)
    at org.hibernate.loader.Loader.doQuery(Loader.java:919)
    at org.hibernate.loader.Loader.doQueryAndInitializeNonLazyCollections(Loader.java:336)
    at org.hibernate.loader.Loader.doList(Loader.java:2610)
    at org.hibernate.loader.Loader.doList(Loader.java:2593)
    at org.hibernate.loader.Loader.listIgnoreQueryCache(Loader.java:2422)
    at org.hibernate.loader.Loader.list(Loader.java:2417)
    at org.hibernate.loader.hql.QueryLoader.list(QueryLoader.java:501)
    at org.hibernate.hql.internal.ast.QueryTranslatorImpl.list(QueryTranslatorImpl.java:371)
    at org.hibernate.engine.query.spi.HQLQueryPlan.performList(HQLQueryPlan.java:216)
[...]

Issue #277, opened 2016-02-05
Agent instrumentation errors with JDBC PreparedStatement


jhsheets:
Here's the stacktrace:

```
org.jboss.weld.interceptor.proxy.InterceptorException: java.lang.VerifyError: StackMapTable format error: bad type array size in method com.microsoft.sqlserver.jdbc.SQLServerPreparedStatement.setTime(ILjava/sql/Time;)V
    at org.jboss.weld.interceptor.proxy.InterceptorInvocationContext.proceed(InterceptorInvocationContext.java:150) ~[weld-servlet.jar:2013-07-17 09:29]
    at org.mybatis.cdi.LocalTransactionInterceptor.invoke(LocalTransactionInterceptor.java:58) ~[mybatis-cdi-1.0.0-beta1.jar:1.0.0-beta1]
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.7.0_79]
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57) ~[na:1.7.0_79]
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.7.0_79]
    at java.lang.reflect.Method.invoke(Method.java:606) ~[na:1.7.0_79]
    at org.jboss.weld.interceptor.proxy.SimpleMethodInvocation.invoke(SimpleMethodInvocation.java:30) ~[weld-servlet.jar:2013-07-17 09:29]
    at org.jboss.weld.interceptor.chain.AbstractInterceptionChain.invokeNext(AbstractInterceptionChain.java:93) ~[weld-servlet.jar:2013-07-17 09:29]
    at org.jboss.weld.interceptor.chain.AbstractInterceptionChain.invokeNextInterceptor(AbstractInterceptionChain.java:78) ~[weld-servlet.jar:2013-07-17 09:2
```

gupele:
Hi,

Thank you for your input

The method signature has the index: "(ILjava/sql/Time;)V", (the 'I') which makes it difficult to understand. 
Is there a way in which you can narrow down the scenario?

jhsheets:
I noticed that the signature looks correct. I don't know much about
bytecode manipulation to know exactly what the issue is. There is an
overloaded setTime method that takes a Calendar object that you're not
instrumenting. Maybe that's the issue?

sfthomas:
[...]

PR #449, opened 2017-10-20
Byte code instrumentation fix

### LINKED ISSUE #428: Agent cannot capture oracle dependency calls to Insert statements
Agent currently cannot capture Oracle Insert statements through statement.exectueUpdate() method. Needs more validation on other methods also.

dhaval24:
Fix #428, #418, #286, #277, #276 , #365 

This pull request fixes the java.lang.Verify error which was being caused in JDK version 1.7 onwards because of improper instrumentation. Changes here facilitate the creation of StackMap frame (a datastructure introduced in JDK 1.7 and mandated in 1.8), fixes constructor instrumentation issue along with improper constructor signature for SQLServer.

#276 mentions issues with HTTPCLIENT 42, however on current testing things work fine for HTTPCLIENT 43 and since its more secure it is advisable to upgrade to this client.

This pull request also upgrades the ASM version to ASM 5.2 

- [x] CHANGELOG.md updated


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -153,6 +153,8 @@ public static String getAgentJarLocation() throws UnsupportedEncodingException {
         }
 
         String path = AgentImplementation.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        int index = path.lastIndexOf('/');
```
Dmitry-Matveev: Is it always in this notation across the platforms or sometimes can be '\\'?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriter.java
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        if ((class1.isInterface()) || (class2.isInterface())) {
+            return "java/lang/Object";
+        }
+
+        do {
+            class1 = class1.getSuperclass();
+        }
+        while (!(class1.isAssignableFrom(class2)));
```
Dmitry-Matveev: Any possible issue with endless cycle or exception here if class1 has no more superClass and class2 happens to be not assignable (e.g. null or something...)?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
```diff
@@ -243,7 +251,8 @@ private ClassVisitorFactory classFactoryForSqlServer() {
             @Override
             public ClassVisitor create(ClassInstrumentationData classInstrumentationData, ClassWriter classWriter) {
                 HashSet<String> ctorSignatures = new HashSet<String>();
-                ctorSignatures.add("(Lcom/microsoft/sqlserver/jdbc/SQLServerConnection;Ljava/lang/String;II)V");
+                ctorSignatures.add("(Lcom/microsoft/sqlserver/jdbc/SQLServerConnection;Ljava/lang/String;" +
+                        "IILcom/microsoft/sqlserver/jdbc/SQLServerStatementColumnEncryptionSetting;)V");
```
Dmitry-Matveev: Is it the same parameter across all SQL calls we'd need to monitor or specific to some version? The previous signature looked more generic.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultByteCodeTransformerTest.java
```diff
@@ -26,11 +26,11 @@
...
 
 public final class DefaultByteCodeTransformerTest {
-    @Test
-    public void noClassInstrumentationDataTest() {
-        DefaultByteCodeTransformer tested = new DefaultByteCodeTransformer(null, false);
-        byte[] mockArray = new byte[1];
-        byte[] result = tested.transform(mockArray, "mock");
-        assertSame(result, mockArray);
-    }
+//    @Test
```
Dmitry-Matveev: Should we delete the test if no longer relevant or uncomment and fix if it still is?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
dhaval24: @Dmitry-Matveev It's a good question. I have an idea for 2 platforms I have tested so far. Java 8 has by default class loader of URLClassLoader type so it won't hit this part of the class .Java 9 has default classloader called "app" which is not of the type URLClassLoader and hence hits this part and only works for this change. I am not sure about other JVM's though.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriter.java
dhaval24: @Dmitry-Matveev to be specific about this, this is the exact implementation as how ASM does by default. Only reason to override this method was because ASM uses Class.getClassLoader() method to load the current class and therefore it gets the ClassLoader which doesn't have all the required classes loaded in it. In this override I pass the ClassLoader from the Callee in order to avoid the lookup and have the classloader which has all the required classes loaded. Therefore I believe that since there is minimal change done from ASM implementation this possibility is very narrow. Is there a way we can test this? How can we know if there is a null being passed or something like that?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
dhaval24: So the default signature of the constructor for JDBC driver version 6.2 is the one I changed. With the previous signature we were missing the constructor instrumentation. Now with the older drivers it might well be possible that signature was one mentioned before but nonetheless I believe enterprise can upgrade to latest JDBC drivers.

dhaval24:
@Dmitry-Matveev  I have fixed the test for DefaultByteCodeTransformerTest and commented the tests which do not have any asserts. They just add to build time as they are not verifying anything for now.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -153,6 +153,8 @@ public static String getAgentJarLocation() throws UnsupportedEncodingException {
         }
 
         String path = AgentImplementation.class.getProtectionDomain().getCodeSource().getLocation().getPath();
```
SergeyKanzhelev: Use https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html instead of string?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -153,6 +153,8 @@ public static String getAgentJarLocation() throws UnsupportedEncodingException {
...
+        int index = path.lastIndexOf('/');
+        path = path.substring(0, index + 1);
```
SergeyKanzhelev: check that index is not `-1`. Just in case so code will not throw. Also see above - using Path instead of string will help

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriter.java
```diff
@@ -0,0 +1,48 @@
+package com.microsoft.applicationinsights.agent.internal.agent;
+
+
+import org.objectweb.asm.ClassWriter;
+
+public class CustomClassWriter extends ClassWriter {
```
SergeyKanzhelev: Please add comment. And make sure you need it to be `public`

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriter.java
```diff
@@ -0,0 +1,48 @@
...
+
+    ClassLoader classLoader;
+
+    public CustomClassWriter(int writerFlag, ClassLoader loader)
+    {
+        super(writerFlag);
+        this.classLoader = loader;
+    }
+
+    protected String getCommonSuperClass(String className1, String className2)
```
SergeyKanzhelev: Add comment for this method.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
```diff
@@ -28,10 +28,10 @@
 
 public final class ClassInstrumentationDataTest {
     private final static String MOCK_CLASS_NAME = "ClassName";
-    private final static String MOCK_METHOD = "Method";
-    private final static String MOCK_SIGNATURE_1 = "Signature1";
-    private final static String MOCK_SIGNATURE_2 = "Signature2";
-    private final static String MOCK_SIGNATURE_3 = "Signature3";
+//    private final static String MOCK_METHOD = "Method";
```
SergeyKanzhelev: remove instead of commenting. BTW, why commenting?

gradle/common-java.gradle
```diff
@@ -40,8 +40,8 @@ if (java6JreDir) {
         bootClasspath += "$archivePath;"
     }
     tasks.withType(JavaCompile) {
-        sourceCompatibility = 1.6
-        targetCompatibility = 1.6
+        sourceCompatibility = 1.8
```
SergeyKanzhelev: So you dropped 1.7 support?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -1,514 +1,520 @@
...
-                    explain.close();
-                } catch (SQLException e) {
-                }
-            }
-        }
-
-        return explainSB;
-    }
-}
+/*
```
SergeyKanzhelev: what changed in this file?

gradle/common-java.gradle
dhaval24: I made it so because sometimes in some situations it the 1.7 target code would not produce the desired exceptions because these changes were mandated from JDK 1.8. Also, we need to decide till what point do we need to support backward compatibility in the SDK. We might not be able to use all the new features of JDK 8 like automatic inference of types in <> operator, passing of functions using lambda expressions etc which is modern java.

dhaval24: If required no harm in reverting it back if things work fine. However at one point we need to make this decision.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
dhaval24: Commenting because this signature was used in tests which were relying on getDecisionMethod() however it turns out that there is no such method in ClassInstrumentationData class.

dhaval24: May be if there is an idea to get this tests running in future, that was the reason.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
dhaval24: @grlima to comment more on this. At-least my intuition here was that while sending telemetry for SqlServer there are only two arguments that come in the methoddata object whereas previously there was a strict condition which checked arguments should be more than or equal to 3

dhaval24:
@SergeyKanzhelev I have addressed most of the concerns including using the Paths class for locating the agent jar, adding comments and also making CustomClassWriter package private instead of public.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -136,24 +139,36 @@ private static void appendJarsToBootstrapClassLoader(Instrumentation inst) throw
...
             InternalAgentLogger.INSTANCE.logAlways(InternalAgentLogger.LoggingLevel.ERROR, "Error while trying to fetch Jar Location, Exception: " + throwable.getMessage());
         }
 
-        String path = AgentImplementation.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        return URLDecoder.decode(path, "UTF-8");
+        String stringPath = AgentImplementation.class.getProtectionDomain().getCodeSource().getLocation().getPath();
+        if (stringPath.charAt(0) == '/') {
+            stringPath = stringPath.substring(1);
+        }
+        Path path = Paths.get(stringPath);
```
littleaj: The `java.nio.*` classes were introduced in 1.7. It's my understanding we still want compatibility with 1.6. If that's still the case, this needs to be reworked. @grlima, can you confirm?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriter.java
littleaj: @dhaval24 To test this method, you could refactor it into a `public static` utility method which takes the ClassLoader along with the other parameters and write tests against that.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
littleaj: What happens if you uncomment these tests?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
littleaj: @dhaval24 Does this limit the supported driver versions?

dhaval24:
@SergeyKanzhelev @littleaj @Dmitry-Matveev @grlima I have pushed changes in this PR which are now addressing all the possible bytecode concerns we have seen till the date. As far as unit tests are concerned I still do not have a way to perform tests and our agent code does not have ones till the date, however this is a blocking issue and we need to get it fixed. I have done end to end testing in several test application and changes work fine.  The CustomClass Loader now performs the commonSuperClass lookup without actually loading the class. The PR also introduces changes in the Application-Insights web gradle file to include shadow jar plugin hence eliminating class path hell issues and also introduces a public constructor in the WebRequestTracking filter to address SpringBoot solutions as discussed.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -136,24 +135,28 @@ private static void appendJarsToBootstrapClassLoader(Instrumentation inst) throw
...
+                        InternalAgentLogger.INSTANCE.logAlways(InternalAgentLogger.LoggingLevel.INFO,"Agent jar found at " + urlPath);
                         int index = urlPath.lastIndexOf('/');
                         urlPath = urlPath.substring(0, index + 1);
                         return urlPath;
                     }
+
                 }
             }
         } catch (Throwable throwable) {
+            InternalAgentLogger.INSTANCE.logAlways(InternalAgentLogger.LoggingLevel.ERROR,"Unable to find the Agent Jar");
```
littleaj: Remove this extra log message. It's going to the same logger, same level and there's no new information here. Feel free to update the wording in the log message on the next line.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriterv2.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * This method returns common super class for both the classes without actually loading
+     * the class using forName(). If no super class is present it returns java/lang/Object class.
+     * @param type1
+     * @param type2
+     * @return The String for the common super class of both the classes
+     */
+    @Override
+    protected String getCommonSuperClass(final String type1, final String type2) {
```
littleaj: At the very least this method should have a unit test, if not the whole class.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriterv2.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class overwrites default class writer of ASM to use the ClassLoader
+ * provided by DefaultByteCode transformer (This loader essentially has all
+ * the required classes already loaded)
+ */
+public class CustomClassWriterv2 extends org.objectweb.asm.ClassWriter {
```
littleaj: I'm not a fan of _SomeClass**V2**_ naming schemes. If the modifications can't be described in one or two words, the go with "branding" it; e.g. _AppInsightsCustomClassWriter_.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassVisitor.java
```diff
@@ -51,7 +52,7 @@ public void visit(int version, int access, String name, String signature, String
     @Override
     public MethodVisitor visitMethod(int access, String name, String desc, String signature, String[] exceptions) {
         MethodVisitor originalMV = super.visitMethod(access, name, desc, signature, exceptions);
-
+        originalMV = new JSRInlinerAdapter(originalMV, access, name, desc, signature, exceptions);
```
littleaj: The next line checks for `null`, which will now always be false. Either that part of the condition is no longer valid and should be removed, or it should be relocated.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
littleaj: Please address these tests. If they are failing, they should be updated to reflect the new functionality. If they are no longer valid, they should be removed (and explain why this is the case).

gradle/common-java.gradle
```diff
@@ -40,8 +40,8 @@ if (java6JreDir) {
         bootClasspath += "$archivePath;"
     }
     tasks.withType(JavaCompile) {
-        sourceCompatibility = 1.6
```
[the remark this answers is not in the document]
littleaj: this.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultClassVisitor.java
dhaval24: well yes I think can be moved one step above may be.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriterv2.java
dhaval24: Sure I will get couple of them.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/CustomClassWriterv2.java
dhaval24: I think this can be just refactored as CustomClassWriter I will do that.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
dhaval24: okay thanks will do this.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/DefaultByteCodeTransformerTest.java
```diff
@@ -27,10 +27,10 @@
 
 public final class DefaultByteCodeTransformerTest {
     @Test
-    public void noClassInstrumentationDataTest() {
+    public void noClassInstrumentationDataTest() throws ClassNotFoundException {
```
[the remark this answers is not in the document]
dhaval24: sure this is by mistake. I think there is any need of throws too.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
dhaval24: @littleaj all the asserts within this tests are commented because there is a method which they intended to write is missing. Therefore there is no use to create unnecessary objects during test run time. It will add overhead only.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
dhaval24: @littleaj this file was edited by @grlima when prepared statements for sqlserver where not working (i.e not capturing the query data) we can no revert this changes back because we no longer need to loose if condition in the sendSQLTelemetry() method.

dhaval24:
@littleaj I have addressed your concerns in the previous review except the diff in CoreAgentNotification Handler. In my IDE git diff shows no difference so will have to figure it out. @grlima had pushed those changes so if you can have a look @grlima would be great.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
littleaj: you can see the difference here: `git diff master -- CoreAgentNotificationsHandler.java > CoreAgentNotificationsHandler.java.diff` (assuming PWD is the file's location). It's the line endings. Your IDE should be set to preserve the existing line ending characters. There may be a way to fix this in IntelliJ, but you may have to load it in Notepad++ an do a find/replace.

If you want to just checkout a clean copy and make the changes again, add use this command to see the "actual" changes: `git diff master -w -- CoreAgentNotificationsHandler.java > CoreAgentNotificationsHandler.java.diff` to ignore all whitespace changes

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/ClassInstrumentationDataTest.java
littleaj: @dhaval24 as discussed, let's just revert this to the master copy and we'll come back to it later. Then, we'll figure out if it needs to be removed or modified so it works.

dhaval24:
Thanks Sergey for your comment. @littleaj and @grlima I am waiting for your approval to merge this into master. Please have a look.

dhaval24:
Folks I have updated the last missing piece on my end which was updating changelog file. If there is anything else which is left let me know so we that we can then merge this.

PR #459, opened 2017-11-01
Introducing Fixed Rate Sampling v2 (Using Processors)

dhaval24:
Introducing Fixed Rate Sampling v2 using TelemetryProcessors architecture. This is in addition to current FixedRateSampler and with then next major version this will be the standard. 

Eg. of creating FixedRateTelemetryProcessor:

In the ApplicationInsights.xml file add the following:

```
<TelemetryProcessors>
        <BuiltInProcessors>
            <Processor type = "FixedRateSamplingTelemetryProcessor">
                <Add name = "SamplingPercentage" value = "50" />
                <ExcludedTypes>
                    <ExcludedType>Request</ExcludedType>
                </ExcludedTypes>
                <IncludedTypes>
                    <IncludedType>Request</IncludedType>
                    <IncludedType>Trace</IncludedType>
                    <IncludedType>Dependency</IncludedType>
                    <IncludedType>Exception</IncludedType>
                </IncludedTypes>
            </Processor>
        </BuiltInProcessors>
    </TelemetryProcessors>

```
The sampling percentage should be between 0 - 100.  If something is specified in IncludeType that means only the particular category of telemetry would be eligible for Sampling. If something is specified in ExcludeType the particular type of telemetry item would not be subject to sampling. Exclude type takes priority over include type.

All correlated telemetry items will be sampled in. With this we plan to sample only using Operation Id. User Id based sampling would be stopped.

Some telemetry type is by default not subject to sampling: PerformanceCounterTelemetry, MetricTelemetry, SessionStateTelemetry.

_This is the part of upcoming changes in all Sampling Architecture in java._

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
@@ -6,25 +6,27 @@ import com.microsoft.applicationinsights.internal.annotation.BuiltInProcessor;
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 import com.microsoft.applicationinsights.telemetry.SupportSampling;
 import com.microsoft.applicationinsights.telemetry.Telemetry;
+import org.omg.CORBA.INTERNAL;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 /**
- * Created by Dhaval Doshi Oct 2017
  * This processor is used to Perform Sampling on User specified sampling rate
- *
+ * <p>
  * How to use in ApplicationInsights Configuration :
- *
+ * <p>
  * <BuiltInProcessors>
-     <Processor type = "FixedRateSamplingTelemetryProcessor">
-         <Add name = "SamplingPercentage" value = "50" />
-         <Add name = "ExcludedTypes" value="Trace" />
-        <Add name = "IncludedTypes" value="Trace;Request" />
-     </Processor>
- </BuiltInProcessors>
+ * <Processor type = "FixedRateSamplingTelemetryProcessor">
+ * <Add name = "SamplingPercentage" value = "50" />
+ * <Add name = "ExcludedTypes" value="Trace" />
+ * <Add name = "IncludedTypes" value="Trace;Request" />
+ * </Processor>
+ * </BuiltInProcessors>
  */
 @BuiltInProcessor("FixedRateSamplingTelemetryProcessor")
 public final class FixedRateSamplingTelemetryProcessor implements TelemetryProcessor {
@@ -39,11 +41,10 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
     private static final String listSeparator = ";";
     private static Map<String, Class> allowedTypes;
 
-    private String excludedTypesString;
-    private Set<Class> excludedTypesHashSet;
+    private Set<Class> excludedTypes;
 
-    private String includedTypesString;
-    private Set<Class> includedTypesHashSet;
+    private Set<Class> includedTypes;
+
 
     /// All sampling percentage must be in a ratio of 100/N where N is a whole number (2, 3, 4, …). E.g. 50 for 1/2 or 33.33 for 1/3.
     /// Failure to follow this pattern can result in unexpected / incorrect computation of values in the portal.
@@ -55,8 +56,8 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
      */
     public FixedRateSamplingTelemetryProcessor() {
         this.samplingPercentage = 100.00;
-        this.includedTypesHashSet = new HashSet<Class>();
-        this.excludedTypesHashSet = new HashSet<Class>();
+        this.includedTypes = new HashSet<Class>();
+        this.excludedTypes = new HashSet<Class>();
         try {
             this.allowedTypes = new HashMap<String, Class>() {{
                 put(dependencyTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry"));
@@ -76,8 +77,8 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
      *
      * @return
      */
-    public Set<Class> getExcludedTypesHashSet() {
-        return excludedTypesHashSet;
+    public Set<Class> getExcludedTypes() {
+        return excludedTypes;
     }
 
     /**
@@ -85,47 +86,22 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
      *
      * @return
      */
-    public Set<Class> getIncludedTypesHashSet() {
-        return includedTypesHashSet;
+    public Set<Class> getIncludedTypes() {
+        return includedTypes;
     }
 
-    /**
-     * This method takes a string of user specified excluded
-     * types and parses it to into set containing respective references
-     *
-     * @param value
-     */
-    public void setExcludedTypes(String value) {
-        this.excludedTypesString = value;
-        Set<Class> newExcludedTypesHashSet = new HashSet<Class>();
-        setIncludedOrExcludedTypes(value, newExcludedTypesHashSet);
-        excludedTypesHashSet = newExcludedTypesHashSet;
-    }
-
-    /**
-     * This method takes a string of user specified included
-     * types and parses it to into set containing respective references
-     *
-     * @param value
-     */
-    public void setIncludedTypes(String value) {
-        this.includedTypesString = value;
-        Set<Class> newIncludedTypesHashSet = new HashSet<Class>();
-        setIncludedOrExcludedTypes(value, newIncludedTypesHashSet);
-        includedTypesHashSet = newIncludedTypesHashSet;
-    }
 
     private void setIncludedOrExcludedTypes(String value, Set<Class> typeSet) {
+
         if (!StringUtils.isNullOrEmpty(value)) {
-            String[] splitList = value.split(listSeparator);
-            for (String item : splitList) {
-                item.trim();
-                if (!StringUtils.isNullOrEmpty(item) && allowedTypes.containsKey(item)) {
-                    typeSet.add(allowedTypes.get(item));
-                } else {
-                    InternalLogger.INSTANCE.error("item is either not allowed to sample or is empty");
-                }
+            value = value.trim();
+            if (!StringUtils.isNullOrEmpty(value) && allowedTypes.containsKey(value)) {
+                typeSet.add(allowedTypes.get(value));
+            } else {
+                InternalLogger.INSTANCE.error("Item is either not allowed to sample or is empty");
             }
+        } else {
+            InternalLogger.INSTANCE.error("Empty types cannot be considered");
         }
     }
 
@@ -191,14 +167,38 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
      */
     private boolean isSamplingApplicable(Class item) {
 
-        if (excludedTypesHashSet.size() > 0 && excludedTypesHashSet.contains(item)) {
+        if (excludedTypes.size() > 0 && excludedTypes.contains(item)) {
             return false;
         }
 
-        if (includedTypesHashSet.size() > 0 && !includedTypesHashSet.contains(item)) {
+        if (includedTypes.size() > 0 && !includedTypes.contains(item)) {
             return false;
         }
 
         return true;
+    }
+
+    /**
+     * This method is invoked during configuration to add one element to the
+     * excluded types set from the xml array list of excluded types
+     * @param value
+     */
+    public void addToExcludedType(String value) {
+
+        setIncludedOrExcludedTypes(value, excludedTypes);
+        InternalLogger.INSTANCE.trace(value + " added as excluded to sampling");
+
+    }
+
+    /**
+     * This method is invoked during configuration to add one element to the
+     * included types set from the xml array list of included types
+     * @param value
+     */
+    public void addToIncludedType(String value) {
+
+        setIncludedOrExcludedTypes(value, includedTypes);
+        InternalLogger.INSTANCE.trace(value + " added as included to sampling");
+
     }
 }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
@@ -15,6 +15,16 @@ import java.util.Set;
 /**
  * Created by Dhaval Doshi Oct 2017
  * This processor is used to Perform Sampling on User specified sampling rate
+ *
+ * How to use in ApplicationInsights Configuration :
+ *
+ * <BuiltInProcessors>
+     <Processor type = "FixedRateSamplingTelemetryProcessor">
+         <Add name = "SamplingPercentage" value = "50" />
+         <Add name = "ExcludedTypes" value="Trace" />
+        <Add name = "IncludedTypes" value="Trace;Request" />
+     </Processor>
+ </BuiltInProcessors>
  */
 @BuiltInProcessor("FixedRateSamplingTelemetryProcessor")
 public final class FixedRateSamplingTelemetryProcessor implements TelemetryProcessor {
@@ -35,7 +45,9 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
     private String includedTypesString;
     private Set<Class> includedTypesHashSet;
 
-    public double samplingPercentage;
+    /// All sampling percentage must be in a ratio of 100/N where N is a whole number (2, 3, 4, …). E.g. 50 for 1/2 or 33.33 for 1/3.
+    /// Failure to follow this pattern can result in unexpected / incorrect computation of values in the portal.
+    private double samplingPercentage;
 
     /**
      * constructor is responsible of initializing this processor
@@ -43,7 +55,6 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
      */
     public FixedRateSamplingTelemetryProcessor() {
         this.samplingPercentage = 100.00;
-        this.samplingPercentage = samplingPercentage;
         this.includedTypesHashSet = new HashSet<Class>();
         this.excludedTypesHashSet = new HashSet<Class>();
         try {
@@ -55,23 +66,33 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
                 put(requestTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RequestTelemetry"));
                 put(traceTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.TraceTelemetry"));
             }};
-        }
-        catch (ClassNotFoundException e) {
+        } catch (ClassNotFoundException e) {
             InternalLogger.INSTANCE.trace("Unable to locate telemetry classes");
         }
     }
 
-    public String getExcludedTypesString() {
-        return excludedTypesString;
+    /**
+     * This method returns a set of classes of excluded types specified by user
+     *
+     * @return
+     */
+    public Set<Class> getExcludedTypesHashSet() {
+        return excludedTypesHashSet;
     }
 
-    public String getIncludedTypesString() {
-        return includedTypesString;
+    /**
+     * This method returns a set of classes of included types specified by user
+     *
+     * @return
+     */
+    public Set<Class> getIncludedTypesHashSet() {
+        return includedTypesHashSet;
     }
 
     /**
      * This method takes a string of user specified excluded
      * types and parses it to into set containing respective references
+     *
      * @param value
      */
     public void setExcludedTypes(String value) {
@@ -84,6 +105,7 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
     /**
      * This method takes a string of user specified included
      * types and parses it to into set containing respective references
+     *
      * @param value
      */
     public void setIncludedTypes(String value) {
@@ -100,17 +122,23 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
                 item.trim();
                 if (!StringUtils.isNullOrEmpty(item) && allowedTypes.containsKey(item)) {
                     typeSet.add(allowedTypes.get(item));
+                } else {
+                    InternalLogger.INSTANCE.error("item is either not allowed to sample or is empty");
                 }
             }
         }
     }
 
-    public double getSamplingPercentage() {
+    /**
+     * Gets the sample rate currently set
+     */
+    double getSamplingPercentage() {
         return samplingPercentage;
     }
 
     /**
      * Sets the user defined sampling percentage
+     *
      * @param samplingPercentage
      */
     public void setSamplingPercentage(String samplingPercentage) {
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+                put(requestTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RequestTelemetry"));
+                put(traceTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.TraceTelemetry"));
+            }};
+        }
+        catch (ClassNotFoundException e) {
+            InternalLogger.INSTANCE.trace("Unable to locate telemetry classes");
+        }
+    }
+
+    public String getExcludedTypesString() {
```
SergeyKanzhelev: why anybody would need to get a string, not the parsed SET?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    private void setIncludedOrExcludedTypes(String value, Set<Class> typeSet) {
+        if (!StringUtils.isNullOrEmpty(value)) {
+            String[] splitList = value.split(listSeparator);
+            for (String item : splitList) {
+                item.trim();
+                if (!StringUtils.isNullOrEmpty(item) && allowedTypes.containsKey(item)) {
+                    typeSet.add(allowedTypes.get(item));
+                }
```
SergeyKanzhelev: else report a problem. Configuration errors are very hard to troubleshoot

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+
+        if (SamplingScoreGeneratorV2.getSamplingScore(telemetry) < samplingPercentage) {
+            return true;
+        }
+
+        else {
+            InternalLogger.INSTANCE.info("Item %s sampled out", telemetry.getClass());
+        }
+
+        return false;
```
SergeyKanzhelev: more fool proof code will be to return `true` globally and only return false when all conditions were met. Danger of this value as default is that some wrong condition will make telemetry items disappear

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/SamplingScoreGeneratorV2.java
```diff
@@ -0,0 +1,61 @@
...
+
+        else {
+            long val = Math.abs(random.nextLong());
+            samplingScore =  ((double)Math.abs(val)/ Long.MAX_VALUE);
+        }
+
+        return samplingScore * 100;
+    }
+
+    private static int getSamplingHashCode(String input) {
```
SergeyKanzhelev: tests for this? Pls copy all the test cases from .NET. We found discrepancy with JavaScript on how numbers are processed. So having those test cases is important.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+    private static final String listSeparator = ";";
+    private static Map<String, Class> allowedTypes;
+
+    private String excludedTypesString;
+    private Set<Class> excludedTypesHashSet;
+
+    private String includedTypesString;
+    private Set<Class> includedTypesHashSet;
+
+    public double samplingPercentage;
```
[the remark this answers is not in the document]
dhaval24: @SergeyKanzhelev 

Addressing 1. There are getters and setters for this however they accept string as input rather than double. This is because our configuration to initialize processors require that. Its public because dot net had kept it public however I don't see it getting accessed directly without getters anywhere so probably can make this private.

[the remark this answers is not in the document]
dhaval24: Addressing 2. Sure I will add this comment makes sense. 100/N should be format.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: Ideally no one should need it. You are right this getters can be removed.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: Understood! Will update accordingly.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/SamplingScoreGeneratorV2.java
dhaval24: Yes for sure. I will add for this.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * constructor is responsible of initializing this processor
+     * to default settings
+     */
+    public FixedRateSamplingTelemetryProcessor() {
+        this.samplingPercentage = 100.00;
+        this.samplingPercentage = samplingPercentage;
```
grlima: What's this do? Is it needed?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+                item.trim();
+                if (!StringUtils.isNullOrEmpty(item) && allowedTypes.containsKey(item)) {
+                    typeSet.add(allowedTypes.get(item));
+                }
+            }
+        }
+    }
+
+    public double getSamplingPercentage() {
+        return samplingPercentage;
```
grlima: Use this.samplingPercentange to be consistent

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+            return true;
+        }
+
+        if (!isSamplingApplicable(telemetry.getClass())) {
+            InternalLogger.INSTANCE.trace("Skip sampling since %s type is not sampling applicable", telemetry.getClass());
+            return true;
+        }
+
+        SupportSampling samplingSupportingTelemetry = ((SupportSampling)telemetry);
+        if (samplingSupportingTelemetry.getSamplingPercentage() != null) {
```
grlima: Is this correct? If  a sampling percentage is available (i.e. different than null), then shouldn't we consider the telemetry for sampling?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/SamplingScoreGeneratorV2.java
```diff
@@ -0,0 +1,61 @@
...
+    public static double getSamplingScore(Telemetry telemetry) {
+
+        double samplingScore = 0.0;
+
+        if (!StringUtils.isNullOrEmpty(telemetry.getContext().getUser().getId())) {
+            samplingScore = (double) (getSamplingHashCode(telemetry.getContext().getUser().getId()) / Integer.MAX_VALUE);
+        }
+
+        else if (!StringUtils.isNullOrEmpty(telemetry.getContext().getOperation().getId())) {
+            samplingScore = (double) (getSamplingHashCode(telemetry.getContext().getOperation().getId()) / Integer.MAX_VALUE);
```
dhaval24: int / int division returns 0

dhaval24:
The build has passed for this.  https://mseng.visualstudio.com/AppInsights/_build/index?buildId=4386457&_a=summary

@SergeyKanzhelev please have a look. I have tried to address most of the change requests. Also I have included the tests for SamplingScoreGenerator. It now does not take into account userId for sampling and I have added relevant test scenarios too which should be implemented in .NET too.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+import com.microsoft.applicationinsights.telemetry.SupportSampling;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Created by Dhaval Doshi Oct 2017
```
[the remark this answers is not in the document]
SergeyKanzhelev: I don't care too much, but suggest to remove it at some point. Many editors have "blame" function that is more reliable than comments in code.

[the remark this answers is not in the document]
dhaval24: @SergeyKanzhelev sure I can remove it from the code I am checking in.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+public final class FixedRateSamplingTelemetryProcessor implements TelemetryProcessor {
+
+    private final String dependencyTelemetryName = "Dependency";
+    private static final String eventTelemetryName = "Event";
+    private static final String exceptionTelemetryName = "Exception";
+    private static final String pageViewTelemetryName = "PageView";
+    private static final String requestTelemetryName = "Request";
+    private static final String traceTelemetryName = "Trace";
+
+    private static final String listSeparator = ";";
```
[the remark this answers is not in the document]
SergeyKanzhelev: what kind of guidance do you need?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,175 @@
...
+    private static Map<String, Class> allowedTypes;
+
+    private String excludedTypesString;
+    private Set<Class> excludedTypesHashSet;
```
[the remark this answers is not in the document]
SergeyKanzhelev: `excludedTypes`

SergeyKanzhelev:
@dhaval24 if you need to ship something quick - your code may work. I'd like to see following some of my suggestions for a better code though... Let me know if you want me to approve to unblock this check in

dhaval24:
@SergeyKanzhelev I do not want to ship out something quickly which won't be robust. I have taken a careful note of your reviews and have made changes to accommodate all of them in my latest commit. Please review the description of the PR for how the configuration of Included and Excluded Types will happen. I have also covered the changes with the relevant test cases and removed couple of test cases which became irrelevant after this change.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ParamExcludedTypeXmlElement.java
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.applicationinsights.internal.config;
+
+import javax.xml.bind.annotation.XmlRootElement;
+import java.util.List;
+
+//This is the class for binding the xml array list of <ExcludedTypes>
```
gavlyukovskiy: It is better to use javadoc here, not just comment.

dhaval24: @gavlyukovskiy yes we can use javadoc instead. I will update. Also we are planning to have a robust style guide to follow for this project in upcoming week and that should help in reducing this kind of inconsistency. Thanks for the suggestion,

dhaval24:
@SergeyKanzhelev This PR is now ready for next set of review and I hope this time close to completion :-) Please let me know if there is anything else you would like to see improved.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
+package com.microsoft.applicationinsights.internal.channel.samplingV2;
+
+import com.microsoft.applicationinsights.agent.internal.common.StringUtils;
+import com.microsoft.applicationinsights.extensibility.TelemetryProcessor;
+import com.microsoft.applicationinsights.internal.annotation.BuiltInProcessor;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.telemetry.SupportSampling;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+import org.omg.CORBA.INTERNAL;
```
SergeyKanzhelev: was this dependency defined in maven package already? I don't see any references updates. Do you really need this new dependency?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+ * This processor is used to Perform Sampling on User specified sampling rate
+ * <p>
+ * How to use in ApplicationInsights Configuration :
+ * <p>
+* <TelemetryProcessors>
+    <BuiltInProcessors>
+        <Processor type = "FixedRateSamplingTelemetryProcessor">
+            <Add name = "SamplingPercentage" value = "50" />
+            <ExcludedTypes>
+                <excludedType>Request</excludedType>
```
SergeyKanzhelev: It is inconsistent to use camel case here when you are using pascal case above.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+    private static final String traceTelemetryName = "Trace";
+
+    private static Map<String, Class> allowedTypes;
+
+    private Set<Class> excludedTypes;
+
+    private Set<Class> includedTypes;
+
+
+    /// All sampling percentage must be in a ratio of 100/N where N is a whole number (2, 3, 4, …). E.g. 50 for 1/2 or 33.33 for 1/3.
```
SergeyKanzhelev: Make is a Javadoc?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Gets the sample rate currently set
+     */
+    double getSamplingPercentage() {
+        return samplingPercentage;
+    }
+
+    /**
+     * Sets the user defined sampling percentage
```
SergeyKanzhelev: Actually - move the comment from private var and place it into this javadoc

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+     *
+     * @param samplingPercentage
+     */
+    public void setSamplingPercentage(String samplingPercentage) {
```
SergeyKanzhelev: why string as an argument?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+        this.samplingPercentage = Double.valueOf(samplingPercentage);
```
SergeyKanzhelev: this thing may throw. Change argument to double or handle exception and do not throw

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
SergeyKanzhelev: mismatch of `get`-er and `set`-er is not the best practice

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -0,0 +1,212 @@
...
+     */
+    @Override
+    public boolean process(Telemetry telemetry) {
+
+        double samplingPercentage = this.samplingPercentage;
+
+        if (telemetry instanceof SupportSampling) {
+
+            if (isSamplingApplicable(telemetry.getClass())) {
+
```
SergeyKanzhelev: what's the deal with so many empty lines?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryProcessorCreator.java
```diff
@@ -46,7 +48,48 @@ public TelemetryProcessor Create(TelemetryProcessorXmlElement confClass) {
             return null;
         }
 
+
+        // If the <ExcludedTypes> tag is not empty
```
SergeyKanzhelev: See how it's done in .NET. Configuration reader can detect that the property you are about to initialize is a list and just adds values to it

SergeyKanzhelev: No hardcoded names

core/src/test/java/com/microsoft/applicationinsights/internal/config/ValidProcessorWithIncludeExcludeType.java
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.applicationinsights.internal.config;
+
+import com.microsoft.applicationinsights.extensibility.TelemetryProcessor;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+public class ValidProcessorWithIncludeExcludeType implements TelemetryProcessor {
```
SergeyKanzhelev: Why do you need this as a public class? Not sure what it does

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
SergeyKanzhelev: Making generic `Add` instead of `excludedType` will allow to generalize this logic in future

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: This might be some unused import I don't need it really would remove it.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: sure would convert to pascal case here.

dhaval24: There is already a separate add tag in the source tag. While I agree with your point that this will add more extensibility but I thought that being specific here adds a lot more ease of understanding and separation of concern. We can discuss on deciding this.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: Yes making javadoc makes sense.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: sure will do that.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: The string argument is because the Add type takes a parameter string and builds the configuration considering the same. Therefore we need to have a setter which takes the string as an argument. I agree its not the best practice but its how our current configuration is and I might fear that doing some might break other things.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: This can throw definitely if the value is some crazy string. I can handle exception here.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: It was just to make code cleaner to look. we can remove them.

core/src/test/java/com/microsoft/applicationinsights/internal/config/ValidProcessorWithIncludeExcludeType.java
dhaval24: This should technically be a part of the test package. Its a stub class which supports testing. I will double check if by mistake the was not was not the part of the test package and move it.

dhaval24:
@SergeyKanzhelev thanks for reviewing, see my comments/opinions here.

SergeyKanzhelev:
@dhaval24 ping again when you'll push updates

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/SamplingScoreGeneratorV2.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Random;
+
+/**
+ * Created by Dhaval Doshi Oct 2017
+ * This class generates the sample using the random number generator.
+ * It also contains the logic to preserve the correlated telemetry items.
+ */
+public class SamplingScoreGeneratorV2 {
+
+    private static Random random = new Random();
```
grlima: Is there potential for multi-threaded contention here? Random is thread-safe, ThreadLocalRandom performs better in multi-threaded environment. 
https://docs.oracle.com/javase/8/docs/api/java/util/Random.html?is-external=true

dhaval24: @grlima yes I can have a look. BTW does this ThreadLocalRandom exist in Java 6 too? Documentation seems to be of Java 8 where as we are currently coding against JDK 1.6

dhaval24:
@SergeyKanzhelev I have addressed some of your changes which were fixable without breaking other parts of the SDK.

PR #461, opened 2017-11-07
Programmatic configuration for TelemetryConfiguration

### LINKED ISSUE #359: Not possible to register WebApp name/key properly when using WebRequestTrackingFilter in Spring boot

gavlyukovskiy:
Fix #359, revisit #411.

Adds a way to configure TelemetryConfiguration from outside application insights library. 
It is a must for spring boot integration because you can retrieve configuration only from spring context itself. 
Whereas the same can be achieved by creating yet another `@Configuration` class, just like `InterceptorRegistry`, and asking user to import that one, it also can be accomplished by creating a [spring boot starter](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-auto-configuration.html) with our full control of context initialization order.
I've prepared [such starter](https://github.com/gavlyukovskiy/azure-application-insights-spring-boot-starter) that links with my forked version of sdk and once is added, requires only single property `azure.application-insights.instrumentation-key` to be set to integrate spring boot application with Application Insights. 
As well allows you to use spring context to manage `TelemetryConfiguration` components - you can define any telemetry module, channel or sampler as `@Bean` and it will be used inside `TelemetryConfiguration`. Also it adds nice spring boot feature such like configuration metadata -
 when IDE shows you the list of available configuration properties.

Conceptually all changes are backward compatible and only concern is that all modules that are using `TelemetryConfiguration#getActive()` must be initialized after `TelemetryConfiguration` is created by spring context, which is not possible for logger appenders, that's why I added lazy initialization inside logback appender (log4j and log4j2 can be modified in the same way) which leads to loss of initial log messages. If you think it is critical I can implement some kind of a buffer that will hold telemetry events until configuration is initialized.

You can look at my starter project to find more details. Also I can upload my sample project that uses forked version of sdk and the starter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Introducing FixedRate Sampling v2 Using Telemetry Processors
 - Fixed issue #436 (TraceTelemetry with Severity is not shown in UI). This fixes a regression issue with `TelemetryClient.trackTrace` and `TelemetryClient.trackException`.
 - Compilation now targets Java 1.7. Java 1.6 is no longer supported.
+- Adding system property `applicationinsights.configurationDirectory` to allow to explicitly set directory containing the config file.
 
 ## Version 1.0.10
 - `track()` method of 'com.microsoft.applicationinsights.TelemetryClient' is now modified. No longer performing pre-sanitization
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
@@ -41,6 +41,10 @@ import com.google.common.base.Strings;
  * Created by gupele on 5/25/2015.
  */
 public final class ConfigurationFileLocator {
+
+    /** name of property containing path to directory with configuration file */
+    public static final String CONFIG_DIR_PROPERTY = "applicationinsights.configurationDirectory";
+
     private final String configurationFileName;
 
     public ConfigurationFileLocator(String configurationFileName) {
@@ -49,22 +53,34 @@ public final class ConfigurationFileLocator {
     }
 
     public InputStream getConfigurationFile() {
-        InputStream inputStream = ConfigurationFileLocator.class.getClassLoader().getResourceAsStream(configurationFileName);
-        if (inputStream != null) {
-            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Configuration file has been successfully found as resource");
-            return inputStream;
-        }
 
-        // Trying to load configuration as a resource.
-        String configurationFile = getConfigurationFromCurrentClassLoader();
+        String configurationFile;
+        InputStream inputStream;
 
-        // If not found as a resource, trying to load from the executing jar directory
-        if (configurationFile == null) {
-            configurationFile = getConfigurationFromLibraryLocation();
+        // first try to get from dir defined explicitly in system property insights.configurationFile
+        String configDirFromProperty = System.getProperty(CONFIG_DIR_PROPERTY);
+        if (configDirFromProperty != null) {
+            configurationFile = getConfigurationAbsolutePath(configDirFromProperty);
+        } else {
 
-            // If still not found try to get it from the class path
+            inputStream = ConfigurationFileLocator.class.getClassLoader().getResourceAsStream(configurationFileName);
+            if (inputStream != null) {
+                InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO,
+                                                  "Configuration file has been successfully found as resource");
+                return inputStream;
+            }
+
+            // Trying to load configuration as a resource.
+            configurationFile = getConfigurationFromCurrentClassLoader();
+
+            // If not found as a resource, trying to load from the executing jar directory
             if (configurationFile == null) {
-                configurationFile = getConfFromClassPath();
+                configurationFile = getConfigurationFromLibraryLocation();
+
+                // If still not found try to get it from the class path
+                if (configurationFile == null) {
+                    configurationFile = getConfFromClassPath();
+                }
             }
         }
 
--- a/core/src/test/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocatorTest.java
+++ b/core/src/test/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocatorTest.java
@@ -56,10 +56,30 @@ public final class ConfigurationFileLocatorTest {
     }
 
     @Test
+    public void testGetConfigurationFileWhereDirIsFromProperty() throws Exception {
+        System.setProperty(ConfigurationFileLocator.CONFIG_DIR_PROPERTY, "src/test/resources");
+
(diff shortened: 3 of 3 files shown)


dhaval24:
@gavlyukovskiy Thank you very much for your contribution towards enabling spring boot support with Application Insights. Amazing work! I have added reviewers to your pull request and we will review it.  Lets have a conversation around this and we shall take it from there. On a brief overview I feel this is a good concept.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -77,6 +80,33 @@ public static TelemetryConfiguration getActive() {
     }
 
     /**
+     * Set {@link TelemetryConfiguration} instance to be 'Active'.
+     * Telemetry configuration must not be set twice explicitly or through calling {@link TelemetryConfiguration#getActive()}.
+     *
+     * @param telemetryConfiguration preconfigured instance of Telemetry Configuration
+     */
+    public static void setActive(TelemetryConfiguration telemetryConfiguration) {
```
SergeyKanzhelev: We found it messy to allow to replace the configuration object. As you mentioned in comments one of the modules may get a hold of an object and will not pick up an update. Would it be better to update properties of an object instead?

With dependency injection we would simply initialize singleton of dependency injection with the value of `telemetryconfiguration.active`

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -77,6 +80,33 @@ public static TelemetryConfiguration getActive() {
...
+        synchronized (s_lock) {
+            if (initialized) {
+                throw new IllegalStateException("Telemetry configuration is already initialized. " +
```
SergeyKanzhelev: Public APIs do not throw in Application Insights SDK

dhaval24: I agree with Sergey's point here. Throwing an exception in public API will crash the customers application and prevent it from starting. It is not a desired behavior.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -77,6 +80,33 @@ public static TelemetryConfiguration getActive() {
...
+                        "Please make sure that TelemetryConfiguration#setActive(TelemetryConfiguration) is not called twice " +
```
dhaval24: Can we some how prevent this behavior programaticallly? If some one tries to call setActive() twice the second call is ignored?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -260,7 +260,11 @@ private String getName(ServletContext context) {
         String name = null;
         try {
             String contextPath = context.getContextPath();
-            if (CommonUtils.isNullOrEmpty(contextPath)) {
+            TelemetryConfiguration telemetryConfiguration = TelemetryConfiguration.getActive();
+            if (telemetryConfiguration.getApplicationName() != null) {
+                name = telemetryConfiguration.getApplicationName();
+            }
+            else if (CommonUtils.isNullOrEmpty(contextPath)) {
                 URL[] jarPaths = ((URLClassLoader) (this.getClass().getClassLoader())).getURLs();
```
dhaval24: This is an issue with JDK 9. The default class loader of JDK 9 is not an instance of URLClassLoader but is instead called "App-Classloader". This call to getURLs() would fail here. Do we have a work around for this ? Have you tried this on a Spring Boot Project with JDK 9?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -260,7 +260,11 @@ private String getName(ServletContext context) {
         String name = null;
         try {
             String contextPath = context.getContextPath();
-            if (CommonUtils.isNullOrEmpty(contextPath)) {
+            TelemetryConfiguration telemetryConfiguration = TelemetryConfiguration.getActive();
+            if (telemetryConfiguration.getApplicationName() != null) {
+                name = telemetryConfiguration.getApplicationName();
```
dhaval24: Do we need a check here if someone has tried setting application name as null?

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
gavlyukovskiy: That's why I added throw to be sure that configuration is initialized only once and no module is initialized with wrong `TelemetryConfiguration`. Probably configuration could be changed once context is loaded but I'm afraid that some module can get properties from the `TelemetryConfiguration` when those properties are not set yet - like `WebRequestTrackingFilter#registerWebApp` that is executed only once.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
gavlyukovskiy: I would agree with you about runtime exceptions, but this one is startup exception, where, from my point of view, is better to get exception when something goes wrong rather than found that telemetry is not working after application is started. But if this is a strong sdk convention, I can just log error into logs.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
gavlyukovskiy: Thanks, I will add check into `TelemetryConfiguration#setApplicationName` that name is not null or empty.

gavlyukovskiy:
I created repo with sample project [here](https://github.com/gavlyukovskiy/azure-application-insights-spring-boot-sample) it shows how it is easy to start with starter in spring boot application.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
gavlyukovskiy: I haven't tried it on JDK 9, but probably `ClassLoader#getSystemResources` will do the job.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
gavlyukovskiy: Answered in previous thread, this is probably an option, but I wouldn't implement it like that, it is too dangerous and hard to debug if something goes wrong, it should not be initialized more than once.

PR #464, opened 2017-11-17
Fix bug462 telemetry naming miss instrumentation key

### LINKED ISSUE #462: Telemetry Naming not align with .net Version
# Issue Description

when I use the Java sdk, the log sent out is named as
 
**"name": "Microsoft.ApplicationInsights.Request"**

but when I use .net sdk, the name contains the instrumentation key, like:

**"name": "Microsoft.ApplicationInsights.aif4614a93a841f4d9dae2754a1d946fc8b.Message",**

# Impact

While using the ApplicaitonInsights Java SDK as FirstParty Partner to onboard Microsoft Asimove Platform, no source node will be created, as the Event Name does not contains the Key

Ref: 
[Xpert Ticket 2777] (http://asimovtickets/helpdesk/Ticket/2777)

# Root Cause
The root cause is that Java SDK hard corded the Telemetry Name, but .net SDK process it with give prefix:

```
        /// <summary>
        /// Normalize instrumentation key by removing dashes ('-') and making string in the lowercase.
        /// In case no InstrumentationKey is available just return empty string.
        /// In case when InstrumentationKey is available return normalized key + dot ('.')
        /// as a separator between instrumentation key part and telemetry name part.
        /// </summary>
        private static string NormalizeInstrumentationKey(string instrumentationKey)
        {
            if (instrumentationKey.IsNullOrWhiteSpace())
            {
                return string.Empty;
            }

            return instrumentationKey.Replace("-", string.Empty).ToLowerInvariant() + ".";
        }
```

XiaoliangHuang:
Fix #462 telemetry name miss instrumentation key.

- [ ] CHANGELOG.md updated

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
@@ -21,7 +21,6 @@
 
 package com.microsoft.applicationinsights.telemetry;
 
-import com.microsoft.applicationinsights.agent.internal.common.StringUtils;
 import com.microsoft.applicationinsights.internal.schemav2.Data;
 import com.microsoft.applicationinsights.internal.schemav2.Domain;
 import com.microsoft.applicationinsights.internal.schemav2.Envelope;
@@ -33,6 +32,8 @@ import java.util.Date;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * Superclass for all telemetry data classes.
@@ -192,7 +193,7 @@ public abstract class BaseTelemetry<T extends Domain> implements Telemetry {
     }
     
     private String normalizeInstrumentationKey(String instrumentationKey){
-    	if (StringUtils.isNullOrEmpty(instrumentationKey)){
+    	if (StringUtils.isEmpty(instrumentationKey) || StringUtils.containsOnly(instrumentationKey, ".- ")){
     		return "";
     	}
     	else{
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
@@ -42,7 +42,7 @@ public abstract class BaseTelemetry<T extends Domain> implements Telemetry {
     private Date timestamp;
     private String sequence;
     
-    private static final String TELEMETRY_NAME_PREFIX = "Microsoft.ApplicationInsights";
+    private final String TELEMETRY_NAME_PREFIX = "Microsoft.ApplicationInsights.";
 
     protected BaseTelemetry() {
     }
@@ -200,11 +200,11 @@ public abstract class BaseTelemetry<T extends Domain> implements Telemetry {
     	}
     }
     
-    private String getTelemetryName(String formattedInstrumentationKey, String envelopType){
+    private String getTelemetryName(String normalizedInstrumentationKey, String envelopType){
     	return String.format(
-    			"{0}{1}{2}",
+    			"%s%s%s",
     			TELEMETRY_NAME_PREFIX,
-    			formattedInstrumentationKey,
+    			normalizedInstrumentationKey,
     			envelopType
     			);
     }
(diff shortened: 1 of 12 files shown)


core/src/test/java/com/microsoft/applicationinsights/telemetry/BaseTelemetryTest.java
```diff
@@ -23,11 +23,17 @@
...
 import com.microsoft.applicationinsights.internal.schemav2.Data;
 import com.microsoft.applicationinsights.internal.schemav2.Domain;
+import com.microsoft.applicationinsights.internal.schemav2.Envelope;
+import com.microsoft.applicationinsights.internal.schemav2.SessionStateData;
+
 import org.junit.Test;
 
 import java.io.IOException;
+import java.io.StringWriter;
+import java.lang.reflect.*;
```
dhaval24: @XiaoliangHuang can we avoid use of wild cards for import here?

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
```diff
@@ -184,4 +190,23 @@ protected String getEnvelopName() {
     protected String getBaseTypeName() {
         throw new UnsupportedOperationException();
     }
+    
+    private String normalizeInstrumentationKey(String instrumentationKey){
+    	if (StringUtils.isNullOrEmpty(instrumentationKey)){
+    		return "";
+    	}
+    	else{
+    		return instrumentationKey.replace("-", "").toLowerCase() + ".";
```
dhaval24: @XiaoliangHuang if in rare case instrumentation key is only "----" (dashes) then this part will introduce additional dot. I don't think this is desirable. Please add a check and a test case also.

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
```diff
@@ -21,6 +21,7 @@
 
 package com.microsoft.applicationinsights.telemetry;
 
+import com.microsoft.applicationinsights.agent.internal.common.StringUtils;
```
dhaval24: @XiaoliangHuang can we avoid agent import in core classes. I believe there is similar method in core also.

XiaoliangHuang: updated, the agent dependent on the core, should not refer agent from core.

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
XiaoliangHuang: Add special case handler, if there are only characters from "-/.", then do not contains ikey.
maybe in the future, should add some schema check on the configuration to restrict the usage of special characters in the ikey

core/src/test/java/com/microsoft/applicationinsights/telemetry/BaseTelemetryTest.java
XiaoliangHuang: this was added as I planed to use reflect to test the private method, but at last not used, so I removed it.

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
```diff
@@ -184,4 +191,23 @@ protected String getEnvelopName() {
     protected String getBaseTypeName() {
         throw new UnsupportedOperationException();
     }
+    
+    private String normalizeInstrumentationKey(String instrumentationKey){
+    	if (StringUtils.isEmpty(instrumentationKey) || StringUtils.containsOnly(instrumentationKey, ".- ")){
```
littleaj: Is it possible for it to have a missing iKey at this point? or is this an error scenario?

SergeyKanzhelev: outcome will be questionable if key is empty, but it is definitely possible. So thins handling is OK

littleaj: cool. thanks for clarifying

dhaval24:
@grlima this should be good for merging. Please have a look at it once before we do so.

dhaval24:
@XiaoliangHuang Thank you very much for your contribution towards the SDK. Your pull request is now merged.

PR #493, opened 2017-12-11
Improved logging

dhaval24:
This PR improves the existing lack of logging issue to ensure that every exception that SDK generates is caught and also logged using SDK Logger. In a nutshell logging is more robust.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -389,7 +389,7 @@ public class TelemetryClient {
     public void track(Telemetry telemetry) {
 
         if (generateCounter.incrementAndGet() % 10000 == 0) {
-            InternalLogger.INSTANCE.info("Total events generated till now "+ generateCounter.get());
+            InternalLogger.INSTANCE.info("Total events generated till now %d", generateCounter.get());
         }
 
         if (telemetry == null) {
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -23,6 +23,7 @@ package com.microsoft.applicationinsights;
 
 import java.util.Date;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.microsoft.applicationinsights.common.CommonUtils;
 import com.microsoft.applicationinsights.extensibility.ContextInitializer;
@@ -58,6 +59,7 @@ public class TelemetryClient {
     private static final Object TELEMETRY_STOP_HOOK_LOCK = new Object();
     private static final Object TELEMETRY_CONTEXT_LOCK = new Object();
 
+    private static AtomicLong generateCounter = new AtomicLong(0);
     /**
      * Initializes a new instance of the TelemetryClient class. Send telemetry with the specified configuration.
      * @param configuration The configuration this instance will work with.
@@ -385,6 +387,11 @@ public class TelemetryClient {
      * @param telemetry The {@link com.microsoft.applicationinsights.telemetry.Telemetry} instance.
      */
     public void track(Telemetry telemetry) {
+
+        if (generateCounter.incrementAndGet() % 10000 == 0) {
+            InternalLogger.INSTANCE.info("Total events generated till now "+ generateCounter.get());
+        }
+
         if (telemetry == null) {
             throw new IllegalArgumentException("telemetry item cannot be null");
         }
--- a/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
@@ -22,6 +22,7 @@
 package com.microsoft.applicationinsights.channel.concrete.inprocess;
 
 import java.io.IOException;
+import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.net.URI;
 import java.util.Map;
@@ -43,6 +44,7 @@ import com.microsoft.applicationinsights.channel.TelemetryChannel;
 
 import com.google.common.base.Strings;
 import com.google.common.base.Preconditions;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 
 /**
  * An implementation of {@link com.microsoft.applicationinsights.channel.TelemetryChannel}
@@ -103,6 +105,7 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
         } catch (Throwable t) {
             developerMode = false;
             InternalLogger.INSTANCE.error("DEVELOPER_MODE_SYSTEM_PROPERTY_NAME is either null or empty");
+            InternalLogger.INSTANCE.trace("stack trace is %s", ExceptionUtils.getStackTrace(t));
         }
         initialize(null,
                 null,
@@ -217,6 +220,7 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
 
         } catch (IOException e) {
             InternalLogger.INSTANCE.error("Failed to serialize Telemetry");
+            InternalLogger.INSTANCE.trace("Stack trace is %s", ExceptionUtils.getStackTrace(e));
             return;
         }
 
@@ -239,7 +243,7 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
             stopped = true;
         } catch (Throwable t) {
             InternalLogger.INSTANCE.error("Exception generated while stopping telemetry transmitter");
-            t.printStackTrace();
+            InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(t));
         }
     }
 
--- a/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
@@ -104,8 +104,8 @@ public final class InProcessTelemetryChannel implements TelemetryChannel {
             }
         } catch (Throwable t) {
             developerMode = false;
-            InternalLogger.INSTANCE.error("DEVELOPER_MODE_SYSTEM_PROPERTY_NAME is either null or empty");
-            InternalLogger.INSTANCE.trace("stack trace is %s", ExceptionUtils.getStackTrace(t));
+            InternalLogger.INSTANCE.trace("DEVELOPER_MODE_SYSTEM_PROPERTY_NAME is either null or empty," +
+                    "stack trace is %s", ExceptionUtils.getStackTrace(t));
         }
         initialize(null,
                 null,
--- a/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
@@ -46,22 +46,22 @@ import com.google.common.base.Preconditions;
 
 /**
  * An implementation of {@link com.microsoft.applicationinsights.channel.TelemetryChannel}
- *
+ * <p>
  * The channel holds two main entities:
- *
+ * <p>
  * A buffer for incoming {@link com.microsoft.applicationinsights.telemetry.Telemetry} instances
  * A transmitter
- *
+ * <p>
  * The buffer is stores incoming telemetry instances. Every new buffer starts a timer.
  * When the timer expires, or when the buffer is 'full' (whichever happens first), the
  * transmitter will pick up that buffer and will handle its sending to the server. For example,
  * a transmitter will be responsible for compressing, sending and activate a policy in case of failures.
- *
+ * <p>
  * The model here is:
- *
+ * <p>
  * Use application threads to populate the buffer
  * Use channel's threads to send buffers to the server
- *
+ * <p>
  * Created by gupele on 12/17/2014.
  */
 public final class InProcessTelemetryChannel implements TelemetryChannel {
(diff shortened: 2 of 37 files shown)


dhaval24:
@littleaj thanks for your inputs on this. I have updated the PR.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -204,6 +211,10 @@ public void send(Telemetry telemetry) {
             String asJson = writer.toString();
             telemetryBuffer.add(asJson);
             telemetry.reset();
+            if (itemsSent.incrementAndGet() % 10000 == 0) {
+                InternalLogger.INSTANCE.info("items sent till now " + itemsSent.get());
```
littleaj: Change this to use the format string method instead of string concat.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -385,6 +387,11 @@ public void trackPageView(PageViewTelemetry telemetry) {
      * @param telemetry The {@link com.microsoft.applicationinsights.telemetry.Telemetry} instance.
      */
     public void track(Telemetry telemetry) {
+
+        if (generateCounter.incrementAndGet() % 10000 == 0) {
+            InternalLogger.INSTANCE.info("Total events generated till now "+ generateCounter.get());
```
littleaj: Use format string here too (instead of string concat)

dhaval24: Ah missed it. Thanks will do!

dhaval24:
Infact about the stack traces I want that those get reported as SDK exception telemetry on the portal. These stack traces are very vital for debugging. We can have this as full feature. @grlima your suggestions?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -99,6 +102,7 @@ public InProcessTelemetryChannel() {
             }
         } catch (Throwable t) {
             developerMode = false;
+            InternalLogger.INSTANCE.error("DEVELOPER_MODE_SYSTEM_PROPERTY_NAME is either null or empty");
```
grlima: Why error? I don't think this system property is required.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -166,6 +172,7 @@ public boolean isDeveloperMode() {
 
     /**
      * Sets value indicating whether this channel is in developer mode.
+     *
```
grlima: empty line. did you mean?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -227,6 +238,8 @@ public synchronized void stop(long timeout, TimeUnit timeUnit) {
             telemetriesTransmitter.stop(timeout, timeUnit);
             stopped = true;
         } catch (Throwable t) {
+            InternalLogger.INSTANCE.error("Exception generated while stopping telemetry transmitter");
+            t.printStackTrace();
```
grlima: "printStackTrace()" writes to std error regardless of whether or not SDKLogger is in use. This may not be desirable. Ideally we print these only when SDKLogger is present.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
dhaval24: I think I generated as error log because last time me and @littleaj discussed on if we should fallback to locating the xml from the project if this was passed incorrectly , we came to conclusion that when someone sets environment variable it is bound to be correct or they should be notified of error. I can reduce the severity to trace if that is not the case

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
dhaval24: This should be by mistake I would remove. Thanks

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
dhaval24: @grlima  Yes printStackTrace() writes to std error stream. It might overflood incase there are too many exceptions. However, my concern here was if there are deep stack traces they might get trimmed by the SDKLogger which is not desirable. I can double check that behavior and if it's not true then I can pass stackTrace as an argument to the logger.

littleaj:
@grlima I agree with your point about the log flood. Gradle has an option `showstacktraces` for debugging build scripts. I like that model. We could mimic that exactly or assume when debug/trace logging is on, we should print the stack traces. 

In either case, I think the stack traces should go through the InternalLogger mechanism. This will cover the case where the client has configured file logging only (nothing to the console). If we always use stdout/stderr then the traces won't end up in the configured log file.

gavlyukovskiy:
Why don't you use slf4j or something? I look through your project and I see that you're reinventing the wheel, you can use appropriate severity for logging and without fear print stacktraces where you need them and customer then can decide what of this he needs. 
Every logging library has ability to configure minimal severity for a logger or a package or just turn it off at all, for some libraries it is also possible to change minimal severity at runtime. As well existing libraries such as log4j2 and logback have much better performance and configuration than your internal loggers mechanism. 
For us it was really odd that application insights logs something to stdout/stderr and we have to configure it to log into file, which is not really good for us because we have specific logging format and file rotating mechanism. We also have ELK stack that collects all logs from every service to elasticsearch and configuring it to catch also application insights logs was not worth our time so we just turned them off.

gavlyukovskiy:
Although it would not be easy change as some customers already configured their systems to work with application insights logs with their current format. Probably all logs can be duplicated into two systems - slf4j and internal loggers, then customer could chose one that they prefer. But I think giving up internal implementation in favor of something standard in Java community is the only way to go.

dhaval24:
@gavlyukovskiy thanks for raising this point and I agree that our implementation of InternalLogger would not be the most efficient logging system out there and it would make sense to move to common logging libraries for the same. However, I agree with you second point too that people might have already configured the application and hence we need to develop incrementally. At this point having more logging in the internal logger makes sense and as we progress further we can move all logging to desired logging framework.

@littleaj @grlima please review my changes here I have addressed at most places though I might have missed some accidentally.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/CpuPerformanceCounterCalculator.java
```diff
@@ -70,6 +70,8 @@ public double getProcessCpuUsage() {
             prevProcessCpuTime = processCpuTime;
         } catch (Exception e) {
             processCpuUsage = Constants.DEFAULT_DOUBLE_VALUE;
+            InternalLogger.INSTANCE.error("Error in getProcessCPUUsage");
+            e.printStackTrace();
```
grlima: +1

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/PerformanceCounterContainer.java
```diff
@@ -224,6 +224,8 @@ public void run() {
                             try {
                                 plugin.preCollection();
                             } catch (Throwable t) {
+                                InternalLogger.INSTANCE.error("Error in thread scheduled for PerformanceCounterContainer");
+                                t.printStackTrace();
```
grlima: +1

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/WindowsPerformanceCounterAsPC.java
```diff
@@ -110,6 +110,8 @@ private void register(String category, String counter, String instance) {
                         setDisplayName(category + " " + counter);
                 pcs.put(key, data);
             } catch (Throwable e) {
+                InternalLogger.INSTANCE.error("Exception while registering windows performance counter as PC");
+                e.printStackTrace();
```
grlima: +1

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmitterImpl.java
```diff
@@ -172,8 +173,12 @@ public void run() {
             return true;
         } catch (Exception e) {
             semaphore.release();
+            InternalLogger.INSTANCE.error("Error in scheduledSend of telemetry items failed. %d items were not sent ", telemetriesFetcher.fetch().size());
+            e.printStackTrace();
```
grlima: +1

core/src/main/java/com/microsoft/applicationinsights/common/CommonUtils.java
```diff
@@ -41,6 +44,8 @@ public static String getHostName(){
         }
         catch (UnknownHostException ex) {
             // optional parameter. do nothing if unresolvable
+            InternalLogger.INSTANCE.trace("Unresolvable host error");
+            InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(ex));
```
littleaj: @dhaval24 There's no reason to call the logger twice on the same level. Feel free to use `%n` to insert a newline between the messages if you want them on separate lines.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -185,6 +181,7 @@ public boolean send(Transmission transmission) {
                 httpClient.dispose(response);
                 // backoff before trying again
                 if (shouldBackoff) {
+                    InternalLogger.INSTANCE.trace("Backing off for 300 seconds");
```
littleaj: Since this is a variable, use the variable value instead of a string literal

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmitterImpl.java
```diff
@@ -31,6 +31,7 @@
 import com.microsoft.applicationinsights.internal.channel.TelemetrySerializer;
```
littleaj: This file has a few cases of the stack trace. Need to convert to the logger call.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -75,6 +76,7 @@ public FixedRateSamplingTelemetryProcessor() {
             }};
         } catch (ClassNotFoundException e) {
             InternalLogger.INSTANCE.trace("Unable to locate telemetry classes");
+            InternalLogger.INSTANCE.trace("stack trace is %s", ExceptionUtils.getStackTrace(e));
```
littleaj: Same as above (don't call the logger on the same level twice; combine the messages into one call)

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/WindowsPerformanceCounterAsMetric.java
```diff
@@ -102,6 +104,8 @@ private void register(Iterable<WindowsPerformanceCounterData> pcsData) {
                     keyToDisplayName.put(key, data.displayName);
                 }
             } catch (Throwable t) {
+                InternalLogger.INSTANCE.trace("error registering %s" , data.displayName);
```
littleaj: Same as above (combine messages to 1 call)

core/src/main/java/com/microsoft/applicationinsights/internal/util/DeviceInfo.java
```diff
@@ -78,6 +80,7 @@ public static String getLocale()
         } catch (Exception e) {
             // Just log - we'll handle it in the fallback path below
             InternalLogger.INSTANCE.trace("Method '%s' could not be found in Locale class - moving to fallback path.", languageTagMethodName);
+            InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(e));
```
littleaj: ^^combine messages to one call

web/src/test/java/com/microsoft/applicationinsights/web/utils/ServletUtils.java
```diff
@@ -21,6 +21,10 @@
```
littleaj: This file has some stacktraces too

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
littleaj: Actually, this needs a few changes. `DEVELOPER_MODE_SYSTEM_PROPERTY_NAME` has the value of the property key. Use the value (not the identifier) in the log message.

I would also change this to a warning. This can only happen if the boolean fails to parse. Change the message to say something along those lines.

core/src/main/java/com/microsoft/applicationinsights/internal/util/ThreadPoolUtils.java
```diff
@@ -58,6 +59,8 @@ public static void stop(ThreadPoolExecutor threadPool, long timeout, TimeUnit ti
         } catch (InterruptedException ie) {
             threadPool.shutdownNow();
             Thread.currentThread().interrupt();
+            InternalLogger.INSTANCE.error("Interrupted exception occured while closing the thread");
```
littleaj: This is not an error. It can be part of normal operation. I wouldn't log anything here; not even trace.

core/src/main/java/com/microsoft/applicationinsights/internal/processor/MetricTelemetryFilter.java
```diff
@@ -55,6 +56,7 @@ public void setNotNeeded(String allNotNeeded) throws Throwable{
             InternalLogger.INSTANCE.trace(String.format("MetricTelemetryFilter: set NotNeeded: %s", notNeeded));
         } catch (Throwable t) {
             InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR, String.format("MetricTelemetryFilter: failed to parse NotNeededNames: %s", allNotNeeded));
+            InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(t));
```
littleaj: Combine with the message on line 56

PR #507, opened 2017-12-27
Fix handling deep stack traces

### LINKED ISSUE #482: ExceptionTelemetry with large / deep stack traces are silently omitted
When using the Logback appender ApplicationInsightsAppender and so indirectly the components of the applicationinsights-core artifact from
   com.microsoft.azure:applicationinsights-logging-logback:1.0.9
   com.microsoft.azure:applicationinsights-core:1.0.9
where is a problem with log statements that include an exception with deep / long stack traces (ExceptionTelemetry).

Deep stack traces are being serialized as event JSON object correctly which has been verified by debugging the applicationinsights-core code.
But when the resulting JSON of the parse stack (shown in Insights Analytics as "details.parsedStack" and stored in Java class ExceptionTelemetry.ExceptionData.ExceptionDetails.parsedStack) exceeds as certain size (which seems to be 32k according to https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Docs/ExceptionData.md), the complete log statement is omitted and not shown in Azure Insights Analytics.

I'm not 100% sure if the ExceptionTelemetry is omitted by applicationinsights-core or by Azure Insights instance itself.
But everything indicates that Azure Insights is receiving the JSON and omitting it remotely.

Omitting an ExceptionTelemetry silently is a very problematic behaviour from a development and operations point of view.
The artifact applicationinsights-core or Application Insights itself should know and respect the Azure Insigts limits and provide a robust behaviour:
If stack traces exceed the given limits, at least the log statement without the deep "parsed stack trace" should be stored and shown in Analytics.
Otherwise application exceptions will never be recognized which is a no-go for a monitoring platform.

dhaval24:
Fix #482 

This pull request handles the deep stack traces, by trimming it down to Maximum of 10 nested exception depth and max size of parsed stack to be equal to 32768. 

In an event of number of nested exceptions exceeding the maximum allowed the first 10 exceptions are considered and a custom exception with a message indicating that maximum depth has exceeded is added as a parent exception.


dhaval24:
@grlima @littleaj could you please have a look to the changes here and comment? This is a critical change that needs to be merged as soon as we can.

dhaval24:
@grlima @littleaj the PR doesn't fix the issue quite so. Please see the referenced issue where the user has tested this out and somehow the issue still remains and is definitely a size related problem. We need to dig deeper. Please do not merge this yet.

dhaval24:
@SergeyKanzhelev could you please provide your opinion on this? I did some math on this and found that the maximum size of ExceptionTelemetry would be 800000 chars. Is this amount of size acceptable on the backend?

SergeyKanzhelev:
@dhaval24 opinion on which part of PR?

dhaval24:
@SergeyKanzhelev we did an experiment on this and the reason is - Our Ingestion endpoint can only accept a payload as big as 2^16 characters. Now since there is no defined rule on trimming out the exceptions. A single exception can be as large as this or nested exceptions can make payload larger than this. Our logic to trim and keep 10 exception is also not sufficient. Further as per the schema each parsed stack can have 32k size and this would make an exception chain of 10 as large as 800000 chars. Our bond schema needs to have some better restriction policy on this.  This can also be a potential issue with .NET SDK. Has any one tried it? Do you know a potential work around for the time being for this issue?

SergeyKanzhelev:
@dhaval24 please post suggestion on required new limits. Thanks for keep digging into it.

dhaval24:
@SergeyKanzhelev what is a good place to post this suggestion ?

SergeyKanzhelev:
@dhaval24 schema issues published at https://github.com/Microsoft/ApplicationInsights-home/issues?q=is%3Aissue+is%3Aopen+label%3Aschema Please file this one there

dhaval24:
@SergeyKanzhelev thanks. Just submitted the issue. https://github.com/Microsoft/ApplicationInsights-Home/issues/181

dhaval24:
Reaching out to community, if anyone has an idea of how long typical java exceptions are, what is the best strategy to trim exceptions and still preserve it's value, please share your knowledge

grlima:
After carefully investigating the issue, we have concluded it would require changes to the backend services to support this change. We will revisit this at a later time if appropriate.

PR #512, opened 2018-01-03
Rethrow threaddeath

### LINKED ISSUE #435: Catching Throwable/Error
I’ve noticed that there are a lot of catch blocks which just catch Throwable and do not rethrow anything. The Java documentation discourages catching and “eating” [Errors](https://docs.oracle.com/javase/6/docs/api/java/lang/Error.html) in general, but this is especially true when  [ThreadDeath](https://docs.oracle.com/javase/6/docs/api/java/lang/ThreadDeath.html) is caught. From the docs: “If `ThreadDeath` is caught by a method, it is important that it be rethrown so that the thread actually dies.”

I'm proposing the following policy moving forward: errors should be allowed to propagate upwards unless there is reason we can recover from the error, e.g. LinkageError.

littleaj:
Fix #435.

For the places where I changed from catching Throwable to catching Exception: It appeared that a Throwable couldn't be generated from this area. Please review these spots carefully and be sure I didn't overlook something.


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlParserUtils.java
```diff
@@ -103,13 +103,17 @@ public static Long getLong(Element element, String elementName) {
...
+            }
+            String strValue = node.getTextContent();
             if (!StringUtils.isNullOrEmpty(strValue)) {
                 Long value = Long.valueOf(strValue);
                 return value;
             }
             return null;
-        } catch (Throwable t) {
+        } catch (Exception e) {
             InternalAgentLogger.INSTANCE.error("Failed to parse attribute '%s' of '%s, default value (true) will be used.'", ENABLED_ATTRIBUTE, elementName);
```
littleaj: Hmm, it appears this error message needs some revision.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlParserUtils.java
```diff
@@ -90,7 +90,7 @@ public static long getLongAttribute(Element element, String elementName, String
                 return value;
             }
             return defaultValue;
-        } catch (Throwable t) {
+        } catch (Exception e) {
             InternalAgentLogger.INSTANCE.error("Failed to parse attribute '%s' of '%s, default value (true) will be used.'", ENABLED_ATTRIBUTE, elementName);
```
littleaj: This also needs to be revised.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlParserUtils.java
littleaj: fixed.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlParserUtils.java
littleaj: fixed.

dhaval24:
@littleaj as a general question on TD, this error occurs when the deprecated method Thread.stop() is called.  I am not aware of other scenarios when this would occur. Can you point me to the instance on the code base when Thread.stop() is called or could be called transitively ?

littleaj:
@dhaval24, `Thread.stop()` is deprecated, but it's not scheduled to be removed. It's unsafe, but (I'm told) can sometimes be useful. See [the article](https://docs.oracle.com/javase/8/docs/technotes/guides/concurrency/threadPrimitiveDeprecation.html) linked in the [javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#stop--).

This is another example of high severity/low priority type bug. It's unlikely to happen, but when it does we need to prevent the application from entering an unexpected and possibly unrecoverable state.

dhaval24:
@arthur I agree that we should absolutely fix it. I was asking to improve my knowledge on this topic.

littleaj:
@dhaval24, I actually don't know of any situations when this would occur. I learned this technique earlier in my career from people who knew Java much better than I did. In "normal" application code, you shouldn't ever need to catch a generic `Throwable`. In javaagent code, however, it is common to catch `Throwable` and rethrowing `ThreadDeath` is part of that pattern to keep the rest of the application working as expected.

dhaval24:
@littleaj thank you very much for the explanation. We should also probably mention this in our contribution guide about best practices.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/logger/InternalAgentLogger.java
```diff
@@ -85,7 +85,7 @@ public synchronized void initialize(String loggerLevel) {
             } else {
                 loggingLevel = LoggingLevel.valueOf(loggerLevel.toUpperCase());
             }
-        } catch (Throwable t) {
+        } catch (Exception e) {
```
dhaval24: @littleaj these were the places you referred that doesn't need throwable correct?

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -507,7 +527,7 @@ private StringBuilder fetchExplainQuery(String commandName, Object object) {
                 }
                 explainSB.deleteCharAt(explainSB.length() - 1);
             }
-        } catch (Throwable t) {
+        } catch (Throwable t) { // FIXME can this be SQLException?
```
dhaval24: @littleaj this could be SQLException. But I would leave it to Throwabl /Exception for now. Might be some string operations can lead to other exceptions not sure.

dhaval24:
@littleaj what is stopping this PR from being merged? There are merge conflicts which could be resolved and then this PR could be completed.

Issue #444, opened 2017-10-09
Request time is not measured correctly for async requests


rpiotrow:
When request returns `CompletableFuture` or is asynchronous in some other way `WebRequestTrackingFilter` will measure only time until `CompletableFuture` is returned, but request is not completed after that. Additional work (defined in `CompletableFuture`) will be done to return response to client - this time won't be measured by `WebRequestTrackingFilter`.

dhaval24:
@rpiotrow Thank you for letting us know the issue. Can you please provide us with additional steps to reproduce this issue?

rpiotrow:
Sure, I'll prepare simple example and attach it here in a day or two.

dhaval24:
@rpiotrow sure this would be great. By the way is this something like a new feature introduced in JDK 9?

rpiotrow:
I'm attaching example application. To run it just unarchive it and invoke `mvn spring-boot:run` (assuming you have instaled Maven). To enable sending metrics to Application Insights you need to `add APPLICATION_INSIGHTS_IKEY` environment variable with instrumentation key from Azure (e.g. `export APPLICATION_INSIGHTS_IKEY="xxxx-xxxx-xxxx-xxxx-xxxx"`)

[sample-app.zip](https://github.com/Microsoft/ApplicationInsights-Java/files/1374396/sample-app.zip)

When making request using curl in console:
```
time curl -s http://localhost:8080/
Hello World!
real    0m2.012s
user    0m0.000s
sys     0m0.004s
```

and in the Application Insights in Azure:
![azure ai time 2017-10-10 22-50-30](https://user-images.githubusercontent.com/355631/31424847-d1687a84-ae5c-11e7-86ca-4862373e4c2d.png)



@dhaval24 asynchronous requests are part of Servlet 3.0 specification (not very new thing), `CompletableFuture` is from Java 8.

dhaval24:
@rpiotrow thank you very much for submitting a repro. We will have a look and assign it to appropriate milestone.

dhaval24:
@rpiotrow do you have a suggested fix for this issue? I know it has been hanging around here for a while and we haven't been able to catch up on this because of High P. issues. Let me know if there is a fix in your mind. We are also very open to accept contributions so please feel free to raise a PR if you wish to

rpiotrow:
[...]

Issue #422, opened 2017-09-11
AI Interceptor overwrite default configuration of Spring MVC


mkaszub:
According to the [documentation ](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-spring-mvc-auto-configuration) of Spring Boot it automaticly configure Spring MVC. 

> If you want to keep Spring Boot MVC features, and you just want to add additional MVC configuration (interceptors, formatters, view controllers etc.) you can add your own @Configuration class of type WebMvcConfigurerAdapter, but without **@EnableWebMvc**

If we take look at the applicationinsights.web pacakge there is a class called InterceptorRegistry

https://github.com/Microsoft/ApplicationInsights-Java/blob/420ace24b444a6da61110fcbec7c98e1a3a63c73/web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java#L33

So it looks like @EnableWebMvc annotation breaks auto configuration feature.

Reproduction :
 * Add static content under resource folder
 * Add reference to the application insights web package

Expected:
 * Spring Boot Autoconfiguration will configure static content for serving

Actual:
 * Static content is not served

Additional links:
 https://stackoverflow.com/questions/46099116/applicationinsights-breaks-default-spring-boot-configuration-for-static-content

dhaval24:
@mkaszub I totally agree with the documentation part that we should not use `@enablewebmvc` here as per the SpringBoot docs. However, the tradition spring frameworks still requires this I believe. Removing this annotation will fix the problem for SpringBoot but at the same time break traditional Spring. We need to think more about this. Do you have a potential solution which can possibly work in both the scenarios ?

cc: @gavlyukovskiy we should look at this too before making we finalize starter.

Apologies for getting so late in the game here @mkaszub

gavlyukovskiy:
I didn't use this class in the starter itself, it won't be applied unless user explicitly imports `InterceptorRegistry`. 

Although inside the starter we can auto-configure `RequestNameHandlerInterceptorAdapter` instead, so that it will be included if user enabled web mvc.

@mkaszub did you explicitly import this class?

dhaval24:
[...]

PR #518, opened 2018-01-09
Added application insights starter for spring boot

gavlyukovskiy:
I want to contribute new module for Spring Boot users.
Initially issue was about wrong setting of application name in Spring Boot applications, but then I thought it would be nice to configure application insights like Spring Boot users used to do it - through [spring boot configuration](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html).

This starter takes care of proper initialization of `TelemetryConfiguration` and `TelemetryClient`. Basically only setting `azure.application-insights.instrumentation-key` is required to start using Application Insights with Spring Boot.

I tried to make tests self explanatory but here is a couple of features I've made:
1. Ability to configure Application Insights using `application.properties`
```
azure.application-insights.enabled=true|false
azure.application-insights.instrumentation-key=<key>
azure.application-insights.quick-pulse.enabled=true|false
azure.application-insights.channel.in-process.developer-mode=true|false
azure.application-insights.channel.in-process.flush-interval-in-seconds=<number>
azure.application-insights.channel.in-process.max-telemetry-buffer-capacity=<number>
azure.application-insights.channel.in-process.max-transmission-storage-files-capacity-in-mb=<number>
azure.application-insights.channel.in-process.throttling=true|false
azure.application-insights.logger.type=console|file
azure.application-insights.logger.level=all|trace|info|warn|error|off
```
2. Adding custom modules by declaring as a bean:
```
@Bean
public TelemetryModule myTelemetryModule() {
    return new TelemetryModule() {
        @Override
        public void initialize(TelemetryConfiguration configuration) {
            // do some staff here
        }
    };
}
```
3. Initializing `WebRequestTrackingFilter` with right application name.
4. Adding spring, spring boot versions as tags and resolving right IP if running with spring cloud infrastructure.

I also have added a test case for which I need #516 to be merged -
ApplicationInsightsTelemetryAutoConfigurationTests#shouldReloadInstrumentationKeyOnTelemetryClient.

As I heard you don't have much experience with Spring Boot so if something is not clear I'm free to answer your questions.

- [x] Design discussion issue in comments of #411 and internal discussions
- [ ] CHANGELOG.md updated


dhaval24:
@gavlyukovskiy this is phenomenal. Though, I haven't looked into details of this, work seems really promising. Indeed it would be a very very valuable contribution for Application Insights Java SDK. I would take time to review it and leave my comments. Adding more reviewers to this PR.

MicNeo:
As a Spring Boot user, which is at the moment integrating Application Insights with it, I'm waiting for this enhancement very much :)

dhaval24:
@gavlyukovskiy I was experimenting with SpringBoot and it's realm of creating Uber jars. I tried an experiment where we can specify Java agent as maven dependency and then modifying some configurations to embed the agent inside the uber jar. However, we still get the class not found error. I will post a seperate issue on the steps taken to reach this stage. Let me know if you have worked with Agents in spring boot. Ideally I would love to have this starter also have java agent inside it so that user can get full end to end Java monitoring experience with just single dependency which can easily then be a part of Uber Jar (deployable units as said in SpringBoot)

gavlyukovskiy:
Added documentation and readme, waiting for review.
@dhaval24 Regarding I don't see way to include agent in starter as spring boot doesn't support that, as I understood from https://github.com/spring-projects/spring-boot/issues/6626 some hacks can be used but only in client's build script.

dhaval24:
Thanks @gavlyukovskiy for pushing the changes. We will do a deep review for this, as this would be a paradigm shift :-) adding more people to review

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+@AutoConfigureAfter(ApplicationInsightsTelemetryAutoConfiguration.class)
+public class ApplicationInsightsWebMvcAutoConfiguration {
+
+    @Bean
+    public FilterRegistrationBean webRequestTrackingFilterRegistrationBean(WebRequestTrackingFilter webRequestTrackingFilter) {
+        FilterRegistrationBean registration = new FilterRegistrationBean();
+        registration.setFilter(webRequestTrackingFilter);
+        registration.addUrlPatterns("/*");
+        registration.setName("webRequestTrackingFilter");
+        registration.setOrder(Ordered.HIGHEST_PRECEDENCE + 10);
```
dhaval24: Why do we need HIGHEST_PRECEDENCE+10. All the AI WebRequestTrackingFilter needs is that it should be first in the Filter Chain. Are there any spring framework filters that run even before the highest precedence? Also a question - is filter the best way to intercept the request execution time and other details? We have noticed an issue with Async request which returns a future and starts execution in different thread. In that case AI behaves weirdly. #444

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+        return registration;
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public WebRequestTrackingFilter webRequestTrackingFilter(@Value("${spring.application.name:application}") String applicationName) {
+        return new WebRequestTrackingFilter(applicationName);
+    }
+
+    @Bean
```
dhaval24: This should also be conditional as far as I know. There should be a way to disable quick pulse if needed. In Standard xml implementation we have QuickPulse XML element which can be set to disabled. We need to have that property here too and then se this bean on conditional with that.

azure-application-insights-spring-boot-starter/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+# Instrumentation key from the Azure Portal
+azure.application-insights.instrumentation-key=00000000-0000-0000-0000-000000000000
+
+# Logging type [console, file]
+azure.application-insights.logger.type=console
+# Logging level [all, trace, info, warn, error, off]
+azure.application-insights.logger.level=info
+
+# Enable/Disable default telemetry modules
```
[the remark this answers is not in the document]
gavlyukovskiy: By default they are all turned on, added these properties just in case user wants to disable those.

azure-application-insights-spring-boot-starter/README.md
```diff
@@ -0,0 +1,140 @@
...
+        return new MyContextInitializer();
+    }
+}
+```
+
+
+#### Configure more parameters using `application.properties`
+```properties
+# Enable/Disable tracking
+azure.application-insights.enabled=true
```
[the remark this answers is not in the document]
gavlyukovskiy: By default it's true and should not be specified to run application. Only single required property is `azure.application-insights.instrumentation-key`

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsProperties.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public Logger getLogger() {
+        return logger;
+    }
+
+    public void setLogger(Logger logger) {
+        this.logger = logger;
+    }
+
+    public static class Logger {
```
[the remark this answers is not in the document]
gavlyukovskiy: yes, there are defaults. Will change to ERROR

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * @author Arthur Gavlyukovskiy
+ */
+@Configuration
+@ConditionalOnProperty(value = "azure.application-insights.instrumentation-key")
+@EnableConfigurationProperties(ApplicationInsightsProperties.class)
+@ConditionalOnClass(TelemetryConfiguration.class)
+@Import({
+        ApplicationInsightsModuleConfiguration.class,
+        ApplicationInsightsWebModuleConfiguration.class
```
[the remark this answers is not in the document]
gavlyukovskiy: We can add `@ConditionalOnWebApplciation` on whole `ApplicationInsightsWebModuleConfiguration`,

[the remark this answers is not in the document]
dhaval24: @gavlyukovskiy Yes that is one way. However, there are instances when the User uses Application Insights only for pushing logs or pushing custom events and doesn't require HTTP tracing. In that scenario, I think this won't work. The better way can be to have a property which says disable/enable WebRequestTracking (By default it is set as enabled). and then condition on it. Thoughts?

azure-application-insights-spring-boot-starter/README.md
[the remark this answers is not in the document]
dhaval24: Sure, it makes sense to specify all the available properties to tweak in the readme, however please make a note there that all these properties are turned on by default so that user can have an understanding.

azure-application-insights-spring-boot-starter/README.md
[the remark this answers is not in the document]
dhaval24: Makes sense.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
gavlyukovskiy: +10 just gives the user the ability to have some interceptor before ours. It may be useful in some cases for example when user want to log request or to measure overhead given by this filter. For instance we have filter called `FinalCatchFilter` which is the very first filter and logs any exception that was not handled in the chain.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
gavlyukovskiy: Oh, probably I missed this one, actually I had a way to disable quick pulse but then I removed it because it seemed to me useless thing, I will return it back.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+    }
+
+    @Bean
+    public TelemetryClient telemetryClient(TelemetryConfiguration configuration) {
+        return new TelemetryClient(configuration);
+    }
+
+    @Bean
```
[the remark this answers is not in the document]
gavlyukovskiy: So this is not the `TelemetrySampler` instance but `TelemetryProcessor`. It actually means that we can't do it `@ConditionalOnMissingBean` and user won't be able to override this one, at least we can do it configurable through `application.properties`

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Bean
+    @ConditionalOnOperatingSystem(OperatingSystem.WINDOWS)
+    @ConditionalOnProperty(value = "azure.application-insights.default-modules.ProcessPerformanceCountersModule.enabled", havingValue = "true", matchIfMissing = true)
+    public ProcessPerformanceCountersModule processPerformanceCountersModule() {
+        try {
+            return new ProcessPerformanceCountersModule();
+        }
+        catch (Exception e) {
+            throw new IllegalStateException("Could not initialize Windows performance counters module", e);
```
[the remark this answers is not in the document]
gavlyukovskiy: I will change this to logging error.

azure-application-insights-spring-boot-starter/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+apply from: "$buildScriptsDir/common-java.gradle"
+apply from: "$buildScriptsDir/publishing.gradle"
+
+archivesBaseName = 'azure-application-insights-spring-boot-starter'
+
+dependencies {
```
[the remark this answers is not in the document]
gavlyukovskiy: I think we have everything needed here to use with spring boot, only new dependency is assertj and junit of newer version and both `testCompile`.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: Okay, however my question was more centered around by providing this kind of facility would it mess up the actual numbers of request execution time? If that is the case we should be very careful with this. I would probably say no to it if it messes up the metrics.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: Yes please do add. There should be ability to disable quick pulse if needed.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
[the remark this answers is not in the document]
dhaval24: Thanks.

azure-application-insights-spring-boot-starter/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    testCompile('org.junit.jupiter:junit-jupiter-api:5.0.2')
+    testCompile('org.junit.jupiter:junit-jupiter-params:5.0.2')
+    testCompile('org.junit.jupiter:junit-jupiter-engine:5.0.2')
+    testCompile('org.springframework.boot:spring-boot-starter-test:1.5.9.RELEASE')
+    testCompile('org.assertj:assertj-core:3.9.0')
+}
+
+// region Publishing properties
+
+projectPomName = project.msftAppInsightsJavaSdk + " Spring Boot starter"
```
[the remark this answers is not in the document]
gavlyukovskiy: I took web module to copy this I intentionally did not include shadowJar because I have no need in it. 

ProjectPomName actually is equal to `Microsoft Application Insights Java SDK Spring Boot starter`, I don't know why other modules use `msftAppInsights` and then add `' Java SDK'`, for example `project.msftAppInsights + " Java SDK Core" == project.msftAppInsightsJavaSdk + ' Core'`

dhaval24:
@gavlyukovskiy have you looked at the retry PR #561 ? Have you addressed additional configuration parameters introduced there? I know at least one introduced there, which allows the user to configure the number of Instant retries the channel can make. Please also modify your code to allow that configuration property and also make sure that whatever properties were allowed to configure using `InProcessTelemetryChannel` are still intact.

azure-application-insights-spring-boot-starter/build.gradle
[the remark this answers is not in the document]
dhaval24: Sure, shadow jar is not needed in this case. Shadowing is usually done only for dependencies that we need to encapsulate like Apache HTTP Client. In this case the SpringBoot Starter jar will be shipped as a jar and it will pull all the transitive dependencies once loaded into the SpringBoot Project, so we do not need that.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
gavlyukovskiy: @dhaval24 should quick pulse depend on web application or it's applies to non web as well?

azure-application-insights-spring-boot-starter/build.gradle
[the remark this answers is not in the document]
dhaval24: @gavlyukovskiy Isn't junit 5.x still in beta?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: @gavlyukovskiy frankly, I do not know about this on the top of my head. I haven't worked on this part of the code base (QuickPulse), however from an end-user perspective QuickPulse / Live View on portal shows information such as Incoming request rate, request duration, request failure rate, CPU usage, Memory usage, Exception Rate, Dependency rate, dependency failure rate and Dependency call rate. Based on this it seems like QuickPulse should not have explicit dependency on web application. Only web related features like Request parameters should not be populated on the portal.

azure-application-insights-spring-boot-starter/build.gradle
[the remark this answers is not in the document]
gavlyukovskiy: no, it was released in September and actually yesterday they did second GA release - 5.1.0 :)

azure-application-insights-spring-boot-starter/README.md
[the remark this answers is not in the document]
gavlyukovskiy: I was thinking about calling them default already makes them turned on by default ;) I will add note that all those enabled by default

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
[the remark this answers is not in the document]
gavlyukovskiy: I remember why I did it, actually you can't just log error here because bean instance must be returned. I'm not quite sure how to fix this properly.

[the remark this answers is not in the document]
gavlyukovskiy: Although I don't see any reason why it would throw an exception. `ProcessBuiltInPerformanceCountersFactory` is instance of `WindowsPerformanceCountersFactory` and if something goes wrong inside this factory it returns empty list.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
gavlyukovskiy: WDYT about `azure.application-insights.web.enabled=true/false` as umbrella for all web modules? But we should state that 
```
azure.application-insights.default-modules.WebRequestTrackingTelemetryModule.enabled=true
azure.application-insights.default-modules.WebSessionTrackingTelemetryModule.enabled=true
azure.application-insights.default-modules.WebUserTrackingTelemetryModule.enabled=true
azure.application-insights.default-modules.WebPerformanceCounterModule.enabled=true
azure.application-insights.default-modules.WebOperationIdTelemetryInitializer.enabled=true
azure.application-insights.default-modules.WebOperationNameTelemetryInitializer.enabled=true
azure.application-insights.default-modules.WebSessionTelemetryInitializer.enabled=true
azure.application-insights.default-modules.WebUserTelemetryInitializer.enabled=true
azure.application-insights.default-modules.WebUserAgentTelemetryInitializer.enabled=true
```
will be implicitly turned off.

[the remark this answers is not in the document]
dhaval24: Umbrella should provide an ease, but thinking more about it how would it affect if the user does some mess up and keeps the umbrella property turned off but enables one of the sub properties in the umbrella ? However, I agree this looks more clean.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
gavlyukovskiy: In general no one cares of order which is `Ordered.LOWEST_PRECEDENCE` by default. I see only two spring filters that is higher than `WebRequestTrackingFilter` -`MetricsFilter` and `OrderedCharacterEncodingFilter`, both of them are `Ordered.HIGHEST_PRECEDENCE` so you can't be sure that even with highest order yours win, also order you your concrete filter becomes non deterministic, which is not good thing.

If user specifies order that he is most probably know what is he doing and leaving some buffer is fine.

Preferable I would not touch it unless someone will give use case where this should be changed.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
gavlyukovskiy: Should be definitely in `application.properties`, `@ConditionalMissingBean` is conditional that checks that bean of this type was not provided by the user. For instance if you have few implementations of `TelemetrySampler` by default `FixedRateTelemetrySampler` will be used, but if user has defined own bean like this
```
@Bean
public TelemetrySampler myTelemetrySampler() {
    return new MyTelemetrySampler();
}
```
our won't be created.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
gavlyukovskiy: I don't think we should merge it somehow, if you say web is disabled then every web module is disabled, if you want to disable some specific web module - go and do it, no need to disable whole web and then enable specific ones. There are only 9 of them :)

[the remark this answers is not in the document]
dhaval24: yes, this looks fine. If you disable the web then every web module is disabled or else, you can disable individual ones as needed.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
dhaval24: Yes, I do get that. In this case we should not keep this bean inside. And let user create a processor if needed. One question, if the user wants to use a built in processor, say fixed rate sampling telemetry processor, how does he do that? Can we specify it in the properties file? If yes can we configure different attributes of the processor too with Properties file?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: @gavlyukovskiy I am not saying that we should provide the handle for user to tweak the filter. I only want to ensure that AI filter gets the highest precedence whenever possible but as you said it can be non deterministic, keeping the highest precedence + 10 should be a good enough guess in most cases.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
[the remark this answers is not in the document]
MicNeo: Seems like very unlikely situation. Can't you return null in such case?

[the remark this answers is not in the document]
gavlyukovskiy: I don't think it's possible, in best case we can either create conditional where we try to create instance and return true only if it was successfully created, or have message with clear action to do, something like 'We are failed to create performance metric, please set `azure.application-insights.default-modules.ProcessPerformanceCountersModule.enabled=false` to avoid this error message'. But in general I don't see any chance for this exception to be thrown in real code. @dhaval24 what do you think?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
gavlyukovskiy: We can add something like `azure.application-insights.telemetry-processor.type` in properties if we have limited number of available processors. The problem is that you may have multiple processors at the same time. We can create this property with allowed values `[fixed_rate, foo, bar, none]` and default will be `fixed_rate`. Tell me what you exactly need and I can find some ideas how we can implement this.

For example data source in spring boot configured like that, first you chose type of data source:
```
spring.datasource.type=<datasource class>
```
and based on the type you chose you can use appropriate set of properties that you can use based on which data source you have chosen. If you chose `org.apache.tomcat.jdbc.pool.DataSource` then you can use `spring.datasource.tomcat.*` to configure tomcat data source parameters; if `com.zaxxer.hikari.HikariDataSource` then properties `spring.datasource.hikari.*` are used.

gavlyukovskiy:
I fixed all issues that we had decision on. Only two things left:
1. Telemetry processors configuration including new sampling module.
2. Exception thrown from `ProcessPerformanceCountersModule`.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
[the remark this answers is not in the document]
MicNeo: @gavlyukovskiy from technical point of view it's possible, such bean declaration is 100% valid and it works:
`@Bean
	public CustomBean customBean(){
		return null;
	}`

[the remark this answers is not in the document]
gavlyukovskiy: You're right, it's valid. But then I should check that module is not null when initializing them in, I can do this as well, but it just seem useless to me to add a lot of checks that verify situation that could not be exist. That's why I throw Illegal**State**Exception here. If there are will be issue with that it indicates bug on our side and you can't save user from all bugs you may ever have. Although user has simple switch to disable this module so it won't be a blocker anyways.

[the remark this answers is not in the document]
dhaval24: @gavlyukovskiy I am fine with failing and asking to set the message  please set azure.application-insights.default-modules.ProcessPerformanceCountersModule.enabled=false to avoid this error message'. Please add this conditional.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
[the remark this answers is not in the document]
dhaval24: Currently there is only FixedRateSampling Telemetry Processor for sampling. In future we would have adaptive sampling telemetry processor. Also you can look into this package `com.microsoft.applicationinsights.internal.processor` for other built in processors we have today. We can alteast create properties like what you mentioned for this built in processors. Does it make sense?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -56,7 +56,9 @@ public FilterRegistrationBean webRequestTrackingFilterRegistrationBean(WebReques
 
     @Bean
     @ConditionalOnMissingBean
-    public WebRequestTrackingFilter webRequestTrackingFilter(@Value("${spring.application.name:application}") String applicationName) {
+    public WebRequestTrackingFilter webRequestTrackingFilter(@Value("${spring.application.name:application}") String applicationName,
+            // we have implicit dependency on configured telemetryConfiguration here
+            @SuppressWarnings("unused") TelemetryConfiguration telemetryConfiguration) {
```
dhaval24: @gavlyukovskiy  Can you explain the need of this? I do understand in the changes above you made this conditional on web application which is auto configured after the default configuration. Is this what you are referring to and believe that TelemetryConfiguration would always be initialized ?

gavlyukovskiy: `@AutoConfigureAfter` and `@AutoConfigureBefore` sets the order of creating bean definitions, but the beans itself are created in different order which controlled by dependency graph. I set explicit dependency on `TelemetryConfiguration` to force it to be created before `WebRequestTrackingFilter` to ensure that call `TelemetryConfiguration#getTelemetryModules` will return all modules.

dhaval24: @gavlyukovskiy thank you very much. Also about the sampling do you have any idea on getting that in?

dhaval24: Yes so if processor returns false then the telemetry is dropped and doesn't hit transmission channel. So processor controls if telemetry should be sent or not. Channel is responsible for ultimately pumping the telemetry and it has several characterstics which tell what is the end point what is the buffer size etc.

gavlyukovskiy: And for what `TelemetrySampler` is used?

dhaval24: Yes please drop it. It is not a recommended way to sample. We should mark it as deprecated also :-)

gavlyukovskiy: What do you think about moving these types of telemetry to enum:
* Dependency
* Event
* Exception
* PageView
* Request
* Trace
? With that we can make autocomplete in properties.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -64,20 +72,8 @@
      */
     public FixedRateSamplingTelemetryProcessor() {
         this.samplingPercentage = 100.00;
-        this.includedTypes = new HashSet<Class>();
-        this.excludedTypes = new HashSet<Class>();
-        try {
-            this.allowedTypes = new HashMap<String, Class>() {{
```
gavlyukovskiy: I little bit refactored two things here, first is double brace initialization which is not considered as good practice because of creating of anonymous class. Second is using .class instead of Class#forName, there is no reason to use this as all classes are present in the core module.

gavlyukovskiy:
Pushed changes regarding in process channel and sampling. Going to sleep, there is almost 4 AM here :)

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: @gavlyukovskiy are you referring to converting this class into ENUMS inside core SDK or specifically in this module? In either cases I think that would be a major refactoring. Could you add more details on why do we need this? What benefits it provides and what would we be loosing here if we keep it to what it is currently. Further all these data types are autogenerated by a C# script I think. We might need to make changes there to do this. Also there might be upcoming change with Schema V 3.0 . Let's discuss more on this.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/conditional/OperatingSystem.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.microsoft.applicationinsights.boot.conditional;
+
+/**
+ * List of available for operating system for condition.
+ *
+ * @author Arthur Gavlyukovskiy
+ */
+public enum OperatingSystem {
```
dhaval24: @gavlyukovskiy what about MacOS ? I have seen certain things very different here. For example some UNIX perf counters are collected by parsing /proc/stats file. But MacOS X doesn't have any such file? Also as of now I haven't been able to successfully set up collectd in Mac OS. What should be the way to differentiate this here. Do we need another Constant here for MacOS. This is not very important for now as most server code is never ran on MacOS but rather on Linux/Unix system but people use mac for dev time and can throw misleading information. What is your thought here?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -126,12 +124,17 @@ public InProcessTelemetryChannel() {
      *                                   Note, value should be between MIN_MAX_TELEMETRY_BUFFER_CAPACITY and MAX_MAX_TELEMETRY_BUFFER_CAPACITY inclusive
      */
     public InProcessTelemetryChannel(String endpointAddress, boolean developerMode, int maxTelemetryBufferCapacity, int sendIntervalInMillis) {
+        this(endpointAddress, null, developerMode, maxTelemetryBufferCapacity, sendIntervalInMillis, true);
+    }
+
+    public InProcessTelemetryChannel(String endpointAddress, String maxTransmissionStorageCapacity, boolean developerMode,
```
dhaval24: @gavlyukovskiy this means we need this public constructor to be introduced in the channel? Is there a way to avoid this? I see that everything else is not configurable without this constructor like throtlling etc parameters. If you it is needed to have a constructor like this we need it's tests also.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsProperties.java
```diff
@@ -103,6 +128,114 @@ public void setLogger(Logger logger) {
         this.logger = logger;
     }
 
+    public static class Channel {
```
dhaval24: @gavlyukovskiy I think we need default values here. If the user does not specify this property the default InProcessTelemetryChannel should be created. Does `@ConditionalOnMissingBean` annotation you added address this, in case user doesn't specify have this property in his properties file will the default channel get created (I hope so it does based on Line 85 in `TelemetryConfigurationFactory` Enum). However if the user does specify this property then you are using a parametrized constructor which implies that user has to now provide all the values and has lost the ability to tweak single property of channel say turn on developer mode or change end point address. Please correct me if I am wrong here, but if that is indeed the case we need to make sure there are default values for this and they should match default values of `InProcessTelemetryChannel` class.

Further we also need to set conditionals on the value the user specifies. I know that limit enforcer does this but setting conditions here would also be very useful. What do you think ?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
```diff
@@ -0,0 +1,19 @@
+package com.microsoft.applicationinsights.internal.channel.samplingV2;
+
+public enum TelemetryType {
```
dhaval24: I am okay with this change. This is what you mentioned enables auto completion correct?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: @gavlyukovskiy I am okay with this unless this is something globally exposed. Autocompletion would be desirable but I would like to understand what could be the implications if this is globally exposed because we won't be able to remove this until next major version and making it public then people can take explicit dependency on this enum and if we want to make it public in any situation then it should not reside in this processor but rather should have it's home somewhere else where it is more easy to locate and used in other parts of the code base.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
gavlyukovskiy: Yes

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/conditional/OperatingSystem.java
gavlyukovskiy: Actually I need only `WINDOWS` for specific performance module to place it in conditional, I added `UNIX` for demonstrating purpose. If you will have modules that should be enabled only for Unix, but not for MacOS and vise versa we can add this one later.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsProperties.java
gavlyukovskiy: I added default values for every property in latest commit. Yes, I added `@ConditionalOnMissingBean` for user to be able to override this bean if needed. Even now with all properties configurable there are could be use-cases for that, for example using custom channel that will log every telemetry into the log or something like this.

gavlyukovskiy: What are the conditions you're asking? Something like that the maximum buffer size should be between 1 and 1000?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/conditional/OperatingSystem.java
dhaval24: Yes there is a module Application Insights CollectD exporter module. Look at this class `com.microsoft.applicationinsights.collectd.ApplicationInsightsWriter`

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsProperties.java
dhaval24: @gavlyukovskiy Yes exactly this is what I meant. If user uses inprocess telemetry channel and specifies different values then what channel expects then we should take the default value. As I mentioned that there is a logic inside the channel using `LimitsEnforcer` class however in order to have our starter to be very self sufficient and robust I think we should have this checks here. What do you think?

azure-application-insights-spring-boot-starter/README.md
```diff
@@ -129,6 +129,26 @@ azure.application-insights.logger.type=console
 # Logging level [all, trace, info, warn, error, off]. Default value: error.
 azure.application-insights.logger.level=error
 
+# Enable/Disable developer mode, all telemetry will be sent immediately without batching. Significantly affects performance and should be used only in developer environment. Default value: false.
+azure.application-insights.channel.in-process.developer-mode=false
+# Endpoint address, Default value: none.
```
dhaval24: I think default this is little bit misleading to users. If the user doesn't specify the endpoint address the default endpoint would be Application Insights backend service which is "https://dc.services.visualstudio.com/v2/track". We should mention something that - if no value is specified traffic would sendto application insights default backend service.

gavlyukovskiy: This was the only property I was not quite sure. Don't you think this is something internal and should not be changed by simply using property value?

dhaval24: @gavlyukovskiy well, if the user wants to send the data to some different end point he or she should have flexibility to do so. What do you think? We can keep this as it is for now and change based on others feedback here. But I think there is no harm personally

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/conditional/OperatingSystem.java
gavlyukovskiy: As I can see this one is not enabled by default. This conditional is used for cases when you want to enable this module by default, but only for specific OS.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
gavlyukovskiy: Added test with default values that are coming from spring-boot.

azure-application-insights-spring-boot-starter/README.md
gavlyukovskiy: Added default value.

gavlyukovskiy:
@dhaval24 do you want starter to throw exception if some value exceeds limits or behave like with xml configuration when closest value is chosen? Now it has no additional logic except type safe binding that you're not able to put a string in property that binds to an integer. 
I can write test for `InProcessTelemetryChannel`, that if you set property `azure.application-insights.channel.in-process.flush-interval-in-seconds=600` it will fall back to 300 seconds because it's maximum allowed value, but this is not my logic, it's how it already works.

gavlyukovskiy:
I would say this is overhead for starter, any value in properties is passed
directly to InProcessTelemetryChannel and it falls back to defaults. From
starter point of view you can not forbid user from writing values behind
the limits. Only way is to add validation that we already have in channel
itself.

2018-02-25 19:47 GMT+02:00 Dhaval Doshi <notifications@github.com>:

dhaval24:
@gavlyukovskiy could you please take out sometime to see the propositions in #573 regarding quickpulse bean. Do you experience the same behavior. It would be worth while to ensure this condition. Thanks

dhaval24:
@gavlyukovskiy changes regarding `@dependsOn` bean looks great. It will always make sure that QuickPulse is initialized after the `telemetryConfiguration` bean is initialized.

gavlyukovskiy:
@dhaval24 I don't see how constants class will solve problem of changing defaults. Once you made some values as public they can be used regardless of its location and you will be in the same situation. Although I don't think changing value of a constant is a breaking change. Removing this constant would be breaking change, but I doubt one will rely on the specific value of some library constant.
As an alternative we can leave them private and just copy into starter defaults.

dhaval24:
@gavlyukovskiy I think let's just keep as it is right now (i.e public). I also think the same that changing value of constant would not be a breaking change, but I would still wait here for other reviewers to sign off on this. If we decide going public then we can refactor to move constants out in seperate package as I mentioned. I can start doing that refactoring if there is an agreement on that, or else we might just keep this as private and copy it into starter defaults for now. thoughts? 

@grlima @littleaj and @Dmitry-Matveev I want your opinion on this too.

dhaval24:
@gavlyukovskiy could you please merge the latest master and push the changes back resolving the merge conflicts. Also what would be the best way if I want to commit on top of this. I think we might be able to create a PR against a seperate branch and then I will also work together to polish this out.

dhaval24:
@gavlyukovskiy we should also take a look at #422

dhaval24:
@gavlyukovskiy could you please change to create this PR against spring-boot-starter branch. That way it would easy for me to add commits on top of this and collaborate to make this final.

azure-application-insights-spring-boot-starter/README.md
```diff
@@ -133,16 +133,16 @@ azure.application-insights.logger.level=error
...
 # Endpoint address, Default value: https://dc.services.visualstudio.com/v2/track.
 azure.application-insights.channel.in-process.endpoint-address=https://dc.services.visualstudio.com/v2/track
-# Maximum count of telemetries that will be batched before sending. Default value: 500.
+# Maximum count of telemetries that will be batched before sending. Must be between 1 and 1000. Default value: 500.
 azure.application-insights.channel.in-process.max-telemetry-buffer-capacity=500
-# Interval to send telemetry. Default value: 5 seconds.
+# Interval to send telemetry. Must be between 1 and 300. Default value: 5 seconds.
 azure.application-insights.channel.in-process.flush-interval-in-seconds=5
-# Size of disk that we can use. Default value: 10 megabytes.
+# Size of disk that we can use. Must be between 1 and 1000. Default value: 10 megabytes.
```
dhaval24: Does't the default and maximum value both seem to be too low. What do you think? Should we bump up this capacity?

azure-application-insights-spring-boot-starter/README.md
```diff
@@ -133,16 +133,16 @@ azure.application-insights.logger.level=error
...
 azure.application-insights.channel.in-process.max-transmission-storage-files-capacity-in-mb=10
 # Enable/Disable throttling on sending telemetry data. Default value: true.
 azure.application-insights.channel.in-process.throttling=true
 
-# Percent of telemetry events that will be sent to Application Insights. Default value: 100% (all telemetry events).
+# Percent of telemetry events that will be sent to Application Insights. Must be between 0.0 and 100.0. Default value: 100 (all telemetry events).
```
dhaval24: We should add this comment here : 
"All sampling percentage must be in a ratio of 100/N where N is a whole number (2, 3, 4, …). E.g. 50 for 1/2 or 33.33 for 1/3.
     *  Failure to follow this pattern can result in unexpected / incorrect computation of values in the portal." 

This is important. We can only set specific sampling percentages for UI to be consistent

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsProperties.java
```diff
@@ -243,7 +243,7 @@ public void setSampling(Sampling sampling) {
 
         static class Sampling {
             /**
-             * Percent of telemetry events that will be sent to Application Insights.
+             * Percent of telemetry events that will be sent to Application Insights. Must be between 0.0 and 100.0.
```
dhaval24: same sampling comment here.

azure-application-insights-spring-boot-starter/README.md
gavlyukovskiy: I didn't understand, can it be 2/3?

dhaval24:
@gavlyukovskiy thank you for the modifying the PR. I would merge this in the branch and would be better to work on. I added few more comments. Nitty gritty ones.

azure-application-insights-spring-boot-starter/README.md
gavlyukovskiy: Actually I don't really know what is this property about at all :) probably it's a question to @littleaj that have bumped maximum value from 100 to 1000 recently.

azure-application-insights-spring-boot-starter/README.md
dhaval24: so 100/N = Sampling ratio. 
Eg 100/2  = 50
100/3 = 33
100/4 = 25
100/5 = 20
100/10 = 10 etc.

azure-application-insights-spring-boot-starter/README.md
dhaval24: @gavlyukovskiy  this property is used for disk persistence. When the network fails the telemetry gets stored in the disk and when it is restored we transmit that telemetry to the cloud.

azure-application-insights-spring-boot-starter/README.md
gavlyukovskiy: probably we can create property for N instead? It would be less error prone if we say that N must be between 1 (100%) and 100 (1%).

what about 0%, is it valid?

gavlyukovskiy: just read the documentation for .NET, it can be 1000 (0.1%) as well. Looks really odd to me, I mean that 1/3 is not really 33 or 33.3 or 33.33, it's 33.(3) which you can properly specify in properties and for this reason per documentation I have to chose closest percentage to 100/N.

azure-application-insights-spring-boot-starter/README.md
gavlyukovskiy: @dhaval24 nice to know, added your explanation to the documentation :)

gavlyukovskiy:
Added note regarding sampling percentage.

Also I think it's important to clarify that users should not copy-paste all the properties unless they need to tweak something. 
I'm seeing such behavior with my library, users just copy everything from readme and it's harder to change default values to something more optimal. E.g. if you change default transmission capacity it will not affect users that just copied properties from readme.

dhaval24:
@gavlyukovskiy @littleaj @grlima I am merging this PR in spring-boot-starter branch and we will continue to iterate there.

gavlyukovskiy:
@dhaval24 don't forget to merge #516 as well, without it appenders will not send data AI because loggers are initialized earlier and, most probably, had send at least single event at the moment when `TelemetryConfiguration` was initialized.

dhaval24:
@gavlyukovskiy thanks for reminding. Sure I will just merge that.

PR #528, opened 2018-01-15
add simple trace data test case

Terryxxx:
@grlima @dhaval24 @littleaj 
add simple trace data test case.


test/smoke/testApps/SimpleTraceData/src/main/java/com/microsoft/ajl/simplecalc/SimpleCalculatorServlet.java
```diff
@@ -0,0 +1,120 @@
...
+		if (bc == null) {
+			System.out.println("No parameters given.");
+			response.setStatus(HttpServletResponse.SC_NO_CONTENT);
+			return;
+		}
+		
+		response.setContentType("text/html;charset=UTF-8");
+		renderHtml(bc, response.getWriter());
+		
+		new TelemetryClient().trackTrace("This is first trace message.");
```
dhaval24: Instead of creating a new instance of TelemetryClient for every trackTrace() call, I would suggest to create a Class variable for TelemtryClient and use it for sending multiple trackTrace() calls. It is advisable to minimize the creating of new instance of telemetry client instances as much as possible.

test/smoke/testApps/SimpleTraceData/src/main/java/com/microsoft/ajl/simplecalc/SimpleCalculatorServlet.java
```diff
@@ -0,0 +1,120 @@
...
+
+		new TelemetryClient().trackTrace("This is second trace message.", SeverityLevel.Error);
```
dhaval24: same here  and for the call below this also.

test/smoke/testApps/SimpleTraceData/src/main/java/com/microsoft/ajl/simplecalc/SimpleCalculatorServlet.java
Terryxxx: @dhaval24 Thanks for your advice, I will do it.

PR #531, opened 2018-01-16
Add Telemetry Test Case

Terryxxx:
@dhaval24 @grlima  @littleaj


test/smoke/testApps/SimpleTelemetryData/src/main/java/com/microsoft/ajl/simplecalc/SimpleCalculatorServlet.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import static com.microsoft.ajl.simplecalc.ParameterConstants.*;
+
+/**
+ * Servlet implementation class SimpleCalculatorServlet
+ */
+@WebServlet(description = "Performs given calculation", urlPatterns = { "/doCalc" })
```
littleaj: Instead of duplicating these each time, you can add another Servlet class to the same test. You shouldn't need to create a new testapp unless you're changing how the app works, e.g. servlet with core only, servlet with a filter, spring boot with filter, jax-rs with agent, etc.

Terryxxx: @littleaj Thanks for your comments. I will change my code. And I have a question. Do you mean to write more servlets send different track data?

littleaj: Yes, a test app should have multiple endpoints/servlets which correspond to different track data calls.

Terryxxx:
@littleaj  Hi Arthur, I have updated the PR, Please help check again. Thanks

PR #534, opened 2018-01-19
Adding smoke/functional test framework using Docker

littleaj:
This is the initial framework for smoke testing the SDK. It will lay the framework for future investments into functional testing. See the documentation in `/test/smoke/README.md` for details on the framework.

Keep in mind this is the initial effort. :)


dhaval24:
Adding @SergeyKanzhelev and @cijothomas to have a look at the work as you were working on similar story for DotNet. May be you would be impressed by this framework!

gavlyukovskiy:
Hi, @littleaj, didn't you look at [Test containers](https://www.testcontainers.org/)? I haven't used this one, but I think it can reduce a lot of code of managing docker containers

dhaval24:
@gavlyukovskiy thank you for the suggestion. BTW is it possible for you to try this framework that @littleaj created and suggest the differences you feel?

gavlyukovskiy:
Unfortunately I can't try this right now due to lack of time and for sure I'm not an expert in docker testing so my opinion costs nothing here :) I just know that many projects started to use it and preferably I would rather use this instead of inventing own framework. Though it's really depends on your use cases.

dhaval24:
@gavlyukovskiy thanks for this input. We might have a look on the Test Containers subject to availability of time :-)

littleaj:
@gavlyukovskiy, Thanks for pointing out Test Containers! I did a short investigation of the library and it looks like it would work great for the docker control inside the tests. As @dhaval24 said, it may take some time, but this looks promising.

Issue #536, opened 2018-01-25
Where is session cookie set?


gondar:
It seems there is code for creating a cookie, but not being called anywhere. This makes session depend on client js creating it in the first place. Not ideal (and not at all similar to .net behaviour). Ideally we would take advantage of session cookie creation code and create it if we didn't receive it from the FE.
Reasons for this are following:
-> Java AI should be able to provide session information without using client JS library
-> For the first call Server Side is involved earlier than Javascript, hence all sessions will miss the first call...
-> Java and .Net implementations could be similar in that matter making it easier to infer knowledge about implementations

gondar:
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8c51d66c08c1baf0844613637fa5d3ccfc7590f9/Src/DependencyCollector/Shared/Implementation/WebRequestDependencyTrackingHelpers.cs#L38 <- this is where .net does it

dhaval24:
@gondar apologies I am not very familier with this part of the SDK but is this issue very similar to the PR #537 where we have stripped off all the date parsing logic and kept only the SessionId as other fields are already deprecated.

gondar:
it was mentioned in above #537 that only if client is there we will have a session. That is a bit of shame but I guess cannot be easily helped.

PR #537, opened 2018-01-25
Fix session date parse issue

khroolick:
Fixed SessionId not sent to appinsight due to miliseconds added to a long date by JS appinsights library


khroolick:
@Yonisha take a look at this please if you are still around 
@dhaval24 this one is a blocker for us

dhaval24:
@khroolick Thank you very much for your contribution. Is this PR a fix for #537 ? As you said this one is blocker for you, does it mean you are not able to use AI java SDK in production without this fix?

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
```diff
@@ -191,8 +190,10 @@ private void parseCookie(Cookie cookie) throws Exception {
      */
     private Date parseDateWithBackwardCompatibility(String dateStr) throws ParseException {
         try {
-            return new Date(Long.parseLong(dateStr));
-        } catch (NumberFormatException e) {
+            Scanner s = new Scanner(dateStr);
+            s.useDelimiter("\\.");
```
dhaval24: Is the date string always going to have separators as "." ? Can date string ever be in format like 01/01/2018 ? If that is the case then this logic will not work. We might need to consider all seperators.

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
```diff
@@ -22,8 +22,7 @@
 package com.microsoft.applicationinsights.web.internal.cookies;
 
 import java.text.ParseException;
-import java.util.Calendar;
-import java.util.Date;
+import java.util.*;
```
dhaval24: Please do not use * imports. Edit it to add individual import statements,

gondar:
We don't have sessionId without it. This renders 70% of our data useless. We cannot do proper funnel analysis nor understand why given POST resulted in slow response (in order to get meaningful data we must correlate it with previous requests).
SessionId is not something that we can overlook when sending data to appInsights, it's a must for us to do analysis required by our management. :-D
Hope that helps.

SergeyKanzhelev:
Date parsing is not required any longer. The field representing it was deprecated.

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
gondar: @dhaval24 can you please check what AppInsights Javascript library sends you? I mean this is contract between to microsoft appinsights codebases one for JS and one for Java. I didn't find any documentation about it (no surprise as this is internal concern). I am sure you can get in touch with relevant team? 

Note that this bug was introduced over a year ago when js team started sending ms in cookie. So this is not about date (date was implementation used even earlier and you have code for backwards compatibility below), now it's format is <long_for_datetime>.<miliseconds> from looking at JS code https://github.com/Microsoft/ApplicationInsights-JS/blob/cac7b363a7f59147adff510f698bc3f4943f2ab0/JavaScript/JavaScriptSDK/ajax/ajaxUtils.ts#L36 (and we see that being a problem in our test env). Is there some ms spec on how ai_session cookie is meant to be?

dhaval24:
@SergeyKanzhelev thank you very much for this. I believe with the new schema we have deprecated the SessionStateData  data structure and therefore none of that gets serialized, and sent to ingestion if I am correct?

gondar:
This is not fix for #536 . This issue is to get sessionId working in situation where we have both appinsights JS in FE and appinsights Java in backend. Java cookie parsing logic is not compatible with JS cookie creation :-D.

SergeyKanzhelev:
You can send SessinoStateData and session acquisition time. But both will be stripped out in our endpoint.

gondar:
We did end to end test in our dev environment. To reproduce issue:
1. put newest JS appinsights snippet on a page
2. put current Java in your java application in standard setup (logging requests via webfilter as described in getting started guide)
3. send any requests to your application from any modern browser (ff, chrome)
4. go to appinsights portal and do following query:
requests
| where isnotempty(sessionId)
| summarize count() by sessionId

the result will be empty, whereas we should see count of requests per each session

:-)
Hope that helps.

gondar:
@SergeyKanzhelev I agree it's not used for anything :-D. It actually duplicates expiry field in standard http.

gondar:
@SergeyKanzhelev so you keep just sessionId and strip all after? @dhaval24 what do u think?

SergeyKanzhelev:
@gondar yes. The code in .NET looks strange https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/47d18691f836ca6509ad2f5014c297930b380685/Src/Web/Web.Shared.Net/SessionTelemetryInitializer.cs#L19-L38 as it also used to have parsing of the date. Now just take a `substring(0, indexOf("|"))`

khroolick:
@dhaval24 I've removed all this unused code altogether so there is nothing to cover with E2E test anymore :-)

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
khroolick: fixed

dhaval24:
@khroolick thank you very much for this. @SergeyKanzhelev  could you please review this changes, as I am not very familier with this part of the code and also probably resembles more the schema. We can take this change then as a part of next Beta release after the stable.

gondar:
Thanks all!
I would just point out that this is bug fix, not enhacement. Not having
sessionid in telemetry is serious limitation of data quality. Also not
working well with js cookie seems more a bug than anything.
Therefore my advise would be to apply it asap rather then with next
release, but its your choice. We decided to use our own fork with fix until
it is sorted.

27.01.2018 19:32 "Dhaval Doshi" <notifications@github.com> napisał(a):

dhaval24:
@gondar appreciate your genuine feedback. I have updated the tag from enhancement to bug. We would discuss around taking this as a part of stable release coming next once @SergeyKanzhelev reviews this PR in depth. I appreciate you taking the pain to use your own fork for the time being. 

@grlima and @littleaj the issue which is described here is for sure a bug which is not desirable for most of the people using JS + Java SDK together. Please have a look and review this.

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/HttpCookieFactory.java
```diff
@@ -46,12 +45,9 @@
      */
     public static Cookie generateSessionHttpCookie(
```
SergeyKanzhelev: So java SDK has a feature to generate session ID cookie if it wasn't set by JavaScript? How does it work around the problems mentioned here http://apmtips.com/blog/2016/07/09/tracking-users-in-api-apps/?

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
```diff
@@ -34,7 +30,6 @@
 
     // region Consts
 
-    private static final int UPDATE_TIMEOUT_IN_MINUTES = 5;
     public static final String COOKIE_NAME = "ai_session";
     public static final int SESSION_DEFAULT_EXPIRATION_TIMEOUT_IN_MINUTES = 30;
```
SergeyKanzhelev: do you need this variable any more?

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
```diff
@@ -103,99 +72,18 @@ public String getSessionId() {
...
-     */
-    private Date parseDateWithBackwardCompatibility(String dateStr) throws ParseException {
-        try {
-            return new Date(Long.parseLong(dateStr));
-        } catch (NumberFormatException e) {
-            return DateTimeUtils.parseRoundTripDateString(dateStr);
-        }
+    private static String parseCookie(Cookie cookie) {
+        String[] split = cookie.getValue().split(RAW_COOKIE_SPLIT_DELIMITER, 2);
+        return split[0];
```
SergeyKanzhelev: substring will work faster

dhaval24:
@khroolick , Sergey has approved your PR but left few comments. Could you please take time to clean up some of the things he mentioned and also respond to one of the questions he asked. Thanks again for the contribution

SergeyKanzhelev:
@dhaval24 I think the question is more for you and @grlima. You'd need to discuss this SDK behavior with usage team to make sure it will not screw up any numbers for them.

dhaval24:
@SergeyKanzhelev thanks. Got it, we will discuss this.

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/SessionCookie.java
khroolick: I agree 
I will change it to use substring

dhaval24:
@khroolick thank you for updating the PR. By the way it has merge conflicts with master. Could you please resolve the conflict so it is ready. @grlima and @littleaj could you please review and sign off this one.

web/src/main/java/com/microsoft/applicationinsights/web/internal/cookies/HttpCookieFactory.java
grlima: This method does seem to be used anywhere but unit tests. As a matter of fact, this factory class is not used at all. We should remove it. Created #577

gondar:
@SergeyKanzhelev can you please notify relevant PMs about this bug. I find this dragging on 

@grlima How much time it will take you to verify? It took us 1h to spot that issue and 1h to fix it. I presume with dev env for app insights java you can spot it and see that it resolves it in 15 minutes, right? If not do you want to do middle of night paring session where I would show you how it works?

SergeyKanzhelev:
CC: @nikmd23

dhaval24:
@grlima this PR has a lot of public methods getting removed / getting refactored in signature also. Before you merge this, please update the readme to reflect this removals and  refactoring.

grlima:
These changes will be merged with #622, which includes additional cleanup and updates to the change log.

PR #540, opened 2018-01-29
Fixing FixedRateSampling to work in old and new version of sampling

### LINKED ISSUE #535: FixedSampling is not reliable since itemCount is set with SamplingScore rather than SamplingPercentage.
When FixedRateSampling is enabled and sampling percentage is set while tracking the events, sampleRate in the JSON is set with SampleScore rather than the sampling percentage.

Ex, If the sampling percentage is set as 100 then the itemCount is supposed to be 1. Since sampling percentage(100) is overwritten with the samplescore (hash of the operationid = 11.104063601747184) , itemCount is showing the value of 100/sample score (100/11=9) which might not makes sense.

The issue is present in the below class and line no is 116.

Class: com.microsoft.applicationinsights.internal.channel.sampling.FixedRateTelemetrySampler

anandsugumaran:
Fix #535 

Fixed both the below issues in this commit.

1. New version of FixedRate Sampling worked only with global sampling percentage and not with user set sampling percentage.

2. Old version of FixedRate Sampling sets sampling score as SamplingPercentage once sampled in. It has to be sampling percentage.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/sampling/FixedRateTelemetrySampler.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/sampling/FixedRateTelemetrySampler.java
@@ -113,7 +113,7 @@ public final class FixedRateTelemetrySampler implements TelemetrySampler {
 
                         return false;
                     }
-                    samplingSupportingTelemetry.setSamplingPercentage(telemetrySamplingScore);
+                    samplingSupportingTelemetry.setSamplingPercentage(testedPercentage);
                 }
             }
         }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
@@ -157,16 +157,21 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
                 if (samplingSupportingTelemetry.getSamplingPercentage() == null) {
 
                     samplingSupportingTelemetry.setSamplingPercentage(samplingPercentage);
-
-                    if (SamplingScoreGeneratorV2.getSamplingScore(telemetry) >= samplingPercentage) {
-
-                        InternalLogger.INSTANCE.info("Item %s sampled out", telemetry.getClass());
-                        return false;
-                    }
+                                        
+                    
                 } else {
                     InternalLogger.INSTANCE.info("Item has sampling percentage already set to :"
                             + samplingSupportingTelemetry.getSamplingPercentage());
+                    
+                    samplingPercentage = samplingSupportingTelemetry.getSamplingPercentage();
                 }
+                
+                if (SamplingScoreGeneratorV2.getSamplingScore(telemetry) >= samplingPercentage) {
+
+                    InternalLogger.INSTANCE.info("Item %s sampled out", telemetry.getClass());
+                    return false;
+                }
+                
             } else {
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/sampling/FixedRateTelemetrySampler.java
```diff
@@ -113,7 +113,7 @@ else if (!includeTypes.isEmpty() && !includeTypes.contains(telemetry.getClass())
 
                         return false;
                     }
-                    samplingSupportingTelemetry.setSamplingPercentage(telemetrySamplingScore);
```
anandsugumaran: telemetrySamplingScore is a dynamically generated value based on operation id. This cannot be set as Sampling percentage after sampled in since this will affect the item count value in the analytics portal. So replaced this with testedPercentage which will have the correct sampling percentage.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -157,16 +157,21 @@ public boolean process(Telemetry telemetry) {
                 if (samplingSupportingTelemetry.getSamplingPercentage() == null) {
 
                     samplingSupportingTelemetry.setSamplingPercentage(samplingPercentage);
-
-                    if (SamplingScoreGeneratorV2.getSamplingScore(telemetry) >= samplingPercentage) {
```
anandsugumaran: As per existing code , sampling will work only when the user didn't set the sampling percentage. To fix this moved the code which samples the telemetry out of the if condition.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -157,16 +157,21 @@ public boolean process(Telemetry telemetry) {
...
-                        InternalLogger.INSTANCE.info("Item %s sampled out", telemetry.getClass());
-                        return false;
-                    }
+                                        
+                    
                 } else {
                     InternalLogger.INSTANCE.info("Item has sampling percentage already set to :"
                             + samplingSupportingTelemetry.getSamplingPercentage());
+                    
+                    samplingPercentage = samplingSupportingTelemetry.getSamplingPercentage();
```
anandsugumaran: Override the local sampling percentage with the user set sampling percentage. Else global sampling percentage will be applied even if the user has set the sampling percentage.

dhaval24:
@anandsugumaran thank you very much for your valuable contribution. We would review it and let you know if there needs to be any changes.

anandsugumaran:
@littleaj @grlima Can you please review this PR ?

dhaval24:
@anandsugumaran by the way just was curious, were you able to do an E2E testing with your change and able to see the right item counts on the Analytics Portal? We would like to analyze that before merging this, and ofcourse as @littleaj said thank you very much for your valuable contribution.

anandsugumaran:
@dhaval24 yeah I did and it looked good in the portal as well.

anandsugumaran:
@dhaval24 can you merge the changes if it's looking good

dhaval24:
@anandsugumaran sure, we will do that by the end of the day.

dhaval24:
@anandsugumaran thank you very much for your contribution. We have merged your changes and it would likely be the part of 2.1.0 version of our SDK (Which is one release after the stable 2.0.0 version)

dhaval24:
@anandsugumaran your changes are being taken to stable release of our SDK which would happen soon. Hopefully this would unblock your team.

anandsugumaran:
@dhaval24 Thanks Dhaval.

PR #541, opened 2018-01-29
Fix null ref check in telemetry correlation Utils

### LINKED ISSUE #506: NULL check missing in agent instrumentation code
When the agent attempts to view the existing correlation headers it does not perform a NULL check [here](https://github.com/Microsoft/ApplicationInsights-Java/blob/36214d5c28eb975e008c5b7c60c4840fdd10ec20/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java#L66).

This can throw an exception during the filter pipeline and lead to an internal server error.

It is recommended that we add a check similar to [this one](https://github.com/Microsoft/ApplicationInsights-Java/blob/36214d5c28eb975e008c5b7c60c4840fdd10ec20/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java#L143-L146) to avoid a nullref exception.

It was see recently on a struts application using thew new 2.0.0-BETA SDK and form based security with j_security_check.

dhaval24:
Fix #506 .

This PR fixes the missing Null Reference checks in the TelemetryCorrelationUtils class which caused NullPointerExceptions.

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Version 2.0.0
+
+- Fix #506 Null Reference Check causing Null Pointer Exception in `TelemetryCorrelationUtils.java`
+
 ## Version 2.0.0-BETA
 - Updating various dependencies to latest version
 - Introducing public class CustomClassWriter in Agent to enable finding common super classes used for Agent instrumentation without loading it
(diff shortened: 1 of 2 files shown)


dhaval24:
@grlima I have merged this as you were already aware of the changes.

PR #547, opened 2018-02-05
Fix usage of format strings in logger and better exception logging.

littleaj:
Got an error from the agent thrown from String.format because 'F' was not a recognized format string token. This is because it was printing a URL encoded string, which had "Program%20Files" in the text. This caused the error.

So, we should _always_ be using format strings in the logger to prevent this kind of issue.

Also, use `Throwable.toString()` instead of `getMessage()` because it provides the exception name in addition to its message. [See here](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/Throwable.java#Throwable.toString%28%29)

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
@@ -60,7 +60,7 @@ public final class AgentImplementation {
             appendJarsToBootstrapClassLoader(inst);
             initializeCodeInjector(inst);
         } catch (Throwable throwable) {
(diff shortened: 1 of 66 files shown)


littleaj:
@dhaval24, you'll have to point me to those places. I agree that by default, catch blocks should log the exception, but this isn't necessarily true for all catch blocks. If you have some blocks in mind, let's review those on a case-by-case basis.

dhaval24:
@littleaj I do not have any specific block. IMO it doesn't hurt to dump stack traces using the highest verbosity level. Doing so can provide us deep dive into debugging when something goes wrong. And It would not affect production because eventually we do have different levels of log settings.

Issue #385, opened 2017-06-09
TransmissionNetworkOutput sending loop should delay for all exceptions and not dispose connection all the time


SergeyKanzhelev:
There are other exceptions that needs delay for the next loop. Otherwise on azure web apps SDK takes out all the available connections.

https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L131

dhaval24:
@SergeyKanzhelev the delay mechanism is now in place for all the exceptions. Please check latest master if interested. Quick question though, I see that current default backoff time is set o 300 sec which can be sometimes too much in high throughput applications. We have seen this with some of the users resulting in lost telemetry. A dirty fix was to have 3 immediate retries before going to 300 seconds backoff. This would be fix unblocking several people though something else could be ideal. Let me know your thoughts and I would be closing this one.

SergeyKanzhelev:
spin-lock's like retries sounds good. I remember I caused this connections starvation when I had some packages mismatch and it kept failing for me with the exception un-expected in code. If this is fixed - great, please close the issue.

dhaval24:
https://github.com/Microsoft/ApplicationInsights-Java/blob/7a5e572b2ffd3256898c6bd05dc3bec6754e3629/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L172

Here is the fix. We now back-off for any kind of exceptions.  Here is the PR for retrying 3 times before Back_off #501

PR #556, opened 2018-02-11
Implementing Retry logic [Reliable Channel]

debugthings:
Added retry logic that is similar to the .NET framework.

* Removes logic from NetworkTransmissionOutput
* Uses observers to handle error conditions
* Added failed transmission retry
* Added partial transmission retry
* Aligned response codes to match .NET
* Uses existing [ExponentialBackOffTimesPolicy](https://github.com/Microsoft/ApplicationInsights-Java/blob/ff7961019d564ecab5f3f3f237cb4844cbddc6a2/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ExponentialBackOffTimesPolicy.java)

> **NOTE** This clobbers the change in #385 referenced here

https://github.com/Microsoft/ApplicationInsights-Java/blob/7a5e572b2ffd3256898c6bd05dc3bec6754e3629/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L172

All exceptions are now handled by `ErrorHandler` in this line.

https://github.com/Microsoft/ApplicationInsights-Java/blob/094383a25b0f98fe6aa2dd6dd58bae10259492e6/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java#L39-L42

- [ ] CHANGELOG.md updated


core/build.gradle
```diff
@@ -65,6 +66,7 @@ shadowJar {
     relocate 'org.apache.commons', 'com.microsoft.applicationinsights.core.dependencies.apachecommons'
     relocate 'com.google.common', 'com.microsoft.applicationinsights.core.dependencies.googlecommon'
     relocate 'javax.annotation', 'com.microsoft.applicationinsights.core.dependencies.javaxannotation'
+    relocate 'com.google.gson', 'com.microsoft.applicationinsights.core.dependencies.gson'
```
dhaval24: Do you use gson for actual code? Or this is only for test? If it is only test I would say do not relocate. Relocation usually should be done when it is absolutely necessary.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/BackendResponse.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.applicationinsights.channel.concrete.inprocess;
+
+public class BackendResponse {
+
+	public int itemsReceived;
```
dhaval24: Adjust the additional space

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java
```diff
@@ -0,0 +1,50 @@
...
+		this.transmissionPolicyManager = policy;
+	}
+	
+	@Override
+	public void onTransmissionSent(TransmissionHandlerArgs args) {
+		
+		validateTransmissionAndSend(args);
+	}
+
+	public boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
```
dhaval24: Does this method need to be public? If it is not accessed from outside, reduce the scope to private or default.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,145 @@
...
+	private BackendResponse getBackendResponse(String response) {
+
+		BackendResponse backend = null;
+		try {
+			// Parse JSON to Java
+			GsonBuilder gsonBuilder = new GsonBuilder();
+			Gson gson = gsonBuilder.create();
+			backend = gson.fromJson(response, BackendResponse.class);
+		} catch (Throwable t) {
+			InternalLogger.INSTANCE.trace("Error deserializing backend response with Gson, message: %s",
```
dhaval24: Please print the stack trace also. ExceptionUtils can be used for this.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,145 @@
...
+
+	public boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
+		if (args.getTransmission() != null && args.getTransmissionDispatcher() != null) {
+			switch (args.getResponseCode()) {
+			case HttpStatus.SC_PARTIAL_CONTENT:
+				BackendResponse beR = getBackendResponse(args.getResponseBody());
+				
+				// In case the 206 was false we can break here
+				if(beR != null && (beR.itemsAccepted == beR.itemsReceived))
+				{
```
dhaval24: I think there should be a log message here, with lowest severity.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,145 @@
...
+				ArrayList<String> originalItems = generateOriginalItems(args);
+				
+				// Somehow the amount of items received and the items sent do not match
+				if(beR != null && (originalItems.size() != beR.itemsReceived))
+				{
+					return false;
+				}
+			
+				
+				ArrayList<String> newTransmission = new ArrayList<String>();
```
dhaval24: A general comment for whole PR. Please prefer to use interfaces. This line can be replaced as follows:
List<String> newTransmission = new ArrayList<>(); 
Dimond Types can now automatically infer the types from Java 7 onwards.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,145 @@
...
+				for (BackendResponse.Error e : beR.errors) {
+					switch (e.statusCode) {
+					case HttpStatus.SC_REQUEST_TIMEOUT:
+					case HttpStatus.SC_INTERNAL_SERVER_ERROR:
+					case HttpStatus.SC_SERVICE_UNAVAILABLE:
+					case 429:
```
dhaval24: So we don't handle 429 ?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,145 @@
...
+				{
+					gis.close();	
+				}
+				if (bufferedReader != null) 
+				{
+					bufferedReader.close();
+				}
+			} catch (IOException e1) {
+				// TODO Auto-generated catch block
+				e1.printStackTrace();
```
dhaval24: Don't print stack trace. Instead dump it using Internal Logger

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
```diff
@@ -0,0 +1,95 @@
...
+				this.transmissionPolicyManager.suspendInSeconds(suspensionPolicy, retryAfterAsSeconds);
+			} catch (Throwable e) {
+				InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.ERROR,
+						"Throttled but failed to block transmission, exception: %s", e.getMessage());
+				this.transmissionPolicyManager.backoff();
+			}
+		
+	}
+
+	private static Date convertToDateToGmt(Date date) {
```
dhaval24: We need java docs for all the methods and their description. It is crucial for all public methods and private methodsl like this where there is core logic.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
```diff
@@ -0,0 +1,95 @@
...
+	{
+		this.transmissionPolicyManager = policy;
+	}
+	
+	@Override
+	public void onTransmissionSent(TransmissionHandlerArgs args) {
+		validateTransmissionAndSend(args);
+	}
+
+	public boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
```
dhaval24: Similar to previous does this method need to be public. More-ever can we create an abstraction for this too, like an interface because it seems like this is being used in different handler classes you created.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -33,294 +34,188 @@
...
+					transmission.setNumberOfSends(0);
+					return false;
+				} else if (code == HttpStatus.SC_OK) {
+					// If we've completed then clear the back off flags as the channel does not need to be throttled
+					transmissionPolicyManager.clearBackoff();
+				}
+				return true;
+					
+			} catch (ConnectionPoolTimeoutException e) {
+				ex = e;
```
dhaval24: All this places we need to dump the stack trace. Please use ExceptionUtils to do so and pass it in logger.

core/build.gradle
debugthings: I added Gson to take care of the parsing of the BackendResponse object that is returned from a partial success. There were no JSON deserializers already included in the solution (at least none that were obvious).

dhaval24: @debugthings makes sense. Gson was used but only for test cases. Since this is now used for production code we do need to shade it. May be we might not need to shade it still. Gradle will pull this as transitive dependency when building. Shading (i.e repacking) to microsoft domain is only needed when it interferes our agent (one example is apache http client). We shade it to avoid instrumenting the calls sdk makes using it. Gson is a widely used library across java and I would rather avoid shading to prevent us from falling into the traps of class path hells.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java
dhaval24: Ah so mockito version 1.10 does not allow mocking private methods. Is this method a part of interface? That's why I was suggesting since this is used widely we can plug it into the interface and then mock it. If that route is not valid for some reason then we might better use reflection. Reducing the publicly exposed methods will improve our abilities to modify it without affecting anyone. Users can take dependency on public methods and modification of signature can thus be very cumbersome.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
debugthings: Eh? In the partial success handler any an all error messages are treated the same. In this instance the partial success will try to resend any of these items in the next transmission.

https://github.com/Microsoft/ApplicationInsights-dotnet/blob/0fe5edf099ecdad37990be0c776b300e7fbcff4a/src/ServerTelemetryChannel/Implementation/PartialSuccessTransmissionPolicy.cs#L107-L111

dhaval24: Ah got it. So this means if partial message was accepted, then we would retry the whole packet correct?

core/build.gradle
debugthings: Fair enough, I also thought that we include it to avoid having to distribute gson with it for instances where this is a drop in addition like a 3rd party.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/BackendResponse.java
debugthings: Fixed

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java
debugthings: I definitely want to keep the tests for this method as it checks the actual logic of the handler(s). I'll do a bit of looking around to see what are some good solutions to keep the scope down. Possibly extending the class as a package private and setting those methods to protected.

debugthings: Fixed this by using the package scope. Only classes defined in the packages can see the method.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
debugthings: Fixed this by using the package scope. Only classes defined in the packages can see the method.

core/build.gradle
dhaval24: Yes, that was one reason but, this creates problems of dependency hells if we miss proper shading, and more over if the user uses dependency management, which most do then transient dependencies get pulled down. Further as the SDK grows and we start using more libraries, it would make sense to just shade the one which interfer the operation and leave the rest

core/src/main/java/com/microsoft/applicationinsights/internal/channel/TransmissionHandlerArgs.java
```diff
@@ -0,0 +1,32 @@
...
+	
+	private Transmission transmission;
+	public void setTransmission(Transmission transmission) { this.transmission = transmission;}
+	public Transmission getTransmission() { return this.transmission;}
+	
+	private int responseCode;
+	public void setResponseCode(int code) { this.responseCode = code;}
+	public int getResponseCode() { return this.responseCode;}
+	
+	private Throwable exception;
```
dhaval24: @debugthings do we ever get a Throwable here? I think most of the time it would get an instance of Exception class. If you know any place where this could be Throwable please let me know

debugthings: From the original implementation they catch Throwable for logging. I also consume it here incase there is an Error and want to retry the transmission.

https://github.com/Microsoft/ApplicationInsights-Java/blob/1e5d72cadd24cb75b46d577ef0f8c3828fc87e19/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L190-L193

Which is then added here
https://github.com/Microsoft/ApplicationInsights-Java/blob/1e5d72cadd24cb75b46d577ef0f8c3828fc87e19/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L209

And then null checked here
https://github.com/Microsoft/ApplicationInsights-Java/blob/1e5d72cadd24cb75b46d577ef0f8c3828fc87e19/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java#L56

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -46,8 +50,16 @@
  *
  * Created by gupele on 6/29/2015.
  */
-public final class TransmissionPolicyManager implements Stoppable {
-
+public final class TransmissionPolicyManager implements Stoppable, TransmissionHandlerObserver {
```
dhaval24: So we had to change the public class also, I think which is fine. However, if some of our users if they had taken dependency on this. we would break them by this change. I am not sure we might need to consider a separate release for this, mentioning that there can be some breaking changes.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,196 @@
...
+	 * Used to parse the original telemetry request in order to resend the failed
+	 * ones.
+	 * 
+	 * @param args
+	 *            The {@link TransmissionHandlerArgs} that contains the
+	 *            {@link Transmission} object.
+	 * @return A List<> of each sent item
+	 */
+	List<String> generateOriginalItems(TransmissionHandlerArgs args) {
+		List<String> originalItems = new ArrayList<String>();
```
dhaval24: Types in <> in Java 7 and onwards can be auto-inferred and is more efficient.  This can be replaced with 
List<String> originalItems = new ArrayList<>();

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,196 @@
...
+	 */
+	boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
+		if (args.getTransmission() != null && args.getTransmissionDispatcher() != null) {
+			switch (args.getResponseCode()) {
+			case HttpStatus.SC_PARTIAL_CONTENT:
+				BackendResponse beR = getBackendResponse(args.getResponseBody());
+				List<String> originalItems = generateOriginalItems(args);
+
+				// Somehow the amount of items received and the items sent do not match
+				if (beR != null && (originalItems.size() != beR.itemsReceived)) {
```
dhaval24: I would suggest to make this condition more robust. This could be more apt
if (beR != null && (originalItems.size()  < beR.itemsReceived))

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,196 @@
...
+					InternalLogger.INSTANCE.trace(
+							"Skipping partial content handler due to itemsReceived being larger than the items sent.");
+					return false;
+				}
+
+				if (beR != null && (beR.itemsAccepted != beR.itemsReceived)) {
```
dhaval24: Similarly for this the condition can be changed to 
if (beR != null && (beR.itemsAccepted < beR.itemsReceived) 
This makes it logically more clear and better readable code. IMO the condition in itself is also more robust.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,196 @@
...
+
+					List<String> newTransmission = new ArrayList<String>();
+					for (BackendResponse.Error e : beR.errors) {
```
dhaval24: Why don't we replace this for-each loop with a for loop like this : 
for (int i = 0; i < ber.errors.length; ++i). This would help simplify the index operation being used mutiple times. We can the just do 
if (i < originalItems.size()) {
     newTransmission.add(originalItems.get(i));
}

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
```diff
@@ -0,0 +1,137 @@
...
+	 * Provides the core logic for the retransmission
+	 * 
+	 * @param args
+	 *            The {@link TransmissionHandlerArgs} for this transmission.
+	 * @return Returns a pass/fail for handling this transmission.
+	 */
+	boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
+		if (args.getRetryHeader() != null && args.getTransmission() != null
+				&& args.getTransmissionDispatcher() != null) {
+			args.getTransmission().incrementNumberOfSends();
```
dhaval24: I see that PartialResponseHandler class doesn't increment the number of sends like this. Is there a reason why it is omitted there? Because eventually we are retrying correct?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
```diff
@@ -0,0 +1,137 @@
...
+		if (Strings.isNullOrEmpty(retryAfterAsString)) {
+			return;
+		}
+
+		try {
+			DateFormat formatter = new SimpleDateFormat(RESPONSE_RETRY_AFTER_DATE_FORMAT);
+			Date date = formatter.parse(retryAfterAsString);
+
+			Date now = Calendar.getInstance().getTime();
+			long retryAfterAsSeconds = (date.getTime() - convertToDateToGmt(now).getTime()) / 1000;
```
dhaval24: Seems like this calculation is always fixed. I might have missed something , please point out to where we are doing exponential backoff in this case.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
```diff
@@ -68,6 +68,15 @@ public SenderThreadLocalBackOffData(long[] backOffTimeoutsInMillis, long addMill
     public boolean isTryingToSend() {
         return currentBackOffIndex != -1;
     }
+
```
dhaval24: Missing Java docs

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
```diff
@@ -92,34 +101,62 @@ public void onDoneSending() {
...
+                return instanceIsActive;
+           } catch (InterruptedException e) {
+               return false;
+           }
+       } finally {
+           lock.unlock();
+       }
+   }
+    
+    public long backOffTimerValue() {
```
dhaval24: Java docs missing for this method

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
```diff
@@ -75,6 +80,11 @@ public synchronized void stopAllSendersBackOffActivities() {
         }
         stopped = true;
     }
+    
+    public long getCurrentBackoffMillis() {
```
dhaval24: docs

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicy.java
```diff
@@ -24,8 +24,9 @@
 /**
  * Created by gupele on 6/29/2015.
  */
-enum TransmissionPolicy {
+public enum TransmissionPolicy {
```
dhaval24: Update java docs for this

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -33,294 +34,198 @@
...
 /**
- * The class is responsible for the actual sending of {@link com.microsoft.applicationinsights.internal.channel.common.Transmission}
+ * The class is responsible for the actual sending of
+ * {@link com.microsoft.applicationinsights.internal.channel.common.Transmission}
  *
  * The class uses Apache's HttpClient framework for that.
  *
  * Created by gupele on 12/18/2014.
  */
 public final class TransmissionNetworkOutput implements TransmissionOutput {
```
dhaval24: There are bunch of methods in this class which do not have java docs. It would be good if we can update those places as we are changing this part.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -46,8 +50,16 @@
...
+
+	private final int DEFAULT_MAX_SECONDS_TO_PAUSE_AFTER_MAX_BACKOFF = 600;
```
dhaval24: How did we came up with this number ? 600 Seconds seems to be too long. In a situation where the network is very restricted this would accumulate bulk of telemetry.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,146 @@
...
+	public void onTransmissionSent(TransmissionHandlerArgs args) {
+		validateTransmissionAndSend(args);
+
+	}
+
+	public boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
+		if (args.getTransmission() != null && args.getTransmissionDispatcher() != null) {
+			switch (args.getResponseCode()) {
+			case HttpStatus.SC_PARTIAL_CONTENT:
+				BackendResponse beR = getBackendResponse(args.getResponseBody());
```
grlima: Nit picking: Please use a more descriptive name instead of beR. "backendResponse" would be fine.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
```diff
@@ -0,0 +1,146 @@
...
+				
+				// Somehow the amount of items received and the items sent do not match
+				if(beR != null && (originalItems.size() != beR.itemsReceived))
+				{
+					return false;
+				}
+			
+				
+				List<String> newTransmission = new ArrayList<String>();
+				for (BackendResponse.Error e : beR.errors) {
```
grlima: null check missing for beR. I'd suggest checking for null right after line 39.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionSendResult.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Created by gupele on 2/10/2015.
  */
-enum TransmissionSendResult {
+public enum TransmissionSendResult {
```
dhaval24: Why do we need to make this enum public? Is it required for test cases? If yes please follow the Java Doc requirements and update the comment to be java doc comment with description.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
debugthings: Agreed, fixed.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
debugthings: Not sure I agree, if these numbers don't match then there was a problem with the transmission and we're in an invalid state I would think. Either the original content was broken or the response is broken

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
debugthings: The index of the items in the errors array is not guaranteed to be the index of the original item sent.

errors[0] = Error() { index = 2, responseCode = 500, message = "Internal Error" }

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
debugthings: In this instance we're creating a brand new transmission to send based on the result of the old one. We are not going to dispatch the entire transmission so no need to increment.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
debugthings: For Throttling requests the server provides the backoff time in the form of the Retry-After header. If that header is missing or invalid (will throw) I call backoff in the catch block

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
debugthings: Not needed, removed

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
debugthings: Agreed, 10 minutes is a long time and it should be configurable. In the .NET implementation it will max out at 30 minutes in the backoff algorithm. Let's talk about this a bit more to see what we really want to do. By the time we reach this point we would have executed the backoff 15 times and probably have been throttled well over 15 minutes. It's possible this is just redundant.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
debugthings: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
debugthings: Removed method, not used anywhere provided no value

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
debugthings: removed

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
debugthings: Maybe. We added four public methods to this but did not remove any; and didn't change the implementation of existing methods. That shouldn't break the contract and the code should execute as expected. Might well warrant a deeper discussion

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
debugthings: Fixed with logic update

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/PartialSuccessHandler.java
dhaval24: Ah this is a catch. Got it

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
dhaval24: ok

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ThrottlingHandler.java
dhaval24: Makes sense. I would say to mention a comment!

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
dhaval24: Okay

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
dhaval24: Hmm

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
dhaval24: @debugthings Yes, ideally I would want that to be configurable too. So instead of keeping that 15 constants for back of we have some sort of iterator pattern implemented which keeps on providing new backoff numbers until user specified limit. Internally iterator could well be implemented as Geometric Progression with ratio of 2. If we want to interleave with a constant time that could also be done.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionSendResult.java
dhaval24: Haha, got it!

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
debugthings: Added docs for public and private methods.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -33,294 +34,228 @@
...
+					// If we've completed then clear the back off flags as the channel does not need
+					// to be throttled
+					transmissionPolicyManager.clearBackoff();
+				}
+				return true;
+
+			} catch (ConnectionPoolTimeoutException e) {
+				ex = e;
+				InternalLogger.INSTANCE.error(
+						"Failed to send, connection pool timeout exception\r\n" + "Stack Trace:\r\n" + "%s",
```
littleaj: Use `%n` and the format string chooses the correct EOL characters. Same with the rest of the log messages.
Also, let's remove the unnecessary concatenations; I don't know if JIT is smart enough to optimize this into one string literal. Let's assume it's not and merge these into one string.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -135,11 +135,13 @@ public TransmissionFileSystemOutput(String folderPath) {
     @Override
     public boolean send(Transmission transmission) {
         if (size.get() >= capacityInKB) {
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "Persistent storage max capcity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", size.get());
```
littleaj: I think this is a `WARNING` level log. It's an exceptional case, but everything is still operational.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -135,11 +135,13 @@ public TransmissionFileSystemOutput(String folderPath) {
...
             return false;
         }
 
         Optional<File> tempTransmissionFile = createTemporaryFile();
         if (!tempTransmissionFile.isPresent()) {
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "Persistent storage unable to create temporary file.");
```
littleaj: I think this is an `ERROR` level log. We can still run (i.e. it's not FATAL), but this component cannot do what it's designed to do.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
```diff
@@ -0,0 +1,192 @@
...
+				}
+
+				if (backendResponse != null && (backendResponse.itemsAccepted < backendResponse.itemsReceived)) {
+					List<String> newTransmission = new ArrayList<String>();
+					for (BackendResponse.Error e : backendResponse.errors) {
+						switch (e.statusCode) {
+						case HttpStatus.SC_REQUEST_TIMEOUT:
+						case HttpStatus.SC_INTERNAL_SERVER_ERROR:
+						case HttpStatus.SC_SERVICE_UNAVAILABLE:
+						case 429:
```
littleaj: For the 429 and 439 case, can you append those lines with comments with a "friendly name" (similar to the constants in the previous cases)

grlima: Should we create constants for 429 and 439?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicy.java
debugthings: Removed public.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
debugthings: Sounds good, eclipse auto format causes this most likely because I'm not using the %n string.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
dhaval24: I think probably we should. It would allow when we extend the support later to support this codes.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
```diff
@@ -92,34 +92,69 @@ public void onDoneSending() {
...
-             try {
-                 long millisecondsToWait = backOffTimeoutsInMillis[currentBackOffIndex];
-                 if (millisecondsToWait > BackOffTimesPolicy.MIN_TIME_TO_BACK_OFF_IN_MILLS) {
-                     millisecondsToWait += addMilliseconds;
-                 }
-                 backOffCondition.await(millisecondsToWait, TimeUnit.MILLISECONDS);
-                 return instanceIsActive;
-            } catch (InterruptedException e) {
-                return false;
+            if (!instanceIsActive) {
```
dhaval24: @debugthings is it true that thread will never starve in the same state (i.e blocked state). If not we might always hit this condition for certain thread.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
dhaval24: I think things are operational, however when this would happen it would indicate loss of telemetry. Imagine a banking application where this occurs and they loss transactional logs which were eventually needed for fraud analysis then this could be critical information. I would keep it as logalways and since this is rare I don't think it would any how effect production loads.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -33,294 +34,228 @@
...
+	 * Private Ctor to initialize class.
+	 * <p>
+	 * Also creates the httpClient using the ApacheSender instance
+	 * @param serverUri The HTTP endpoint to send our telemetry too.
+	 * @param transmissionPolicyManager
+	 */
+	private TransmissionNetworkOutput(String serverUri, TransmissionPolicyManager transmissionPolicyManager) {
+		Preconditions.checkNotNull(serverUri, "serverUri should be a valid non-null value");
+		Preconditions.checkArgument(!Strings.isNullOrEmpty(serverUri), "serverUri should be a valid non-null value");
+		Preconditions.checkNotNull(transmissionPolicyManager,
```
dhaval24: What are scenarios where this could be null. How does code flow happen in case of this event?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
littleaj: I looked it up and JIT _is_ smart enough to turn those into one literal, so I'm less concerned with the concatenation. :)

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
littleaj: `HttpStatus` is part of the Apache HttpClient. Will a comment be sufficient?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
littleaj: The Preconditions statements are for us. Callers of this method should make sure `serverUri` is not null before calling.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
littleaj: I'm suggesting changing it to `InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, ...` I also think it should be logAlways.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
debugthings: I think it makes more sense to have this as a constant. I modified `TransmissionSendResult` as it wasn't used anywhere else and was internal to the `com.microsoft.applicationinsights.internal.channel.common` package

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
debugthings: I am removing this as it's redundant. There is a logger call in `createTemporaryFile()`

littleaj: Sounds good.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
debugthings: So, this is an artifact of the blocking backoff implementation (which I left intact). The only way for this to get set is if someone calls stop(). It's highly unlikely that it will be called since it's not referenced anywhere in the code.

It's in place to make sure if someone wants to disable the backoff they can do so with stop. Not surprisingly there is no start method so once the manager is stopped that's it

dhaval24: If there is a feature like this to disable the backoff it should be documented. And if there is no way to start if back then stop also doesn't make sense to me. Either case I think if we want to keep it just comment on it , regarding it is not advised to use this at any point or just remove the method if we do not plan to have this functionality.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
debugthings: Agreed, WARN makes more sense as it's something that is actionable but not ex

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
debugthings: Fixed with constants mentioned in other change.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java
debugthings: I am all for deleting code that is not used.

debugthings: I left this code intact as there are some old tests that are dependent on it and that is too much to tackle in this PR. Created task #559 to follow up

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -83,8 +95,31 @@ public void run() {
...
+    public void backoff() {
+    	policyState.setCurrentState(TransmissionPolicy.BACKOFF);
+    	long backOffMillis = backoffManager.backOffCurrentSenderThreadValue();
+        if (backOffMillis > 0)
+        {
+        	long backOffSeconds = backOffMillis / 1000;
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "App is throttled, telemetry will be blocked for %s seconds.", backOffSeconds);
+        	this.suspendInSeconds(TransmissionPolicy.BACKOFF, backOffSeconds);
+        } else {
+        	// TODO Remove Static Time
```
AlexKlimovMS: Is this TODO still needed?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -83,8 +95,31 @@ public void run() {
...
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "Backoff has been maxed out, will suspend thread for %s seconds.", DEFAULT_MAX_SECONDS_TO_PAUSE_AFTER_MAX_BACKOFF);
+        	this.suspendInSeconds(TransmissionPolicy.BACKOFF, DEFAULT_MAX_SECONDS_TO_PAUSE_AFTER_MAX_BACKOFF);
+        }
+    }
+    
+    public void clearBackoff() {
+    	policyState.setCurrentState(TransmissionPolicy.UNBLOCKED);
+        backoffManager.onDoneSending();
```
AlexKlimovMS: Minor: extra spaces.

grlima:
Replaced by https://github.com/Microsoft/ApplicationInsights-Java/pull/561

Issue #480, opened 2017-12-05
Upgrading gradle wrapper to latest version breaks win native build


littleaj:
Just something I found while trying to upgrade the gradle wrapper. The current wrapper version is 3.5 and the latest gradle version is 4.6. We should upgrade eventually, but need to understand why this is happening before we do.

dhaval24:
@littleaj is this still relevant?

littleaj:
Technically, it is an issue that I found; but we have no drivers at this time to upgrade. I'm going to close it and we'll tackle it if upgrading becomes necessary.

PR #557, opened 2018-02-12
upgrade wrapper to 4.1 which includes a bug fix

littleaj:
There's a bug in gradle which affects the output from exec. It appears to be fixed here in version 4.1. Here's a link to the fix: 
https://github.com/gradle/gradle/blob/941559e020f6c357ebb08d5c67acdb858a3defc2/subprojects/logging/src/main/java/org/gradle/internal/logging/text/AbstractLineChoppingStyledTextOutput.java#L112


littleaj:
the failed tests are due to a build server issue. merging

dhaval24:
@littleaj we should close #480 as I believe this PR fixes the issue. Though the latest gradle version as mentioned there is 4.3.x . If we are still planning to go there we may keep it. Please do accordingly.

PR #561, opened 2018-02-14
Implementing Retry logic [Reliable Channel] [STABLE Branch]

debugthings:
Completion of PR #556 into Stable Branch

Added retry logic that is similar to the .NET framework.

* Removes logic from NetworkTransmissionOutput
* Uses observers to handle error conditions
* Added failed transmission retry
* Added partial transmission retry
* Aligned response codes to match .NET
* Uses existing [ExponentialBackOffTimesPolicy](https://github.com/Microsoft/ApplicationInsights-Java/blob/ff7961019d564ecab5f3f3f237cb4844cbddc6a2/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ExponentialBackOffTimesPolicy.java)

> **NOTE** This clobbers the change in #385 referenced here

https://github.com/Microsoft/ApplicationInsights-Java/blob/7a5e572b2ffd3256898c6bd05dc3bec6754e3629/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java#L172

All exceptions are now handled by `ErrorHandler` in this line.

https://github.com/Microsoft/ApplicationInsights-Java/blob/094383a25b0f98fe6aa2dd6dd58bae10259492e6/core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/ErrorHandler.java#L39-L42

- [ ] CHANGELOG.md updated


core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -47,313 +45,339 @@
...
+		LimitsEnforcer maxTelemetryBufferCapacityEnforcer = createDefaultMaxTelemetryBufferCapacityEnforcer(null);
+
+		LimitsEnforcer sendIntervalInSecondsEnforcer = createDefaultSendIntervalInSecondsEnforcer(null);
+
+		boolean throttling = true;
+		if (namesAndValues != null) {
+			throttling = Boolean.valueOf(namesAndValues.get("Throttling"));
+			developerMode = Boolean.valueOf(namesAndValues.get(DEVELOPER_MODE_NAME));
+			try {
+				maxInstantRetries = Integer.parseInt(INSTANT_RETRY_NAME);
```
grlima: I think this will blow. Shouldn't it be Integer.parseInt(namesAndValues.get(INSTANT_RETRY_NAME)); ?

dhaval24: Yes good catch. This will blow

dhaval24:
@debugthings just wanted to understand did you changed / added anything into this PR which is different from the previous one which we reviewed extensively? Do I need to go through this PR again?

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -47,313 +45,339 @@
...
-                        MAX_FLUSH_BUFFER_TIMEOUT_IN_SECONDS,
-                        DEFAULT_FLUSH_BUFFER_TIMEOUT_IN_SECONDS,
-                        currentValue == null ? DEFAULT_FLUSH_BUFFER_TIMEOUT_IN_SECONDS : currentValue);
-
-        return sendIntervalInSecondsEnforcer;
-    }
+	
+	private final static String INSTANT_RETRY_NAME = "MaxInstantRetry";
+	private final static int DEFAULT_MAX_INSTANT_RETRY = 3;
+	private final static int DEFAULT_MAX_TELEMETRY_BUFFER_CAPACITY = 500;
```
dhaval24: This number also seems low to me. I think default max capacity can be double

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -47,313 +45,339 @@
...
+	private final static int MIN_MAX_TELEMETRY_BUFFER_CAPACITY = 1;
+	private final static int MAX_MAX_TELEMETRY_BUFFER_CAPACITY = 1000;
```
dhaval24: This is also definitely less. I think this should be definitely increased.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -47,313 +45,339 @@
...
+	 */
+	@Override
+	public void send(Telemetry telemetry) {
+		Preconditions.checkNotNull(telemetry, "Telemetry item must be non null");
+
+		if (isDeveloperMode()) {
+			telemetry.getContext().getProperties().put("DeveloperMode", "true");
+		}
+
+		if (telemetrySampler != null) {
```
dhaval24: Is this just kept for maintaining backward compatibility with Sampling V1? Nonetheless, with this major version we should definitely scrap off the old sampling algorithm as it uses user id based sampling which is not desirable for sure.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -47,313 +45,339 @@
...
+
+	/**
+	 * Sets an optional Sampler that can sample out telemetries Currently, we don't
+	 * allow to replace a valid telemtry sampler.
+	 *
+	 * @param telemetrySampler
+	 *            - The sampler
+	 */
+	@Override
+	public void setSampler(TelemetrySampler telemetrySampler) {
```
dhaval24: This should be deprecated. It is the old sampling strategy which we do not encourage any more. If we are not scrapping it off entirely then atleast please use @depracted tag here

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -160,4 +192,28 @@ public Thread newThread(Runnable r) {
...
+	}
+	
+	@Override
+	public void addTransmissionHandler(TransmissionHandler handler) {
+		if(handler != null) {
+			this.transmissionHandlers.add(handler);
+		}
+	}
+	
+	public void setMaxInstantRetries(int maxInstantRetries) {
```
dhaval24: Java docs are still missing here!

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -160,4 +192,28 @@ public Thread newThread(Runnable r) {
...
+		if (maxInstantRetries >= 0 && maxInstantRetries < INSTANT_RETRY_MAX) {
+			INSTANT_RETRY_AMOUNT = maxInstantRetries;
+		}
+	}
+	
+	public int getMaxInstantRetries() {
```
dhaval24: same here please update java docs here.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
grlima: It's the same as before, so we should not change it in this PR.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
grlima: It's the same as before, so we should not change it in this PR.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
grlima: @dhaval24  - Again, nothing new here. The same code was there before. We can address unrelated changes as separate PR's.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
grlima: @dhaval24 - same here. Unrelated to this PR.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
debugthings: Yes very good catch. Fixed and added null check.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -46,8 +50,17 @@
  *
  * Created by gupele on 6/29/2015.
  */
-public final class TransmissionPolicyManager implements Stoppable {
-
+public final class TransmissionPolicyManager implements Stoppable, TransmissionHandlerObserver {
+
+	private int INSTANT_RETRY_AMOUNT = 3; // Should always be set by the creator of this class
```
littleaj: this is using the java naming convention for constants. For member variables, `camelCase` is preferred, but I'm not going to block the PR for this :)

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
debugthings: Integer.parseInt does throw NumberFormatException which I already had a catch for. It should only log if we have ERROR tracing on.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
debugthings: Fair enough.

debugthings:
@dhaval24 I did add the ability to set the max retry amount from the XML so it will be available if that number is too high or low. Outside of that, it was adding javadocs or simple refactor

dhaval24:
@debugthings for me it's approved :-) I do not have any additional concerns

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
```diff
@@ -0,0 +1,192 @@
...
+				String line;
+				while ((line = bufferedReader.readLine()) != null) {
+					originalItems.add(line);
+				}
+				if (gis != null) {
+					gis.close();
+				}
+				if (bufferedReader != null) {
+					bufferedReader.close();
+				}
```
AlexKlimovMS: This should be closed in finally statement?

dhaval24: Yes this if should definitely go in finally!

littleaj: lines 125--130 should be wrapped in the finally block below

debugthings: Yup, moved to finally and added try/catch around those close() statements as well.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -80,11 +93,45 @@ public void run() {
...
+    /** 
+     * Suspend the transmission thread according to the current back off policy.
+     */
+    public void backoff() {
+    	policyState.setCurrentState(TransmissionPolicy.BACKOFF);
+    	long backOffMillis = backoffManager.backOffCurrentSenderThreadValue();
+        if (backOffMillis > 0)
+        {
+        	long backOffSeconds = backOffMillis / 1000;
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "App is throttled, telemetry will be blocked for %s seconds.", backOffSeconds);
```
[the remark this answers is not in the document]
debugthings: Possibly, I can remove the log always, The backoff is called in large increments so it would become chatty if there were multiple threads and if we were unable to send data for a large amount of time.

PR #563, opened 2018-02-16
minor fixes for specifying custom applciationinsights jars for smoke tests

littleaj:
Overview of issues:
_variables were defined twice:_ opted to use parent project definitions. this scenario is also dependent on it being defined in one place.
_guava missing from test dependencies:_ Since we re-namespace the guava dependency, it needs to be named here also so these still work when custom jars are used.

### DIFF
--- a/test/smoke/build.gradle
+++ b/test/smoke/build.gradle
@@ -10,9 +10,18 @@ allprojects {
 
 subprojects {
 	ext {
+		/* 	
+			By default, the Applciation Insights jars are built and used by the smokeTests.
+			To use different jars for smokeTests, reassign these variables to point the path to the jar using this 'files' function.
+			For example,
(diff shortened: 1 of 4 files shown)


littleaj:
@dhaval24 That's still possible. However, the purpose of the smokeTests is to test a specific version of the SDK. If you vary the version of the SDK in a smokeTest run, the results won't make sense.

dhaval24:
@littleaj I totally agree with your point. However, keeping this facility will provide atleast in my opinion the flexibility to build test apps with with specific jars/version of SDK if needed to pinpoint some issues during local development phases. Also it might help in catching regression bugs (just my thoughts though!)

littleaj:
@dhaval24 This does not remove the ability to do that.

dhaval24:
@littleaj yes I get that :-) Thanks for clarifying that!

PR #568, opened 2018-02-22
Fix thread leak

### LINKED ISSUE #513: Memory leak in SDKShutdownActivity - JVM shutdown hook
I noticed AI uses a JVM shutdownhook  to close various resources/threadpools created by AI. However this won't run until the application server is shutdown and will cause a classloader leak during redeploys. Please remove this shutdownhook when AI is running in an application server environment.

littleaj:
Fix #513 .

**_NOTE: This fix changes the lowest version of the Servet API we support from 2.5 to 3.0!_**

Here's the gist of the issue:
1. When our TelemetryClient loads it starts a bunch of threads.
2. It registers a shutdown hook with the JVM
3. When an application using our TelemetryClient is undeployed, the app server expects our threads to stop, but since the JVM isn't stopping, the threads continue to run.

Here's how I fixed it:
1. I added names to the threads we create. Added AtomicInteger for tracking thread pool names.
2. I changes the logging format to include the thread name after the thread id.
3. I added a WebListener which kills our threads when the webapp is shutdown/undeployed.

Here's some other things you'll find in this PR:
* added some additional scripts for smoketests to finish debugging.
* Annotated our filter with the Servlet 3.0 annotations for filters

- [ ] Test this in SpringBoot App
- [x] CHANGELOG.md updated


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
```diff
@@ -37,13 +37,14 @@
  */
```
dhaval24: I have a fundamental question. Since we are internally using Apache HTTP Client 4.5.3 to send the data, why do we need this sender in the first place? I think if there is anywhere still we are using HTTP Client 4.2 we should remove those. For eg `ApacheSenderFactory` looks for old jar on the class path and creates an instance of ApacheSender42. I think we do not need it at all. @littleaj can you take a look as a part of this improvements and trim the unnecessary code there ?

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -118,25 +123,51 @@ private void stopStoppables() {
...
+
+    public void stopAll() {
+        getShutdownAction().run();
     }
 
-    private SDKShutdownThread getShutdownThread() {
-        if (shutdownThread == null) {
+    private SDKShutdownAction getShutdownAction() {
+        if (shutdownAction == null) {
             synchronized (this) {
```
dhaval24: @littleaj  As a one cent advise, I would prefer to start a practice of using `Reentrant Lock` instead of synchronizing on `this`. There are numerous benefits . I would like everyone to go over. 
https://stackoverflow.com/questions/11821801/why-use-a-reentrantlock-if-one-can-use-synchronizedthis
Also it is much more cleaner and less prone to errors.

web/build.gradle
```diff
@@ -36,7 +36,7 @@ dependencies {
     compile ([group: 'org.apache.httpcomponents', name: 'httpasyncclient', version: '4.1.3'])
     provided 'com.opensymphony:xwork:2.0.4' // Struts 2
     provided 'org.springframework:spring-webmvc:3.1.0.RELEASE'
-    provided group: 'javax.servlet', name: 'servlet-api', version: '2.5'
+    provided group: 'javax.servlet', name: 'javax.servlet-api', version: '3.0.1'
```
dhaval24: Please make sure to register this on OSS before merging.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/AppProfileFetcher.java
```diff
@@ -21,14 +21,16 @@
...
 import java.io.IOException;
 import java.util.concurrent.ExecutionException;
+
 import org.apache.http.ParseException;
 
 /**
  * Retrieves the application profile from storage
  */
- public interface AppProfileFetcher {
+ public interface AppProfileFetcher extends Closeable {
```
dhaval24: For this @littleaj could you please confirm if this doesn't break spring boot scenario. Specifically I would like to see that with this change do we still need to register the Bean for WebRequestTrackingFilter or the framework can automatically pick it?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -39,7 +46,7 @@
 
 public class CdsProfileFetcher implements AppProfileFetcher {
 
-	private HttpAsyncClient httpClient;
+	private CloseableHttpAsyncClient httpClient;
```
dhaval24: why shouldn't this variable be final? Do we ever need to do multiple initialization for this?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
littleaj: That's something we can address separately. If it's worth exploring we should file an issue.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
littleaj: I don't see any benefits of using `RentrantLock`. IMO, Java's lock classes are much less readable and harder to debug. I'm happy to listen to evaluate some evidence why `synchronized` is inadequate for this case.

However, (quoting the best answer from the linked stackoverflow article):
> ... use it when you actually need something it provides that synchronized doesn't ... but remember that the vast majority of synchronized blocks hardly ever exhibit any contention, let alone high contention. I would advise developing with synchronization until synchronization has proven to be inadequate, rather than simply assuming "the performance will be better" if you use ReentrantLock. Remember, these are advanced tools for advanced users. (And truly advanced users tend to prefer the simplest tools they can find until they're convinced the simple tools are inadequate.) ...

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
littleaj: The setter is public

littleaj:
@dhaval24 Regarding your comment about `AtomicInteger`. It does appear to be possible to have multiple instances of some of these classes, so I've refactored the `ThreadFactory` creation to include and instance identifier.

web/build.gradle
littleaj: Pending review

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/AppProfileFetcher.java
littleaj: Sounds good. I'll see what I can find out.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
dhaval24: Please file an issue to keep an eye on. We can then close if not needed.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
dhaval24: Ok, I have no objections. I personally feel writing synchronized is two step process whereas reentrant lock just makes it more readable and easy.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
dhaval24: Sure in this case it cannot be fine but atleast @littleaj the setter can be set to package private (aka default access). This setter is only used in CdsProfileFetcherTests.java which resides in the same package. I know this is not part of your PR but it's good if we can address if possible in this PR.

dhaval24:
@littleaj yes it doesn't appear to have the multiple instances of the classes, but there is no restriction in code base that it would prevent a developer to make that mistake in the future. Either we should add those restrictions on make sure that ThreadIds you are providing are unique always. May be the logic you added in your recent commit will ensure that.

littleaj:
@dhaval24 The instance identifier does this.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
littleaj: @dhaval24 I don't think we can reduce the visibility of a public method in a public class w/o going through the deprecation procedure.

dhaval24:
@littleaj really sorry for the later reply. But I found an issue with SpringBoot.I was experimenting more with this and when I instrumented it with Agent I found that we won't be able to use the parameterized constructor in web request tracking filter to bind the name to the filter. 

Therefore, this won't eliminate creating the two beans I mentioned in my above comment, but would be mandatory. 

This won't block your PR though, just there would be no advantage in case of SpringBoot. Apologies again.

PR #572, opened 2018-02-23
Reliable channel: close resources in finally block.

grlima:
PR #561 

- [ ] CHANGELOG.md updated

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
@@ -114,27 +114,36 @@ public class PartialSuccessHandler implements TransmissionHandler {
 
 		if (args.getTransmission().getWebContentEncodingType() == "gzip") {
 
+			GZIPInputStream gis = null;
+			BufferedReader bufferedReader = null;
+
 			try {
-				GZIPInputStream gis = new GZIPInputStream(
+				gis = new GZIPInputStream(
 						new ByteArrayInputStream(args.getTransmission().getContent()));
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/PartialSuccessHandler.java
```diff
@@ -114,27 +114,36 @@ boolean validateTransmissionAndSend(TransmissionHandlerArgs args) {
...
 				String line;
 				while ((line = bufferedReader.readLine()) != null) {
 					originalItems.add(line);
 				}
+			} catch (IOException ex) {
+				InternalLogger.INSTANCE.error("IOException: Error while reading the GZIP stream.%nStack Trace:%n%s", ExceptionUtils.getStackTrace(ex));
+			} catch (Throwable t) {
+				InternalLogger.INSTANCE.error("Error while reading the GZIP stream.%nStack Trace:%n%s",	ExceptionUtils.getStackTrace(t));
+			} finally {
 				if (gis != null) {
```
littleaj: If you want, you can inline `gis` and remove this block. `InputStreamReader` and `BufferedReader` both close thier respective reader/stream in each of their `close` methods.

dhaval24: @littleaj yes I would like to see this change. Thanks for pointing out. It's not blocking but these are good performance constructs!

grlima: @littleaj - thanks, will follow up with you in person

PR #576, opened 2018-02-26
Bugfix against retry logic

debugthings:
When reviewing this code to implement a few refactored items I noticed that there was still a retry hard code in the logic. This PR fixes that.

As it would be there are still a mix of tabs and spaces in the files that cause some additional lines for the commit that don't need to be there.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
@@ -189,10 +189,8 @@ public final class TransmissionNetworkOutput implements TransmissionOutput {
 				respString = EntityUtils.toString(respEntity);
 				retryAfterHeader = response.getFirstHeader(RESPONSE_THROTTLING_HEADER);
 
-				// After the third time through this dispatcher we should reset the counter and
-				// then fail to second TransmissionOutput
-				if (code > HttpStatus.SC_PARTIAL_CONTENT && transmission.getNumberOfSends() >= MAX_RESEND) {
-					transmission.setNumberOfSends(0);
+				// After we reach our instant retry limit we should fail to second TransmissionOutput
+				if (code > HttpStatus.SC_PARTIAL_CONTENT && transmission.getNumberOfSends() > this.transmissionPolicyManager.getMaxInstantRetries()) {
 					return false;
 				} else if (code == HttpStatus.SC_OK) {
 					// If we've completed then clear the back off flags as the channel does not need
@@ -231,7 +229,7 @@ public final class TransmissionNetworkOutput implements TransmissionOutput {
 				}
 				httpClient.dispose(response);
 
-				if (code != HttpStatus.SC_OK && transmission.getNumberOfSends() < MAX_RESEND) {
+				if (code != HttpStatus.SC_OK) {
 					// Invoke the listeners for handling things like errors
 					// The listeners will handle the back off logic as well as the dispatch
 					// operation
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
@@ -50,8 +50,7 @@ import java.util.concurrent.TimeUnit;
  *
  * Created by gupele on 12/18/2014.
  */
-public final class TransmissionNetworkOutput implements TransmissionOutput {
-	private static final int MAX_RESEND = 3;
+public final class TransmissionNetworkOutput implements TransmissionOutput {	
 	private final static String CONTENT_TYPE_HEADER = "Content-Type";
 	private final static String CONTENT_ENCODING_HEADER = "Content-Encoding";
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -104,32 +104,32 @@ public TransmissionPolicyManager(boolean throttlingIsEnabled) {
...
-    	policyState.setCurrentState(TransmissionPolicy.BACKOFF);
-    	long backOffMillis = backoffManager.backOffCurrentSenderThreadValue();
-        if (backOffMillis > 0)
+        policyState.setCurrentState(TransmissionPolicy.BACKOFF);
+        long backOffMillis = backoffManager.backOffCurrentSenderThreadValue();
+        if (backOffMillis > 0) 
         {
             long backOffSeconds = backOffMillis / 1000;
-            InternalLogger.INSTANCE.info("App is throttled, telemetry will be blocked for %s seconds.", backOffSeconds);
+            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.TRACE, "App is throttled, telemetry will be blocked for %s seconds.", backOffSeconds);
```
grlima: Will this flood the log and make users unhappy?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -189,10 +189,8 @@ public boolean send(Transmission transmission) {
 				respString = EntityUtils.toString(respEntity);
 				retryAfterHeader = response.getFirstHeader(RESPONSE_THROTTLING_HEADER);
 
-				// After the third time through this dispatcher we should reset the counter and
-				// then fail to second TransmissionOutput
-				if (code > HttpStatus.SC_PARTIAL_CONTENT && transmission.getNumberOfSends() >= MAX_RESEND) {
```
grlima: The variable MAX_RESEND is no longer needed

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
dhaval24: Well I think it has more consequences than just making user unhappy. Writing logs to disk (i.e file) or console is expensive application and during peak load there can be some sort of throttling in sending some telemetry but because of extensive logging this slow production application down and then can lead to missing important transactions (Say in high throughput e-commerce platform). This should not be log always.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -162,10 +162,10 @@ public TransmissionPolicyStateFetcher getTransmissionPolicyState() {
...
     private synchronized void doSuspend(TransmissionPolicy policy, long suspendInSeconds) {
         try {
-            if (policy == TransmissionPolicy.UNBLOCKED ) {
+            if (policy == TransmissionPolicy.UNBLOCKED) {
                 return;
             }
-                       
+
             Date date = Calendar.getInstance().getTime();
             date.setTime(date.getTime() + 1000 * suspendInSeconds);
```
dhaval24: Can we add explicit braces here.  date.setTime(date.getTime() + (1000 * suspendInSeconds))
I know that multiplication takes precedence over addition but I would still prefer to be explicit to avoid any malfunctions in situations.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
```diff
@@ -214,36 +214,36 @@ public Thread newThread(Runnable r) {
...
+        if (handler != null) {
+            this.transmissionHandlers.add(handler);
+        }
+    }
+
+    /**
+     * Set the number of retries before performing a back off operation.
+     * @param maxInstantRetries Number of retries
+     */
+    public void setMaxInstantRetries(int maxInstantRetries) {
```
dhaval24: I think we should always keep minimum instant retires to 3. Allowing instant retries to be set to 0 will again invite us for similar troubles in constrained networks. I would suggest that we change the condition to reflect that.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
littleaj: I agree with @dhaval24, this shouldn't be `logAlways`. I would go with `debug`.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java
debugthings: instantRetries defaults to 3 if it is not set with this method, changed logic to only work if value is range [1..10]. I agree that 0 will put us in to a condition where we are backing off too soon, but disagree that we should always have this at 3 and give no option to lower.

PR #578, opened 2018-02-27
Fixed javadocs tags, that caused build errors

Mart-Bogdan:
Fixed javadocs tags, that caused build errors while executing `javadoc` gradle task

**P.S.** do I need issue and CHANGELOG for this tiny fix?

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/TransmissionHandler.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/TransmissionHandler.java
@@ -3,10 +3,9 @@ package com.microsoft.applicationinsights.internal.channel;
 /**
  * An interface that is used to create a concrete class that is called by the the {@link TransmissionHandlerObserver}
  * <p>
- * This is used to implement classes like {@link ErrorHandler} and {@link PartialSuccessHandler}.
+ * This is used to implement classes like {@link com.microsoft.applicationinsights.internal.channel.common.ErrorHandler}
+ * and {@link com.microsoft.applicationinsights.internal.channel.common.PartialSuccessHandler}.
  * @author jamdavi
- * 
- * 
  */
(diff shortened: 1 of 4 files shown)


littleaj:
@dhaval24 This is just what is needed. It either needs `import` statements or FQ name so the javadoc generator can find the symbols.

Also, import statements only affect compile time. There unused imports have no effect on bytecode.

Mart-Bogdan:
Hello, @dhaval24 , I uderstand why you are against imports, while it's not affecting generated code.

Issue I may think of is that IDE shows full class name in Doc explorer
image
In contrast of: 
image

But I've decided to double-check, generated HTML documentation shows short names, so it seems ok:
image


It's no problem for me to change imports.

Just thought, perhaps some IDE unaware of doc coments removed "unused imports"?

So what reolution would be, I could specify types exactly,
but doing fillowing is unlikely `{@link com.microsoft.applicationinsights.internal.channel.common.PartialSuccessHandler PartialSuccessHandler}`

dhaval24:
@littleaj I disagree on this. I do not think that it is appropriate to keep the unused import statements in the code base. I think this is absolutely a bad practice and be it any phase (compile or runtime optimization should be emphasized). If we need to replace this by FQ names we should replace it with fully qualify names. Also I thought that this bug is more or less fixed.

https://bugs.java.com/view_bug.do?bug_id=4652655

dhaval24:
@Mart-Bogdan thank you for understanding the concerns here. I would personally much appreciate if you can change this to fully qualified names instead. I believe this was a bug way back and all modern industry IDEs including Intellij and Eclipse should have this in their latest releases fixed :)

Mart-Bogdan:
@dhaval24 I've changed imports, as you suggested.

By the way, I'm pushing to `2.0.0-STABLE` branch, don't know is that OK.

PR #582, opened 2018-02-28
Fix customizing pom.xml in Gradle build

### LINKED ISSUE #581: Published POM for 2.0.0 is invalid
Resolving from jcenter gives this error: 
```
> Could not resolve com.microsoft.azure:applicationinsights-core:2.0.0.
      > Could not get resource 'https://jcenter.bintray.com/com/microsoft/azure/applicationinsights-core/2.0.0/applicationinsights-core-2.0.0.pom'.
         > Could not GET 'https://jcenter.bintray.com/com/microsoft/azure/applicationinsights-core/2.0.0/applicationinsights-core-2.0.0.pom'. Received status code 409 from server: Conflict
```

https://jcenter.bintray.com/com/microsoft/azure/applicationinsights-core/2.0.0/applicationinsights-core-2.0.0.pom shows the error message:
"The target deployment path 'com/microsoft/azure/applicationinsights-core/2.0.0/applicationinsights-core-2.0.0.pom' does not match the POM's expected path prefix 'com/microsoft/azure/core/2.0.0'. Please verify your POM content for correctness and make sure the source path is a valid Maven repository root path."

lhotari:
Fixes #581 and improves the way how `pom.xml` files are customized .


dhaval24:
@lhotari thank you very much for your valuable contribution and standing up to help in our critical time. I really appreciate it. Hope you give application insights a try and let us know the experience. We would constantly try to take feedback and improve. Looking forward to see even more contribution in future :) @littleaj will review your build changes and suggest any updates or questions he may have as he primarily manages it! I will glance my eyes through it also.

dhaval24:
@lhotari briefly looking over this, we do have currently one issue in our POM. It mentions some of the dependencies as guava, httpclient etc that we shade as 1st order compile dependencies. This can lead to potential class path hells. We do not really need them to be mentioned as first order dependencies in the POM as they are shaded. #583 for more details. Does this PR fixes that also?

lhotari:
@dhaval24 I now added a commit to this PR to also fix #583

gradle/publishing.gradle
```diff
@@ -58,9 +58,41 @@ dependencies {
     mavenDeployer "org.apache.maven.wagon:wagon-ftp:2.8"
 }
 
-afterEvaluate {
-    assert (!projectPomName.isEmpty() && !projectPomDescription.isEmpty())
-    pom {
+uploadArchives { task ->
```
dhaval24: Where do these parameters(username, pass, url)  get passed through now?  I see you trimmed down the the hard coded urls.

lhotari: I didn't change the way the parameters are passed. The mavenRepositoryUrl, mavenUsername and mavenUserPassword parameters are assumed to be passed by gradle properties, just like before. To explain the change that you have commented on:
Instead of adding the uploadArchives task in afterEvaluate, I moved the repository config to the doFirst block of uploadArchives task. I also changed the way that the existence of mavenRepositoryUrl, mavenUsername and mavenUserPassword parameters are checked. These parameters are only required when the uploadArchives task is part of the Gradle task graph and will be executed during the build. The [same approach was used in Gradle core build](https://github.com/gradle/gradle/blob/80be49b/gradle/publish.gradle#L47-L73).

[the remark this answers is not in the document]
lhotari: @littleaj Thanks for the clarification. I added a commit bed9829 to this PR that makes #592 obsolete. That commit adds a task "generatePom" that generates a pom beside the artifact jar file. This task is a now a dependency of copyLibsToGlobalArtifactsFolder task, so it should cover your way to publish. 
I think it makes sense to keep "uploadArchives" and "install" tasks since those help users of the library to work with local forks and patches. Users would typically want to start using a change internally before the changes have been merged upstream.

PR #584, opened 2018-03-01
Merge 2.0.0 changes to master

littleaj:
This brings the branches back together after the fork for 2.0.0.


gradle.properties
```diff
@@ -1,4 +1,4 @@
 // Project properties
-version=2.0.0-BETA
+version=2.0.0
```
grlima: There are missing STABLE release commits in this PR. Stable branch is at version 2.0.1, for example.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - Fixed PageView telemetry data not being reported. 
 - Fixed Issue #526 (NPE in MapUtil.copy())
 
+## Version 2.0.0
+
+- Fix #506 Null Reference Check causing Null Pointer Exception in `TelemetryCorrelationUtils.java`
+
```
grlima: There should be more items in this changelog for 2.0.0 - i believe this PR is missing the last commits from Stable branch

gradle.properties
dhaval24: @littleaj did you pulled the latest stable before merging locally?

dhaval24:
Also FYI don't delete the stable branch. We might need if we want to do a hot fix.

littleaj:
@grlima @dhaval24 Fixed. I needed to update my local 2.0.0-STABLE apparently.

Also, Since we have a tag on the commit we released, we can delete the branch; but I'll leave it for convenience. If we ever need to do this again, we can just merge the _version_ branch (e.g. 2.0.0-STABLE) straight into master after the release is tagged. We won't lose the commit as long as something is pointing to it. See this post: https://stackoverflow.com/questions/33283350/what-happen-to-git-tags-pointing-to-a-removed-commit

gradle.properties
littleaj: @dhaval24 nope :) fixed.

lhotari:
@littleaj please review #582 since that's a PR for 2.0.0-STABLE branch.

dhaval24:
@lhotari thanks for pointing out. Yes we should review this before merge.

littleaj:
@grlima I think all the issues have been addressed.

PR #594, opened 2018-03-07
Adding test case for PageView

littleaj:
In addition to the test:

* This also includes adding tomcat 8.5 to test envs (missing from previous PR)
* Included `STARTED` event in smoketests so it's clear when the test method has started in the logs
* Fixed deserialization code in smoke tests to support PageViewData


grlima:
Nice. Thanks for adding this.

PR #602, opened 2018-03-10
Enable Cold Start

### LINKED ISSUE #590: Revisit All logAlways() Logs in the SDK
We have several places in the SDK where logAlways() messages have been used and sometimes it is just not necessary to have them. Let;'s revisit them and start pruning or converting them to ERROR level verbosity as needed.

This would facilitate the cold start of our SDK.

dhaval24:
This PR Enables the cold start of the SDK by removing most of the logAlways() calls. Only throttling statements now have logAlways() calls in the SDK. Every other place this call has been removed to reduce tracing.

Fix #590


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilder.java
```diff
@@ -110,9 +111,10 @@ public AgentConfiguration parseConfigurationFile(String baseFolder) {
                 try {
                     boolean debugMode = Boolean.valueOf(debugModeAsString);
                     agentConfiguration.setDebugMode(debugMode);
-                    InternalAgentLogger.INSTANCE.logAlways(InternalAgentLogger.LoggingLevel.ERROR, "Instrumentation debug mode set to '%s'", debugMode);
+                    InternalAgentLogger.INSTANCE.warn("Instrumentation debug mode set to '%s'", debugMode);
```
littleaj: IMO, this is an info message.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
```diff
@@ -86,24 +85,26 @@ public InputStream getConfigurationFile() {
...
 
         if (configurationFile != null) {
-            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.INFO, "Configuration file has been successfully found in: '%s'", configurationFile);
+            InternalLogger.INSTANCE.info("Configuration file has been successfully found in: '%s'", configurationFile);
             try {
                 return new FileInputStream(configurationFile);
             } catch (FileNotFoundException e) {
-                InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Configuration file '%s' could not be opened for reading", configurationFile);
-                InternalLogger.INSTANCE.trace("stack trace is %s", ExceptionUtils.getStackTrace(e));
+                InternalLogger.INSTANCE.warn("Configuration file '%s' could not be opened for reading %s", configurationFile,
```
littleaj: generally, i think it's a good idea to have a colon (`:`) before the stack trace printout.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
```diff
@@ -86,24 +85,26 @@ public InputStream getConfigurationFile() {
...
+                        ExceptionUtils.getStackTrace(e));
             }
         } else {
-            InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Configuration file '%s' could not be found", configurationFileName);
+            InternalLogger.INSTANCE.error("Configuration file '%s' could not be found", configurationFileName);
```
littleaj: Why error instead of warn?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilder.java
dhaval24: Sure should be info. I dropped one level down (it was ERROR before). I will drop it down to INFO. Thanks for pointing out.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
dhaval24: Sure will add a colon before stacktrace statements.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ConfigurationFileLocator.java
dhaval24: I thought that AI configuration file is required. I am aware that there is certain sort of default configuration but still I thought that config file has iKey. unless user passes it as an environment variable it needs to be present here.

dhaval24:
@littleaj just added  a commit to better decorate the log messages.

dhaval24:
@littleaj I have merged the latest master and have resolved all the merge conflicts. Please take a look once and then I would merge this.

dhaval24:
@littleaj thanks I ran `gradle prepare` and it succeeded without issues.

PR #603, opened 2018-03-11
Fix Status Code in RequestTelemetry when Servlet Throws Exception

### LINKED ISSUE #600: Filter doesn't see request result code when Servlet throws exceptions
When a Servlet throws an exception, the filter properly reports the exception. However, when the request telemetry is decorated with information from the HttpServletResponse, the response at this time still has response code set to 200.

In the cloud, this shows two bits of telemetry: a request and an exception; but the request is marked successful when it should not.

From my initial analysis, the exception is still "on its way up" through the filter chain. Nothing has indicated to the HttpServletResponse object something has failed. We want to report the result code that the caller ultimately receives. This seems to only be incorrect when the Servlet throws an exception.

dhaval24:
This PR fixes the issue where the where the `WebRequestTelemetryFilter` stamps the response code as 200 OK when the Servlet Throws. 

The fix includes explicitly setting the response code as 500 on the `RequestTelemetry` item when filter catches `RunTimeException`, `IOException`or `ServletException`.  This needs to be set explicitly because the response code officially get's set from `org.apache.catalina.core. StandardWrapparValue` class's `invoke()` method in it's catch block. This method is called way later after the FilterChain completes it's execution.

This PR also Deprecates the `ApplicationInsightsHttpResponseWrapper` class and all the places where casting was done to this, as we no longer require this class as of Servlet API 3.0.1 onwards. It already has `getStatus()` method for which this wrapper was designed. 

Fix #600

- [x] CHANGELOG.md updated

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
@@ -21,30 +21,29 @@
 
 package com.microsoft.applicationinsights.web.internal;
 
+import com.microsoft.applicationinsights.TelemetryClient;
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import com.microsoft.applicationinsights.agent.internal.coresync.impl.AgentTLS;
+import com.microsoft.applicationinsights.common.CommonUtils;
+import com.microsoft.applicationinsights.internal.agent.AgentConnector;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.util.ThreadLocalCleaner;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.annotation.WebFilter;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Date;
 import java.util.LinkedList;
-
-import javax.servlet.Filter;
-import javax.servlet.ServletRequest;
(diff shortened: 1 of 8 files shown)


dhaval24:
All broken tests are fixed. It needed Servlet API as testCompile also.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -75,28 +74,35 @@
...
-            onException(re, req, response, isRequestProcessedSuccessfully);
-            throw re;
-        } finally {
+        } catch (ServletException | IOException | RuntimeException ex) {
+
+            //Explicitly set the response to 500 as the above three exceptions converts to 500
+            //later in the the chain inside the catch block of org.apache.catalina.core.StandardWrapperValue class
+            //The call to this method happens after the Filter Chain has been completed and thus response object
+            //can never have this value at this time.
+            response.setStatus(500);
```
littleaj: We can't do this. We must discover what the app is returning to the client.

dhaval24: @littleaj if you have better alternatives I am happy to take that. But in all my test I did not found any way to control this behavior. Filter chain gets released way before 500 is set.

dhaval24: In case of non-tomcat app servers where some other class will govern setting this value, it would also follow the same patter of setting 500. Now ordering might change but the value of response code should be the same in any app server regardless for this three exceptions.

dhaval24:
Please take a look at the last comment here : https://stackoverflow.com/questions/20772837/java-servlet-throw-http-500-error-on-purpose

dhaval24:
@littleaj I got your explanation now. As a work around as discussed we can set the resultcode to un known and the success = false. Ideally as suggested we might need to hook at lower level then filters.

We can rest of the goodies from the branch like deprecating ApplicationInsightsHttpResponseWrapper

dhaval24:
Closing as we are researching on better alternatives.

DineshAyyapillai:
@littleaj @dhaval24 Do we find any alternative solution for this issue? we are also looking for fix similar to this issue

PR #604, opened 2018-03-11
add MaxInstantRetry XML Element in Channel configuration

### LINKED ISSUE #591: Provide an XML Configuration to configure the value of MaxInstantRetry
Currently the `InProcessTelemetryChannel` has the capacity to programatically set the value of MaxInstantRetries between [1.,10]. However, this value should be configurable via XML Configuration.

Possible solution :  
Add XML attribute `MaxInstantRetry` in the `ChannelXMLElement` class for configuration to pick this up.

dhaval24:
Fix #591 

This PR adds ability to configure MAX_INSTANT_RETRIES from the channel configuration.


dhaval24:
@debugthings don't worry about this! I was playing around the code and added this small enhancement :-)

dhaval24:
@littleaj thanks for reviewing. So James' PR #561 uses this configuration value. It's just that the value was never getting passed from XML. I just added it in this PR :-)

littleaj:
@dhaval24 ok cool. Thanks!

PR #613, opened 2018-03-19
Update README.md

dhaval24:
Since we already released stable 2.0.1 version of the SDK, updated the readme to remove that section. Please comment if there is anything else.

cc: @nikmd23


nikmd23:
LGTM

PR #614, opened 2018-03-20
Change JmxMetricPerformanceCounter

### LINKED ISSUE #609: JMX Performance Counters are reported with same time stamp
JMX Performance counters are getting reported with the same time stamp. Apparently it seems the issue is with `JmxMetricPerformanceCounter` class which holds a final reference to `MetricTelemetry` object. 

Time stamp of Telemetry reference is set in the `TelemetryClient` track() method based on null check on Timestamp field. Therefore this reference get's time stamped once and there after never updates its's value.

Ideally we should reset the Telemetry Item / send a new instance of the MetricTelemetry.

brunolins16:
This change allow **JmxMetricPerformanceCounter** to send a new instance of MetricTelemetry instead of reuse it.

Fix #609

- [ ] CHANGELOG.md updated

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
@@ -34,16 +34,17 @@ import com.microsoft.applicationinsights.telemetry.MetricTelemetry;
  * Created by gupele on 3/15/2015.
  */
 public final class JmxMetricPerformanceCounter extends AbstractJmxPerformanceCounter {
-    private final MetricTelemetry telemetry = new MetricTelemetry();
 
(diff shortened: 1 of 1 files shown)


dhaval24:
@brunolins16 have you tested this change locally? Do you see the problem fixed?

brunolins16:
@dhaval24, sure! The following table, attached, is the raw telemetry after the updated jar.
raw_telemetry

dhaval24:
@brunolins16 thanks a lot for the contribution.

dhaval24:
@littleaj and @grlima I think this can be merged. Let me know!

MicNeo:
@dhaval24 how long it will take to have that fix available in maven repo? I'm waiting for it, because I experiencing the same issue :)

dhaval24:
@MicNeo we expect to release a new version of SDK on Maven shortly. If everything goes smoothly you can expect a release of 2.0.2 version by the end of March to earliest April 1st week.

PR #619, opened 2018-03-23
too many root element annotations. there's only one root element

littleaj:
While working on our configuration XMLs' schema documentation, I noticed all of these annotations. This marks the root element of an xml document. We only have one, so I removed all the ones that were wrong.


littleaj:
@dhaval24, the smoke tests use the config file (I'm pretty sure some unit tests do too), so if that passes, I'm satisfied.

dhaval24:
I trust you 👍

PR #621, opened 2018-03-23
added waitForItems methods so to not rely on guessing the right time

littleaj:
This expands the smoke tests, adding a method which waits for a number of telemetry items matching a particular condition.

This is intended for testing perf counters.


dhaval24:
I like the comment messages. (seriously this is the last time), however you can adjust the default import settings for Intellij and then use CTRL + ALT + O to optimize import. I always keep it to do not use * settings.

PR #623, opened 2018-03-26
Adding testAutoExceptionWithFailedRequest case

Terryxxx:
@littleaj @grlima @dhaval24

### DIFF
--- a/test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
+++ b/test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
@@ -259,4 +259,21 @@ public class CoreAndFilterTests extends AiSmokeTest {
         long expected = (new Duration(0, 0, 0, 20, 0).getTotalMilliseconds());
         assertTrue(actual >= expected);
     }
+
+    @Test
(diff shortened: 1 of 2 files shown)


test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
```diff
@@ -259,4 +259,21 @@ public void testRequestSlowWithResponseTime() {
         long expected = (new Duration(0, 0, 0, 20, 0).getTotalMilliseconds());
         assertTrue(actual >= expected);
     }
+
+    @Test
+    @TargetUri("/autoExceptionWithFailedRequest")
+    public void testAutoExceptionWithFailedRequest() {
```
littleaj: @Terryxxx, please mark this test with `@Ignore` and uncomment lines 270--271. I want to be sure we see that this test is being ignored.

Also, make a comment referencing bug #600. When we fix that bug, this test should pass.

Terryxxx: @littleaj Thanks for your comment. I will finish it today.

Terryxxx:
@littleaj I have finished it. Please check it again.

PR #631, opened 2018-03-31
HeartBeat Feature Implementation

dhaval24:
We would like to introduce HeartBeat feature. 

## What is Heartbeat?
Heartbeat in computer science is a periodic signal generated by software or hardware to indicate the normal operation or perform some synchronization. (https://en.wikipedia.org/wiki/Heartbeat_(computing))

### HeartBeat in Application Insights
HeartBeat in Application Insights would send out some minimal amount of information on a scheduled interval to assist in high level monitoring of application health and would provide some meta data like OS type, Java version, Application Insights SDK version, Website Name and Host (If running on Azure WebApps) and is extensible to add more information should your application have heartbeat specific requirements. 

### HeartBeat Default Payload
Default Heartbeat payload will contain following information : 
HeartBeat would be MetricTelemetry Item, a successful healthy heartbeat would have Metric Value = 0
It would contain following custom Dimensions - 
OS Name - Indicates the name of operating system on which application is running
Java Version - Indicates the JDK version which is used by the application.
SDK Version - Indicates the version of ApplicationInsights Java SDK being used
processSessionId - a random guid which represents each running session of application

If Application is running on Azure WebApps it would also have following information -
Website Name - Name of the website 
Host Stamp Name - Name of the host

### Core Components
`HeartBeatProviderInterface` - Defines the implementation of concrete HeartBeat Provider. Should a user want to create a custom heartbeat provider they should implement the contract of this interface.

`HearBeatProvider` - Default Concrete Implementation of HeartBeatProviderInterface for transmitting HeartBeat Metric

`HeartBeatDefaultPayloadProviderInterface` - Interface which defines HeartBeatPayloadProvider and mechanism to add heartbeat property via separate callable. If a user would like to add custom properties which belong to specific group, implementing this interface is recommended.

`BaseDefaultHeartbeatPropertyProvider` - Concrete implementation of HeartBeatDefaultPayloadProviderInterface for default properties like SDK version, OS name, Java Version and processSessionId

`WebAppsDefaultHeartbeatProvider` - Concrete implementation of HeartBeatDefaultPayloadProviderInterface for default azure webapps properties.

`HeartBeatModule` - Concrete Implementation of TelemetryModule interface which composes HeartBeatProvider class to transmit heartbeat periodically. This module is registered by default even with bare minimum telemetry configuration.

`HeartBeatPropertyPayload` - A container for holding heartbeat property value and meta data about if it has updated and health.

### Can you turn of HeartBeat ?
Yes, heartbeat can be turned off but it is highly not recommended to do this as it provides vital overview of your application health periodically.

### What is the default interval at which heatbeat is sent?
The, default interval at which heartbeat will be sent is 15 minutes.

### What is the minimum interval which can be configured to send heartbeat?
The, minimum interval which can be configured is 5 seconds.

### Can we override default heartbeat properties?
No, you cannot override default heartbeat properties. The are default and vital.

### Can you exclude default heartbeat providers?
Yes, you can but it is not recommended

- [ ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 2.1.0
+- Introduced Heartbeat support to monitor application health periodically.
+
 ## Version 2.0.2 
 - Fix incorrect success flag set when capturing HTTP Dependency.
 - Fix [#577](https://github.com/Microsoft/ApplicationInsights-Java/issues/577), removed HttpFactory class as it was not being used.
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
@@ -21,6 +21,8 @@
 
 package com.microsoft.applicationinsights.internal.config;
 
+import com.microsoft.applicationinsights.internal.heartbeat.HeartBeatModule;
+import com.microsoft.applicationinsights.telemetry.Telemetry;
 import java.io.InputStream;
 import java.util.Map;
 import java.util.List;
@@ -119,7 +121,9 @@ public enum TelemetryConfigurationFactory {
     private void setMinimumConfiguration(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
         setInstrumentationKey(userConfiguration, configuration);
         configuration.setChannel(new InProcessTelemetryChannel());
+        addHeartBeatModule(configuration);
         setContextInitializers(null, configuration);
+        initializeComponents(configuration);
     }
 
     private void setInternalLogger(SDKLoggerXmlElement sdkLogger, TelemetryConfiguration configuration) {
@@ -173,7 +177,13 @@ public enum TelemetryConfigurationFactory {
             ReflectionUtils.loadComponents(TelemetryModule.class, modules, configurationModules.getAdds());
         }
 
+        //if heartbeat module is not loaded, load heartbeat module
+        if (!heartBeatModuleAdded(modules)) {
+            addHeartBeatModule(configuration);
+        }
+
         List<TelemetryModule> pcModules = getPerformanceModules(appConfiguration.getPerformance());
+
         modules.addAll(pcModules);
     }
 
@@ -500,6 +510,18 @@ public enum TelemetryConfigurationFactory {
                         "Failed to initialized telemetry module " + module.getClass().getSimpleName() + ". Exception");
             }
         }
+    }
+
+    private void addHeartBeatModule(TelemetryConfiguration configuration) {
+        HeartBeatModule module = new HeartBeatModule(new HashMap<String, String>());
+        configuration.getTelemetryModules().add(module);
+    }
+
+    private boolean heartBeatModuleAdded(List<TelemetryModule> module) {
+        for (TelemetryModule mod : module) {
+            if (mod instanceof HeartBeatModule) return true;
+        }
+        return false;
     }
 
     void setPerformanceCountersSection(String performanceCountersSection) {
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
@@ -22,7 +22,6 @@
 package com.microsoft.applicationinsights.internal.config;
 
 import com.microsoft.applicationinsights.internal.heartbeat.HeartBeatModule;
-import com.microsoft.applicationinsights.telemetry.Telemetry;
 import java.io.InputStream;
 import java.util.Map;
 import java.util.List;
@@ -178,7 +177,7 @@ public enum TelemetryConfigurationFactory {
         }
 
         //if heartbeat module is not loaded, load heartbeat module
-        if (!heartBeatModuleAdded(modules)) {
+        if (!isHeartBeatModuleAdded(modules)) {
             addHeartBeatModule(configuration);
         }
 
@@ -512,12 +511,21 @@ public enum TelemetryConfigurationFactory {
         }
     }
 
+    /**
+     * Adds heartbeat module with default configuration
+     * @param configuration TelemetryConfiguration Instance
+     */
     private void addHeartBeatModule(TelemetryConfiguration configuration) {
         HeartBeatModule module = new HeartBeatModule(new HashMap<String, String>());
         configuration.getTelemetryModules().add(module);
     }
 
-    private boolean heartBeatModuleAdded(List<TelemetryModule> module) {
+    /**
+     * Checks if heartbeat module is present
+     * @param module List of modules in current TelemetryConfiguration Instance
+     * @return true if heartbeat module is present
+     */
+    private boolean isHeartBeatModuleAdded(List<TelemetryModule> module) {
         for (TelemetryModule mod : module) {
             if (mod instanceof HeartBeatModule) return true;
         }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
@@ -10,12 +10,34 @@ import java.util.UUID;
 import java.util.concurrent.Callable;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 
+/**
+ * <h1>Base Heartbeat Property Provider</h1>
+ *
+ * <p>
+ *   This class is a concrete implementation of {@link com.microsoft.applicationinsights.internal.heartbeat.HeartBeatDefaultPayloadProviderInterface}
+ *   It enables setting SDK Metadata to heartbeat payload.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
 public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPayloadProviderInterface {
 
+  /**
+   * Collection holding default properties for this default provider.
+   */
   private final Set<String> defaultFields;
 
-  private UUID uniqueProcessId;
+  /**
+   * Random GUID that would help in analysis when app has stopped and restarted. Each restart will
+   * have a new GUID. If the application is unstable and goes through frequent restarts this will help
+   * us identify instability in the analytics backend.
+   */
+  private static UUID uniqueProcessId;
 
+  /**
+   * Name of this provider.
+   */
   private final String name = "Base";
 
   public BaseDefaultHeartbeatPropertyProvider() {
@@ -29,10 +51,16 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
   }
 
   @Override
+  public boolean isKeyWord(String keyword) {
+    return defaultFields.contains(keyword);
+  }
+
+  @Override
   public Callable<Boolean> setDefaultPayload(final List<String> disableFields,
       final HeartBeatProviderInterface provider) {
     return new Callable<Boolean>() {
 
+      // using volatile here to avoid caching in threads.
       volatile boolean hasSetValues = false;
       volatile Set<String> enabledProperties = MiscUtils.except(defaultFields, disableFields);
       @Override
@@ -57,6 +85,7 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
                 hasSetValues = true;
                 break;
               default:
+                //We won't accept unknown properties in default providers.
                 InternalLogger.INSTANCE.trace("Encountered unknown default property");
                 break;
             }
@@ -71,6 +100,10 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
     };
   }
 
+  /**
+   * This method initializes the collection with Default Properties of this provider.
+   * @param defaultFields collection to hold default properties.
+   */
   private void initializeDefaultFields(Set<String> defaultFields) {
 
     if (defaultFields == null) {
@@ -82,10 +115,19 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
     defaultFields.add("processSessionId");
   }
 
+  /**
+   * Gets the JDK version being used by the application
+   * @return String representing JDK Version
+   */
   private String getJdkVersion() {
     return System.getProperty("java.version");
   }
 
+  /**
+   * Returns the Application Insights SDK version user is using to instrument his application
+   * @return returns string prefixed with "java" representing the Application Insights Java
+   * SDK version.
+   */
   private String getSdkVersion() {
 
     String sdkVersion = "java";
@@ -99,13 +141,20 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
     return sdkVersion + ":unknown-version";
   }
 
-
+  /**
+   * Gets the OS version on which application is running.
+   * @return String representing OS version
+   */
   private String getOsType() {
     return System.getProperty("os.name");
   }
 
+  /**
+   * Returns the Unique GUID for the application's current session.
+   * @return String representing GUID for each running session
+   */
   private String getProcessSessionId() {
-    if (this.uniqueProcessId == null) {
+    if (uniqueProcessId == null) {
       uniqueProcessId = UUID.randomUUID();
     }
     return uniqueProcessId.toString();
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
@@ -2,23 +2,24 @@ package com.microsoft.applicationinsights.internal.heartbeat;
 
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 import com.microsoft.applicationinsights.internal.util.PropertyHelper;
-import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 
 public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPayloadProviderInterface {
 
-  private final List<String> defaultFields;
+  private final Set<String> defaultFields;
 
   private UUID uniqueProcessId;
 
   private final String name = "Base";
 
   public BaseDefaultHeartbeatPropertyProvider() {
-    defaultFields = new ArrayList<>();
+    defaultFields = new HashSet<>();
     initializeDefaultFields(defaultFields);
   }
 
@@ -28,17 +29,12 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
   }
 
   @Override
-  public boolean isKeyword(String keyword) {
-    return containsIgnoreCase(keyword, defaultFields);
-  }
-
-  @Override
   public Callable<Boolean> setDefaultPayload(final List<String> disableFields,
       final HeartBeatProviderInterface provider) {
     return new Callable<Boolean>() {
 
       volatile boolean hasSetValues = false;
-      volatile List<String> enabledProperties = except(defaultFields, disableFields);
+      volatile Set<String> enabledProperties = MiscUtils.except(defaultFields, disableFields);
       @Override
       public Boolean call() {
         for (String fieldName : enabledProperties) {
@@ -60,6 +56,9 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
                 provider.addHeartBeatProperty(fieldName, getProcessSessionId(), true);
                 hasSetValues = true;
                 break;
+              default:
+                InternalLogger.INSTANCE.trace("Encountered unknown default property");
+                break;
             }
           }
           catch (Exception e) {
@@ -69,57 +68,18 @@ public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPay
         }
         return hasSetValues;
       }
-
-      private List<String> except(List<String> list1, List<String> list2) {
-        try {
-          if (list1 == null || list2 == null) throw new IllegalArgumentException("Input is null");
-          List<String> union = new ArrayList<>(list1);
-          union.addAll(list2);
-          List<String> intersection = new ArrayList<>(list1);
-          intersection.retainAll(list2);
-          union.removeAll(intersection);
-          return union;
-        }
-        catch (Exception e) {
-          InternalLogger.INSTANCE.warn("stack trace is %s", ExceptionUtils.getStackTrace(e));
-        }
-        finally{
-          if (list1 != null) return list1;
-          return list2;
-        }
-      }
     };
   }
 
-  private void initializeDefaultFields(List<String> defaultFields) {
+  private void initializeDefaultFields(Set<String> defaultFields) {
 
     if (defaultFields == null) {
-      defaultFields = new ArrayList<>();
+      defaultFields = new HashSet<>();
     }
     defaultFields.add("jdkVersion");
     defaultFields.add("sdk-version");
     defaultFields.add("osType");
     defaultFields.add("processSessionId");
-  }
-
-  private boolean containsIgnoreCase(String keyword, List<String> inputList) {
-    try {
-      if (keyword == null) throw new IllegalArgumentException("Keyword to compare is null");
-      if (inputList == null) throw new IllegalArgumentException("List to compare is null");
-      for (String key : inputList) {
-        if (key.equalsIgnoreCase(keyword)) return true;
-      }
-      return false;
-    }
-    catch (Exception e) {
-      InternalLogger.INSTANCE.warn("exception while comparision, stack trace is %s",
-          ExceptionUtils.getStackTrace(e));
-    }
-    finally{
-      //return true so we don't add property when comparison exception
-      return true;
-    }
-
   }
 
   private String getJdkVersion() {
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
@@ -1,0 +1,153 @@
+package com.microsoft.applicationinsights.internal.heartbeat;
+
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.internal.util.PropertyHelper;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+import java.util.UUID;
+import java.util.concurrent.Callable;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+
+public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPayloadProviderInterface {
+
+  private final List<String> defaultFields;
+
+  private UUID uniqueProcessId;
+
+  private final String name = "Base";
+
+  public BaseDefaultHeartbeatPropertyProvider() {
+    defaultFields = new ArrayList<>();
+    initializeDefaultFields(defaultFields);
+  }
+
+  @Override
+  public String getName() {
+    return this.name;
+  }
+
+  @Override
+  public boolean isKeyword(String keyword) {
+    return containsIgnoreCase(keyword, defaultFields);
+  }
+
+  @Override
+  public Callable<Boolean> setDefaultPayload(final List<String> disableFields,
+      final HeartBeatProviderInterface provider) {
+    return new Callable<Boolean>() {
+
+      volatile boolean hasSetValues = false;
+      volatile List<String> enabledProperties = except(defaultFields, disableFields);
+      @Override
+      public Boolean call() {
+        for (String fieldName : enabledProperties) {
+          try {
+            switch (fieldName) {
+              case "jdkVersion":
+                provider.addHeartBeatProperty(fieldName, getJdkVersion(), true);
+                hasSetValues = true;
+                break;
+              case "sdk-version":
+                provider.addHeartBeatProperty(fieldName, getSdkVersion(), true);
+                hasSetValues = true;
+                break;
+              case "osType":
+                provider.addHeartBeatProperty(fieldName, getOsType(), true);
+                hasSetValues = true;
+                break;
+              case "processSessionId":
+                provider.addHeartBeatProperty(fieldName, getProcessSessionId(), true);
+                hasSetValues = true;
+                break;
+            }
+          }
+          catch (Exception e) {
+           InternalLogger.INSTANCE.warn("Failed to obtain heartbeat property, stack trace"
+               + "is: %s", ExceptionUtils.getStackTrace(e));
+          }
+        }
+        return hasSetValues;
+      }
+
+      private List<String> except(List<String> list1, List<String> list2) {
+        try {
+          if (list1 == null || list2 == null) throw new IllegalArgumentException("Input is null");
+          List<String> union = new ArrayList<>(list1);
+          union.addAll(list2);
+          List<String> intersection = new ArrayList<>(list1);
+          intersection.retainAll(list2);
+          union.removeAll(intersection);
+          return union;
+        }
+        catch (Exception e) {
+          InternalLogger.INSTANCE.warn("stack trace is %s", ExceptionUtils.getStackTrace(e));
+        }
+        finally{
+          if (list1 != null) return list1;
+          return list2;
+        }
+      }
+    };
+  }
+
+  private void initializeDefaultFields(List<String> defaultFields) {
+
+    if (defaultFields == null) {
+      defaultFields = new ArrayList<>();
+    }
+    defaultFields.add("jdkVersion");
+    defaultFields.add("sdk-version");
+    defaultFields.add("osType");
+    defaultFields.add("processSessionId");
+  }
+
+  private boolean containsIgnoreCase(String keyword, List<String> inputList) {
+    try {
+      if (keyword == null) throw new IllegalArgumentException("Keyword to compare is null");
+      if (inputList == null) throw new IllegalArgumentException("List to compare is null");
+      for (String key : inputList) {
+        if (key.equalsIgnoreCase(keyword)) return true;
+      }
+      return false;
+    }
+    catch (Exception e) {
+      InternalLogger.INSTANCE.warn("exception while comparision, stack trace is %s",
+          ExceptionUtils.getStackTrace(e));
+    }
+    finally{
+      //return true so we don't add property when comparison exception
+      return true;
+    }
+
+  }
+
+  private String getJdkVersion() {
+    return System.getProperty("java.version");
+  }
+
+  private String getSdkVersion() {
+
+    String sdkVersion = "java";
+
+    Properties sdkVersionProps = PropertyHelper.getSdkVersionProperties();
+    if (sdkVersionProps != null) {
+      String version = sdkVersionProps.getProperty("version");
+      sdkVersion = String.format("java:%s", version);
+      return sdkVersion;
+    }
+    return sdkVersion + ":unknown-version";
+  }
+
+
+  private String getOsType() {
+    return System.getProperty("os.name");
+  }
+
+  private String getProcessSessionId() {
+    if (this.uniqueProcessId == null) {
+      uniqueProcessId = UUID.randomUUID();
+    }
+    return uniqueProcessId.toString();
+  }
+}
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
@@ -1,0 +1,14 @@
+package com.microsoft.applicationinsights.internal.heartbeat;
+
+import java.util.List;
+import java.util.concurrent.Callable;
+
+public interface HeartBeatDefaultPayloadProviderInterface {
+
+  String getName();
+
+  boolean isKeyword(String keyword);
+
+  Callable<Boolean> setDefaultPayload(List<String> disableFields, HeartBeatProviderInterface provider);
+
+}
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
@@ -3,10 +3,41 @@ package com.microsoft.applicationinsights.internal.heartbeat;
 import java.util.List;
 import java.util.concurrent.Callable;
 
+/**
+ * <h1>Interface for setting default properties</h1>
+ *
+ * <p>
+ *   This interface is used to set the default payload of a provider and defines implementation for
+ *   some helper methods to assist it.
+ *
+ *   The default concrete implementations are {@link com.microsoft.applicationinsights.internal.heartbeat.BaseDefaultHeartbeatPropertyProvider}
+ *   and {@link com.microsoft.applicationinsights.internal.heartbeat.WebAppsDefaultHeartbeatProvider}
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
 public interface HeartBeatDefaultPayloadProviderInterface {
 
+  /**
+   * Returns the name of the heartbeat provider.
+   * @return Name of the heartbeat provider
+   */
   String getName();
 
+  /**
+   * Tells if the input string is a reserved property.
+   * @param keyword string to test
+   * @return true if input string is reserved keyword
+   */
+  boolean isKeyWord(String keyword);
+
+  /**
+   * Returns a callable which can be executed to set the payload based on the parameters.
+   * @param disableFields List of Properties to be excluded from payload
+   * @param provider The current heartbeat provider
+   * @return Callable which can be executed to add the payload
+   */
   Callable<Boolean> setDefaultPayload(List<String> disableFields, HeartBeatProviderInterface provider);
 
 }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
@@ -7,8 +7,6 @@ public interface HeartBeatDefaultPayloadProviderInterface {
 
   String getName();
 
-  boolean isKeyword(String keyword);
-
   Callable<Boolean> setDefaultPayload(List<String> disableFields, HeartBeatProviderInterface provider);
 
 }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
@@ -1,0 +1,100 @@
+package com.microsoft.applicationinsights.internal.heartbeat;
+
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import com.microsoft.applicationinsights.extensibility.TelemetryModule;
+import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+public class HeartBeatModule implements TelemetryModule {
+
+  private final HeartBeatProviderInterface heartBeatProviderInterface;
+
+  public HeartBeatModule(Map<String, String> properties) {
+
+    if (properties != null) {
+      for (Map.Entry<String, String> entry : properties.entrySet()) {
+        switch (entry.getKey()) {
+          case "HeartBeatInterval":
+            try {
+              setHeartBeatInterval(Long.valueOf(entry.getValue()));
+            } catch (Exception e) {
+              //add log
+            }
+          case "isHeartBeatEnabled":
+            try {
+              setHeartBeatEnabled(Boolean.parseBoolean(entry.getValue()));
+            }
+            catch (Exception e) {
+              //log
+            }
+          case "ExcludedHeartBeatPropertiesProvider":
+            try {
+              List<String> excludedHeartBeatPropertiesProviderList = parseStringToList(entry.getValue());
+              setExcludedHeartBeatPropertiesProvider(excludedHeartBeatPropertiesProviderList);
+            }
+            catch (Exception e) {
+              //log
+            }
+          case "ExcludedHeartBeatProperties":
+            try {
+              List<String> excludedHeartBeatPropertiesList = parseStringToList(entry.getValue());
+              setExcludedHeartBeatProperties(excludedHeartBeatPropertiesList);
+            }
+            catch (Exception e) {
+              //log
+            }
+        }
+      }
+    }
+
+    heartBeatProviderInterface = new HeartBeatProvider();
+  }
+
+  public long getHeartBeatInterval() {
+    return this.heartBeatProviderInterface.getHeartBeatInterval();
+  }
+
+  public void setHeartBeatInterval(long heartBeatInterval) {
+    this.heartBeatProviderInterface.setHeartBeatInterval(heartBeatInterval);
+  }
+
+  public List<String> getExcludedHeartBeatProperties() {
+    return heartBeatProviderInterface.getExcludedHeartBeatProperties();
+  }
+
+  public void setExcludedHeartBeatProperties(List<String> excludedHeartBeatProperties) {
+    this.heartBeatProviderInterface.setExcludedHeartBeatProperties(excludedHeartBeatProperties);
+  }
+
+  public List<String> getExcludedHeartBeatPropertiesProvider() {
+    return heartBeatProviderInterface.getExcludedHeartBeatPropertyProviders();
+  }
+
+  public void setExcludedHeartBeatPropertiesProvider(
+      List<String> excludedHeartBeatPropertiesProvider) {
+      this.heartBeatProviderInterface.setExcludedHeartBeatPropertyProviders(excludedHeartBeatPropertiesProvider);
+  }
+
+  public boolean isHeartBeatEnabled() {
+    return this.heartBeatProviderInterface.isHeartBeatEnabled();
+  }
+
+  public void setHeartBeatEnabled(boolean heartBeatEnabled) {
+    this.heartBeatProviderInterface.setHeartBeatEnabled(heartBeatEnabled);
+  }
+
+  @Override
+  public void initialize(TelemetryConfiguration configuration) {
+    InternalLogger.INSTANCE.info("heartbeat is enabled");
+  }
+
+  private List<String> parseStringToList(String value) {
+    if (value == null || value.length() == 0) return new ArrayList<>();
+    List<String> valueList = Arrays.asList(value.split(";"));
+    return valueList;
+  }
+
+}
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
@@ -110,7 +110,7 @@ public class HeartBeatModule implements TelemetryModule {
       isEnabled = true;
     }
     finally{
-      lock.lock();
+      lock.unlock();
     }
 
   }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
@@ -7,12 +7,20 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 public class HeartBeatModule implements TelemetryModule {
 
   private final HeartBeatProviderInterface heartBeatProviderInterface;
 
+  private Lock lock = new ReentrantLock();
+
+  private boolean isEnabled = false;
+
   public HeartBeatModule(Map<String, String> properties) {
+
+    heartBeatProviderInterface = new HeartBeatProvider();
 
     if (properties != null) {
       for (Map.Entry<String, String> entry : properties.entrySet()) {
@@ -20,12 +28,14 @@ public class HeartBeatModule implements TelemetryModule {
           case "HeartBeatInterval":
             try {
               setHeartBeatInterval(Long.valueOf(entry.getValue()));
+              break;
             } catch (Exception e) {
               //add log
             }
           case "isHeartBeatEnabled":
             try {
               setHeartBeatEnabled(Boolean.parseBoolean(entry.getValue()));
+              break;
             }
             catch (Exception e) {
               //log
@@ -34,6 +44,7 @@ public class HeartBeatModule implements TelemetryModule {
             try {
               List<String> excludedHeartBeatPropertiesProviderList = parseStringToList(entry.getValue());
               setExcludedHeartBeatPropertiesProvider(excludedHeartBeatPropertiesProviderList);
+              break;
             }
             catch (Exception e) {
               //log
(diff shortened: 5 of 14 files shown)


dhaval24:
A general comment on the heartbeat interval. Currently there is no maximum limit on heartbeat interval. I believe C# implementation also does not have the same.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/WebAppsDefaultHeartbeatProvider.java
```diff
@@ -0,0 +1,136 @@
...
+      final HeartBeatProviderInterface provider) {
+    return new Callable<Boolean>() {
+      volatile boolean hasSetValues = false;
+      volatile Set<String> enabledProperties = MiscUtils.except(defaultFields, disableFields);
+      @Override
+      public Boolean call() {
+        for (String fieldName : enabledProperties) {
+          try {
+            switch (fieldName) {
+              case "website_site_name":
```
Dmitry-Matveev: Do you read it for every heartbeat? In .NET SDK, Telemetry Initializers cached the value, and, therefore, were missing slot swap when the name changes. It looks like you read it every time here, so heartbeat will pick new name up.

dhaval24: No, I am not reading it everytime, but this is a good point, if this is a scenario then we might have to read always or if not atleast every X minutes.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -173,7 +176,13 @@ private void setTelemetryModules(ApplicationInsightsXmlConfiguration appConfigur
             ReflectionUtils.loadComponents(TelemetryModule.class, modules, configurationModules.getAdds());
         }
 
+        //if heartbeat module is not loaded, load heartbeat module
+        if (!isHeartBeatModuleAdded(modules)) {
+            addHeartBeatModule(configuration);
```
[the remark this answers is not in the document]
dhaval24: @Dmitry-Matveev in .net SDK DiagnosticModule is used to enable HeartBeat module and I believe DiagnosticModule is turned on by default. However, in Java SDK there is no diagnostic module and if we want to enable heartbeat module by default if not specified in the configuration file I think this would be the way. If you have better option I would like to hear that. 
As far as behavior will go, both the SDKs would be consistent that means unless user has turned the HeartBeat module off explicitly in configuration it will be turned on by default with default settings.

dhaval24:
Somehow Github doesn't find @d3r3kk in reviewers list. Please feel free to add your points here Derek.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+      try {
+        if (heartbeatProperties.containsKey(propertyName) && !HeartbeatDefaultPayload.isDefaultKeyWord(propertyName)) {
+          HeartBeatPropertyPayload payload = new HeartBeatPropertyPayload();
+          payload.setHealthy(isHealthy);
+          payload.setPayloadValue(propertyValue);
+          heartbeatProperties.put(propertyName, payload);
+          setResult = true;
+        }
+        else {
+          throw new Exception("heartbeat property cannot be set without adding it first, or heartbeat"
```
d3r3kk: Given the chance to do this a second time, I would allow Add *or* Set to actually add new fields. Perhaps issue a logger warning if a user is Adding new values via Set, if there is any real danger of doing this.

Originally, I was concerned that allowing adds to occur at any time might encourage the abuse of the Heartbeat for what is more appropriate to do with TrackEvent calls. And with the heartbeats properties being around for the entire duration of a program's runtime, I was slightly concerned with memory fragmentation should any abuse occur. However, [I doubt there is much of a real problem there](https://stackoverflow.com/questions/5240860/what-causes-memory-fragmentation-in-net).

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/WebAppsDefaultHeartbeatProvider.java
d3r3kk: See Microsoft/ApplicationInsights-dotnet-server#872 for the issue tracking the slot-swap problem.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @d3r3kk  Actually, I would like to keep the convention of add by default for adding and set by default for modifying. So keeping this in mind logically add should come before modifying (aka set in this case). I agree this is a bit opinionated and restricting the users but setting the convention I think will prove to be better in long run. Lot's of tunable knobs sometimes lead to spaghetti :)

dhaval24:
@d3r3kk thank you very much for reviewing and comments. I have addressed one of it and left my opinion on another. I am glad you enjoyed reading through this.

dhaval24:
In Recent commit optimizes the performance of heartbeat module when it is disabled and did some name refactoring to better represent the functionality.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
```
cijothomas: This exception will be thrown if 1. Property is not added first. 2. propertyname is a keyword.
Would be ideal to separate

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+   */
+  private Telemetry gatherData() {
+
+    MetricTelemetry heartbeat = new MetricTelemetry(HEARTBEAT_SYNTHETIC_METRIC_NAME, 0.0);
+    Map<String, String> property = heartbeat.getProperties();
+    for (Map.Entry<String, HeartBeatPropertyPayload> entry : heartbeatProperties.entrySet()) {
+      property.put(entry.getKey(), entry.getValue().getPayloadValue());
+      double currentValue = heartbeat.getValue();
+      currentValue += entry.getValue().isHealthy() ? 0 : 1;
+      heartbeat.setValue(currentValue);
```
cijothomas: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/705 This issue is addressed right?  :)

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/WebAppsDefaultHeartbeatProvider.java
```diff
@@ -0,0 +1,147 @@
...
+   * Populates the default Fields with the properties
+   * @param defaultFields
+   */
+  private void initializeDefaultFields(Set<String> defaultFields) {
+    if (defaultFields == null) {
+      defaultFields = new HashSet<>();
+    }
+
+    defaultFields.add("website_site_name");
+    defaultFields.add("website_home_name");
```
cijothomas: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/c61fb45b5d31f5910320ea9d788de528e16d8ff0/Src/WindowsServer/WindowsServer.Shared/AppServicesHeartbeatTelemetryModule.cs#L21

Should we use the same set used by .NET?
If you are using slot, then it need to be periodically refreshed as well.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 2.1.0
+- Introduced Heartbeat support to monitor application health periodically.
```
cijothomas: rephrase to something like 'Heartbeat feature which sends periodic hearbeats with basic info about the app/runtime to Application Insights'

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+    defaultFields.add("osType");
+    defaultFields.add("processSessionId");
+  }
+
+  /**
+   * Gets the JDK version being used by the application
+   * @return String representing JDK Version
+   */
+  private String getJdkVersion() {
+    return System.getProperty("java.version");
```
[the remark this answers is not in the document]
dhaval24: This tells the JRE on which application is running. No there is no distinction as such. So in your above example it would report 1.8

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+      return sdkVersion;
+    }
+    return sdkVersion + ":unknown-version";
+  }
+
+  /**
+   * Gets the OS version on which application is running.
+   * @return String representing OS version
+   */
+  private String getOsType() {
```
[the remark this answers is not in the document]
dhaval24: Makes sense. I can do this.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
```diff
@@ -0,0 +1,185 @@
...
+    this.heartBeatProviderInterface.setHeartBeatEnabled(heartBeatEnabled);
+  }
+
+  @Override
+  public void initialize(TelemetryConfiguration configuration) {
+    if (!isInitialized) {
+      synchronized (lock) {
+        if (!isInitialized) {
+          this.heartBeatProviderInterface.initialize(configuration);
+          InternalLogger.INSTANCE.info("heartbeat is enabled");
```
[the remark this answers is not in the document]
dhaval24: I renamed this variable in the latest comment. I agree it was misleading.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+        if (!heartbeatProperties.containsKey(propertyName)) {
+             HeartBeatPropertyPayload payload = new HeartBeatPropertyPayload();
+             payload.setHealthy(isHealthy);
+             payload.setPayloadValue(propertyValue);
+             heartbeatProperties.put(propertyName, payload);
+             isAdded = true;
+             InternalLogger.INSTANCE.trace("added heartbeat property");
+        }
+        else {
+          throw new Exception("heartbeat property cannot be added twice");
```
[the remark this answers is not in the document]
dhaval24: Could be done.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+    boolean isAdded= false;
+    if (!StringUtils.isEmpty(propertyName)) {
+      try {
+        if (!heartbeatProperties.containsKey(propertyName)) {
+             HeartBeatPropertyPayload payload = new HeartBeatPropertyPayload();
+             payload.setHealthy(isHealthy);
+             payload.setPayloadValue(propertyValue);
+             heartbeatProperties.put(propertyName, payload);
+             isAdded = true;
+             InternalLogger.INSTANCE.trace("added heartbeat property");
```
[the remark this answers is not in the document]
dhaval24: We can log name-value pair. Not common practice in this SDK but good to start may be

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @cijothomas I am not sure. I will have to check this. Most likely not, but I think we can take this in the later iteration given the release time constrain by build. Let me know your thoughts.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/WebAppsDefaultHeartbeatProvider.java
dhaval24: Yes I need to change the names to match .NET. Unfortunately this were added recently in .NET, not when I implemented this.

CHANGELOG.md
dhaval24: I think this sentence is sufficient. I can link the PR for description.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * This method initializes the collection with Default Properties of this provider.
+   * @param defaultFields collection to hold default properties.
+   */
+  private void initializeDefaultFields(Set<String> defaultFields) {
+
+    if (defaultFields == null) {
+      defaultFields = new HashSet<>();
+    }
+    defaultFields.add("jdkVersion");
```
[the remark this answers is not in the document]
dhaval24: @cijothomas  Might be a good idea, but since there are different classes for each Different kind of providers I think doesn't add a lot of value since the separation of concerns is already addressed.

dhaval24:
@cijothomas I have left my opinions on your review. Thanks for taking a look. I will address some of them :)

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+      @Override
+      public Boolean call() {
+        for (String fieldName : enabledProperties) {
+          try {
+            switch (fieldName) {
+              case "jdkVersion":
+                provider.addHeartBeatProperty(fieldName, getJdkVersion(), true);
+                hasSetValues = true;
+                break;
+              case "sdk-version":
```
littleaj: Normally, I would be consistent with the naming scheme (hyphenated-case vs. camelCase). Is this in the HB event spec?

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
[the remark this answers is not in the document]
littleaj: It's generally good practice to refactor these as constants; especially if you have string "keys" that are consistent (especially if you use them in tests).

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
```diff
@@ -0,0 +1,43 @@
...
+   */
+  boolean isKeyWord(String keyword);
+
+  /**
+   * Returns a callable which can be executed to set the payload based on the parameters.
+   * @param disableFields List of Properties to be excluded from payload
+   * @param provider The current heartbeat provider
+   * @return Callable which can be executed to add the payload
+   */
+  Callable<Boolean> setDefaultPayload(List<String> disableFields, HeartBeatProviderInterface provider);
```
littleaj: Nitpick: I would consider renaming this method since _get**Xyz**_ and _set**Xyz**_ are generally recognized as field accessors.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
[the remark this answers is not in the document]
littleaj: I would go one step further:
Why initialize the provider if it's disabled? I would refactor so the responsibility of "being enabled" lives in this class. If this module is disabled, we don't need any of its downstream dependencies.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatPropertyPayload.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * <p>
+ *  Defines the Payload class to store and send heartbeat properties and allowing to keep track of updates
+ *  to them.
+ *  </p>
+ *
+ *  @author Dhaval Doshi
+ *  @since 03-30-2018
+ */
+public class HeartBeatPropertyPayload {
```
littleaj: I find most of the javadoc for the fields and methods of this class confusing.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * Telemetry client instance used to send heartbeat.
+   */
+  private TelemetryClient telemetryClient;
+
+  /**
+   * ThreadPool used for adding properties to concurrent dictionary
+   */
+  private ExecutorService executorService = Executors.newCachedThreadPool();
```
littleaj: instead of `executorService`, please name this so it gives some indication of what the threads will be doing. For example, `propertyUpdateService`.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * Threadpool used to send data heartbeat telemetry
+   */
+  private ScheduledExecutorService scheduledExecutorService = Executors.newScheduledThreadPool(1);
```
littleaj: Same here; e.g. `heartbeastSenderService`.

littleaj: If you're going to initialize this with only 1 thread allowed make sure the Runnables allowed in it will always complete before the next one is scheduled to run. This also depends if this pool works on a timer or a delay (it could support both).

What happens if heartbeat@t_0 is stuck in the pool and heartbeat@t_1 gets scheduled?

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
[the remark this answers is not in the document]
littleaj: I agree with @cijothomas. I don't see value in this log statement _unless_ the kvp is included.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: @dhaval24, I'm inclined to agree with @d3r3kk. See the methods defined in Java's Map class. There could be places in the code where you won't know if a key has been added or not. I'm not a fan of the _if containsKey **set(x=y)** else **add(x=y)**_ pattern when you can have a simple _**set(x=y)**_.

Just as in Java's Map, putIfAbsent can be nice, if there's a precedent.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
```diff
@@ -0,0 +1,56 @@
+package com.microsoft.applicationinsights.internal.heartbeat;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+public class HeartbeatDefaultPayload {
```
littleaj: I would think this class is the Payload, but really its delegating things to PayloadProviders. Consider renaming.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * This class contains some misc. functions used in heartbeat module.
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
+public class MiscUtils {
+
+  /**
+   * Returns a list which contains Items present in list2 but not in the set.
```
littleaj: Careful with capitalization. "Items" implies there is a class named "Item".

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
```diff
@@ -0,0 +1,69 @@
...
+   * @param set
+   * @param list2
+   * @return
+   */
+  public static Set<String> except(Set<String> set, List<String> list2) {
+    try {
+      if (set == null || list2 == null) throw new IllegalArgumentException("Input is null");
```
littleaj: Separate precondition checks

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class MiscUtils {
+
+  /**
+   * Returns a list which contains Items present in list2 but not in the set.
+   * @param set
+   * @param list2
+   * @return
+   */
+  public static Set<String> except(Set<String> set, List<String> list2) {
```
littleaj: This is `L - S`. I would reverse the parameter order.

Also, it may be easier to do this: `return new HashSet<String>(list).removeAll(set)`

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
```diff
@@ -0,0 +1,69 @@
...
+      Set<String> result = new HashSet<>();
+      for (String s : list2) {
+        if (!set.contains(s)) {
+          result.add(s);
+        }
+      }
+      return result;
+    }
+    catch (Exception e) {
+      InternalLogger.INSTANCE.warn("stack trace is %s", ExceptionUtils.getStackTrace(e));
```
littleaj: utility methods shouldn't log anything

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
```diff
@@ -0,0 +1,69 @@
...
+    try {
+      if (keyword == null) throw new IllegalArgumentException("Keyword to compare is null");
+      if (inputList == null) throw new IllegalArgumentException("List to compare is null");
+      for (String key : inputList) {
+        if (key.equalsIgnoreCase(keyword)) return true;
+      }
+      return false;
+    }
+    catch (Exception e) {
+      InternalLogger.INSTANCE.warn("exception while comparision, stack trace is %s",
```
littleaj: Same here. Utility methods shouldn't log.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: Please specify a ThreadPool which names the threads. See another class to show you the pattern.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: This also needs a thread pool which names threads.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
dhaval24: @littleaj I agree with your point. I still need to match the naming scheme of the .NET. So there can be something off in the names here.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
dhaval24: @littleaj this was my thought, but I wanted to keep API very similar to C# implementation in this case following java camel case for consistency. Let me know if you still think that we should rename.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: I will address both of this. Thanks

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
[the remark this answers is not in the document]
dhaval24: Great idea, I should probably add this check. It is usless to perform operation of initialization if the module is disabled. I think this should checked in conjucation with isHeartBeatEnabled. I will do the updates.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
littleaj: makes sense

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * <h1>Base Heartbeat Property Provider</h1>
+ *
+ * <p>
+ *   This class is a concrete implementation of {@link com.microsoft.applicationinsights.internal.heartbeat.HeartBeatDefaultPayloadProviderInterface}
+ *   It enables setting SDK Metadata to heartbeat payload.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
```
littleaj: I would remove the `@since` line. It's for specifying the version where an entity first appeared. We don't use it anywhere else. If you feel it's necessary, replace the date with the release version number where this feature will appear.

The same for the other 7 occurrences of `@since`.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  public Callable<Boolean> setDefaultPayload(final List<String> disableFields,
+      final HeartBeatProviderInterface provider) {
+    return new Callable<Boolean>() {
+
+      // using volatile here to avoid caching in threads.
+      volatile boolean hasSetValues = false;
+      volatile Set<String> enabledProperties = MiscUtils.except(defaultFields, disableFields);
```
littleaj: This doesn't need to be volatile. The reference is not updated.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+    return defaultFields.contains(keyword);
+  }
+
+  @Override
+  public Callable<Boolean> setDefaultPayload(final List<String> disableFields,
+      final HeartBeatProviderInterface provider) {
+    return new Callable<Boolean>() {
+
+      // using volatile here to avoid caching in threads.
+      volatile boolean hasSetValues = false;
```
littleaj: This doesn't need to be volatile. You're reasoning in the comment is incorrect. No other thread can change this value. In fact, these should just be local to the `call()` method to make it a bit easier for GC to clean it up.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * <p>
+ *   This class is a concrete implementation of {@link com.microsoft.applicationinsights.internal.heartbeat.HeartBeatDefaultPayloadProviderInterface}
+ *   It enables setting SDK Metadata to heartbeat payload.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
+public class BaseDefaultHeartbeatPropertyProvider implements HeartBeatDefaultPayloadProviderInterface {
```
littleaj: I'd suggest renaming this class to `DefaultHeartbeatPropertyProvider`. "Base" implies this is an abstract class and superclass to other providers built atop this one.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
```diff
@@ -0,0 +1,162 @@
...
+    defaultFields.add("sdk-version");
+    defaultFields.add("osType");
+    defaultFields.add("processSessionId");
+  }
+
+  /**
+   * Gets the JDK version being used by the application
+   * @return String representing JDK Version
+   */
+  private String getJdkVersion() {
```
littleaj: It's not the _JdkVersion_ it's the _JreVersion_

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
littleaj: Same comment as below: _jreVersion_

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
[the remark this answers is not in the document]
littleaj: You could consider also checking ` Runtime.class.getPackage().getImplementationVersion()`

dhaval24:
@littleaj I have tried to address maximum of your comments in my latest commit. Take a look hopefully those which are not addressed are not very critical at the moment.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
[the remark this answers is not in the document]
littleaj: I don't think this change will work. `isHeartbeatEnabled()` uses the `heartbeatProviderInterface`, but the provider doesn't get the config values until it's initialized. Right?

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -133,10 +135,10 @@ public boolean addHeartBeatProperty(String propertyName, String propertyValue,
              payload.setPayloadValue(propertyValue);
              heartbeatProperties.put(propertyName, payload);
              isAdded = true;
-             InternalLogger.INSTANCE.trace("added heartbeat property");
+             InternalLogger.INSTANCE.trace("added heartbeat property %s - %s", propertyName, propertyValue);
```
littleaj: Hyphen is an interesting choice. Usually you see `%s:%s` or `%s=%s` to indicate key=value.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/DefaultHeartbeatPropertyProvider.java
```diff
@@ -145,7 +152,7 @@ private String getSdkVersion() {
    * Gets the OS version on which application is running.
    * @return String representing OS version
    */
-  private String getOsType() {
+  private String getOsVersion() {
```
littleaj: Do we want the OS version? If so, you should explore the difference between `os.name` and `os.version` (i.e. `os.name` does not contain version information).

I also found a few articles stating that `os.name` doesn't have the correct value when on Windows 10. Perhaps there is a different way to find this information.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -88,6 +90,8 @@ public HeartBeatProvider() {
     this.heartbeatProperties = new ConcurrentHashMap<>();
     this.isEnabled = true;
     this.heartbeatsSent = new AtomicLong(0);
+    this.propertyUpdateService = Executors.newCachedThreadPool(ThreadPoolUtils.createDaemonThreadFactory(HeartBeatProvider.class));
```
littleaj: This makes these two thread pools indistinguishable. You may need another method in ThreadPool utils to specify a custom name.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+      //results to come out as some I/O bound properties may take time.
+      propertyUpdateService.submit(HeartbeatDefaultPayload.populateDefaultPayload(getExcludedHeartBeatProperties(),
+          getExcludedHeartBeatPropertyProviders(), this));
+
+      heartBeatSenderService.scheduleAtFixedRate(heartBeatPulse(), interval, interval, TimeUnit.SECONDS);
+    }
+  }
+
+  @Override
+  public boolean addHeartBeatProperty(String propertyName, String propertyValue,
```
littleaj: I must say, I have strong feelings against this add/set property pattern. I don't understand why we need to throw if someone is overwriting a property with `add` or setting a new property with `set`. Please explain why we are doing it this way.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+        if (!heartbeatProperties.containsKey(propertyName)) {
+             HeartBeatPropertyPayload payload = new HeartBeatPropertyPayload();
+             payload.setHealthy(isHealthy);
+             payload.setPayloadValue(propertyValue);
+             heartbeatProperties.put(propertyName, payload);
+             isAdded = true;
+             InternalLogger.INSTANCE.trace("added heartbeat property %s - %s", propertyName, propertyValue);
+        }
+        else {
+          throw new Exception("heartbeat property cannot be added twice. Please use setHeartBeatProperty instead to modify the value");
```
littleaj: Please use a more specific Exception subclass.

littleaj:
@dhaval24, also address the failing unit test.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @littleaj I believe it adds more readability and easier to distinguish operations. I know that it is not common in Java's map class however at the same time I also do feel that being some what opinionated here is not causing any blockage. Unless there is a mistake in documentation I don't think user will play around with this.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: This is nice to have but I don't think this should block the PR.

core/src/test/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProviderMock.java
```diff
@@ -0,0 +1,114 @@
...
+
+import com.microsoft.applicationinsights.TelemetryConfiguration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class HeartBeatProviderMock implements HeartBeatProviderInterface {
```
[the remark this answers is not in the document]
dhaval24: We can do this later on too. I would like to get this in and since this is a test class and also doesn't affect performance I don't believe it is a blocker.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -119,7 +120,9 @@ public final void initialize(TelemetryConfiguration configuration) {
     private void setMinimumConfiguration(ApplicationInsightsXmlConfiguration userConfiguration, TelemetryConfiguration configuration) {
```
littleaj: Will the heartbeat module always be added? It didn't look like it the way this method is called.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
```diff
@@ -0,0 +1,185 @@
...
+ * <p>
+ *   This module is the core module which is is configured by default with default settings
+ *   when ApplicationInsights SDK boots up. It is used to transmit diagnostic heartbeats to
+ *   Application Insights backend.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
+public class HeartBeatModule implements TelemetryModule {
```
littleaj: It appears this needs a no arg constructor as well.

core/src/test/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatTests.java
```diff
@@ -0,0 +1,330 @@
...
+    Assert.assertTrue(module.getExcludedHeartBeatPropertiesProvider().contains("webapps"));
+  }
+
+  @AfterClass
+  public static void tear() throws Exception{
+    tearDown();
+  }
+
+  private static void tearDown() throws Exception{
+    Method method = TelemetryConfiguration.class.getDeclaredMethod("setActiveAsNull");
```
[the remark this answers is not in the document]
dhaval24: @littleaj unfortunately this won't work. The tests code is in a different package. Package private provides access to the files in only the immediate pacakage where the java class resides.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatModule.java
dhaval24: We can have that ... but not needed. We can pass the null inside the parameterized constructor. There is a check there which handles it. I don't think we need to expose other constructor here if we can avoid.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: One knob is simpler than two knobs; and there's no reason to be this strict. That's all I'm saying.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+  @Override
+  public void initialize(TelemetryConfiguration configuration) {
+    if (isEnabled) {
+      if (this.telemetryClient == null) {
+        this.telemetryClient = new TelemetryClient(configuration);
+      }
+
+      //Submit task to set properties to dictionary using separate thread. we do not wait for the
+      //results to come out as some I/O bound properties may take time.
+      propertyUpdateService.submit(HeartbeatDefaultPayload.populateDefaultPayload(getExcludedHeartBeatProperties(),
```
littleaj: Which properties are I/O bound? I wasn't able to find any. We may be able to remove the thread pool.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * Telemetry client instance used to send heartbeat.
+   */
+  private TelemetryClient telemetryClient;
+
+  /**
+   * ThreadPool used for adding properties to concurrent dictionary
+   */
+  private ExecutorService propertyUpdateService;
```
littleaj: Since this is single use, I don't think it's needed as a field. Local variable in constructor should be fine.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+  /**
+   * Send the heartbeat item synchronously to application insights backend.
+   */
+  private void send() {
+
+    synchronized (lock) {
+      MetricTelemetry telemetry = (MetricTelemetry)gatherData();
+      telemetry.getContext().getOperation().setSyntheticSource(HEARTBEAT_SYNTHETIC_METRIC_NAME);
+      telemetryClient.trackMetric(telemetry);
+      heartbeatsSent.incrementAndGet();
```
littleaj: This can be in the log line below. It returns the value after incrementing. Also, since it's in a lock it doesn't need to be atomic

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: There are no properties which are IO bound today. But later on there will be when we will query ARM to get VM ID. I would rather keep it.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/DefaultHeartbeatPropertyProvider.java
dhaval24: We can amend it, if there is a better way though I am not aware of one. This is meta data which is not very very vital though.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @littleaj fair enough. I was under impression to change the api to have only set() method instead of add() and set() both. But from the comment now I understand that the only ask here is to not throw the exception but log the message accordingly which is acceptable. I can make this changes too.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
dhaval24: @littleaj look into the method `setTelemetryModules()`. It adds HB module explicitly and also look at`setMininmumConfiguration()`. This module would be present by default.

dhaval24:
@littleaj all of your concerns must have been addressed with my recent changes. Please take a look and would be great if we can get this merged if no blockers are left.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+        if (!heartbeatProperties.containsKey(propertyName)) {
+             HeartBeatPropertyPayload payload = new HeartBeatPropertyPayload();
+             payload.setHealthy(isHealthy);
+             payload.setPayloadValue(propertyValue);
+             heartbeatProperties.put(propertyName, payload);
+             isAdded = true;
+             InternalLogger.INSTANCE.trace("added heartbeat property %s - %s", propertyName, propertyValue);
+        }
+        else {
+          throw new IllegalStateException("heartbeat property cannot be added twice. Please use setHeartBeatProperty instead to modify the value");
```
littleaj: If you are throwing and catching in the same method based on a conditional, you can refactor it so that the statements in the catch block are moved to inside the else block (in place of throw). Another option would be to throw the exception out of this method.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
littleaj: @dhaval24 , Ok, I see. 

I did notice that you can reduce some redundancy if you remove the call to `addHeartbeatModule` from `setMinimumConfiguration`. Since `setTelemetryModules` is always called and `setMinimumConfiguration` is only called when the XML config is missing, you don't need the add call in `setMinimumConfiguration`.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: Makes sense.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+      InternalLogger.INSTANCE.trace("No of heartbeats sent, %s", heartbeatsSent.get());
+    }
+
+  }
+
+  /**
+   * Creates and returns the heartbeat telemetry.
+   * @return Metric Telemetry which represent heartbeat.
+   */
+  private Telemetry gatherData() {
```
littleaj: This should just return `MetricTelemetry` so you don't have to cast it.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+
+  /**
+   * Runnable which is responsible for calling the send method to transmit telemetry
+   * @return Runnable which has logic to send heartbeat.
+   */
+  private Runnable heartBeatPulse() {
+    return new Runnable() {
+      @Override
+      public void run() {
+        if (isEnabled) {
```
littleaj: I don't follow this conditional. Will the thread start even when it's disabled?

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
```diff
@@ -0,0 +1,304 @@
...
+          + "encountered and exception with stacktrace, %s", ExceptionUtils.getStackTrace(e));
+    }
+  }
+
+  /**
+   * Send the heartbeat item synchronously to application insights backend.
+   */
+  private void send() {
+
+    synchronized (lock) {
```
littleaj: What is the lock guarding? The ConcurrentMap keeps the CRUD to it thread-safe. I think you can just remove the lock and keep the atomic int

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
littleaj: See other comment regarding the lock. If you remove the lock, you can keep the atomic int.

core/src/test/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatTests.java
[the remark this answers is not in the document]
littleaj: You should write tests for a class in the same package as the class under test.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: Fair, Okay this should be addressed.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: Statements and returns should just be fine. I will make this changes.

dhaval24:
@littleaj I have done another round of edits for your comments. I think everything is now addressed and this should be good to go. Kindly please approve after checking.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
dhaval24: @littleaj I need to add this for SpringBoot scenarios. Here it is not called.

core/src/test/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatTests.java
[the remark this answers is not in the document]
littleaj: Oh, I see. This is fine, then

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
littleaj: Ahh, I see. I missed the return statements above.

---
In reply to: [181910995](https://github.com/Microsoft/ApplicationInsights-Java/pull/631#discussion_r181910995) [](ancestors = 181910995)

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @littleaj my intention here is to send heartBeats synchronously.  Telemetry client's track() method sends Telemetry asynchronously which is not desired here. I wanted to guard against an untoward situation like this. I have removed the Atomic version to primitive as you mentioned we are already locking and therefore avoid overhead.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/MiscUtils.java
dhaval24: Fixed

dhaval24:
@littleaj fixed couple of remaining unaddressed feedback also. Let me know if there are anything else.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @littleaj this is addressed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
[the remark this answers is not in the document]
dhaval24: @littleaj this is addressed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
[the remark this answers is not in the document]
dhaval24: Addressed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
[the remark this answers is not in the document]
dhaval24: @littleaj this is addressed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/WebAppsDefaultHeartbeatProvider.java
dhaval24: This is addressed.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: @littleaj this is addressed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: In a way you are correct. This can happen, considering the situation --> Thread A (starts sending heartbeat) --->current-state(blocked in wire) ---> ThreadPool gets instruction to schedule a new heartbeat and it would be in blocking state unless it gets a free thread to perform the task. This is practically unavoidable since there is no ScheduledCacheThreadPool. 

However, performing the math, considering the retry time of 15 minutes and the worst case scenario where heartbeat is scheduled for every 30 minutes if we set the maxPoolSize of 30 it would suffice for our scenario and ensure that every heartbeat gets a worker in any case. After that retries would be over and we would have a free worker thread to pick the new HeartBeat payload. I have pushed this changes, let me know your thoughts here.

dhaval24: One more thing I found that ScheduledThreadPool doesn't have the concept of corePool and maxPool size. So in this case we might need to specify 'X' threads by default which will be always available to complete the task.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
dhaval24: It is infact a container of default heartbeat providers and I think should be fine.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
```diff
@@ -0,0 +1,43 @@
...
+   * @return Name of the heartbeat provider
+   */
+  String getName();
+
+  /**
+   * Tells if the input string is a reserved property.
+   * @param keyword string to test
+   * @return true if input string is reserved keyword
+   */
+  boolean isKeyWord(String keyword);
```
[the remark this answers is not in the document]
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatDefaultPayloadProviderInterface.java
```diff
@@ -0,0 +1,43 @@
...
+ *   some helper methods to assist it.
+ *
+ *   The default concrete implementations are {@link com.microsoft.applicationinsights.internal.heartbeat.BaseDefaultHeartbeatPropertyProvider}
+ *   and {@link com.microsoft.applicationinsights.internal.heartbeat.WebAppsDefaultHeartbeatProvider}
+ * </p>
+ *
+ * @author Dhaval Doshi
+ * @since 03-30-2018
+ */
+public interface HeartBeatDefaultPayloadProviderInterface {
```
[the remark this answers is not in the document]
dhaval24: I think it should be `HeartBeat` as they are seperate words.  Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/BaseDefaultHeartbeatPropertyProvider.java
dhaval24: Fixed

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: This is nit: I think should be fine :)

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartBeatProvider.java
dhaval24: Fixed

d3r3kk:
Please note that there is a relevant change in the dotnet repo here today:
Known sdk versions (added heartbeat): Microsoft/ApplicationInsights-Home#213
Update heartbeat properties to include internal sdk version: Microsoft/ApplicationInsights-dotnet#782

This change will affect the 2.7.0-stable release of .Net in July, so you might want to track an issue separate from this PR.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,4 +1,3 @@
-#Fri Feb 09 18:53:31 PST 2018
```
littleaj: @dhaval24 The gradle related files don't need to be updated. (gradle-wrapper.properties, gradle-wrapper.jar, gradlew)

PR #634, opened 2018-04-04
added jetty9 test appServer

littleaj:
Adds jetty to smoketests


test/smoke/appServers/Jetty.9/jre.excludes.txt
```diff
@@ -0,0 +1,2 @@
+openjdk:7-jre
+azul/zulu-openjdk:7
```
dhaval24: I believe you mentioned you were planning to remove zulu-jdk. Has that changed or are you addressing it in separate PR?

littleaj: I'm going to add that to a different one: #642

Issue #643, opened 2018-04-09
Spring Boot Starter for Java SDK: auto-configure cloud_roleName


grlima:
#632

dhaval24:
#646 should address this.

dhaval24:
This is addressed with Application Insights SpringBoot starter which auto configures cloud_RoleName

Issue #579, opened 2018-02-27
Remove Misleading Log Messages in WebRequestTrackingFilter


dhaval24:
The log message in the catch block inside `initialize(FilterConfig filterConfig)` method in `WebRequestTrackingFilter.java`  class is very misleading.  This is emitted even in situations when user doesn't use the agent and filter is successfully registered. 

PS: It is general behavior of the filter to test if agent is present and throw. The catch statements should be specific and use different logs for different exceptions.

dhaval24:
#646 will address this.

littleaj:
Which log message? I don't see one that is obviously misleading.

Regarding your PS, catch statements don't always need to be specific.

dhaval24:
This has been resolved.

PR #646, opened 2018-04-12
Azure Application Insights SpringBoot Starter

dhaval24:
This PR introduces a final version of Azure Application Insights SpringBoot starter. It extends on the work done in #518 by @gavlyukovskiy. 

The starter closely has all the required functionalities and tuning knobs that ApplicationInsights.xml file provides to the user using application.properties and spring dependency injection. 

The core additional functionality which are added on the Top of #518 are as follows:
1. Autoconfiguration of performance counter properties (collection interval etc.)
2. Autoconfiguration of JVM Counters
3. Autoconfiguration of JMX counters
4. Autoconfiguration of Heartbeat
5. Autoconfiguration of Cloud_RoleName (helps in logical representation of micro-service) #643 

It also removes all field injection and instead now uses constructor injection and setter injection as recommended by Spring. 

Thank you @gavlyukovskiy  for your very vital contribution to our project.

Note: This should not be merged until #631 is merged. Once #631 is merged -> merge master in this branch resolve API changes and then merge.

- [ ] CHANGELOG.md updated


dhaval24:
@gavlyukovskiy please feel free to review this, and suggest changes. Thank you very much for all your help.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+            if (heartBeat.getExcludedHeartBeatPropertiesList().size() > 0) {
+                heartBeatModule.setExcludedHeartBeatProperties(heartBeat.getExcludedHeartBeatPropertiesList());
+            }
+            if (heartBeat.getExcludedHeartBeatProviderList().size() > 0) {
+                heartBeatModule.setExcludedHeartBeatPropertiesProvider(heartBeat.getExcludedHeartBeatProviderList());
+            }
+            return heartBeatModule;
+        }
+        catch (Exception e) {
+            throw new IllegalStateException("could not configure Heartbeat, please disable 'azure.application-insights.default.modules.HearBeat.enabled'"
```
gavlyukovskiy: I think it's better to change wording `please disable` to `please set`.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+        catch (Exception e) {
+            throw new IllegalStateException("Could not initialize Jvm Performance Counters module "
+                + "please set the property 'azure.application-insights.default.modules.JvmPerformanceCountersModule.enabled=false' to "
+                + "avoid this error message", e);
+        }
+    }
+
+    @Bean
+    @ConditionalOnProperty(value = "azure.application-insights.default.modules.HearBeat.enabled", havingValue = "true", matchIfMissing = true)
```
gavlyukovskiy: Typo `HearBeat` to `HeartBeat`.
Also isn't it web module? If so it's better to move it under web auto-configuration.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -0,0 +1,339 @@
...
+            processAndLoadJmxCounters(jmx.getJmxCounters());
+        }
+        return PerformanceCounterContainer.INSTANCE;
+    }
+
+    /**
+     * This method is used to process and load list of JmxCounters provided in the configuration.
+     * @param jmxCounterList
+     */
+    private void processAndLoadJmxCounters(List<String> jmxCounterList) {
```
gavlyukovskiy: I think it would be better to move all this logic below into `ApplicationInsightsProperties`, to keep auto-configuration class simpler.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
dhaval24: @gavlyukovskiy yes I will fix this typo. No it is not web module. It is core module and is expected to work even in worker role scenarios. I think this is a right place for it.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
dhaval24: Yes this refactoring can be done.

dhaval24:
@littleaj @gavlyukovskiy I have included the heartbeat changes here and done the required refactoring suggested by @gavlyukovskiy to make the code a bit cleaner. It would be great if you can review the changes and raise any concerns if present.

dhaval24:
By the way, when you do code reviews here, you will see some of the diffs in classes which are not changed logically (Eg InProcessTelemetryChannel). However, I have created this enabling CheckStyle internally in the IDE so that later on when it's enabled there are minimal files to edit, and why not since this is a new feature :)

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+@ConditionalOnProperty(value = "azure.application-insights.web.enabled", havingValue = "true", matchIfMissing = true)
+@AutoConfigureAfter(ApplicationInsightsTelemetryAutoConfiguration.class)
+public class ApplicationInsightsWebMvcAutoConfiguration {
+
+    @Bean
+    public FilterRegistrationBean webRequestTrackingFilterRegistrationBean(WebRequestTrackingFilter webRequestTrackingFilter) {
+        FilterRegistrationBean registration = new FilterRegistrationBean();
+        registration.setFilter(webRequestTrackingFilter);
+        registration.addUrlPatterns("/*");
+        registration.setName("webRequestTrackingFilter");
```
dhaval24: This should not be hardcoded.

dhaval24: @littleaj and @grlima I looked into this method. This is method is usually used to set the name of the RegistrationBean, however it is not necessary to set the name and the default behavior here is that if this is not set explicitly the name of the bean would be taken into the account. 

Further, a surprising thing this method no longer seems to be present in the latest SpringBoot framework however no mention anywhere in the change logs too. 

https://github.com/spring-projects/spring-boot/commit/9cb5f3da89b26cc60725a7e0a0c300a53d38f91f#diff-4c7df2b424db8ecb8e18ac9485ddcc8d

Therefore just to be safe i would remove it. Anyways, it doesn't affect the behavior either.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -102,16 +99,12 @@ public FixedRateSamplingTelemetryProcessor() {
...
-            } else {
-                InternalLogger.INSTANCE.error("Item is either not allowed to sample or is empty");
-            }
+        value = value.trim();
+        TelemetryType telemetryType = TelemetryType.valueOfOrNull(value);
+        if (telemetryType != null) {
+            typeSet.add(allowedTypes.get(telemetryType));
         } else {
-            InternalLogger.INSTANCE.error("Empty types cannot be considered");
+            InternalLogger.INSTANCE.error("Telemetry type " + value + " is either not allowed to sample or is empty");
```
littleaj: use format string, not concat

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -41,14 +47,17 @@
 @BuiltInProcessor("FixedRateSamplingTelemetryProcessor")
 public final class FixedRateSamplingTelemetryProcessor implements TelemetryProcessor {
 
-    private final String dependencyTelemetryName = "Dependency";
-    private static final String eventTelemetryName = "Event";
-    private static final String exceptionTelemetryName = "Exception";
-    private static final String pageViewTelemetryName = "PageView";
-    private static final String requestTelemetryName = "Request";
-    private static final String traceTelemetryName = "Trace";
+    public static final double DEFAULT_SAMPLING_PERCENTAGE = 100.;
```
littleaj: trailing dot is syntactically correct? if so, there's nothing wrong here. I just didn't know that 😄

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -41,14 +47,17 @@
...
+    private static Map<TelemetryType, Class> allowedTypes = new HashMap<>();
 
-    private static Map<String, Class> allowedTypes;
+    static {
```
littleaj: I'm fine with moving this to the static constructor, but changing the map type is unnecessary. We don't need to introduce parsing into this. String.hashCode does a fine job here.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
```diff
@@ -0,0 +1,19 @@
+package com.microsoft.applicationinsights.internal.channel.samplingV2;
```
littleaj: See comment in FixedRateSamplingTelemetryProcessor. This should be removed.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/HeartBeatProvider/SpringBootHeartBeatProvider.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * <h1>SpringBoot Heartbeat Property Provider</h1>
+ * <p>
+ *   This class is a concrete implementation of {@link HeartBeatPayloadProviderInterface}
+ *   It enables setting SpringBoot Metadata to heartbeat payload.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ */
+public class SpringBootHeartBeatProvider implements HeartBeatPayloadProviderInterface {
```
littleaj: Is it possible to extend the implementation in core? or make it a delegate? You shouldn't need to reimplement the whole class.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -74,6 +74,20 @@ public static TelemetryConfiguration getActive() {
...
 
+    /**
+     * This method provides the new instance of TelmetryConfiguration without loading the configuration
+     * from configuration file. This will just give a plain bare bone instance. Typically used when
+     * performing configuration programatically by creating beans, using @Beans tags. This is a common
+     * scenario in SpringBoot.
+     * @return {@link com.microsoft.applicationinsights.TelemetryConfiguration}
+     */
+    public static TelemetryConfiguration getActiveWithoutInitializingConfig() {
+        synchronized (s_lock) {
```
littleaj: null check?

littleaj:
My main concern is that we've essentially reimplemented configuration loading for the spring boot case. 
If there's a way to refactor this so we don't need as much "populate the config object" code and use what is already in core, that would be beneficial for future maintenance and support of this code.

If not, there needs to be a test which creates a configuration object the normal way and an equivalent object in the spring boot way to make sure the spring boot code path makes the same object.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/initializer/SpringBootTelemetryInitializer.java
```diff
@@ -0,0 +1,31 @@
...
+ * <h1>TelemetryInitializer to set the CloudRoleName Instance</h1>
+ *
+ * <p>
+ *  This Telemetry Initializer is used to auto-configure cloud_RoleName field
+ *  to get a logical component on AppMap.
+ * </p>
+ *
+ * @author Dhaval Doshi
+ */
+public class SpringBootTelemetryInitializer implements TelemetryInitializer {
```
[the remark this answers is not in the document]
dhaval24: I would break it into two parts:
1. Would someone like to use AI with SpringBoot without starter? : Possibility, relatively very slim though.
2. Can we move this to core? - I don't think so because we cannot distinguish if the app is of type springboot in the core and then downstream methods would fail. Further, I think this is nice separation of logics. I think spring related stuff should be at one place.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
dhaval24: Good catch :) will add it.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
```diff
@@ -21,11 +21,9 @@
 
 package com.microsoft.applicationinsights.extensibility.initializer;
 
-import com.microsoft.applicationinsights.internal.logger.InternalLogger;
+import com.microsoft.applicationinsights.extensibility.ContextInitializer;
```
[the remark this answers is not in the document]
dhaval24: I think this change is only about removing unused import. Why it's unnecessary? They are overhead.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: Doesn't fail, but probably typo. I don't know either. I will make it 100.0 :)

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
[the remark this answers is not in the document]
littleaj: We can do style clean-up in a separate PR.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
littleaj: Probably for the best. It's an easy character to miss if you're skimming.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: By doing this IDE can perform auto-complete when users are typing this properties. That was the reason of this change.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: Sure.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/inprocess/InProcessTelemetryChannelTest.java
```diff
@@ -44,18 +45,24 @@ public void testStringIntegerMaxInstanceRetry() {
...
     @Test()
     public void testInvalidIntegerMaxInstanceRetry() {
         HashMap<String, String> map = new HashMap<String, String>();
         map.put(INSTANT_RETRY_NAME, "-1");
         new InProcessTelemetryChannel(map);
     }
+
+    @Test
+    public void testInProcessTelemetryChannelWithDefaultSpringBootParameters() {
+        new InProcessTelemetryChannel("https://dc.services.visualstudio.com/v2/track", "10",
```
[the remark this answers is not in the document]
dhaval24: @littleaj I think this can be refactored but then we might need to use reflection to test the values of InProcessTelemetryChannel. Do you prefer it?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -230,33 +232,20 @@ public WebRequestTrackingFilter() {
...
-                try {
-                    InternalLogger.INSTANCE.info("Agent was not found. Skipping the agent registration");
-                    return;
-                } catch (ThreadDeath td) {
-                    throw td;
-                } catch (Throwable t2) {
-                    // chomp
-                }
-            }
+        //If Agent Jar is not present in the class path skip the process
```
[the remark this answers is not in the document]
dhaval24: Well this is a bug in-itself. We write unnecessary log message and discover the agent in a very poor way. I think we discussed about this in past. Not sure if you remember it though and we had debugged through it too. 

#579

core/src/test/java/com/microsoft/applicationinsights/internal/channel/inprocess/InProcessTelemetryChannelTest.java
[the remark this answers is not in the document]
littleaj: No, don't worry about it. This is fine for now. I didn't realize it didn't have accessors for most fields. Perhaps we can add accessors in the future or make the class not final.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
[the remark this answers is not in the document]
littleaj: Detecting the agent is an important part of our current functionality. It could be better, but it's a separate issue. It should be a separate PR.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
littleaj: makes sense

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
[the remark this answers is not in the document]
littleaj: We should address it separately. I made some comments in #579. I'd like to have a separate discussion regarding this change.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -21,15 +21,11 @@
 
 package com.microsoft.applicationinsights.channel.concrete.inprocess;
```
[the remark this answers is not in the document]
dhaval24: Sure in my recent commit I have tried to do that.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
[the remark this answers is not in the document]
dhaval24: I think there is no harm touching some files here. I agree big changes can go separately, this is just a import removal :)

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
dhaval24: See my response above for it.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
littleaj: @dhaval24, let's discuss offline

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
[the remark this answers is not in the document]
dhaval24: #resolved

core/src/test/java/com/microsoft/applicationinsights/internal/channel/inprocess/InProcessTelemetryChannelTest.java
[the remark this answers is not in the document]
dhaval24: #resolved

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: #resolved

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: #resolved

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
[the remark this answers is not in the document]
littleaj: Sure. I'm fine with fixing some style issues here and there, but try to keep it contained to files which are related to the feature.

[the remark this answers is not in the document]
littleaj: That way, if we want a list of files having to do with this feature, we can simply look at files changed in this PR or the squashed commit. Having this here makes the git history look like this file has something to do with the spring boot starter.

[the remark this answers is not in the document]
dhaval24: @littleaj I agree with you on this. I have tried to reduce majority of styling things at other places too.
#resolved

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/HeartBeatProvider/SpringBootHeartBeatProvider.java
dhaval24: Well I think know. The design for this is such that,  each separate functionality has it's own HB provider. For example springboot has it's own, Azure VM will have it's own and WebApps have it's own. This provides the flexibility for user to turn off the whole provider in itself incase if they do not need it rather then handpicking each properties to disable. The contract each Provider has to follow is to implement the interface. I think this provides nice abstraction, separates the module in itself from core and lets us be able to freely modify it without affecting others.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java
dhaval24: Done

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
dhaval24: Done

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: Done

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: @littleaj I have given some deep thoughts on this, and I think I would agree with you on this. Here are the reasons: This enum is only used for the sampling, not required elsewhere. It adds an overhead of maintaining it and ensuring that all new types are also added to it. A person can easily miss this unless there is an automated way to do it. Further more this enum is public which then adds the overhead of not deprecating it before a major version. The degree of advantage it provides is not absolute necessity and overhead is pretty high.

I will make this changes :) Thanks for drawing the attention.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
dhaval24: Will be amended based on my response above.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
```diff
@@ -0,0 +1,155 @@
...
+     * Bean for HeartBeatModule. This also sets the properties for HeartBeatModule
+     * @param environment
+     * @return initialized instance with user specified properties of {@link HeartBeatModule}
+     */
+    @Bean
+    @ConditionalOnProperty(value = "azure.application-insights.default.modules.HeartBeat.enabled", havingValue = "true", matchIfMissing = true)
+    public HeartBeatModule heartBeatModule(Environment environment) {
+        try {
+            HeartBeatModule heartBeatModule = new HeartBeatModule();
+            HeartbeatDefaultPayload.addDefaultPayLoadProvider(new SpringBootHeartBeatProvider(environment));
```
gavlyukovskiy: If it's expected from users to be able to provide own `HeartBeatProvider` implementation it's worth to create bean:
```
@Bean
@ConditionalOnMissingBean
public HeartBeatPayloadProviderInterface heartBeatPayloadProviderInterface(Environment environment) {
    return new SpringBootHeartBeatProvider(environment);
}
```
and inject `HeartBeatPayloadProviderInterface` into `HeartBeatModule`. It's important to have interface declaration instead of implementation - `SpringBootHeartBeatProvider` because only this way users are able to override the bean.

dhaval24: In all cases we would like to have the properties of the current provider for better assessment. However it makes sense to do this more elegantly using bean. I would refactor this.

dhaval24: @gavlyukovskiy this is now addressed.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/TelemetryType.java
dhaval24: #resolved

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
dhaval24: #resolved

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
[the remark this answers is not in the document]
dhaval24: I reverted this changes. I will submit another PR for this changes. #resolved

dhaval24:
I wanted to point out the following: By reverting the changes of how agent is detected to the original implementation, the user will see this misleading exception message when the Agent Jar is not present.

AI: ERROR 04-05-2018 09:08:53.150, 31(localhost-startStop-1): Could not find Agent: 'java.lang.NoClassDefFoundError: com/microsoft/applicationinsights/agent/internal/coresync/AgentNotificationsHandler
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:338)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at com.microsoft.applicationinsights.internal.agent.AgentConnector.register(AgentConnector.java:79)
	at com.microsoft.applicationinsights.web.internal.WebRequestTrackingFilter.registerWebApp(WebRequestTrackingFilter.java:287)
	at com.microsoft.applicationinsights.web.internal.WebRequestTrackingFilter.initialize(WebRequestTrackingFilter.java:263)
	at com.microsoft.applicationinsights.web.internal.WebRequestTrackingFilter.init(WebRequestTrackingFilter.java:145)
	at org.apache.catalina.core.ApplicationFilterConfig.initFilter(ApplicationFilterConfig.java:285)
	at org.apache.catalina.core.ApplicationFilterConfig.<init>(ApplicationFilterConfig.java:112)
	at org.apache.catalina.core.StandardContext.filterStart(StandardContext.java:4590)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5233)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1419)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1409)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: com.microsoft.applicationinsights.agent.internal.coresync.AgentNotificationsHandler
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:338)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 27 more

IMO, this is very confusing and I can remember couple of tickets when users pointed out to this weired behavior. 

My suggestion : We can take two approaches - 
1. Either avoid dumping stackTrace for this scenario.  (Caution: We will also miss any other exceptions which might come up here).
2. Change the way how agent is detected. I will submit another PR as requested.

My opinion will hinge towards second suggestion, because effectively when a runtime exception is thrown compiler has to jump instructions which makes the program terribly slow. You can look at more details on this thread (https://stackoverflow.com/questions/299068/how-slow-are-java-exceptions) 
Further, users expect exceptions when a bad behavior happens in the program, not for what is expected to be usual behavior. This is also considered seriously bad practice. I would strongly urge to have a better way for detecting agent. I propose one in another PR, if there is a better alternative I am happy to evaluate it.

dhaval24:
Modularized the way to specify version number for starter and added the Starer version in HeartBeat payload for better diagnosis.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/HeartBeatProvider/SpringBootHeartBeatProvider.java
littleaj: Ok, i see. This is sending spring boot specific data; a decorator of heartbeat events. Makes sense.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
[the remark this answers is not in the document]
littleaj: @dhaval24, What has functionally changed here? I'm only seeing whitespace changes. If nothing has changed related to spring boot starter, please revert this.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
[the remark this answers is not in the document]
littleaj: This is not resolved.... please revert since it's has no relevance to this feature.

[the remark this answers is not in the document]
dhaval24: @littleaj I thought we had discussed that since this was only an import removal, it was fine. If this opinion has changed and you think that we should keep the additional import then I can bring it back. In my opinion, we rarely get chance to work on code cleaning and small changes (which do not affect the logic should be fine). While I do understand your concern for clean history this should not be something which is too big of change. @grlima what are your thoughts? If you both agree on reverting this small change I can do that.

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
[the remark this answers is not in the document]
littleaj: As much as I dislike the additional changes in this file....this should be fine. If you can find time to fix this, please do.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/SdkVersionContextInitializer.java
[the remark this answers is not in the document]
dhaval24: @littleaj I have reverted this file. Thanks fore your concerns on separating the logic.

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
dhaval24: #resolved

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -41,14 +46,24 @@
...
+
+    private static final String dependencyTelemetryName = "Dependency";
     private static final String eventTelemetryName = "Event";
     private static final String exceptionTelemetryName = "Exception";
     private static final String pageViewTelemetryName = "PageView";
     private static final String requestTelemetryName = "Request";
     private static final String traceTelemetryName = "Trace";
 
-    private static Map<String, Class> allowedTypes;
+    static {
```
dhaval24: @littleaj dual initialization is now fixed.

dhaval24:
@littleaj @grlima I have added tests discussed which shows similarity between traditional XML based configuration and Springboot configuration. Let me know if there is anything else which needs to be taken care here.

PR #514, opened 2018-01-05
Adding support for Filters, Layouts and ignoreException parameters for Log4j2 Application Insights appender

dhaval24:
Fix #348 

This PR enhances the Application Insights Log4j2 Appender by adding facility to support basic parameters of Log4j2 Appenders including Filters, Layouts and ignoreExceptions using the builder pattern to create the Plugin as suggested in Log4j2 Documentation.

It also fixes the faulty package name of the test case file which prevented the execution of Log4j2 tests and adds specific test cases for this particular feature.

E2E testing performed and verified logs are being reported to Application Insights Azure Portal.

For significant contributions please make sure you have completed the following items:
- [x] CHANGELOG.md updated

Issue #348, opened 2016-11-19
Enable filter in log4j ApplicationInsightsAppender


tobiasdiez:
The configuration 
````xml
<ApplicationInsightsAppender name="applicationInsightsAppender">
   <ThresholdFilter level="ERROR" onMatch="ACCEPT" onMismatch="DENY" />
</ApplicationInsightsAppender>
````
results in the error
> appender ApplicationInsightsAppender has no parameter that matches element ThresholdFilter.

----

I think, as a fix the appender needs to add a builder
````java
@PluginBuilderFactory
public static <B extends Builder<B>> B newBuilder() {
    return new Builder<B>().asBuilder();
}
́public static class Builder<B extends Builder<B>> extends AbstractAppender.Builder<B>
            implements org.apache.logging.log4j.core.util.Builder<ApplicationInsightsAppender> { ... }
````

dhaval24:
@tobiasdiez could you please shed some more details on this issue? Also, I see you suggesting a fix in the issue discussion. We are very open to contributions and would be really great if you could send a PR if it is possible ?

tobiasdiez:
Sadly, I don't have time right now to create a PR but, of course, I'm happy to provide more details.

The default log4j2 appender accept the options `filter, layout, ignoreExceptions` (see [AbstractAppender](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/appender/AbstractAppender.html#AbstractAppender(java.lang.String,%20org.apache.logging.log4j.core.Filter,%20org.apache.logging.log4j.core.Layout,%20boolean))). The `ApplicationInsightsAppender` should also support these basic settings. For this, the factory method
https://github.com/Microsoft/ApplicationInsights-Java/blob/36214d5c28eb975e008c5b7c60c4840fdd10ec20/logging/log4j2/appenderSrc/main/java/com/microsoft/applicationinsights/log4j/v2/ApplicationInsightsAppender.java#L81-L83
needs to be extended by the arguments `                                              @PluginAttribute("ignoreExceptions") boolean ignoreExceptions,
                                              @PluginElement("Layout") Layout layout,
                                              @PluginElement("Filters") Filter filter` (see [log4j documentation](http://logging.apache.org/log4j/2.x/manual/extending.html#Appenders) for an example. What is not shown in the example is that you simply pass these arguments to the parent constructor.).
[...]

PR #648, opened 2018-04-19
Adding trace log4j1.2 test case

Terryxxx:
@littleaj @grlima @dhaval24 I have finished trace log4j1.2 test case and update performance counter test case appservers. Please review it.


test/smoke/testApps/TraceLog4j1_2/src/main/resources/ApplicationInsights.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dhaval24: Why can't be a common XML used?  This introduces too much redundancy. I was not following every PR's but is it common in the framework to have each test case as an application. If answer is "no" there is no need for separate application. Infact, i would take one step more and reduce the number of test servlets also, club the relevant tests together and reduce the clutter.

test/smoke/testApps/TraceLog4j1_2/src/main/resources/log4j.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<!DOCTYPE log4j:configuration SYSTEM "log4j.dtd">
+
+<log4j:configuration xmlns:log4j="http://jakarta.apache.org/log4j/">
+    <appender name="test" class="com.microsoft.applicationinsights.log4j.v1_2.ApplicationInsightsAppender">
```
dhaval24: Lags critical feature tests. See overall comment for the PR.

test/smoke/testApps/TraceLog4j1_2/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+
+dependencies {
+    compile aiCoreJar
+    compile aiWebJar
+    compile aiLog4j1_2Jar
+    compile 'com.google.guava:guava:20.0'
+    
+    providedCompile 'javax.servlet:javax.servlet-api:3.0.1'
+
+    providedRuntime 'mysql:mysql-connector-java:5.1.44'
```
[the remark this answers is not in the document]
littleaj: That's an artifact from other build files. It's not needed.

test/smoke/testApps/TraceLog4j1_2/src/main/resources/ApplicationInsights.xml
littleaj: We do need a separate app because of the log4j library. We want a separate app for each logging framework.

It would be more work that it's worth to try an factor out common XML configs. It's by design to have a separate config for each application. Each application will vary either by config or its compile time dependencies.

dhaval24: Makes sense. I get that different logging frameworks will need separate applications as each app can have 1 logging framework. #ByDesign

littleaj:
@dhaval24, since the test for the #348 / #514 fix has a different config, right now that would require a different test application; or a way to refresh the logging config w/o restarting the application server (stopping the app server kills the docker container because docker watches the entrypoint process).

test/smoke/testApps/TraceLog4j1_2/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-app xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://java.sun.com/xml/ns/javaee" xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd" id="SimpleCalculator_1" version="3.0">
+  <display-name>SimpleTelemetry</display-name>
+  <filter>
```
littleaj: You can remove the `<filter>` and `<filter-mapping>` tags here (lines 4 -- 11). Since we have other tests for the filter, and the filter isn't needed for logging.

It doesn't hurt, though.

test/smoke/testApps/AutoPerfCounters/src/smokeTest/resources/appServers.txt
```diff
@@ -1,5 +1,6 @@
+jbosseap6
 jbosseap7
 tomcat7
 tomcat8
 tomcat85
-jbosseap6jetty9
\ No newline at end of file
+jetty9
```
littleaj: Good catch :)

dhaval24:
@Terryxxx , @littleaj explanation gives me enough justification :) to approve this.

test/smoke/testApps/TraceLog4j1_2/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/TraceLog4j1_2Test.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+
+public class TraceLog4j1_2Test extends AiSmokeTest {
+    @Test
+    @TargetUri("/traceLog4j1_2")
+    public void testTraceLog4j1_2() {
+        // this doesn't work with jbosseap6; under investigation
+        Assume.assumeFalse(currentImageName.contains("jbosseap6"));
+
+        assertEquals(1, mockedIngestion.getCountForType("RequestData"));
```
littleaj: Same thing here. Since the filter isn't under test in this class, we don't need to check for RequestData. (but again, it doesn't hurt 😄 )

test/smoke/testApps/TraceLog4j1_2/src/main/webapp/WEB-INF/web.xml
Terryxxx: Thanks for your comments, I will remove those useless code in next PR.

PR #650, opened 2018-04-23
Optimizations for Filter/TelemetryClient startup

littleaj:
This limits the use of reflection during initialization.
This also makes a few other expensive initialization asynchronous.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/annotation/AnnotationPackageScanner.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/annotation/AnnotationPackageScanner.java
@@ -33,13 +33,14 @@ import java.util.List;
  * Created by gupele on 3/15/2015.
  */
 public final class AnnotationPackageScanner {
+    private AnnotationPackageScanner(){}
     /**
      * The method will scan packages searching for classes that have the needed annotations.
      * @param annotationsToSearch The annotations we need.
      * @param packageToScan The packages to scan from, note that all sub packages will be scanned too.
      * @return A list of class names that are under the package we asked and that carry the needed annotations
      */
-    public List<String> scanForClassAnnotations(final Class<? extends Annotation>[] annotationsToSearch, String packageToScan) {
+    public static List<String> scanForClassAnnotations(final Class<? extends Annotation>[] annotationsToSearch, String packageToScan) {
         final ArrayList<String> performanceModuleNames = new ArrayList<String>();
         AnnotationDetector.TypeReporter reporter = new AnnotationDetector.TypeReporter() {
             @Override
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -22,9 +22,14 @@
 package com.microsoft.applicationinsights.internal.channel.common;
 
 import java.io.IOException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 
+import com.microsoft.applicationinsights.internal.shutdown.SDKShutdownActivity;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
@@ -39,22 +44,30 @@ import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
  */
 final class ApacheSender43 implements ApacheSender {
 
-    private final CloseableHttpClient httpClient;
-
+    private final AtomicReference<CloseableHttpClient> httpClientRef = new AtomicReference<>();
+    private volatile boolean isClientInitialized = false;
+    private final ExecutorService initializer = Executors.newSingleThreadExecutor();
     public ApacheSender43() {
-        PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
-        cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
-        cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+        initializer.execute(new Runnable() {
 
-        httpClient = HttpClients.custom()
-                .setConnectionManager(cm)
-                .useSystemProperties()
-                .build();
-         }
+            @Override
+            public void run() {
+                PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
+                cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
+                cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+
+                httpClientRef.compareAndSet(null, HttpClients.custom()
+                        .setConnectionManager(cm)
+                        .useSystemProperties()
+                        .build());
+            }
+        });
+        SDKShutdownActivity.INSTANCE.register(initializer);
+     }
 
     @Override
     public HttpResponse sendPostRequest(HttpPost post) throws IOException {
-        return httpClient.execute(post);
+        return getHttpClient().execute(post);
     }
 
     @Override
@@ -71,7 +84,7 @@ final class ApacheSender43 implements ApacheSender {
     @Override
     public void close() {
         try {
-            httpClient.close();
+            ((CloseableHttpClient)getHttpClient()).close();
         } catch (IOException e) {
             InternalLogger.INSTANCE.error("Failed to close http client, exception: %s", e.toString());
         }
@@ -79,7 +92,21 @@ final class ApacheSender43 implements ApacheSender {
 
     @Override
     public HttpClient getHttpClient() {
-        return httpClient;
+        if (!isClientInitialized) {
+            synchronized (this) {
+                if (!isClientInitialized) {
+                    while (httpClientRef.get() == null) {
+                        try {
+                            TimeUnit.MILLISECONDS.sleep(3);
+                        } catch (InterruptedException e){
+                        }
+                    }
+                    isClientInitialized = true;
+                }
+            }
+        }
+
+        return httpClientRef.get();
     }
 
     @Override
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmitterImpl.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmitterImpl.java
@@ -22,7 +22,8 @@
 package com.microsoft.applicationinsights.internal.channel.common;
 
 import java.util.Collection;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -120,7 +121,7 @@ public final class TransmitterImpl implements TelemetriesTransmitter {
 
     private final TelemetrySerializer serializer;
 
-    private final ScheduledThreadPoolExecutor threadPool;
+    private final ScheduledExecutorService threadPool;
 
     private final TransmissionsLoader transmissionsLoader;
 
@@ -138,8 +139,7 @@ public final class TransmitterImpl implements TelemetriesTransmitter {
 
         semaphore = new Semaphore(MAX_PENDING_SCHEDULE_REQUESTS);
 
-        threadPool = new ScheduledThreadPoolExecutor(2);
-        threadPool.setThreadFactory(ThreadPoolUtils.createDaemonThreadFactory(TransmitterImpl.class, instanceId));
+        threadPool = Executors.newScheduledThreadPool(2, ThreadPoolUtils.createDaemonThreadFactory(TransmitterImpl.class, instanceId));
 
         this.transmissionsLoader = transmissionsLoader;
         this.transmissionsLoader.load(false);
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
@@ -70,14 +70,14 @@ public final class FixedRateSamplingTelemetryProcessor implements TelemetryProce
         this.excludedTypes = new HashSet<Class>();
         try {
             this.allowedTypes = new HashMap<String, Class>() {{
-                put(dependencyTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry"));
-                put(eventTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.EventTelemetry"));
-                put(exceptionTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.ExceptionTelemetry"));
-                put(pageViewTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.PageViewTelemetry"));
-                put(requestTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RequestTelemetry"));
-                put(traceTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.TraceTelemetry"));
+                put(dependencyTelemetryName, com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry.class);
+                put(eventTelemetryName, com.microsoft.applicationinsights.telemetry.EventTelemetry.class);
+                put(exceptionTelemetryName, com.microsoft.applicationinsights.telemetry.ExceptionTelemetry.class);
+                put(pageViewTelemetryName, com.microsoft.applicationinsights.telemetry.PageViewTelemetry.class);
+                put(requestTelemetryName, com.microsoft.applicationinsights.telemetry.RequestTelemetry.class);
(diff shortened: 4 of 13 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/annotation/AnnotationPackageScanner.java
```diff
@@ -33,13 +33,14 @@
...
 public final class AnnotationPackageScanner {
+    private AnnotationPackageScanner(){}
     /**
      * The method will scan packages searching for classes that have the needed annotations.
      * @param annotationsToSearch The annotations we need.
      * @param packageToScan The packages to scan from, note that all sub packages will be scanned too.
      * @return A list of class names that are under the package we asked and that carry the needed annotations
      */
-    public List<String> scanForClassAnnotations(final Class<? extends Annotation>[] annotationsToSearch, String packageToScan) {
+    public static List<String> scanForClassAnnotations(final Class<? extends Annotation>[] annotationsToSearch, String packageToScan) {
```
littleaj: Given the way this class is being used, a static method makes more sense. This is one change that is independent of the performance improvements.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -39,22 +44,30 @@
...
+                PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
+                cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
+                cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
 
-        httpClient = HttpClients.custom()
-                .setConnectionManager(cm)
-                .useSystemProperties()
-                .build();
-         }
+                httpClientRef.compareAndSet(null, HttpClients.custom()
```
littleaj: This sets the httpClient atomically. If two initialization procedures are running at the same time (somehow), the compareAndSet call will let first writer win.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -71,15 +84,29 @@ public void dispose(HttpResponse response) {
...
+            ((CloseableHttpClient)getHttpClient()).close();
         } catch (IOException e) {
             InternalLogger.INSTANCE.error("Failed to close http client, exception: %s", e.toString());
         }
     }
 
     @Override
     public HttpClient getHttpClient() {
-        return httpClient;
+        if (!isClientInitialized) {
```
littleaj: First call will always find this condition true (isClientInitialized==false) and then it will "block" until the initializer thread has finished.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmitterImpl.java
```diff
@@ -138,8 +139,7 @@ public TransmitterImpl(TransmissionDispatcher transmissionDispatcher, TelemetryS
 
         semaphore = new Semaphore(MAX_PENDING_SCHEDULE_REQUESTS);
 
-        threadPool = new ScheduledThreadPoolExecutor(2);
-        threadPool.setThreadFactory(ThreadPoolUtils.createDaemonThreadFactory(TransmitterImpl.class, instanceId));
+        threadPool = Executors.newScheduledThreadPool(2, ThreadPoolUtils.createDaemonThreadFactory(TransmitterImpl.class, instanceId));
```
littleaj: Another change not related to perf. Just cleaning up.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -216,6 +243,15 @@ private void setTelemetryProcessors(ApplicationInsightsXmlConfiguration appConfi
         }
     }
 
+    private void addDefaultBuiltInProcessors(List<String> p) {
+        p.add(FixedRateSamplingTelemetryProcessor.class.getCanonicalName());
```
littleaj: maybe I should use the FQCNs here....

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -290,18 +320,20 @@ private void setInstrumentationKey(ApplicationInsightsXmlConfiguration userConfi
 
         ArrayList<TelemetryModule> modules = new ArrayList<TelemetryModule>();
 
-        final List<String> performanceModuleNames =
-                new AnnotationPackageScanner().scanForClassAnnotations(new Class[]{PerformanceModule.class}, performanceCountersSection);
+        List<String> performanceModuleNames = new ArrayList<>();
+        if (System.getProperty(PERFORMANCE_MODULES_SCANNING_ENABLED_PROPERTY, "false").equalsIgnoreCase("true")) {
```
littleaj: This does change behavior, but I think this is ok. I don't think there are customers using the annotations and I would like to deprecate them (we can do this now, if you like). If someone does want these turned on, they can use the property.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -463,39 +489,6 @@ private void loadProcessorComponents(
...
-     *
-     * Note that currently we 'swallow' all exceptions and simply return null if we fail
-     *
-     * @param className The class we create an instance of
-     * @param interfaceClass The class' parent interface we wish to work with
-     * @param <T> The class type to create
-     * @return The instance or null if failed
-     */
-    @SuppressWarnings("unchecked")
-    private <T> T createInstance(String className, Class<T> interfaceClass) {
```
littleaj: These methods are redundant...

web/src/main/java/com/microsoft/applicationinsights/internal/config/WebReflectionUtils.java
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.applicationinsights.internal.config;
+
+import static com.microsoft.applicationinsights.internal.config.ReflectionUtils.addClass;
+
+public class WebReflectionUtils {
+    public static void initialize() {
```
littleaj: This is called in `<clinit>` of the Filter. It should add all these classes so they are initialized properly.

Tests for this are pending.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: I see so you are using a SingleThreadExecutor to ensure that the initialization runs on one and only one thread and then using AtomicReference to ensure that this gets setted atomically . AtomicReference is nice when you want to share the state of the same immutable object and might want to change that shared object's value from one thread to another. This ensures atomicity and is slightly better in terms of performance. But here I am not sure you are really sharing between threads because you are using SingleThreadExecutor which I believe is a factory method and essentially runs on Single Thread. Even classical locking with Volatile, should also work here in my understanding. I can be wrong though, happy to understand your thought process.

https://stackoverflow.com/questions/3964211/when-to-use-atomicreference-in-java

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -71,15 +84,29 @@ public void dispose(HttpResponse response) {
     @Override
     public void close() {
         try {
-            httpClient.close();
+            ((CloseableHttpClient)getHttpClient()).close();
```
dhaval24: Why not change the interface ApacheSender itself and modify to get the CloseableHttpClient. I am more inclined towards deprecating the support for ApacheHttpClient42 also which I believe is very old and even our SDK doesn't use it anywhere. I would like to avoid this cast.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: This makes sense, just would like to know the scenario where this could happen. I was under impression that the channel first calls the constructor or initializer method to instantiate the instance of sender even before calling send().

core/src/main/java/com/microsoft/applicationinsights/internal/channel/samplingV2/FixedRateSamplingTelemetryProcessor.java
```diff
@@ -70,14 +70,14 @@ public FixedRateSamplingTelemetryProcessor() {
         this.excludedTypes = new HashSet<Class>();
         try {
             this.allowedTypes = new HashMap<String, Class>() {{
-                put(dependencyTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry"));
-                put(eventTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.EventTelemetry"));
-                put(exceptionTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.ExceptionTelemetry"));
-                put(pageViewTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.PageViewTelemetry"));
-                put(requestTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.RequestTelemetry"));
-                put(traceTelemetryName, Class.forName("com.microsoft.applicationinsights.telemetry.TraceTelemetry"));
+                put(dependencyTelemetryName, com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry.class);
```
dhaval24: This was changed in SpringBoot PR I think but good that you modified : ) here as well.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ReflectionUtils.java
```diff
@@ -56,9 +88,14 @@
                 return null;
             }
 
-            Class<?> clazz = Class.forName(className).asSubclass(interfaceClass);
+            Class<?> clazz = builtInMap.get(className);
+            if (clazz == null) {
+                InternalLogger.INSTANCE.trace(" :( class lookup failed: %s", className);
+                clazz = Class.forName(className).asSubclass(interfaceClass);
```
dhaval24: Please add a comment to mention this as a fallback when class is not loaded. It would be difficult to understand later on.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -193,8 +216,12 @@ private void setTelemetryProcessors(ApplicationInsightsXmlConfiguration appConfi
         if (configurationProcessors != null) {
             ArrayList<TelemetryProcessorXmlElement> b = configurationProcessors.getBuiltInTelemetryProcessors();
             if (!b.isEmpty()) {
-                final List<String> processorsBuiltInNames =
-                        new AnnotationPackageScanner().scanForClassAnnotations(new Class[]{BuiltInProcessor.class}, performanceCountersSection);
+                List<String> processorsBuiltInNames = new ArrayList<>();
+                if (System.getProperty(BUILTIN_PROCESSORS_SCANNING_ENABLED_PROPERTY, "false").equalsIgnoreCase("true")) {
```
dhaval24: Please add a comment about this property too. It is hard to remember later on.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
dhaval24: I am not sure if it causes significant over head. FQCNs are difficult to understand. My focus would be to keep this as it is and focus on simplicity :) . Please also add java doc comments even though method is private with short description for better understanding.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
dhaval24: I am fine with reducing as much as code as possible and focus on simple and elegant API.  I have no objections and since we would be releasing a feature version 2.1.0, it might be a good time.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -66,6 +78,17 @@
 
     private AppInsightsConfigurationBuilder builder = new JaxbAppInsightsConfigurationBuilder();
 
+    private static final Set<String> defaultPerformaceModuleClassNames = new HashSet<>();
+
+    static {
```
dhaval24: Are you sure that this works. I think the static block executes before even the constructor of the program is called and it is when the constructor is called the memory is allocated to the variables. I might be wrong, please correct me if so.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -65,6 +68,15 @@ public synchronized void register(Closeable closeable) {
             closeables.add(closeable);
         }
 
+        public synchronized void register(final ExecutorService service) {
```
dhaval24: Why registration needs to be synchronized? I think at the end what what matters is that all the stoppables are present in the collection. I don't think there is dependency on specific operation or the order of registration. It might be wise to instead make the list atomic in this case if it is after all needed.

web/src/main/java/com/microsoft/applicationinsights/internal/config/WebReflectionUtils.java
dhaval24: I am suspecting this scenario can break springboot when we add filter and it's initialize method is called. Here is the reason : SpringBoot adds the beans for all these modules using DI. That happens when the framework initializes and it also adds the Filter as a a bean. Now when this init method is called it will try to add all this things back into the class path when they are already present which can cause potential conflicts. We need to test this with Springboot app before merging.

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
littleaj: The static initializer (aka <clinit>) is executed whenever the class name is dereferenced. It runs exactly once.

What's the concern here?

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
littleaj: sure

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
littleaj: I'm asking myself this because these are "depedencies" of this class and clutter the import statements, IMO.  It doesn't really matter, though.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
littleaj: Mainly, I was following the pattern of the other `register` methods. Should be fine

core/src/main/java/com/microsoft/applicationinsights/internal/config/ReflectionUtils.java
```diff
@@ -56,9 +88,14 @@
                 return null;
             }
 
-            Class<?> clazz = Class.forName(className).asSubclass(interfaceClass);
+            Class<?> clazz = builtInMap.get(className);
+            if (clazz == null) {
+                InternalLogger.INSTANCE.trace(" :( class lookup failed: %s", className);
```
littleaj: oops. need to remove this

web/src/main/java/com/microsoft/applicationinsights/internal/config/WebReflectionUtils.java
littleaj: I didn't follow what you meant here. This will be called before the initialize method in WebRequestTrackingFilter. This doesn't add anything to the classpath. If this is "already on the classpath" it's not causing conflicts, it's referencing what's already there.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
dhaval24: Sounds good. #Resolved for consistency.

core/src/main/java/com/microsoft/applicationinsights/internal/config/ReflectionUtils.java
littleaj: I think the conditional itself is pretty clear. What would you have the comment say?

web/src/main/java/com/microsoft/applicationinsights/internal/config/WebReflectionUtils.java
dhaval24: sounds good. We should definitely test it though before shipping with a SpringBootApp to make sure that it works smoothly.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: The AtomicReference is to guarantee the reads/writes to this reference are guarded without the need for another lock.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: Resolved

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: Yup thanks for clarification. Makes sense. Should we use as a global practice throughout SDK?

PR #651, opened 2018-04-24
removed usage of -1 as a default value for some measurements

### LINKED ISSUE #647: Negative CPU values on Server Details
Moved from https://github.com/Microsoft/ApplicationInsights-Home/issues/212

Reported by @rstirling 

>My application is presenting negative CPU values at Server>Server Details panel. This behavior is normal?
>image
>The application is deployed on development machine. The default language is English. It's a Java application running on Tomcat Server using version 2.0.2 AI agent/libs.

littleaj:
This just causes confusion when seen in the cloud. Now, if we can't measure something, we won't report it as some garbage value.

Fix #647 .


dhaval24:
@littleaj wondering if changelog was updated for this :) if not we should do that before release.

PR #654, opened 2018-04-26
remove useless code and add trace log4j1.2 with exception test case

Terryxxx:
@littleaj  @dhaval24  @grlima


littleaj:
thanks @Terryxxx !

PR #655, opened 2018-04-26
IkeyResolver: support system properties and limit retries.

grlima:
Enable system properties for the HTTP client used in the ikey resolver. Also, limit retries in the resolver to save on performance in case of constant failures.


web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -133,6 +152,15 @@ public void setEndpointAddress(String endpoint) throws MalformedURLException {
         return this.httpClient.execute(request, null);
     }
 
+    private synchronized void incrementFailureCount(String instrumentationKey) {
```
dhaval24: We could possibly remove synchornization from this method if we make the failureCounters HashMap to be a thread safe concurrent HashMap because we only need atomicity in adding the values to the map.

PR #664, opened 2018-05-09
Adding HeartBeat test case

Terryxxx:
@dhaval24 @littleaj @grlima 

Hi guys, I have completed heartbeat test case. Please review it.

### DIFF
--- a/test/smoke/testApps/HeartBeat/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HeartBeatTest.java
+++ b/test/smoke/testApps/HeartBeat/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HeartBeatTest.java
@@ -1,0 +1,43 @@
+package com.microsoft.applicationinsights.smoketest;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nullable;
+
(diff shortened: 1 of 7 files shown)


test/smoke/testApps/HeartBeat/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HeartBeatTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.junit.Test;
+
+public class HeartBeatTest extends AiSmokeTest {
+
+    @Test
+    @TargetUri(value = "/index.jsp")
+    public void testHeartBeat() throws Exception {
+        List<Envelope> metrics = mockedIngestion.waitForItems(getMetricPredicate("HeartbeatState"), 2, 70, TimeUnit.SECONDS);
+        assertEquals(2, metrics.size());
```
littleaj: We should assert on some of the field values. This just tests that the config works.

@dhaval24, what should be checked?

dhaval24: We can assert on basic fields are added into the custom dimensions of HeartBeat payload should have values for : jreVersion, sdkVersion, osVersion, processSessionId.

PR #666, opened 2018-05-12
Smoketest support for dependency containers

littleaj:
TODO
- [x] update smoketest readme with instructions on dependency container usage.
- [x] Add stop all, prune stopped and prune networks to test prep (in build script?)
- [x] add logs for create/delete of docker network
- [x] fix issue with container names. They need something generated from the image. It needs to be unique across all tests.


test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -225,19 +234,38 @@ public void run() {
 	}
 
 	@ClassRule
-	public static TestWatcher agentModeLoader = new TestWatcher() {
+	public static TestWatcher classOptionsLoader = new TestWatcher() {
 		@Override
 		protected void starting(Description description) {
+			System.out.println("Configuring test class...");
 			UseAgent ua = description.getAnnotation(UseAgent.class);
```
dhaval24: Please refrain from using the variable names like "ua". Understandable that it is test code, but more of it is a framework where people will contribute and it makes harder to read code and contribute.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -376,11 +405,62 @@ protected static void checkMockedIngestionHealth() throws Exception {
 		assertEquals(MockedAppInsightsIngestionServlet.PONG, postResponse);
 	}
 
-	protected static void startDockerContainer() throws Exception {
+	private static void createDockerNetwork() throws Exception {
+		try {
+			networkId = docker.createNetwork(networkName);
+		} catch (Exception e) {
```
dhaval24: This only throws IOException and InterruptedException. Please make Exception type specific.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -376,11 +405,62 @@ protected static void checkMockedIngestionHealth() throws Exception {
...
+	}
+
+	private static void cleanUpDockerNetwork() throws Exception {
+		if (networkId == null) {
+			System.out.println("No network id....nothing to clean up");
+			return;
+		}
+		try {
+			docker.deleteNetwork(networkName);
+		} catch (Exception e) {
```
dhaval24: Same exception type can be made specific.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/docker/AiDockerClient.java
```diff
@@ -184,4 +188,21 @@ public boolean isContainerRunning(String id) throws IOException, InterruptedExce
...
+
+	public String createNetwork(String name) throws IOException, InterruptedException {
+	    Process p = buildProcess(dockerExePath, "network", "create", "--driver", "bridge", name).start();
+	    waitAndCheckCodeForProcess(p, 10, TimeUnit.SECONDS, "creating network");
+	    return getFirstLineOfProcessOutput(p);
+    }
+
+	private static String getFirstLineOfProcessOutput(Process p) throws IOException {
+		List<String> lines = CharStreams.readLines(new InputStreamReader(p.getInputStream()));
+		return lines.get(0);
```
dhaval24: This is not needed. You are essentially reading the whole stream to just get the first line. It can be expensive operation. Instead just use, readFirstLine() method of CharStreams.

https://google.github.io/guava/releases/15.0/api/docs/com/google/common/io/CharStreams.html#readFirstLine(com.google.common.io.InputSupplier)

littleaj: That particular method was removed in guava 18.0. smoketests use version 20.0

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
littleaj: It's common practice to throw Exception in test code. Everything written thus far follows that convention. I'd only catch something specific when a specific action based on exception type is needed.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -121,12 +125,16 @@ protected static void stopContainer(ContainerInfo info) throws Exception {
 	}
 
 	protected static short currentPortNumber = BASE_PORT_NUMBER;
-	
+
+	private static List<DependencyContainer> dependencyImages = new ArrayList<>();
 	protected static ContainerInfo currentContainerInfo = null;
+	protected static Stack<ContainerInfo> allContainers = new Stack<>();
```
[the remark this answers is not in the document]
littleaj: makes sense

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
dhaval24: ok. This was a nitpick :) My only point was to ensure consistency. If this is a practice which is followed through out the smoketest framework, then I would suggest that you outline it in contributing.md so that there is a better chance of consistency when others contribute to it.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
dhaval24: Same as above. It might be common within our team / or group, but when people want to contribute over on GitHub they might come from different work place and different culture. Clear instructions would help them and save our time too.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/docker/AiDockerClient.java
dhaval24: If you have reservations on using Guava for this purpose, I would not object you because it's test code and has less direct impact on customers. However, if you wish to you can use the Java's native method from BufferedReader to read the first line.

https://docs.oracle.com/javase/7/docs/api/java/io/BufferedReader.html#readLine()

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -225,19 +235,38 @@ public void run() {
...
+			System.out.println("Configuring test class...");
 			UseAgent ua = description.getAnnotation(UseAgent.class);
 			if (ua != null) {
 				agentMode = ua.value();
 				System.out.println("AGENT MODE: "+agentMode);
 			}
+			WithDependencyContainers wdc = description.getAnnotation(WithDependencyContainers.class);
+			if (wdc != null) {
+				for (DependencyContainer container : wdc.value()) {
+					if (container.value() == null || container.value().trim().isEmpty()) {
```
dhaval24: I would prefer removing the long if check with isNotBlank() method from apache-commons-lang. I think we already use it in our project.

https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#isNotBlank(java.lang.String)

littleaj: thanks for the tip.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/docker/AiDockerClient.java
littleaj: In the cases where this is used, there is always one line of output. The overhead from this function will be minimal, but I'll keep the methods you mentioned in mind. Thanks for the tips

dhaval24: np :)

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -225,19 +236,38 @@ public void run() {
...
 		protected void starting(Description description) {
+			System.out.println("Configuring test class...");
 			UseAgent ua = description.getAnnotation(UseAgent.class);
 			if (ua != null) {
 				agentMode = ua.value();
 				System.out.println("AGENT MODE: "+agentMode);
 			}
+			WithDependencyContainers wdc = description.getAnnotation(WithDependencyContainers.class);
+			if (wdc != null) {
+				for (DependencyContainer container : wdc.value()) {
```
dhaval24: @littleaj Can we do this in parallel? How about using a thread pool to add it in a ConcurrentList? If we have many dependencies in future this should help speeding up according to me.

dhaval24: @littleaj Can we do this in parallel? How about using a thread pool to add it in a ConcurrentList? If we have many dependencies in future this should help speeding up according to me. Just a thought though!

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -376,11 +407,76 @@ protected static void checkMockedIngestionHealth() throws Exception {
...
+			System.out.printf("Starting container: %s%n", imageName);
+			final String containerId = docker.startContainer(imageName, dc.portMapping(), networkId);
+			assertFalse("'containerId' was null/empty attempting to start container: "+imageName, Strings.isNullOrEmpty(containerId));
+			System.out.printf("Dependency container started: %s (%s)%n", imageName, containerId);
+
+			String containerName = docker.getRunningContainerName(containerId);
+			if (containerName == null) {
+				String message = String.format("Could not get container name for id=%s. ", containerId);
+				if (docker.isContainerRunning(containerId)) {
+					message += "It appears to be running.";
```
dhaval24: Do you explicitly depend on ContainerName? If yes, then this message seems to be bit misleading. Shouldn't the container stop in this case or forced stop?

Ideally I believe either name or id of the container is sufficient correct?

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/WithDependencyContainers.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface WithDependencyContainers {
+    DependencyContainer[] value();
```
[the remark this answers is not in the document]
littleaj: This is equivalent to varargs for annotations. This allows the user to not have to create the array object which is not an important part of using this annotation.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -421,13 +540,49 @@ public void resetMockedIngestion() throws Exception {
 
 	@AfterWithParams
 	public static void tearDownContainer(final String appServer, final String os, final String jreVersion) throws Exception {
```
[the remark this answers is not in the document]
littleaj: It could, but I didn't want to over complicate this during the first phase. If performance needs to be improved, this is something we can investigate. At this point, I don't think the smoke test perf is too bad.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
littleaj: This is just checking for inconsistencies to assist in debugging. Both should be defined

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
littleaj: This part is just reading the configuration, so doing it in parallel wouldn't really help here.

Later, when the config is used to start the containers, I want ordering preserved in case one container depends on another being started. I think there are tests which could be done in parallel, but I don't want to make this change for all tests. We could make it configurable in the future.

dhaval24:
@littleaj thanks for your answers. If you confident go ahead and merge.

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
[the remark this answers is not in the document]
littleaj: There is some order that needs to be preserved:
1. Start dependencies (assume no inter-dependencies; otherwise it would need to be serial)
2. Start test app
3. Do Test
4. Stop test app
5. Stop dependencies

As long as there are no inter-dependencies, the order of stopping dependencies is arbitrary; but I use a stack so everything is stopped in the reverse order of it being started.

PR #667, opened 2018-05-14
adding field validations

dhaval24:
This PR adds the additional additional fields to validate in HeartBeat payload.


dhaval24:
Merging it as it is straight forward field validation.

PR #675, opened 2018-05-31
scan logs for exceptions as last test on container

littleaj:
This creates a test which runs last (thanks to the test sorter which runs tests in alphabetical order). Since it's in the smoketest superclass, it runs on all current tests. This is before a container is stopped and the next one is selected.

There is only one pattern right now which checks for exceptions and one suppression "rule".

Going forward, we'll need to keep an eye on these results and continue to populate the search rules and suppression rules as needed.


littleaj:
Too many false positives with this test. Cutting for now.

PR #683, opened 2018-06-08
Fix guava thirdparty shadow

### LINKED ISSUE #661: Relocate Guava package com.google.thridparty
I noticed that the core library exports the package `com.google.thirdparty` of Guava.
Is there any reason why this package is not relocated, like all other packages of dependencies?

When this package is relocated, the SDK also seems to be compatible with Java 9.

littleaj:
Fix #661.

- [x] CHANGELOG.md updated

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -59,14 +59,16 @@ dependencies {
     testCompile group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
 }
 
+ext.dependenciesPrefix = 'com.microsoft.applicationinsights.core.dependencies'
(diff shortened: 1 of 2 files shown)


core/build.gradle
```diff
@@ -59,14 +59,16 @@ dependencies {
     testCompile group: 'com.google.code.gson', name: 'gson', version: '2.8.2'
 }
 
+ext.dependenciesPrefix = 'com.microsoft.applicationinsights.core.dependencies'
```
dhaval24: This is nice!

PR #688, opened 2018-06-14
added timezone to dateformat for logs

littleaj:
to make debugging customer issues easier when point-in-time matters; i.e. to correlate customer log events with events in the application insights services.

TODO:
- [ ] check smoketest logs for timezone and any errors finding timezone by id.


SergeyKanzhelev:
Would UTC work better?

dhaval24:
@SergeyKanzhelev I and @littleaj had this discussion. Java's datetime by default take the local time zone, but our analytics report it in UTC. It might make sense to have UTC to do better correlation of incident on backend to lost telemetry, which was the purpose for this. Could you please tell what other SDKs do here so we are consistent as well.

littleaj:
UTC makes sense actually. Updating...

PR #694, opened 2018-06-28
Fundamentals : Checkstylework

dhaval24:
This PR is to enforce checkstyle rules on the entire project. The style guide is under /config/checkstyle/checkstyle.xml

I have currently used Google Java Style for this and integration in the build is done using the Gradle Checkstyle plugin.

This PR is an initial commit, and we can adopt to and relax some of the style rules here.

@littleaj @grlima please take a look and feel free to add commits on top to update the style as per our usecase here.


dhaval24:
@grlima @littleaj did any one of you had a chance to take a look at this. I would suggest we close on this soon :)

dhaval24:
@littleaj yep /config/checkstyle/checkstyle.xml is the file for the checkstyle config. And it is added to the build script in common-java.gradle file.

core/native.gradle
```diff
@@ -30,14 +30,8 @@ buildscript {
...
 
+
 import org.apache.commons.io.FilenameUtils
-import org.gradle.nativeplatform.BuildType
-import org.gradle.nativeplatform.NativeBinarySpec
-import org.gradle.nativeplatform.NativeLibrarySpec
-import org.gradle.nativeplatform.SharedLibraryBinarySpec
-import org.gradle.nativeplatform.StaticLibraryBinarySpec
-import org.gradle.nativeplatform.platform.NativePlatform
-import org.gradle.nativeplatform.toolchain.VisualCpp
```
littleaj: These are used below. Removing this breaks the script.

dhaval24: ah okay didn't carefully looked at it after checkstyle ran

dhaval24: Fixed

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,247 @@
...
+      <message key="ws.followed"
+        value="GenericWhitespace ''{0}'' is followed by whitespace."/>
+      <message key="ws.preceded"
+        value="GenericWhitespace ''{0}'' is preceded with whitespace."/>
+      <message key="ws.illegalFollow"
+        value="GenericWhitespace ''{0}'' should followed by whitespace."/>
+      <message key="ws.notPreceded"
+        value="GenericWhitespace ''{0}'' is not preceded with whitespace."/>
+    </module>
+    <module name="Indentation">
```
littleaj: IMO, I think all these values should be doubled.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/ClassVisitorFactory.java
```diff
@@ -24,9 +24,7 @@
 import org.objectweb.asm.ClassVisitor;
 import org.objectweb.asm.ClassWriter;
 
-/**
- * Created by gupele on 7/27/2015.
- */
+/** Created by gupele on 7/27/2015. */
```
littleaj: I don't like this style. Javadoc should always be 3 lines like before.

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,247 @@
...
+        value="\\u00(09|0(a|A)|0(c|C)|0(d|D)|22|27|5(C|c))|\\(0(10|11|12|14|15|42|47)|134)"/>
+      <property name="message"
+        value="Consider using special escape sequence instead of octal value or Unicode escaped value."/>
+    </module>
+    <module name="AvoidEscapedUnicodeCharacters">
+      <property name="allowEscapesForControlCharacters" value="true"/>
+      <property name="allowByTailComment" value="true"/>
+      <property name="allowNonPrintableEscapes" value="true"/>
+    </module>
+    <module name="LineLength">
```
littleaj: I'd prefer to remove this module. A lot of the changes done are because of this rule and most of them make the code look worse, IMO.

config/checkstyle/checkstyle.xml
littleaj: sounds good

dhaval24:
@littleaj thanks let's use one triage meeting to finalize the rules and then it would be easy to apply! This was just to see how it works with build

config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,247 @@
...
+      <property name="minLineCount" value="2"/>
+      <property name="allowedAnnotations" value="Override, Test"/>
+      <property name="allowThrowsTagsForSubclasses" value="true"/>
+    </module>
+    <module name="MethodName">
+      <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9_]*$"/>
+      <message key="name.invalidPattern"
+        value="Method name ''{0}'' must match pattern ''{1}''."/>
+    </module>
+    <module name="SingleLineJavadoc">
```
littleaj: (Mentioned in another file where this was applied) I'd like to remove this.

littleaj:
@dhaval24 Sounds good.

dhaval24:
@littleaj Abandoning the PR. I will create a new one with only `checkstyle.xml` for us to review the rules.

PR #696, opened 2018-07-05
Adding request jsp test case

Terryxxx:
@littleaj @dhaval24 @grlima 

Hi guys, I have completed request jsp test case. Please review it. Thanks.


Terryxxx:
@dhaval24  This is the last automation test case in "Test Cases for Automation.doc". I confirmed with Arthur. This case want to ensured that we can see the request even though it's jsp.

PR #698, opened 2018-07-11
fix channel capacity check to be actually mb, fix associated tests

### LINKED ISSUE #697: TransmissionFileSystemOutput not working
Hello, 

I'm trying to stress Appinsights tracking mechanism, and I figured out that there's a bug inside the class TransmissionFileSystemOutput.
I tried to track 300,000 telemetry (traces, requests, dependencies), but I ran into error messages : "Persistent storage max capacity has been reached", even if the "MaxTransmissionStorageFilesCapacityInMB" was set to 1000MB.
I looked into the transmission folder inside the system temporary folder and saw that the error was raised when I had only a hundred files of 8KB each.
So, looking at the Java code (TransmissionFileSystemOutput.java:137) : 
```java
if (size.get() >= capacityInKB) {
    InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", size.get());
    return false;
}
```
If I'm correct, the variable "size" is updated with temporary file size in **bytes**, and the maximum capacity is in **MB**.
I rebuilt with the following condition `(size.get() >= capacityInKB * 1024)` and it worked perfectly.

Could you also please tell me where to find detailed information about the parameters of the "Channel" element of the ApplicationInsights.xml file ?
Does this configuration work for Java projects or are there still some issues like this one  ?

Regards,
Sebastien

dhaval24:
Fix #697 .

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
@@ -137,7 +137,10 @@ public final class TransmissionFileSystemOutput implements TransmissionOutput {
 
         long currentSizeInBytes = size.get();
         if (currentSizeInBytes >= capacityInBytes) {
-        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", (currentSizeInBytes / 1024));
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; "
+                + "currently at %.3f KB. Telemetry will be lost, "
+                + "please consider increasing the value of MaxTransmissionStorageFilesCapacityInMB property in the configuration file.",
+                (currentSizeInBytes / 1024.0));
             return false;
         }
 
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
@@ -134,8 +134,10 @@ public final class TransmissionFileSystemOutput implements TransmissionOutput {
 
     @Override
     public boolean send(Transmission transmission) {
-        if (size.get() >= capacityInKB) {
-        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", size.get());
+
+        long currentSizeInKb = size.get();
+        if (currentSizeInKb >= capacityInKB * 1024) {
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", currentSizeInKb);
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


dhaval24:
Seems like the test `testFetchOldestFiles()` : this works fine in all the runs on the local environment. However, when running on build machine it fails for some reason. @littleaj do you have any clue on this? Further, do you think manually running the tests on the build machine and validating could help eliminate this?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -134,8 +134,10 @@ public TransmissionFileSystemOutput(String folderPath) {
 
     @Override
     public boolean send(Transmission transmission) {
-        if (size.get() >= capacityInKB) {
-        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", size.get());
+
+        long currentSizeInKb = size.get();
+        if (currentSizeInKb >= capacityInKB * 1024) {
```
littleaj: Let's define a constant for `capacityInKB`.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
```diff
@@ -21,23 +21,21 @@
...
+import java.util.Collection;
+import org.apache.commons.io.FileUtils;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
 public final class TransmissionFileSystemOutputTest {
     private final static String TRANSMISSION_FILE_EXTENSION = "trn";
-    private final static int SIZE_OF_TRANSMISSION_CONTENT = 100;
+    private final static int SIZE_OF_TRANSMISSION_CONTENT = 349525;
```
littleaj: Please show where this number came from, i.e. `100 * 1024 *1024; // 100 KB`. I'm not really sure what this number is.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -134,8 +134,10 @@ public TransmissionFileSystemOutput(String folderPath) {
 
     @Override
     public boolean send(Transmission transmission) {
-        if (size.get() >= capacityInKB) {
-        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", size.get());
+
+        long currentSizeInBytes = size.get();
+        if (currentSizeInBytes >= capacityInBytes) {
+        	InternalLogger.INSTANCE.logAlways(InternalLogger.LoggingLevel.WARN, "Persistent storage max capacity has been reached; currently at %s KB. Telemetry will be lost, please set the MaxTransmissionStorageFilesCapacityInMB property in the configuration file.", (currentSizeInBytes / 1024));
```
littleaj: Sorry i didn't notice this earlier, change this to `%.3f`; and change `1024` to `1024.0` so the quotient is a double.

Also, nitpicking, the user doesn't need to "`please set`" the property; the should "`consider changing`" it. i.e. hitting this message doesn't necessarily mean something is wrong.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
```diff
@@ -50,6 +53,9 @@ public TransmissionFileSystemOutputTest() {
         workingFolder = new File(LocalFileSystemUtils.getTempDir(), TEMP_TEST_FOLDER).getAbsolutePath();
```
littleaj: I think `workingFolder` is obsolete and can be removed. Right?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
dhaval24: Thanks done!

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
dhaval24: Removed!

PR #701, opened 2018-07-12
Update issue templates

littleaj:
Added templates for bug reports and feature requests

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,7 +14,7 @@ Describe what actually happened that was unexpected.
 Steps to reproduce the behavior:
 
 **System information**
-Please complete the following information:
+Please provide the following information:
  - SDK Version: 
  - OS type and version:
  - Application Server type and version (if applicable):
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,16 +4,19 @@ about: Create a report to help us improve
 
 ---
 
-**Expected behavior**
+### Expected behavior
 Describe what you expected to happen.
 
-**Actual behavior**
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,28 @@
...
+**System information**
+Please provide the following information:
+ - SDK Version: 
+ - OS type and version:
+ - Application Server type and version (if applicable):
+ - Using spring-boot?
+ - Additional relevant libraries (with version, if applicable):
+
+**Logs**
+[Turn on SDK logs](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-java-troubleshoot#debug-data-from-the-sdk) and attach/paste them to the issue. If using an application server, also attach any relevant server logs.
```
Dmitry-Matveev: We probably need to advise on removing any sensitive information from those logs before attaching to public repo on GitHub.

dhaval24: +1 yes ... also when sometimes user uploads a repro we should advise to remove all the PIDs

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+**Expected behavior**
+Describe what you expected to happen.
+
+**Actual behavior**
+Describe what actually happened that was unexpected.
+
+**To Reproduce**
+Steps to reproduce the behavior:
```
dhaval24: I would suggest if possible a trimmed down version of Application with exact configuration. (This could be possible next step) but it greatly reduces time to investigate.

.github/ISSUE_TEMPLATE/bug_report.md
littleaj: Ah, good catch. Will do

PR #715, opened 2018-08-10
added proto files and protobuf/grpc build plugin

littleaj:
Not much else to say here. The files aren't used yet so this won't have an impact on the current functionality. 

Do review the changes to the shadowing configuration with more rigor. I changed the paths to some of our dependencies. This shouldn't affect anything unless someone is using the dependency out of our package (which shouldn't be happening).


core/build.gradle
```diff
@@ -54,6 +86,9 @@ dependencies {
     compile ([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3'])
     compile ([group: 'com.google.guava', name: 'guava', version: '20.0'])
     compile ([group: 'com.google.code.gson', name: 'gson', version: '2.8.2'])
+    compile ([group: 'com.google.protobuf', name:'protobuf-java', version:'3.6.1'])
```
dhaval24: @littleaj I hope we have the required permission for all these three libraries from OSS before shipping. Further just a heads-up I might have to soon change the MetricTelemetry and it's associated data types (the Schema has been changed a bit there, and would need to run the bond again). At that time we might need Proto files to be regenerated. Do they generate on every build now? Also, what are the implications if we need to regenerate proto files. Do we need to release LF and SDK again at same time ?

littleaj: I'm working on approval for the library. The gradle tool is already approved.

core/build.gradle
```diff
@@ -65,10 +100,15 @@ shadowJar {
...
-    relocate 'com.google.common', "${dependenciesPrefix}.googlecommon"
+//    relocate 'com.google.common', "${dependenciesPrefix}.googlecommon"
     relocate 'javax.annotation', "${dependenciesPrefix}.javaxannotation"
-    relocate 'com.google.gson', "${dependenciesPrefix}.gson"
-    relocate 'com.google.thirdparty', "${dependenciesPrefix}.google.thirdparty"
+//    relocate 'com.google.gson', "${dependenciesPrefix}.gson"
+//    relocate 'com.google.thirdparty', "${dependenciesPrefix}.google.thirdparty"
+    relocate 'com.google', "${dependenciesPrefix}.google"
+    relocate 'io.grpc', "${dependenciesPrefix}.io.grpc"
+    relocate 'io.opencensus', "${dependenciesPrefix}.io.opencensus"
```
dhaval24: Indeed gRPC core has OpenCensus embedded in it. http://central.maven.org/maven2/io/grpc/grpc-core/1.14.0/grpc-core-1.14.0.pom (strange but seems like a force :))

core/src/main/proto/Metric.proto
```diff
@@ -0,0 +1,17 @@
...
+option java_package = "com.microsoft.localforwarder.library.inputs.contracts";
+option java_multiple_files = true;
+
+import "DataPoint.proto";
+
+message Metric {
+	int32 ver = 1;
+    
+	// only one metric in the list is currently supported by Application Insights storage. If multiple data points were sent only the first one will be used.
+	repeated DataPoint metrics = 2;
```
dhaval24: I still believe that this List<DataPoint> should be depreciated from the MetricData class which is creating this non-sense. Indeed we can bring in all the fields in the MetricData class inside the DataPoint class and just deprecate the whole class itself. It is of no use IMO and is just taking Memory footprint. Anyways, can be a conversation for seperate PR.

PR #717, opened 2018-08-23
Updated trackMetric to conform to bond spec

littleaj:
Original description
---
The stddev should be non-null in aggregate metrics. Maybe it should be -1 since that's an invalid value for Standard Deviation.

See [MetricTelemetry.updateKind](https://github.com/Microsoft/ApplicationInsights-Java/blob/0e2fe37143ad78b600d52189bc1994ccf70ac99c/core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java#L188). If stddev isn't set (or any of the statistic fields), then the metric won't be marked as kind=Aggregate. Perhaps this logic is flawed since it also excludes the possibility of an aggregation on a data set with size=1.

---
According to the bond spec, stddev can be null; as well as the other statistic fields. Updated TelemetryClient and MetricTelemetry to reflect the [bond spec](https://github.com/Microsoft/ApplicationInsights-Home/tree/master/EndpointSpecs/Schemas/Bond) (this PR is just about Metric).

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -270,17 +270,6 @@ public class TelemetryClient {
 
     /**
      * Sends a numeric metric to Application Insights. Appears in customMetrics in Analytics, and under Custom Metrics in Metric Explorer.
-     * @param name The name of the metric. Max length 150.
-     * @param value The value of the metric.
-     * @param properties Named string values you can use to search and classify trace messages.
-     * @throws IllegalArgumentException if name is null or empty.
-     */
-    public void trackMetric(String name, double value, Map<String, String> properties) {
-        trackMetric(name, value, null, null, null, null, properties);
-    }
-
-    /**
-     * Sends a numeric metric to Application Insights. Appears in customMetrics in Analytics, and under Custom Metrics in Metric Explorer.
      * @param telemetry The {@link com.microsoft.applicationinsights.telemetry.Telemetry} instance.
      */
     public void trackMetric(MetricTelemetry telemetry) {
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -39,6 +39,7 @@ import com.microsoft.applicationinsights.internal.util.MapUtil;
 import com.microsoft.applicationinsights.channel.TelemetryChannel;
 
 import com.google.common.base.Strings;
+import com.sun.istack.internal.Nullable;
 
 // Created by gupele
 /**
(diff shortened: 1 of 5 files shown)


dhaval24:
Setting it to 0.0 makes the most sense. Also in most of the case Stddev is not used. In case of the metric with 1 data point (gauge type metric) they won't have min, max, standdev anything. These statistics are usually for metrics which have multiple values over aggregated period.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -237,6 +237,7 @@ public void trackMetric(String name, double value, int sampleCount, double min,
         if (sampleCount > 1) {
             mt.setMin(min);
             mt.setMax(max);
+            mt.setStandardDeviation(0.0);
```
dhaval24: As I commented in the PR: This change should happen in the MetricTelemetry class. Basically there is no need of the following logic : 

```if ((metric.getCount() != null) && metric.getCount() == 1) {
            // Singular data point. This is not an aggregation.
            isAggregation = false;
        }
```
as aggregation is always true irrespective of count and this is how backend treats it. If you would like you can confirm with Greg, but I am pretty sure about this based on my ongoing work on metrics' conversation with him.

littleaj:
@dhaval24, I was trying to make the following point:

* You will always have 1 point if you are making a Measurement
* You will have 1..N points if you are sampling/aggregating a collection of measurements. There could still exist a thread in charge of measuring some set of values and in some time window only one point is collected and reported. You've still sampled and computed aggregations, but the data set just happens to be of size=1 in this time window. Also, in this case you couldn't compute StdDev using the sample std deviation formula.

Now that , I think it is valid to report standard deviation as null.

This all depends on what we do with Metric.kind={Measurement|Aggregation. I think we'll need to talk to Greg about the proper handling here.

littleaj:
@dhaval24 , if they're handing us a `MetricTelemetry` object it's up to them to fill in all the fields. This came up because of the Local Forwarder work. Let's discuss in more detail tomorrow

dhaval24:
@littleaj I agree. If we report this as null, backend will automatically set it to 0 which is also eventually the same result. For kind yes we can double check. 

my point was only that the change that you made should happen inside MetricTelemetry class

macrogreg:
We have (sadly, for historical reasons) a bad treatment for aggregated vs. non-aggregated values. @littleaj  is right that there are two logically different cases:
- An aggregate that just happened to have count = 1. Then SdtDev = 0.
- A non-aggregated metric point (something that should not exist, but it does for historical reasons). Then there is no StdDev. But the backend will set it to 0. :)

I t is true that 0 is a valid value for StdDev even for an aggregate over many points while -1 clearly says "no StdDev". But in the context of the recent discussion, I would prefer to see one and the numerical value for whenever we fill in "null", and 0 is just more intuitive there.

littleaj:
I updated the title and description with current information. It turns out stddev doesn't need to be non-null

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -217,42 +217,66 @@ public void trackTrace(TraceTelemetry telemetry) {
...
+    }
+
+    /**
+     * Sends a numeric metric to Application Insights. Appears in customMetrics in Analytics, and under Custom Metrics in Metric Explorer.
+     * @param name The name of the metric. Max length 150.
+     * @param value The value of the metric.
+     * @param properties Named string values you can use to search and classify trace messages.
+     * @throws IllegalArgumentException if name is null or empty.
+     */
+    public void trackMetric(String name, double value, Map<String, String> properties) {
```
dhaval24: Is this the new public API we want to expose? The method looks pretty handy but we should just make sure, that other SDKs also have it. Even if they don't I am fine with introducing this method, just good to check and suggest them too. I was looking for this method too and was surprised didn't existed!

core/src/main/java/com/microsoft/applicationinsights/internal/util/MapUtil.java
```diff
@@ -35,14 +35,20 @@
...
+     * @param source the source map. Cannot be null.
+     * @param target the target map. Cannot be null.
+     * @param <Value> The type of the values in both maps
+     * @throws IllegalArgumentException if either {@code source} or {@code target} are null.
+     */
     public static <Value> void copy(Map<String, Value> source, Map<String, Value> target) {
         if (source == null) {
-            Preconditions.checkArgument(source != null, "source must not be null");
-            return;
+            throw new IllegalArgumentException("source must not be null");
```
dhaval24: Generally, we don't want the customer's app to crash because of exception in tracing library. Now I do understand that this would happen most likely because of users fault and in that case they would like to know immediately. But I would like to discuss this a bit more. May be we can add try catch and call trackException() to report to AI?

dhaval24:
@littleaj may be we should add one unit test too in MetricTelemetryTests.java

core/src/main/java/com/microsoft/applicationinsights/internal/util/MapUtil.java
littleaj: Look into `Preconditions.checkArgument`. This doesn't change how the method behaves.

dhaval24: Alright makes sense. Thanks for explanation.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
littleaj: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/39a5ef23d834777eefdd72149de705a016eb06b0/src/Microsoft.ApplicationInsights/TelemetryClient.cs#L228

littleaj:
One thing I didn't change in Java: https://github.com/Microsoft/ApplicationInsights-dotnet/blob/6e7d6d78f966846e524c39c22664271fb45b2a27/src/Microsoft.ApplicationInsights/DataContracts/MetricTelemetry.cs#L203

Should getCount() return 1 if MetricTelemetry.count==null? I didn't change it because of this:
```java
MetricTelemetry mt = new MetricTelemetry();
mt.setCount(null);
boolean isCountNull = mt.getCount() == null; // If we returned 1 by default, this would be false
```

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
dhaval24: Cool makes sense.

macrogreg: Let's please not make any more overloads that take non-aggregates (like this one).
In the mid-term we want to deprecate the existing ones too :)

core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java
```diff
@@ -186,17 +195,13 @@ protected MetricData getData() {
     }
 
     private void updateKind() {
```
dhaval24: @macrogreg just for my information, does the backend still has the distinction between aggregate and measurement? I thought all of the metrics are now treated as aggregate. Am I not following anything here?

PR #718, opened 2018-08-23
SpringBoot Starter Improvements : Micrometer adjustments, addressing feedback from Pivotal

dhaval24:
This PR makes necessary adjustments to starter to incorporate integration with Micrometer metrics in a smooth way. 

Please not this is work in progress and there would be some minor tweaks required when wiring with Boot 2 will be done (essentially our Autoconfiguration should work before boot2's autoconfiguration)

Fixes the thread shutdown issue by registering ApplicationInsightsServletContextListener.

Ensures smooth ikey integration by allowing environment variable use also.

Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x

- [x] Needs  #717 to be merged

- [x] Update Release notes

- [x] Review public exposure changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 - Introduces SpringBoot Starter 1.0.2-BETA
 - Fixes the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
 - SpringBoot Starter now supports reading iKey using all the variable names as core sdk.
-- Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x
+- Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x.
+- `InterceptorRegistry` class no longer has `@EnableWebMvc` annotation as it breaks springboot autoconfig.
 
 # Version 2.1.2
 - Fix [#676](https://github.com/Microsoft/ApplicationInsights-Java/issues/676). This fixes the HTTP dependency collection when using NetFlix Zuul Library.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+# Version 2.2.0
+- Introduces SpringBoot Starter 1.0.2-BETA
+- Fixes the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
+- SpringBoot Starter now supports reading iKey using all the variable names as core sdk.
+- Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x
+
 # Version 2.1.2
 - Fix [#676](https://github.com/Microsoft/ApplicationInsights-Java/issues/676). This fixes the HTTP dependency collection when using NetFlix Zuul Library.
 - Remove the method `httpMethodFinishedWithPath` from the interface `ImplementationCoordinator.java` as the associated instrumentation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 # Version 2.2.0
 - Introduces SpringBoot Starter 1.0.2-BETA
 - Starter now respects autoconfiguration for Micrometer metrics.
-- Fixes the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
+- Fix [#712](https://github.com/Microsoft/ApplicationInsights-Java/issues/712) the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
 - SpringBoot Starter now supports reading iKey using all the variable names as core sdk.
 - Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x.
 - `InterceptorRegistry` class no longer has `@EnableWebMvc` annotation as it breaks springboot autoconfig.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Version 2.2.0
 - Introduces SpringBoot Starter 1.0.2-BETA
+- Starter now respects autoconfiguration for Micrometer metrics.
 - Fixes the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
 - SpringBoot Starter now supports reading iKey using all the variable names as core sdk.
 - Starter would no longer support relaxed binding of ikey property due to complex conditional need and backport problems with RelaxedBinder from Boot 2 to 1.5.x.
--- a/azure-application-insights-spring-boot-starter/build.gradle
+++ b/azure-application-insights-spring-boot-starter/build.gradle
@@ -20,6 +20,9 @@ import com.microsoft.applicationinsights.build.tasks.PropsFileGen
  * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
  * DEALINGS IN THE SOFTWARE.
  */
+plugins {
+    id 'com.github.johnrengelman.shadow' version '2.0.1'
+}
 
 apply from: "$buildScriptsDir/common-java.gradle"
 apply from: "$buildScriptsDir/publishing.gradle"
@@ -44,8 +47,9 @@ sourceSets {
 }
 
 dependencies {
-    compile (project(':core'))
-    compile (project(':web'))
+    compile(project(':core'))
+    compile(project(':web'))
+    compile ([group: 'org.apache.commons', name: 'commons-lang3', version: '3.7'])
     provided('org.springframework.boot:spring-boot:1.5.9.RELEASE')
     provided('org.springframework.boot:spring-boot-autoconfigure:1.5.9.RELEASE')
     provided('org.springframework.boot:spring-boot-starter-web:1.5.9.RELEASE')
@@ -53,6 +57,28 @@ dependencies {
     testCompile('junit:junit:4.12')
     testCompile('org.springframework.boot:spring-boot-starter-test:1.5.9.RELEASE')
     testCompile('org.assertj:assertj-core:2.6.0')
+}
+
+ext.dependenciesPrefix = 'com.microsoft.applicationinsights.boot.dependencies'
+shadowJar {
+    dependencies {
+        exclude(project(':core'))
+        exclude(project(':web'))
+        exclude(dependency([group: 'eu.infomas', name: 'annotation-detector', version: '3.0.5']))
+        exclude(dependency([group: 'commons-io', name: 'commons-io', version: '2.6']))
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3']))
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpasyncclient', version: '4.1.3']))
+        exclude(dependency([group: 'com.google.guava', name: 'guava', version: '20.0']))
+        exclude(dependency([group: 'com.google.code.gson', name: 'gson', version: '2.8.2']))
+    }
+    classifier=''
+    relocate 'org.apache.commons', "${dependenciesPrefix}.apachecommons"
+    relocate 'org.apache.http', "${dependenciesPrefix}.http"
+}
+
+jar {
+    dependsOn shadowJar
+    enabled = false
 }
 
 compileJava.dependsOn(processResources)
--- a/azure-application-insights-spring-boot-starter/build.gradle
+++ b/azure-application-insights-spring-boot-starter/build.gradle
@@ -67,13 +67,18 @@ shadowJar {
         exclude(dependency([group: 'eu.infomas', name: 'annotation-detector', version: '3.0.5']))
         exclude(dependency([group: 'commons-io', name: 'commons-io', version: '2.6']))
         exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3']))
+        // Transitive dependencies of httpclient that needs to be excluded seperately as shadow jar doesn't exclude them
+        // automatically.
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpcore', version: '4.4.6']))
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpcore-nio', version: '4.4.6']))
+        // end excluding transitive dependencies
+
         exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpasyncclient', version: '4.1.3']))
         exclude(dependency([group: 'com.google.guava', name: 'guava', version: '20.0']))
         exclude(dependency([group: 'com.google.code.gson', name: 'gson', version: '2.8.2']))
     }
     classifier=''
     relocate 'org.apache.commons', "${dependenciesPrefix}.apachecommons"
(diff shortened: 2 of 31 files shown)


dhaval24:
@tokaplan just added you to start understanding!

dhaval24:
@snicoll could you please confirm that the changes to `InterceptorRegistry` will eliminate the autoconfigure issues that were raised? I followed spring's official documentation for making this change.

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
```diff
@@ -21,18 +21,27 @@
...
+ * that needs to be configured consider removing the @EnableWebMvc annotation and extending directly from
+ * WebMvcConfigurationSupport or DelegatingWebMvcConfiguration.
+ *
+ * @see <a href="https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/EnableWebMvc.html">
+ *   https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/EnableWebMvc.html</a>
  */
-@EnableWebMvc
 @Configuration
-public class InterceptorRegistry extends WebMvcConfigurerAdapter {
+public class InterceptorRegistry extends WebMvcConfigurationSupport {
```
dhaval24: @snicoll could you please confirm that the changes to InterceptorRegistry will eliminate the autoconfigure issues that were raised? I followed spring's official documentation for making this change.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/conditionals/InstrumentationKeyCondition.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Conditional to check if instrumentation key is either set in application.properties
+ * or as system property or environment variable.
+ *
+ * @author Dhaval Doshi
+ */
+public class InstrumentationKeyCondition implements Condition {
+
+  @Override
+  public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
```
snicoll: `@ConditionalOnProperty` already searches in those location. Is there a reason why you can't use it?

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
snicoll: I don't remember the context, sorry. What autoconfigure isssues?

This code is much better than the previous situation. It should be noted that the doc you've pasted applies primarily to end users and applications. Using `@EnableWebMvc` in a starter is simply not acceptable as it switches off everything Spring Boot does for the user.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
```diff
@@ -91,7 +93,17 @@ public void setEnabled(boolean enabled) {
   }
 
   public String getInstrumentationKey() {
-    return instrumentationKey;
+    // First try getting ikey from application.properties
```
snicoll: I don't understand this code. If the key is present in `application.properties` or as a System property or OS Env variables, it will be injected. Doing this manually either means you're post-processing the key, or something else is wrong.

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
bclozel: Extending `WebMvcConfigurationSupport` is like using `@EnableWebMvc`, it will override all Spring Boot auto-configuration for Spring MVC. In this case, implementing `WebMvcConfigurer` or extending `WebMvcConfigurerAdapter` should solve the problem.

dhaval24: @snicoll @bclozel thanks this is a very good insight! I kind of felt this when taking a look at the doc but wasn't explicit. I will either implement `WebMvcConfigurer` or `WebMvcConfigurerAdapter`. By the way it already was using `WebMvcConfigurerAdapter`. In this case I can just remove `@EnableWebMvc` annotation and leave this upto end user to put it on one of their configuration classes. Let me know if this makes more sense!

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/conditionals/InstrumentationKeyCondition.java
dhaval24: @snicoll due to historic reasons users of Application could set instrumentation key with three different names either in environment variables or properties which is : `APPLICATION_INSIGHTS_IKEY`and `APPINSIGHTS_INSTRUMENTATIONKEY` therefore I needed to verify if the instrumentation key is present as either of the three and therefore I had to come up with this custom conditional instead of `@ConditionalOnProperty` annotation.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
dhaval24: As I mentioned in the comment above instrumentation key can come from any of the three locations and therefore I had the need to perform this logic. Let me know if you think there is a better way to do it.

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
snicoll: I don't think you should mention `@EnableWebMvc` anywhere at all. We'll do something sensible automatically for the user and adding this annotation manually is a corner case.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/conditionals/InstrumentationKeyCondition.java
snicoll: You should extends from `SpringBootCondition` then because your custom condition doesn't show up in the auto-config report. I also find the doc a bit misleading: it should describe the fact the keys are different.

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
dhaval24: Makes sense. The only question is that what happens when user is using plain Spring MVC instead of springboot ?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/conditionals/InstrumentationKeyCondition.java
dhaval24: Sounds good. Thanks for the suggestion here, and I will improve the comments to make it more relevant.

web/src/main/java/com/microsoft/applicationinsights/web/spring/internal/InterceptorRegistry.java
snicoll: If they're using plain Spring MVC, none of this apply as the auto-configuration support is in Spring Boot.

dhaval24:
@littleaj @grlima could you please take a look at it? I would like to get this reviewed and merged!

dhaval24:
Ping .... folks could you please start review of my PR too :) @grlima @tokaplan @littleaj

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/conditionals/InstrumentationKeyCondition.java
```diff
@@ -0,0 +1,55 @@
...
+public class InstrumentationKeyCondition extends SpringBootCondition {
+
+  @Override
+  public ConditionOutcome getMatchOutcome(ConditionContext context, AnnotatedTypeMetadata metadata) {
+    String iKey = context.getEnvironment().getProperty("azure.application-insights.instrumentation-key");
+    if (StringUtils.isNoneBlank(iKey)) {
+      return new ConditionOutcome(true, ConditionMessage.of("instrumentation key found"));
+    }
+    iKey = IkeyResolver.getIkeyFromEnvironmentVariables();
+    if (StringUtils.isNoneBlank(iKey)) {
```
littleaj: nitpick: `isNotBlank` is a better choice here.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/helpers/IkeyResolver.java
```diff
@@ -0,0 +1,58 @@
...
+ * @author Dhaval Doshi
+ */
+public class IkeyResolver {
+
+  private static final String EXTERNAL_PROPERTY_IKEY_NAME = "APPLICATION_INSIGHTS_IKEY";
+  private static final String EXTERNAL_PROPERTY_IKEY_NAME_SECONDARY = "APPINSIGHTS_INSTRUMENTATIONKEY";
+
+    public static String getIkeyFromEnvironmentVariables() {
+    String v = System.getProperty(EXTERNAL_PROPERTY_IKEY_NAME);
+    if (StringUtils.isNoneBlank(v)) {
```
littleaj: same nitpick as before: `isNotBlank` is for a single value, `isNoneBlank` is for varargs.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/initializer/SpringBootTelemetryInitializer.java
```diff
@@ -1,4 +1,4 @@
-package com.microsoft.applicationinsights.boot.initializer;
```
littleaj: This comment applies to all relevant files: I don't get the package rename. If it were me, I'd go with `com.microsoft.applicationinsights.web.spring.boot.*`, modeling after the other spring related classes we have.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/helpers/IkeyResolver.java
dhaval24: Done

azure-application-insights-spring-boot-starter/build.gradle
```diff
@@ -44,8 +47,9 @@ sourceSets {
 }
 
 dependencies {
-    compile (project(':core'))
-    compile (project(':web'))
+    compile(project(':core'))
```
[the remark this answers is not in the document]
dhaval24: So I checked the provided configuration and it says that client is required to add this dependency on the classpath which is not desired and hence I don't think it would work but I may be wrong and hence convinced.

azure-application-insights-spring-boot-starter/build.gradle
```diff
@@ -55,6 +59,28 @@ dependencies {
...
+        exclude(dependency([group: 'eu.infomas', name: 'annotation-detector', version: '3.0.5']))
+        exclude(dependency([group: 'commons-io', name: 'commons-io', version: '2.6']))
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3']))
+        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpasyncclient', version: '4.1.3']))
+        exclude(dependency([group: 'com.google.guava', name: 'guava', version: '20.0']))
+        exclude(dependency([group: 'com.google.code.gson', name: 'gson', version: '2.8.2']))
+    }
+    classifier=''
+    relocate 'org.apache.commons', "${dependenciesPrefix}.apachecommons"
+    relocate 'org.apache.http', "${dependenciesPrefix}.http"
```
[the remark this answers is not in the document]
littleaj: Try this:
https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#sub:disabling_resolution_transitive_dependencies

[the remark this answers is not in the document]
dhaval24: I am not sure what you mean here. I already have following dependencies excluded:
```
exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.3']))
        exclude(dependency([group: 'org.apache.httpcomponents', name: 'httpasyncclient', version: '4.1.3']))
```
I want all it's transitive dependencies to also be excluded. This is more of a question of filtering in shadow jar. I think, the doc you point to is how to import a gradle dependency without all it's transitive dependencies.

http://imperceptiblethoughts.com/shadow/#filtering_dependencies

Shadow jar docs say that transitive dependencies cannot be filtered out.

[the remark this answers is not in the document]
dhaval24: Anyway, I have excluded those transitive dependencies manually.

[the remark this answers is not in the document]
littleaj: I see. Still, I don't think we should repackage the http packages.

use `gradlew :azure-application-insights-spring-boot-starter:dependencies` to get a graph of how dependencies are resolved.

After running, it looks like you just need
```gradle
exclude(dependency('org.apache.httpcomponents:httpcore:4.4.6'))
exclude(dependency('org.apache.httpcomponents:httpcore-nio:4.4.6'))
```
and you can remove
```gradle
relocate 'org.apache.http', "${dependenciesPrefix}.http"
```

There's probably a way to do this automatically using gradle's project dependencies DSL, but this should be fine.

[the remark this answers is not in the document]
littleaj: Also, you can use a wildcard on the version.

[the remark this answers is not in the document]
dhaval24: Thanks, see my latest commit. I removed it already before you commented :) so we are good. Appreciate you bringing this up.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -65,11 +73,40 @@ public FilterRegistrationBean webRequestTrackingFilterRegistrationBean(WebReques
         return registration;
     }
 
+    /**
+     * Programmatically registers an AppliactionInsightsServletContextListner to destroy all the running threads.
```
tokaplan: Two typos in 'AppliactionInsightsServletContextListner'.
Happened to notice while learning the codebase :)

dhaval24:
@grlima could you please take a look and approve this PR. I have addressed all the comments from Arthur and Alex.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+# Version 2.2.0
+- Introduces SpringBoot Starter 1.0.2-BETA
+- Starter now respects autoconfiguration for Micrometer metrics.
+- Fixes the thread shutdown issue in SpringBoot Starter by registering `ApplicationInsightsServletContextListener`.
```
grlima: can you please add a link to the GH issue with the shutdown?

dhaval24: Done

PR #726, opened 2018-08-30
Deprecate ai.device.role* tags, replace with ai.cloud.role*

littleaj:
`ai.device.roleName` and `ai.device.roleInstance` are deprecated and replaced with `ai.cloud.role` and `ai.cloud.roleInstance`, respectively.

The .NET SDK uses a CloudContext object for tracking these values. This makes the most sense to me, so I implemented this in our SDK.

Let me know if you agree with the new Initializers.

This should go in the next release along side the LocalForwarder changes.

Before Merge:
- [x] CHANGELOG.md updated, need notes about deprecating accessors in DeviceContext and their replacements.
- [x] pending test for sending both tags

Post Release:
- [ ] Update documentation: sample config should show `WebAppNameContextInitializer` along side the other Web* initializers.


azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/initializer/SpringBootTelemetryInitializer.java
```diff
@@ -23,9 +23,8 @@
 
   @Override
   public void initialize(Telemetry telemetry) {
-
-    DeviceContext device = telemetry.getContext().getDevice();
-    device.setRoleName(appName);
+    CloudContext device = telemetry.getContext().getCloud();
```
dhaval24: well device variable name doesn't make sense now. may be something related to cloud as name ?:)

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/CloudContext.java
```diff
@@ -0,0 +1,29 @@
+package com.microsoft.applicationinsights.extensibility.context;
+
+import com.microsoft.applicationinsights.internal.util.MapUtil;
+
+import java.util.concurrent.ConcurrentMap;
```
dhaval24: Can we have a javadoc for this class?

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/ContextTagKeys.java
```diff
@@ -863,7 +880,9 @@ protected void reset(String name, String qualifiedName) {
         InternalSdkVersion = "ai.internal.sdkVersion";
         InternalAgentVersion = "ai.internal.agentVersion";
         SyntheticSource = "ai.operation.syntheticSource";
-        InternalNodeName = "ai.internal.nodename";
+        InternalNodeName = "ai.internal.nodeName";
```
dhaval24: was the camel casing of nodeName intentional or this is typo?

core/src/main/java/com/microsoft/applicationinsights/internal/util/DeviceInfo.java
```diff
@@ -55,20 +56,8 @@ public static String getOperatingVersionArchitecture()
...
-            return ip.getCanonicalHostName();
-        }
-        catch (UnknownHostException e)
-        {
-            InternalLogger.INSTANCE.error("Failed to get canonical host name, exception: %s", e.toString());
-            InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(e));
-        }
-        return null;
+    public static String getHostName() {
+        return CommonUtils.getHostName();
```
dhaval24: like this reuse  👍

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/CloudInfoContextInitializer.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.applicationinsights.extensibility.initializer;
+
+import com.microsoft.applicationinsights.common.CommonUtils;
+import com.microsoft.applicationinsights.extensibility.ContextInitializer;
+import com.microsoft.applicationinsights.extensibility.context.CloudContext;
+import com.microsoft.applicationinsights.telemetry.TelemetryContext;
+
+public final class CloudInfoContextInitializer implements ContextInitializer {
```
dhaval24: Please add java doc to the class.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -178,11 +182,21 @@ public void init(FilterConfig config) {
         }
     }
 
+    private void configureWebAppNameContextInitializer(String appName, TelemetryConfiguration configuration) {
+        for (ContextInitializer ci : configuration.getContextInitializers()) {
+            if (ci instanceof WebAppNameContextInitializer) {
+                ((WebAppNameContextInitializer)ci).setAppName(appName);
```
dhaval24: Please correct me if I am thinking write. 

ContextInitializers -> TelemetryInitializers -> Processors (This is the pipeline correct) in this case it should be fine. If this is not the case and ContextInitializers fall after TelemetryInitializers than this will screw up the springboot app name (set by TelemetryInitializer) which is not the same as the servlet context parsed name and should be set actually in case of springboot instead of servlet context parsed name. Double check and confirm this.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/initializer/SpringBootTelemetryInitializer.java
littleaj: correct. I'll fix that

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/CloudContext.java
littleaj: yes

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/ContextTagKeys.java
littleaj: The correct version is "nodeName". See the same class in the .NET SDK.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/CloudInfoContextInitializer.java
littleaj: Will do

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: Yup make sense to update it to context initializer. Now the question is, will boot initializer override the one in the core sdk, then it should be perfectly fine. We can try and double check that it is the case !

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsModuleConfiguration.java
```diff
@@ -84,6 +85,15 @@ public DeviceInfoContextInitializer deviceInfoContextInitializer() {
         return new DeviceInfoContextInitializer();
     }
 
+    /**
+     * Bean for CloudInfoContextInitializer
+     * @return instance of {@link CloudInfoContextInitializer}
+     */
+    @Bean
+    public CloudInfoContextInitializer cloudInfoContextInitializer() {
```
dhaval24: By the way @littleaj reduce the visibility of this to default instead of public. See #718 PR where I have reduced the visibility of all beans as they are not needed outside.

littleaj: will do.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: That's a good question, but I don't think they'll conflict. The `WebAppNameContextInitializer` doesn't need to be added to a spring-boot app, but I think it's possible for the `WebAppNameContextInitializer` to replace the spring-boot one.

As it stands now, with no changes, you have to explicitly add the WebAppNameContextInitializer so it should function properly. Let's have a discussion about the best approach here when you're back.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/initializer/SpringBootTelemetryInitializer.java
littleaj: [ ] testing checkbox...

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -72,7 +73,7 @@
         WebReflectionUtils.initialize();
     }
     // region Members
-    private final static String FILTER_NAME = "ApplicationInsightsWebFilter";
+    final static String FILTER_NAME = "ApplicationInsightsWebFilter";
```
dhaval24: Was this made non private for testing purpose? Ok It's fine I think. Class is final so no one will extend it. By the way, cam protected work here?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -173,6 +175,11 @@ public void init(FilterConfig config) {
 
             telemetryClient = new TelemetryClient(configuration);
             webModulesContainer = new WebModulesContainer(configuration);
+
+            if (StringUtils.isNotEmpty(config.getFilterName())) {
```
dhaval24: Is this the fix that you are attempting to make for starter scenario when we inject filter?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: Yes, it's `@VisibleForTesting`...I should've added that

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: No, this is to fix the log messages.

littleaj:
@dhaval24 @grlima , I'm thinking the `DeviceInfoContextInitializer` should still send the old tag as long as `DeviceContext` still has setters for it. I'm checking how ingestion will handle both tags. I _think_ it will favor the `ai.cloud.roleInstance` over `ai.device.roleInstance`, but I want to be sure.

dhaval24:
@littleaj I think as @SergeyKanzhelev suggested that ingestion automatically transforms deviceInfo to cloud. You can double check though.

littleaj:
@dhaval24 it does, but I want to be sure it will favor the value in the new tag if we send both. It should be doing this.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: added comment

dhaval24:
@littleaj is it possible to do this verification and then instead only use one of the Initializers. I am not personally a fan of duplication of same logic. If possible we should adhere to one.

littleaj:
Confirmed sending both tags will work as expected: the `cloud_roleInstance` dimension gets the value from the new tag.

dhaval24:
What does old tag give then? The same thing?

dhaval24:
@littleaj as of today I think `SdkVersionContextInitializer` and `DeviceContextInitializers` are added by default. Look into 'ContextInitializersInitializers.java` class. What you can do there is just add the CloudContextInitializers too and remove the logic of cloud_RoleName setting from DeviceContextInitializer. The end user will see no change in behavior on updating the SDK.

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/DeviceInfoContextInitializer.java
```diff
@@ -40,5 +40,11 @@ public void initialize(TelemetryContext context)
         device.setOperatingSystemVersion(DeviceInfo.getOperatingSystemVersion());
         device.setId(DeviceInfo.getHostName());
         device.setLocale(DeviceInfo.getLocale());
+
+        // remove this when setRoleInstance is removed from DeviceContext
+        String hostName = CommonUtils.getHostName();
```
dhaval24: My specific point was for this revert.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: I don't see it :)

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -84,6 +85,7 @@
     private String appName;
     private static final String AGENT_LOCATOR_INTERFACE_NAME = "com.microsoft.applicationinsights."
         + "agent.internal.coresync.AgentNotificationsHandler";
+    private String filterName = FILTER_NAME;
```
dhaval24: Where is this field used for? Your commit message says - added fitlerName field "just in case someone tries to add tow instances"

littleaj: line 289 and 295. set on line 180

core/src/main/java/com/microsoft/applicationinsights/extensibility/initializer/DeviceInfoContextInitializer.java
littleaj: Right. Since they're default initializers, I'll remove this.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: Thanks makes sense. Just a question will this field ever be null? For example if user forgets to specify filterName?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: I think this would work fine. I double checked the code in the starter. Essentially we create a barebone TelemetryConfiguration() object there and inject all the initializers and modules etc using DI. 

https://github.com/Microsoft/ApplicationInsights-Java/blob/cf042c35fdae23479e84765962580e307a8aadef/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/ApplicationInsightsTelemetryAutoConfiguration.java#L114

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/boot/initializer/SpringBootTelemetryInitializer.java
```diff
@@ -23,9 +23,8 @@
 
   @Override
   public void initialize(Telemetry telemetry) {
-
-    DeviceContext device = telemetry.getContext().getDevice();
-    device.setRoleName(appName);
+    CloudContext cloud = telemetry.getContext().getCloud();
+    cloud.setRole(appName);
```
dhaval24: AppMap uses this to create a new Role (element) in UX. Essentially this represents a Microservice in it self. I will come back to why I made this comment here, in another comment in this PR.

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/initializers/WebAppNameContextInitializer.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public void setAppName(String appName) {
+        this.appName = appName;
+    }
+
+    @Override
+    public void initialize(TelemetryContext context) {
+        if (!CommonUtils.isNullOrEmpty(appName)) {
+            CloudContext cloud = context.getCloud();
+            cloud.setRole(appName);
```
dhaval24: This part is specifically related to what I commented about AppMap component name. When you set this from the Filter this would override the appName which springboot gives. 
Specifically when we think about converting TelemetryInitializer to ContextInitializer (the one in Starter)

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -250,30 +272,28 @@ private void setKeyOnTLS(String key) {
...
 
-    private synchronized void initialize(FilterConfig filterConfig) {
-
+    private synchronized void initializeAgentIfAvailable(FilterConfig filterConfig) {
         StopWatch sw = StopWatch.createStarted();
 
         //If Agent Jar is not present in the class path skip the process
-        if (!CommonUtils.isClassPresentOnClassPath(AGENT_LOCATOR_INTERFACE_NAME,
-            this.getClass().getClassLoader())) {
+        if (!CommonUtils.isClassPresentOnClassPath(AGENT_LOCATOR_INTERFACE_NAME, this.getClass().getClassLoader())) {
```
dhaval24: Why this style change is part of this PR?

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: I don't think it'll be null, possibly empty or whitespace (it's up to the J2EE spec), but I can add a wrapper when it's used to guard against that.

littleaj: Actually, String.format handles nulls, so it's fine.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: Just to see if you'd notice :P

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: As discussed, the flow is a bit different and the spring boot initializer overwrites the app name from the WebAppNameContextInitializer, if both are enabled. The latter initializer is required in config to work, so it should be rare to see both initializer enabled.

littleaj:
@dhaval24 , I think I've addressed all your concerns

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: I didn't but appreciate that you changed it back! These were nit picks but I learned from you in my last PRs. Nits go unnoticed and slowly start creeping in :)

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/ContextTagKeys.java
```diff
@@ -185,7 +189,7 @@ public final String getApplicationId() {
     /**
      * @param value new value of ApplicationId property
      */
-    public final void setApplicationId(String value) {
```
grlima: Why are we removing the public accessor? Does it break back compat?

core/src/main/java/com/microsoft/applicationinsights/common/CommonUtils.java
```diff
@@ -32,15 +32,18 @@ public static boolean isNullOrEmpty(String string) {
...
     try {
-      InetAddress addr;
-      addr = InetAddress.getLocalHost();
+      InetAddress addr = InetAddress.getLocalHost();
       return addr.getCanonicalHostName();
     } catch (UnknownHostException ex) {
-      // optional parameter. do nothing if unresolvable
-      InternalLogger.INSTANCE.trace(
-          "Unresolvable host error. Stack trace generated is %s", ExceptionUtils.getStackTrace(ex));
+      InternalLogger.INSTANCE.warn("Error resolving hostname:%n%s", ExceptionUtils.getStackTrace(ex));
```
grlima: Thanks for fixing this.

core/src/main/java/com/microsoft/applicationinsights/extensibility/context/ContextTagKeys.java
littleaj: that shouldn't have been changed. 

Technically, these keys should not have setters. The keys are predetermined and are given special treatment when ingested; they key name shouldn't be mutable. I was thinking of doing this to the setters until I realized it's a bond generated file.

PR #733, opened 2018-09-04
Adding test case http dependency 

Terryxxx:
This PR is to validate Apache Http Client and OKHttp Client dependency.
@littleaj @dhaval24 @grlima Please review it.


test/smoke/testApps/HttpDependency/src/main/java/com/microsoft/ajl/simplecalc/SimpleTestHttpDependencyServlet.java
```diff
@@ -0,0 +1,64 @@
...
+            GetHttpUseOkHttpClient("https://www.microsoftabc.com");
+        } catch (Exception e) {
+            // TODO: handle exception
+        }
+
+        // Get false http use apache http client
+        GetHttpUseApacheHttpClient("https://www.bingxxxxx.com");
+    }
+
+    public void GetHttpUseApacheHttpClient(String url) throws IOException {
```
dhaval24: Make this function private.  Also follow java methods convention. Use came casing, remane this to `getHttpUseApacheHttpClient`.

test/smoke/testApps/HttpDependency/src/main/java/com/microsoft/ajl/simplecalc/SimpleTestHttpDependencyServlet.java
```diff
@@ -0,0 +1,64 @@
...
+        HttpClient client = HttpClientBuilder.create().build();
+        HttpGet get = new HttpGet(url);
+        HttpResponse httpResponse = client.execute(get);
+    }
+
+    public void GetHttpUseOkHttpClient(String url) throws IOException {
```
dhaval24: Same comment as above.

test/smoke/testApps/HttpDependency/build.gradle
```diff
@@ -0,0 +1,24 @@
+apply plugin: 'war'
+
+dependencies {
+    compile aiCoreJar
+    compile aiWebJar
+    compile 'com.google.guava:guava:20.0'
+
+    providedCompile 'javax.servlet:javax.servlet-api:3.0.1'
+
+    compile 'org.apache.httpcomponents:httpclient:4.5.3'
```
dhaval24: We would need to check for other versions of ApacheHttpClient too. May be a seperate test or whatever better way @littleaj knows of. ApacheHttpClient 4.5.6 in particular.

test/smoke/testApps/HttpDependency/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HttpDependencyTest.java
```diff
@@ -0,0 +1,60 @@
...
+@UseAgent
+public class HttpDependencyTest extends AiSmokeTest {
+
+    @Test
+    @TargetUri("/httpDependency")
+    public void testHttpDependency() {
+        assertEquals(1, mockedIngestion.getCountForType("RequestData"));
+
+        //Now only can get 3 remote dependency. Can not collect non-exist url failed http depnedncy if used Apache http client
+        //assertEquals(4, mockedIngestion.getCountForType("RemoteDependencyData"));
```
dhaval24: Do you mean we do not collect failed dependencies when using ApacheHttpClient?

test/smoke/testApps/HttpDependency/build.gradle
Terryxxx: Thanks for your remind. I will discuss with @littleaj  about this.

test/smoke/testApps/HttpDependency/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HttpDependencyTest.java
Terryxxx: I think that when giving a non-exist URL, it should produce a Dependency and an Exception.
But it only produce an exception using follow code:
"GetHttpUseApacheHttpClient("https://www.bingxxxxx.com")"

dhaval24: I don't think so. In this case the dependency call doesn't complete itself and throws the exception in middle before the instrumentation call completes. So I think this is the right behavior. We should see failed HTTP dependency when we get something like 500 or similar result codes. @littleaj and @grlima can comment on this too.

Terryxxx: I don't know why using OKHttp Client can get ?

Terryxxx: In .NET SDK, we can collect failed http dependency if used non-exist URL.

dhaval24: very interesting. @grlima are you aware about this behavior in our sdk?

littleaj:
Confirmed that we should be showing a dependency event as well as an exception (as long as the client throws one).

Terryxxx:
@littleaj  I will create a new PR for different Http Client(4.5.6). Can I get the failed dependency if i used newest master code?

Terryxxx:
I have add 'ignore' for this test. @littleaj

littleaj:
This test was included in #959

PR #736, opened 2018-09-12
Fix possible open streams which could be causing the failing tests.

littleaj:
<del>Not sure if we want to merge this yet. Let's let the build run and see if this fixes it.</del>

Yup, this was the issue. Let's merge this.


littleaj:
It worked! 🕺

PR #738, opened 2018-09-13
Tests for WebAppNameContextInitializer

littleaj:
I noticed this was missing tests. They're pretty straightforward.

I did optimize the filter test class a bit: using mockito instead of class override to give it more flexibility.

### DIFF
--- a/web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
+++ b/web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
@@ -37,7 +37,9 @@ import com.microsoft.applicationinsights.web.utils.ServletUtils;
 import java.io.IOException;
 import java.lang.reflect.Field;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.mockito.Mockito.*;
 
 /**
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
I doubt Java SDK is better then .NET in handling slot swap. If it is - add test case for it. See https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/648 for details

SergeyKanzhelev:
@littleaj which is exactly the problem with the slot swap. As I said .NET is not better than that now... we entertained an idea to periodically check the environment variable for update which probably what we will end up implementing. Perhaps add ignored test for now =)

dhaval24:
@SergeyKanzhelev could a better way be to have some sort of callback when the slot change happens? Checking of environment variable would still have some delay between the when slot change happens and SDK picks up. Till then App Map would surface incorrect information I believe thoughts ?

web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
```diff
@@ -74,7 +64,7 @@ public void testFilterInitializedSuccessfullyFromConfiguration() throws ServletE
         Filter filter = createInitializedFilter();
         WebModulesContainer container = ServletUtils.getWebModuleContainer(filter);
 
-        Assert.assertNotNull("Container shouldn't be null", container);
+        assertNotNull("Container shouldn't be null", container);
         Assert.assertTrue("Modules container shouldn't be empty", container.getModulesCount() > 0);
```
dhaval24: I would like consistency if possible, can we stay consistent in importing asserts. Here Assert.assertFalse is for example imported and Assert.assertTrue isn't.

web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
```diff
@@ -156,6 +146,18 @@ public void testUnhandledIOExceptionWithTelemetryClientThatThrows() throws Illeg
...
+        FilterChain chain = mock(FilterChain.class);
+
+        mocks.filter.doFilter(ServletUtils.generateDummyServletRequest(), ServletUtils.generateDummyServletResponse(), chain);
+        String role = mocks.mockTelemetryClient.getContext().getCloud().getRole();
+        assertNotNull(role);
+        assertEquals(this.mockContextPath.replace("/", ""), role);
+
+    }
+
     // region Private methods
```
dhaval24: This kind of C# stuff also we should consider cleaning :)

SergeyKanzhelev:
@dhaval24 there is no callback. Callback can be implemented by inspecting environment variables. This is how it works =(

web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
littleaj: `// region` is recognized by IntelliJ.

web/src/test/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilterTests.java
littleaj: We can clean up consistency issue like this in a style related PR.

PR #742, opened 2018-09-17
Removed Servlet 3 annotations

littleaj:
<del>Waiting on smoketests before merging.</del>


web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -204,7 +202,6 @@ private void configureWebAppNameContextInitializer(String appName, TelemetryConf
      */
     public void destroy() {
         cleanup();
-        ThreadContext.remove();
```
grlima: What did it use to do?

littleaj: it's redundant. It's called in `cleanup()`

PR #756, opened 2018-09-30
Set up JaCoCo Code Coverage, update Gradle wrapper

dhaval24:
This PR sets up JaCoCo code coverage, updates Gradle wrapper to 4.2 (updating beyond this fails native linking), fixes build.gradle file for web module and cleans some imports in tests.


dhaval24:
Currently this creates separete coverage reports. I am experimenting to ultimately have one single  report generated for all projects

core/src/test/java/com/microsoft/applicationinsights/channel/concrete/localforwarder/LocalForwarderModelTransformerUtilityTests.java
```diff
@@ -21,20 +22,14 @@
 import com.microsoft.localforwarder.library.inputs.contracts.SeverityLevel;
 import com.microsoft.localforwarder.library.inputs.contracts.StackFrame;
 import com.microsoft.localforwarder.library.inputs.contracts.Telemetry;
-import org.apache.commons.lang3.RandomStringUtils;
-import org.junit.*;
-import org.mockito.internal.matchers.Null;
-import sun.misc.Perf;
```
littleaj: Thanks for removing the unused imports

web/build.gradle
```diff
@@ -46,6 +46,7 @@ dependencies {
     testCompile group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.1'
     testCompile group: 'org.json', name:'json', version:'20090211'
     testCompile group: 'com.microsoft.azure', name: 'azure-storage', version: '2.1.0'
+    testCompile group: 'com.google.guava', name: 'guava', version: '20.0'
```
littleaj: What's this?

dhaval24: I added this because I could not run individual tests without this

dhaval24:
@littleaj tried to play around a bit after you left, seems like xml is not getting generated somehow by moving to the subprojects. Also it's not picking the specified version of Jacoco. I think that subprojects closure is not being used at all. Its just picking up the default configuration of jacoco. We might for now just keep it into commons-java file and later experiment.

dhaval24:
@littleaj as mentioned, in the comment above, I have reverted your commit to ensure that xml's are being generated. I think we can be satisfied with this for now and later on move experiment more for combined report. Let me know your thoughts.

littleaj:
That makes sense. I guess we'll use the import file for now.

PR #757, opened 2018-10-02
Revert "Set up JaCoCo Code Coverage, update Gradle wrapper"

littleaj:
Reverts Microsoft/ApplicationInsights-Java#756

This implementation breaks the smoke tests. Let's see if we can make it work with what we have


dhaval24:
@littleaj no problem.

dhaval24:
@littleaj we fixed this with the revert to gradle version.

PR #766, opened 2018-10-25
Fix the filter order in starter

dhaval24:
This fixes the order of Filter injection using SpringBoot starter. It allows to capture the correct response code for failed requests.

It turns out that :  Ordered.HIGHEST_PRECEDENCE = Integer.MIN_VALUE and we were adding 10 to it which reduced the priority eventually ;)

@grlima @littleaj could you please review this change. It is small and there is request to make hot fix for this. 

Deployed the changes in the test app and here is the distributed transaction. 

image

### DIFF
--- a/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebMvcAutoConfiguration.java
+++ b/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebMvcAutoConfiguration.java
@@ -69,7 +69,7 @@ public class ApplicationInsightsWebMvcAutoConfiguration {
         FilterRegistrationBean registration = new FilterRegistrationBean();
         registration.setFilter(webRequestTrackingFilter);
         registration.addUrlPatterns("/*");
-        registration.setOrder(Ordered.HIGHEST_PRECEDENCE + 10);
+        registration.setOrder(Ordered.HIGHEST_PRECEDENCE);
         return registration;
     }
 
--- a/azure-application-insights-spring-boot-starter/gradle.properties
+++ b/azure-application-insights-spring-boot-starter/gradle.properties
@@ -1,1 +1,1 @@
-spring.boot.starter.version-number=1.1.0-BETA
+spring.boot.starter.version-number=1.1.1-BETA
--- a/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
+++ b/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
@@ -73,7 +73,8 @@ import org.springframework.core.env.Environment;
         ApplicationInsightsWebModuleConfiguration.class
 })
 @AutoConfigureBefore(name = {
-    "io.micrometer.spring.autoconfigure.export.azuremonitor.AzureMonitorMetricsExportAutoConfiguration"
+    "io.micrometer.spring.autoconfigure.export.azuremonitor.AzureMonitorMetricsExportAutoConfiguration",
+    "com.microsoft.azure.spring.autoconfigure.metrics.AzureMonitorMetricsExportAutoConfiguration"
 })
 public class ApplicationInsightsTelemetryAutoConfiguration {
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 // Project properties
-version=2.2.0
+version=2.2.1
 group=com.microsoft.azure
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 // Project properties
-version=2.2.1
+version=2.2.0
 group=com.microsoft.azure
 
--- a/test/smoke/testApps/SpringBootTest/src/main/java/com/springbootstartertest/controller/TestController.java
+++ b/test/smoke/testApps/SpringBootTest/src/main/java/com/springbootstartertest/controller/TestController.java
@@ -36,4 +36,9 @@ public class TestController {
 		client.trackEvent("EventDataPropertyTest", properties, metrics);
 		return "hello";
 	}
+
+	@GetMapping("/throwsException")
+	public void resultCodeTest() throws Exception {
+		throw new Exception("This is an exception");
+	}
 }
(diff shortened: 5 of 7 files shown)


azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebMvcAutoConfiguration.java
```diff
@@ -69,7 +69,7 @@ public FilterRegistrationBean webRequestTrackingFilterRegistrationBean(WebReques
         FilterRegistrationBean registration = new FilterRegistrationBean();
         registration.setFilter(webRequestTrackingFilter);
         registration.addUrlPatterns("/*");
-        registration.setOrder(Ordered.HIGHEST_PRECEDENCE + 10);
+        registration.setOrder(Ordered.HIGHEST_PRECEDENCE);
```
dhaval24: Ordered.HIGHEST_PRECEDENCE = Integer.MIN_VALUE. Adding 10 to it eventually reduced our filter priority :(

grlima: Does having Highest Precedence mess up other scenarios? It seems @gavlyukovskiy had concerns with doing this when he first introduced the starter in #518 
Specifically, he mentioned it could prevent users or the framework itself from running code before ours. Is it a problem? Can we double check?

dhaval24: @grlima For reference : https://www.tuturself.com/posts/view?menuId=3&postId=1141
If any other filter runs before our Filter, it would just not enable us to capture the right timing information, the correct response code etc. It is critical to have our filter as the first filter to get the correct information. For example - this filter was running after our Filter : https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/filter/OncePerRequestFilter.java
after which the response code in the response object changed.

littleaj: It may be ok to order it as highest as long as it can be changed by the user. @dhaval24 , is this possible?

littleaj:
@dhaval24, have you tested this with spring exception handlers?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebMvcAutoConfiguration.java
dhaval24: @littleaj currently there is no such mechanism in starter, but it would be fairly easy to bring this if needed. However, in case of a conflict it is upto the Spring to decide which order it wants to execute the filters. Also there are multiple filters in Spring with highest precedence. It is fine for them to execute in relative order among themselves. 

The important point here is that our filter executes above any other filter whose priority is less that HIGHEST_ORDER. @grlima this should also explain your concern.

dhaval24:
> @dhaval24, have you tested this with spring error/exception handlers?

@littleaj can you elaborate a bit on this.

dhaval24:
@grlima and @littleaj ok, I did some more research: https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-embedded-container-servlets-filters-listeners

It says to avoid HIGHEST_PRECEDENCE. But I think +10 is making it too late in the chain. May +2 so that it just comes after the metrics filter should be okay.

dhaval24:
Spring’s WebMVC Metrics filter : https://github.com/spring-projects/spring-boot/blob/5a4220c77383741205203b615d40de82b890e153/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcMetricsFilter.java#L166  is responsible to set the status code for the request ultimately. 

The priority of this filter is : Ordered.HIGHEST_PRECEDENCE + 1
Based on SpringBoot docs here: https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-embedded-container-servlets-filters-listeners 

Therefore, if we would not have the Ordered.HIGHEST_PRECEDENCE, it would be impossible to capture the correct response code from the Filter Based Method we have today. We need to be first in the filter chain in one way or the other.

dhaval24:
> Spring’s WebMVC Metrics filter : https://github.com/spring-projects/spring-boot/blob/5a4220c77383741205203b615d40de82b890e153/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcMetricsFilter.java#L166 is responsible to set the status code for the request ultimately.
> 
> The priority of this filter is : Ordered.HIGHEST_PRECEDENCE + 1
> Based on SpringBoot docs here: https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-embedded-container-servlets-filters-listeners
> 
> Therefore, if we would not have the Ordered.HIGHEST_PRECEDENCE, it would be impossible to capture the correct response code from the Filter Based Method we have today. We need to be first in the filter chain in one way or the other.

@gavlyukovskiy please consider the above!! and let me know your thoughts

littleaj:
@dhaval24 [ErrorPageFilter](https://github.com/spring-projects/spring-boot/blob/ce9c053cbfc41f4753830bbb6c55ee7b2487856b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/support/ErrorPageFilter.java) can also set the status

dhaval24:
> @dhaval24 [ErrorPageFilter](https://github.com/spring-projects/spring-boot/blob/ce9c053cbfc41f4753830bbb6c55ee7b2487856b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/support/ErrorPageFilter.java) can also set the status

Sure @littleaj that makes sense. It's precedence is also Ordered.HIGHEST_PRECEDENCE + 1 (as from SpringBoot docs). Therefore I think we would need to have the filter at HIGHEST_PRECEDENCE if we want to stick with the filter based solution.

littleaj:
@dhaval24 regarding error handlers, in j2ee you can specify error pages to handle certain exceptions. These pages can change the result code sent back to the user (which don't make it back to the filter). I was wondering if Spring has the same issue. Spring lets you define error handlers which do the same thing.

littleaj:
@dhaval24 What happens if you define more filters with HIGHEST_PRECEDENCE? Is the order determinate by some other factor?

gavlyukovskiy:
@littleaj order of beans with order value is non deterministic, though there is interface `PriorityOrdered` which is used in exceptional cases and those beans have higher precedence over the beans that don't implement it.

dhaval24:
Thanks @littleaj , I queued a new build.

dhaval24:
https://dev.azure.com/mseng/AppInsights/_build/results?buildId=7619743&view=logs

dhaval24:
@grlima and @littleaj I also verified the correlation and it works good with this change too. 
image

Here is a node app communicating with java app. I think this change is good to go.

PR #769, opened 2018-11-09
added check for open port before creating test server.

littleaj:
This makes the tests a bit more reliable.

### DIFF
--- a/web/src/test/java/com/microsoft/applicationinsights/web/utils/JettyTestServer.java
+++ b/web/src/test/java/com/microsoft/applicationinsights/web/utils/JettyTestServer.java
@@ -26,14 +26,40 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.servlet.FilterMapping;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 
+import java.io.IOException;
+import java.net.ServerSocket;
+
 /**
  * Created by yonisha on 2/3/2015.
  */
 public class JettyTestServer {
     private Server server;
-    private int Min = 1050;
-    private int Max = 15000;
-    private int portNumber = Min + (int)(Math.random() * ((Max - Min) + 1));
+    private final int Min = 1050;
+    private final int Max = 15000;
+    private final int portNumber;
+
+    public JettyTestServer() {
+        // try 256 times for ports
+        int initialPortNumber = Min + (int)(Math.random() * ((Max - Min) + 1));
+        for (int i = 0; i < 256; i++) {
+            ServerSocket ss = null;
+            try {
+                ss = new ServerSocket(initialPortNumber);
+            } catch (IOException e) {
+                System.out.printf("port '%d' in use. Trying next one.%n", initialPortNumber);
+                initialPortNumber++;
+            } finally {
+                if (ss != null) {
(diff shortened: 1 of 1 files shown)


web/src/test/java/com/microsoft/applicationinsights/web/utils/JettyTestServer.java
```diff
@@ -26,14 +26,40 @@
...
+    private final int Max = 15000;
+    private final int portNumber;
+
+    public JettyTestServer() {
+        // try 256 times for ports
+        int initialPortNumber = Min + (int)(Math.random() * ((Max - Min) + 1));
+        for (int i = 0; i < 256; i++) {
+            ServerSocket ss = null;
+            try {
+                ss = new ServerSocket(initialPortNumber);
```
dhaval24: @littleaj I am bit confused! I thought the purpose was to try different ports until we find an open port. If this is true I find 2 concerns:
1. the `initialPortNumber` is created outside for loop and I don't see where it is getting updated in the loop.
2. The loop never breaks after finding the successful port thus always running for 256 times. 
Either, I am fully lost or may be you need to update this.

littleaj: 1. I don't want to generate a random number each time because it could choose the same one twice.
2. lol...yes, it does need a `break`.

web/src/test/java/com/microsoft/applicationinsights/web/utils/JettyTestServer.java
```diff
@@ -26,14 +26,41 @@
...
+
 /**
  * Created by yonisha on 2/3/2015.
  */
 public class JettyTestServer {
     private Server server;
-    private int Min = 1050;
-    private int Max = 15000;
-    private int portNumber = Min + (int)(Math.random() * ((Max - Min) + 1));
+    private final int Min = 1050;
```
dhaval24: @littleaj can you please adhere to camelCase for varable names here. I know you didn't made this variable but since you touched this part of the code better that we can clean up if possible.

web/src/test/java/com/microsoft/applicationinsights/web/utils/JettyTestServer.java
```diff
@@ -26,14 +26,41 @@
...
+    private final int Max = 15000;
+    private final int portNumber;
+
+    public JettyTestServer() {
+        // try 256 times for ports
+        int initialPortNumber = Min + (int)(Math.random() * ((Max - Min) + 1));
```
dhaval24: A standard way / best practice in java to get random integer within the range is this:

```
import java.util.concurrent.ThreadLocalRandom;

// nextInt is normally exclusive of the top value,
// so add 1 to make it inclusive
int randomNum = ThreadLocalRandom.current().nextInt(min, max + 1);
```
Reference: https://stackoverflow.com/questions/363681/how-to-generate-random-integers-within-a-specific-range-in-java 

@littleaj it would be great if you can change this one as bonus!

littleaj: Since this is test code, so I'm not really worried about this.

PR #774, opened 2018-11-21
Performance optimization/use concurrent collection

dhaval24:
This PR is the first step to increase the performance of the SDK by removing the low level locking mechanism to add telemetries in the buffer.

It uses `ConcurrentLinkedQueue` which uses optimized lock free algorithm (http://www.cs.rochester.edu/~scott/papers/1996_PODC_queues.pdf) and is Thread safe.

It also removes the primitive `long` used to check the Telemetry container and instead uses `AtomicLong` so the `fetch()` method can be made lock free.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
@@ -21,18 +21,19 @@
 
 package com.microsoft.applicationinsights.internal.channel.common;
 
-import java.util.*;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
-
 import com.google.common.base.Preconditions;
 import com.microsoft.applicationinsights.internal.channel.TelemetriesTransmitter;
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 import com.microsoft.applicationinsights.internal.util.LimitsEnforcer;
 import com.microsoft.applicationinsights.telemetry.Telemetry;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * The class is responsible for getting instances of {@link com.microsoft.applicationinsights.telemetry.Telemetry}
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
@@ -21,11 +21,12 @@
 
 package com.microsoft.applicationinsights.internal.channel.common;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.google.common.base.Preconditions;
 import com.microsoft.applicationinsights.internal.channel.TelemetriesTransmitter;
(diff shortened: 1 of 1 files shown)


dhaval24:
@reyang can you take a brief look at this ?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -220,12 +234,14 @@ public void flush() {
...
+    private Collection<T> prepareTelemetriesForSend() {
+        generation.incrementAndGet();
 
-        final List<T> readyToBeSent = telemetries;
+        // Create an unmodifiable view of the collection so that the items cannot be edited
+        final Collection<T> readyToBeSent = new ArrayList<>(Collections.unmodifiableCollection(telemetries));
 
-        telemetries = new ArrayList<T>(maxTelemetriesInBatch);
+        // Clear telemetry for next batch
+        telemetries.clear();
```
reyang: This seems to be broken, what if another thread pushed data to telemetries between L241 and L244?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -55,25 +57,32 @@
...
 
-        private final long expectedGeneration;
+        private final AtomicLong expectedGeneration;
 
-        private TelemetryBufferTelemetriesFetcher(long expectedGeneration) {
+        private TelemetryBufferTelemetriesFetcher(AtomicLong expectedGeneration) {
             this.expectedGeneration = expectedGeneration;
         }
 
+        // No need for low-level synchronization. Atomic long cover cover the generation check consistency
```
reyang: "cover cover" - typo

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -55,25 +57,32 @@
...
+            // override equals and hash code
+            if (expectedGeneration.get() != generation.get()) {
+                return Collections.emptyList();
             }
+
+            generation.incrementAndGet();
+
+            // Create an unmodifiable view of the collection so that the items cannot be edited
+            Collection<T> readyToBeSent = new ArrayList<>(Collections.unmodifiableCollection(telemetries));
+
```
reyang: This doesn't seem to be the correct way of using concurrent collections.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
dhaval24: will fix.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
dhaval24: This might be a valid point. I will double check some references.

dhaval24:
@reyang sadly Java out of the box doesn't provide a ConcurrentBag or similar non ordered data structure. The closest one is ConcurrentQueue. It surely has the overhead of LIFO pattern, but I would rather stay to what java provides by default rather than writing my own class for current purposes. Thoughts?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -220,12 +234,14 @@ public void flush() {
...
      * @return The list of {@link Telemetry} instances that are ready to be sent
      */
-    private List<T> prepareTelemetriesForSend() {
-        ++generation;
+    private Collection<T> prepareTelemetriesForSend() {
+        generation.incrementAndGet();
 
-        final List<T> readyToBeSent = telemetries;
+        // Create an unmodifiable view of the collection so that the items cannot be edited
+        final Collection<T> readyToBeSent = new ArrayList<>(Collections.unmodifiableCollection(telemetries));
```
littleaj: You don't need a unmodifiableCollection here.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
littleaj: You don't need unmodifiableCollection here.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
littleaj: Also, this is potentially an unbounded operation.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
littleaj: I think you still want the reference swap, but use `AtomicReference`. Using `clear` could unintentionally remove some items

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -55,25 +57,32 @@
...
-        private TelemetryBufferTelemetriesFetcher(long expectedGeneration) {
+        private TelemetryBufferTelemetriesFetcher(AtomicLong expectedGeneration) {
             this.expectedGeneration = expectedGeneration;
         }
 
+        // No need for low-level synchronization. Atomic long cover cover the generation check consistency
+        // ConcurrentLinkedList handles other parts for collection operation.
         @Override
         public Collection<T> fetch() {
-            synchronized (lock) {
```
littleaj: If we don't lock on the collection anymore, what is it guarding? Should it be removed?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TelemetryBuffer.java
```diff
@@ -55,25 +57,32 @@
...
-        private final long expectedGeneration;
+        private final AtomicLong expectedGeneration;
 
-        private TelemetryBufferTelemetriesFetcher(long expectedGeneration) {
+        private TelemetryBufferTelemetriesFetcher(AtomicLong expectedGeneration) {
             this.expectedGeneration = expectedGeneration;
         }
 
+        // No need for low-level synchronization. Atomic long cover cover the generation check consistency
+        // ConcurrentLinkedList handles other parts for collection operation.
```
littleaj: ConcurrentLinked*Queue*

dhaval24:
@littleaj and @reyang thanks for your comments. Let's hold the PR for now until we have a better / consistent implementation. I will update the code accordingly.

dhaval24:
abandoning this PR for now

Issue #716, opened 2018-08-11
Support W3C distributed tracing format


lmolkova:
[W3C distributed tracing protocol](https://github.com/w3c/distributed-tracing/tree/master/trace_context) is going to replace current `Request-Id` format.

The implementation should be split into several phases
1. Start emitting w3c-compatible operation ids, without supporting W3C headers.
**This phase should be implemented first and phase 2 could follow in several months.**

2. Opt-in support for W3C and Request-Id
Once we are sure p2 works well across different SDKs and don't break customers, we could switch to p3 

3. Supporting W3C enabled by default, in back-compat mode with Request-Id
4. Deprecating and retiring Request-Id (in several years)

# 1. W3C compatible ids
In order to enable customer transition story and backward compatibility, the first step is to start emitting  w3c-compatible ids:
32 [a-z][0-9] lowercase characters (i.e. lowercase guid without dashes or 16 random bytes encoded into hex string). All zeros is considered invalid. 
Example of the valid id: `0af7651916cd43dd8448eb211c80319c`. 

The rest: header names and hierarchy in Request-Id could still be preserved.

If request-id was received from upstream, the SDK should continue to behave the same - propagate and do not attempt to change the format.

So, from Java SDKs perspective, the only change here seems to be  `TelemetryCorrelationUtils.generateRootId` function

https://github.com/Microsoft/ApplicationInsights-Java/blob/master/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TelemetryCorrelationUtils.java#L392

# 2. Supporting W3C (opt-in) and Request-Id. 
there are several existing implementations that could be used as reference: .NET AI SDK, Python OpenCensus.
 
There should be a setting that enables W3C support.

When W3C is enabled:
   1. If incoming request contains `traceparent` only: parse it, set operation, parent and telemetry id accordingly. `traceparent` should be injected into outgoing requests. request-id should be injected in back-compat format `|traceId.spanId.` so that older SDKs could understand it.
   2. If incoming request contains only `Request-Id`, new traceId and spanId should be started, Request-Id value should be used as a parent. Root part of Request-Id should be stored in custom dimension on the request telemetry if root part is different from traceId. On the outgoing request side, Request-Id should be emitted in the `|traceId.spanId.` format
   3. if request contains both: request-id and traceparent, request-id should be ignored, so it becomes the same as p1
   4. if request does not contain any correlation headers, see case p2

lmolkova:
[...]

PR #775, opened 2018-11-28
W3C : Tracecontext classes 

dhaval24:
This PR brings in the required classes for integrating W3C correlation protocol. 
Related #716 

@reyang @lmolkova take a look. 

- [ ] CHANGELOG.md updated

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
@@ -2,12 +2,36 @@ package com.microsoft.applicationinsights.web.internal.correlation.tracecontext;
 
 import com.google.common.annotations.VisibleForTesting;
 import java.util.concurrent.ThreadLocalRandom;
+import org.apache.http.annotation.Experimental;
 
+/**
+ * This class represents the Traceparent data structure based on
+ * @link https://github.com/w3c/trace-context/blob/master/trace_context/HTTP_HEADER_FORMAT.md
+ *
+ * @author Reily Yang
+ * @author Dhaval Doshi
+ */
+@Experimental
 public class Traceparent {
 
+    /**
+     * Version number between range [0,255] inclusive
+     */
     @VisibleForTesting final int version;
+
+    /**
+     * 16 byte trace-id that is used to uniquely identify a distributed trace
+     */
     @VisibleForTesting final String traceId;
+
+    /**
+     * It is a 8 byte ID that represents the caller span
+     */
     @VisibleForTesting final String spanId;
+
+    /**
+     * An 8-bit field that controls tracing flags such as sampling, trace level etc.
+     */
     @VisibleForTesting final int traceFlags;
 
     private Traceparent(int version, String traceId, String spanId, int traceFlags, boolean check) {
@@ -20,12 +44,24 @@ public class Traceparent {
         this.traceFlags = traceFlags;
     }
 
+    /**
+     * The constructor that tries to create Traceparent Object from given version, traceId, spanID
+     * and traceFlags.
+     * @param version
+     * @param traceId
+     * @param spanId
+     * @param traceFlags
+     */
     public Traceparent(int version, String traceId, String spanId, int traceFlags) {
         this(version, traceId != null ? traceId : randomHex(16),
             spanId != null ? spanId : randomHex(8),
             traceFlags, true);
     }
 
+    /**
+     * This constructor creates a new Traceparent object having new traceId.
+     * It should only be used if the call is the starting point of distributed trace.
+     */
     public Traceparent() {
         this(0, randomHex(16), randomHex(8), 0, false);
     }
@@ -42,6 +78,14 @@ public class Traceparent {
         return spanId;
     }
 
+    /**
+     * Validates the given input based on W3C specifications.
+     * @param version
+     * @param traceId
+     * @param spanId
+     * @param traceFlags
+     * @throws IllegalArgumentException
+     */
     private static void validate(int version, String traceId, String spanId, int traceFlags)
         throws IllegalArgumentException {
         if (version < 0 || version > 254) {
@@ -64,12 +108,21 @@ public class Traceparent {
         }
     }
 
-
+    /**
+     * Converts the Traceparent object to header format
+     * Eg: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
+     * @return traceparent
+     */
     @Override
     public String toString() {
         return String.format("%02x-%s-%s-%02x", version, traceId, spanId, traceFlags);
     }
 
+    /**
+     * Helper method to create a random hexadecimal string of n bytes.
+     * @param n
+     * @return n byte hexadecimal string
+     */
     @VisibleForTesting static String randomHex(int n) {
         byte[] bytes = new byte[n];
         ThreadLocalRandom.current().nextBytes(bytes);
@@ -80,6 +133,12 @@ public class Traceparent {
         return sb.toString();
     }
 
+    /**
+     * Helper method to check if a given string of n bytes is hexadecimal
+     * @param s
+     * @param n
+     * @return boolean
+     */
     private static boolean isHex(String s, int n) {
         if (s == null || s.length() == 0) {
             return false;
@@ -100,6 +159,11 @@ public class Traceparent {
         return true;
     }
 
+    /**
+     * Marshals the traceparent from String to Traceparent object
+     * @param s
+     * @return Traceparent
+     */
     public static Traceparent fromString(String s) {
         if (s == null || s.length() == 0) {
             return null;
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
@@ -1,0 +1,125 @@
+package com.microsoft.applicationinsights.web.internal.correlation.tracecontext;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class Traceparent {
+
+    @VisibleForTesting final int version;
+    @VisibleForTesting final String traceId;
+    @VisibleForTesting final String spanId;
+    @VisibleForTesting final int traceFlags;
+
+    private Traceparent(int version, String traceId, String spanId, int traceFlags, boolean check) {
+        if (check) {
+            validate(version, traceId, spanId, traceFlags);
+        }
+        this.version = version;
+        this.traceId = traceId;
+        this.spanId = spanId;
+        this.traceFlags = traceFlags;
+    }
+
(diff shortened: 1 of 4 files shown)


web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Class that represents Tracestate header based on
+ * @link https://github.com/w3c/trace-context/blob/master/trace_context/HTTP_HEADER_FORMAT.md
+ *
+ * According to W3C specification, implementations should pass the tracestate header to downstream
+ * calls if received.
+ *
+ * Implementations can add vendor specific details here.
+ *
+ * @author Reiley Yang
```
reyang: Curious, why do we put author name here?

reyang: Last time I've seen this was when I work on the C++ compiler, at that time there was no "git blame".

dhaval24: Most Java libraries do this :) Its a common pattern in Java projects.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,189 @@
+package com.microsoft.applicationinsights.web.internal.correlation.tracecontext;
+
+import com.google.common.annotations.VisibleForTesting;
```
reyang: better to keep import order consistent with other files

dhaval24: done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.http.annotation.Experimental;
+
+/**
+ * Class that represents Tracestate header based on
+ * @link https://github.com/w3c/trace-context/blob/master/trace_context/HTTP_HEADER_FORMAT.md
+ *
+ * According to W3C specification, implementations should pass the tracestate header to downstream
+ * calls if received.
+ *
+ * Implementations can add vendor specific details here.
```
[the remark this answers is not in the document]
dhaval24: done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,189 @@
...
+            if ('a' <= c && c <= 'f') {
+                continue;
+            }
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * Marshals the traceparent from String to Traceparent object
```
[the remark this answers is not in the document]
dhaval24: done

reyang:
Please merge this to a working branch (e.g. w3c) instead of master.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Helper method to check if a given string of n bytes is hexadecimal
+     * @param s
+     * @param n
+     * @return boolean
+     */
+    private static boolean isHex(String s, int n) {
+        if (s == null || s.length() == 0) {
```
littleaj: One option is to `Long.parseLong(s, 16)` and if it throws NumberFormatException, return false.

Sometimes try/catch can be more performant, but if that's not a consideration at this point, I wouldn't worry about it. Also, I'd write a test to confirm prove it's more performant if you do choose that method.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    public static Traceparent fromString(String s) {
+        if (s == null || s.length() == 0) {
+            return null;
+        }
+        String[] arr = s.split("-");
+        if (arr.length != 4) {
+            return null;
+        }
+        if (!isHex(arr[0], 2)) {
```
littleaj: Depending on the usage here, you should consider replacing the `isHex` checks and instead assign the `Character.digit` calls to local variables, then check for `-1` since the `.digit` method is doing these checks already.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Class that represents Tracestate header based on
+ *
+ * @author Reiley Yang
+ * @author Dhaval Doshi
+ * @link https://github.com/w3c/trace-context/blob/master/trace_context/HTTP_HEADER_FORMAT.md
+ *
+ * Implementations can add vendor specific details here.
+ */
+@Experimental
```
littleaj: Is something looking for these annotations?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * It is a 8 byte ID that represents the caller span
+     */
+    @VisibleForTesting
+    final String spanId;
+
+    /**
+     * An 8-bit field that controls tracing flags such as sampling, trace level etc.
+     */
+    @VisibleForTesting
```
littleaj: Perhaps `@VisibleForTesting` isn't necessary since there are public accessors.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+            throw new IllegalArgumentException("invalid traceId");
+        }
+        if (!isHex(spanId, 16)) {
+            throw new IllegalArgumentException("invalid spanId");
+        }
+        if (spanId.equals("0000000000000000")) {
+            throw new IllegalArgumentException("invalid spanId");
+        }
+        if (traceFlags < 0 || traceFlags > 255) {
+            throw new IllegalArgumentException("traceFlags must be within range [0, 255]");
```
littleaj: Either the error message is wrong or the condition is wrong: `[0, 255]` means between `0` and `255`, inclusive.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+        return spanId;
+    }
+
+    /**
+     * Validates the given input based on W3C specifications.
+     */
+    private static void validate(int version, String traceId, String spanId, int traceFlags)
+        throws IllegalArgumentException {
+        if (version < 0 || version > 254) {
+            throw new IllegalArgumentException("version must be within range [0, 255)");
```
littleaj: Either the condition or the message is incorrect: `[0, 255)` means between 0 inclusive and 255 exclusive.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+        this.version = version;
+        this.traceId = traceId;
+        this.spanId = spanId;
+        this.traceFlags = traceFlags;
+    }
+
+    /**
+     * The constructor that tries to create Traceparent Object from given version, traceId, spanID
+     * and traceFlags.
```
littleaj: It may be good to include what happens when `traceId` or `spanId` is null.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
littleaj: nvm...I read the condition wrong.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
littleaj: I misread the conditional, this is correct.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
dhaval24: No I have marked this just because, ultimately this classes would be part of tracecontext libraray under different package name. Therefore, I just want to ensure no one depends on it.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
dhaval24: I think the isHex() method covers this check. It will throw an exception, when they are null.

dhaval24:
@littleaj as this is classes are not perhaps the final version, and there will be some refinements, merging this branch into feature branch and we shall iterate on it.

littleaj:
@dhaval24 sounds good

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    public Traceparent(int version, String traceId, String spanId, int traceFlags) {
```
lmolkova: `int traceFlags` why int? it's byte

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Traceparent.java
```diff
@@ -0,0 +1,185 @@
...
+            return null;
+        }
+        if (!isHex(arr[0], 2)) {
+            return null;
+        }
+        if (!isHex(arr[3], 2)) {
+            return null;
+        }
+
+        return new Traceparent(
```
lmolkova: so in some cases, when traceparent is not valid we return null and in some cases we throw. this is inconsistent. I'd prefer we never throw. 
The problem is that if you've got bad traceparent - it's no event your fault and you cannot do anything about it. 

Of course you can ak everyone what creates traceparent to catch exceptions and check for  nulls, but it's easier to do this once in traceparent

And if it's not valid return a new one - with brand new values - that's what spec also states

reyang:
This will be removed and replace by the actual implementation :)

PR #776, opened 2018-11-29
Initial integration of W3C protocol for incoming request

dhaval24:
Initial implementation for integration of W3C protocol for Incoming Request. 
Related #775 #716 

Please note this is not a complete PR yet. There are some tests which are missing and I will add it. Just pushing the code so that, people can take a look and start commenting on approach.

@lmolkova @reyang please review.

- [ ] CHANGELOG.md updated

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
@@ -21,7 +21,9 @@
 
 package com.microsoft.applicationinsights.web.extensibility.modules;
 
+import com.microsoft.applicationinsights.web.internal.correlation.TraceContextCorrelation;
 import java.util.Date;
+import java.util.Map;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
@@ -48,9 +50,33 @@ public class WebRequestTrackingTelemetryModule implements WebTelemetryModule, Te
     private TelemetryClient telemetryClient;
     private boolean isInitialized = false;
 
+    private boolean isW3CEnabled = false;
+
+    /**
+     * Field to indicate if W3C tracing protocol is enabled.
+     */
+    private final String W3C_CONFIGURATION_PARAMETER = "W3CEnabled";
+
     // endregion Members
 
     // region Public
+
+    public WebRequestTrackingTelemetryModule() {
+
+    }
+
+    /**
+     * Ctor that parses incoming configuration.
+     * @param configurationData
+     */
+    public WebRequestTrackingTelemetryModule(Map<String, String> configurationData) {
+        assert configurationData != null;
+
+        if (configurationData.containsKey(W3C_CONFIGURATION_PARAMETER)) {
+            isW3CEnabled = Boolean.valueOf(configurationData.get(W3C_CONFIGURATION_PARAMETER));
+        }
+    }
+
 
     /**
      * Begin request processing.
@@ -94,7 +120,13 @@ public class WebRequestTrackingTelemetryModule implements WebTelemetryModule, Te
 
             // Look for cross-component correlation headers and resolve correlation ID's
             HttpServletResponse response = (HttpServletResponse) res;
-            TelemetryCorrelationUtils.resolveCorrelation(request, response, telemetry);
+            if (isW3CEnabled) {
+                TraceContextCorrelation.resolveCorrelation(request, response, telemetry);
+            } else {
+                // Default correlation experience
+                TelemetryCorrelationUtils.resolveCorrelation(request, response, telemetry);
+            }
+
 
         } catch (Exception e) {
             String moduleClassName = this.getClass().getSimpleName();
@@ -132,7 +164,11 @@ public class WebRequestTrackingTelemetryModule implements WebTelemetryModule, Te
             telemetry.setDuration(new Duration(endTime - context.getRequestStartTimeTicks()));
             
             String instrumentationKey = this.telemetryClient.getContext().getInstrumentationKey();
-            TelemetryCorrelationUtils.resolveRequestSource((HttpServletRequest) req, telemetry, instrumentationKey);
+            if (isW3CEnabled) {
+                TraceContextCorrelation.resolveRequestSource((HttpServletRequest) req, telemetry, instrumentationKey);
+            } else {
+                TelemetryCorrelationUtils.resolveRequestSource((HttpServletRequest) req, telemetry, instrumentationKey);
+            }
 
             telemetryClient.track(telemetry);
         } catch (Exception e) {
--- a/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
@@ -21,6 +21,7 @@
 
 package com.microsoft.applicationinsights.web.extensibility.modules;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.microsoft.applicationinsights.web.internal.correlation.TraceContextCorrelation;
 import java.util.Date;
 import java.util.Map;
@@ -50,7 +51,7 @@ public class WebRequestTrackingTelemetryModule implements WebTelemetryModule, Te
     private TelemetryClient telemetryClient;
     private boolean isInitialized = false;
 
-    private boolean isW3CEnabled = false;
+    @VisibleForTesting boolean isW3CEnabled = false;
 
     /**
      * Field to indicate if W3C tracing protocol is enabled.
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
@@ -104,7 +104,7 @@ public class TraceContextCorrelation {
             addTracestateInResponseHeader(response);
 
         } catch (Exception e) {
-            InternalLogger.INSTANCE.trace("unable to perform correlation :%s", ExceptionUtils.
+            InternalLogger.INSTANCE.error("unable to perform correlation :%s", ExceptionUtils.
                 getStackTrace(e));
         }
     }
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
@@ -23,7 +23,9 @@ public class TraceContextCorrelation {
 
     public static final String TRACEPARENT_HEADER_NAME = "traceparent";
     public static final String TRACESTATE_HEADER_NAME = "tracestate";
+    public static final String REQUEST_CONTEXT_HEADER_NAME = "Request-Context";
     public static final String AZURE_TRACEPARENT_COMPONENT_INITIAL = "az";
+    public static final String REQUEST_CONTEXT_HEADER_APPID_KEY = "appId";
 
     /**
      * Private constructor as we don't expect to create an object of this class.
@@ -58,11 +60,11 @@ public class TraceContextCorrelation {
             }
 
 
-            String traceId = request.getHeader(TRACEPARENT_HEADER_NAME);
+            String traceparent = request.getHeader(TRACEPARENT_HEADER_NAME);
 
             Traceparent outGoingTraceParent = null;
 
-            if (traceId == null || traceId.isEmpty()) {
+            if (traceparent == null || traceparent.isEmpty()) {
                 outGoingTraceParent = new Traceparent();
 
                 // represents the id of the current request.
@@ -74,7 +76,7 @@ public class TraceContextCorrelation {
                 // set parentId as null because this is the is the originating request
                 requestTelemetry.getContext().getOperation().setParentId(null);
             } else {
-                Traceparent incomingTraceparent = Traceparent.fromString(traceId);
+                Traceparent incomingTraceparent = Traceparent.fromString(traceparent);
 
                 // create outgoing traceParent using the incoming header
                 outGoingTraceParent = new Traceparent(0, incomingTraceparent.getTraceId(),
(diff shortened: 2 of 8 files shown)


web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+     */
+    public static String getAppId() {
+
+        String instrumentationKey = TelemetryConfiguration.getActive().getInstrumentationKey();
+        String appId = InstrumentationKeyResolver.INSTANCE.resolveInstrumentationKey(instrumentationKey);
+
+        //it's possible the appId returned is null (e.g. async task is still pending or has failed). In this case, just
+        //return and let the next request resolve the ikey.
+        if (appId == null) {
+            InternalLogger.INSTANCE.trace("Application correlation Id could not be retrieved (e.g. task may be pending or failed)");
```
littleaj: On the fence: Should this be a warning?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+                getStackTrace(e));
+        }
+    }
+
+    /**
+     * This adds the tracestate in response header so that the Callee can know what is the caller's AppId.
+     */
+    private static void addTracestateInResponseHeader(HttpServletResponse response) {
+
+        if (response.containsHeader(CORRELATION_COMPANION_HEADER_NAME)) {
```
littleaj: What's the rationale here? Are we opting not to overwrite if something else added the correlation header?

If that's the case, we may want to log something or it could be difficult to determine why correlation isn't working when this evals to true.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+            String[] keyval = s.split("=");
+            properties.put(keyval[0], keyval[1]);
+        }
+        return properties;
+    }
+
+    public static String retriveTracestate() {
+        //check if context is null - no correlation will happen
+        if (ThreadContext.getRequestTelemetryContext() == null) {
+            InternalLogger.INSTANCE.warn("No correlation wil happen, Thread context is null");
```
littleaj: typo: "wil" -> "will"

Also, shouldn't it read "RequestTelemetryContext" instead of "Thread context"?

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
```diff
@@ -94,7 +120,13 @@ public void onBeginRequest(ServletRequest req, ServletResponse res) {
 
             // Look for cross-component correlation headers and resolve correlation ID's
             HttpServletResponse response = (HttpServletResponse) res;
-            TelemetryCorrelationUtils.resolveCorrelation(request, response, telemetry);
+            if (isW3CEnabled) {
+                TraceContextCorrelation.resolveCorrelation(request, response, telemetry);
```
lmolkova: we should be backward compatible and if w3c headers are not present, check for legacy headers

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+                return;
+            }
+
+            if (requestTelemetry == null) {
+                InternalLogger.INSTANCE.error("Failed to resolve correlation. requestTelemetry is null.");
+                return;
+            }
+
+
+            String traceId = request.getHeader(CORRELATION_HEADER_NAME);
```
lmolkova:
```suggestion
            String traceparent = request.getHeader(CORRELATION_HEADER_NAME);
```

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * A class that is responsible for performing correlation based on W3C protocol.
+ * This is a clean implementation of W3C protocol and doesn't have the backward
+ * compatibility with AI-RequestId protocol.
+ *
+ * @author Dhaval Doshi
+ */
+public class TraceContextCorrelation {
+
+    public static final String CORRELATION_HEADER_NAME = "traceparent";
```
lmolkova:
```suggestion
    public static final String TRACEPARENT_HEADER_NAME = "traceparent";
```

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+    public static final String CORRELATION_COMPANION_HEADER_NAME = "tracestate";
```
lmolkova:
```suggestion
    public static final String TRACESTATE_HEADER_NAME = "tracestate";
```

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+
+            String traceId = request.getHeader(CORRELATION_HEADER_NAME);
+
+            Traceparent outGoingTraceParent = null;
+
+            if (traceId == null || traceId.isEmpty()) {
+                outGoingTraceParent = new Traceparent();
+
+                // represents the id of the current request.
+                requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
```
lmolkova: |outGoingTraceParent.getTraceId().outGoingTraceParent.getSpanId().

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+                requestTelemetry.getContext().getOperation().setParentId(null);
+            } else {
+                Traceparent incomingTraceparent = Traceparent.fromString(traceId);
+
+                // create outgoing traceParent using the incoming header
+                outGoingTraceParent = new Traceparent(0, incomingTraceparent.getTraceId(),
+                    null, incomingTraceparent.getTraceFlags());
+
+                // set id of this request
+                requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
```
lmolkova: |outGoingTraceParent.getTraceId().outGoingTraceParent.getSpanId().

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+
+                // represents the trace-id of this distributed trace
+                requestTelemetry.getContext().getOperation().setId(outGoingTraceParent.getTraceId());
+
+                // represents the parent-id of this request which is combination of traceId and incoming spanId
+                requestTelemetry.getContext().getOperation().setParentId(outGoingTraceParent.getTraceId() + "-" +
```
lmolkova: |outGoingTraceParent.getTraceId().incomingTraceparent.getSpanId().

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+            String tracestate = request.getHeader(CORRELATION_COMPANION_HEADER_NAME);
+            Tracestate tracestateObject = Tracestate.fromString(tracestate);
+            Map<String, String> tracestatePropertiesMap = getPropertiesMap(tracestateObject);
+
+            // create outbound tracestate to be propagated to downstream calls
+            Tracestate outboundTracestate = createOutboundTracestate(tracestatePropertiesMap, getAppId());
+            ThreadContext.getRequestTelemetryContext().setTracestate(outboundTracestate);
+
+            // Let the callee know the caller's AppId
+            addTracestateInResponseHeader(response);
```
lmolkova: we should not set tracestate in response header

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+                // represents the id of the current request.
+                requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
+
+                // represents the trace-id of this distributed trace
+                requestTelemetry.getContext().getOperation().setId(outGoingTraceParent.getTraceId());
+
+                // set parentId as null because this is the is the originating request
+                requestTelemetry.getContext().getOperation().setParentId(null);
+            } else {
+                Traceparent incomingTraceparent = Traceparent.fromString(traceId);
```
lmolkova: depending on the way you chose to validate - you might need to check incomingTraceparent for null and probably catch exceptions.

If you follow comments I've left in the previous PR (never throw, return brand new traceid and spanid on the invalid parent), this is the right usage, but Traceparent.fromString needs to be fixed to use it like that

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: @lmolkova, what should we pass in the response header to achieve the same functionality? Should we just use the previous header? And why not tracestate?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Yes, this makes sense. incomingTraceparent can be null and in that case we need to create new TraceId and SpanId. I think it's place resides in `Traceparent.fromString()`. As I said, there are tests I need to implement :)

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+
+            // create outbound tracestate to be propagated to downstream calls
+            Tracestate outboundTracestate = createOutboundTracestate(tracestatePropertiesMap, getAppId());
+            ThreadContext.getRequestTelemetryContext().setTracestate(outboundTracestate);
+
+            // Let the callee know the caller's AppId
+            addTracestateInResponseHeader(response);
+
+        } catch (Exception e) {
+            InternalLogger.INSTANCE.trace("unable to perform correlation :%s", ExceptionUtils.
```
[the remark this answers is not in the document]
dhaval24: agreed. Will update.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Trace is good enough here I think. ID should be resolved eventually. It's just that async task might take some time to complete.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Naming suggestion makes sense. Will update.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: I know these are backward compatibility comments. Understand it's importance. The question I have is, should we make changes in UX to understand correct format too or should SDK set it as UX expects? 
My understanding is that the protocol is what defines the contract between UX and SDK.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
lmolkova: W3C does not define response headers. we should continue responding with Request-Context to everything

dhaval24:
@lmolkova @reyang To clarify, does Python and Go also set the request Id and Parent Id in the same format as .NET?

lmolkova:
>  To clarify, does Python and Go also set the request Id and Parent Id in the same format as .NET?

LocalForwarder does.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: That is correct. However looking here, I don't see any case when this would happen. As this is the only method which sets the Header in the response. I think it is safe to remove.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: so the `ThreadContext` is the wrapper class which stores the RequestTelemetry in InherritableThreadLocal. When the threads change we want to retrive the context from ThreadLocal.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: That makes sense. I will update this one. The question comes in mind then what happens when one wants to use OpenCensus SDKs and just change the end-point to go to AI endpoint. This wouldn't work definitely. Do we want to introduce this legacy in the brand new implementation. 
I am aware .NET has done this! @reyang what are your thoughts here? 
PS : The work is minimal, but it's like once we ship something it's difficult to revert back.

dhaval24:
> > To clarify, does Python and Go also set the request Id and Parent Id in the same format as .NET?
> 
> LocalForwarder does.

Ok, however I don't think most customers even use Local Forwarder as of today (based on telemetry). The direct channel should be as compatible as Local Forwarder. 

We can do whatever we want to do and support the UX as of now. However my $0.02 would be, please lets release this as experimental work. UX needs to adhere to the common protocol. That is how essentially systems work. In the final version I would like to see a clean implementation of this protocol which can be treated as a reference.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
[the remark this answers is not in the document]
dhaval24: done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
littleaj: I know, but the ThreadContext (call the static instance `T`) provides the RequestTelemetryContext (call the instance `R`)... `T.get -> R` and this checks if `R` is null but the messages says `T` is null.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,288 @@
...
+     * @param request
+     * @param response
+     * @param requestTelemetry
+     */
+    public static void resolveCorrelation(HttpServletRequest request, HttpServletResponse response,
+        RequestTelemetry requestTelemetry) {
+
+        try {
+            if (request == null) {
+                InternalLogger.INSTANCE.error("Failed to resolve correlation. request is null.");
```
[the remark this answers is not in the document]
dhaval24: I am not fully convinced. I think if the websdk is enabled than correlation should be there by default in every request. Further, if request object is somehow null I think it's justified to have error level log.

dhaval24:
Closing this PR in favor of #782 and #785

PR #777, opened 2018-11-29
Fix test assertions

### LINKED ISSUE #440: Several tests' asserts have expected/actual parameters reversed
JUnit's `assertEquals` method has the form `assertEquals(X expected, X actual)`, but most of the tests here have expected and actual in the wrong parameter slot. The result _should_ be the same (since equality is symmetric), but the test output will be incorrect. For example, assertEquals produces the output "java.lang.AssertionError: expected: <E>, but was: <A>" where _E_ is the value you expect and _A_ is the actual value read from the code under test. Current tests will have those values reversed which can cause confusion while debugging.

littleaj:
Fix #440.

Addresses the `assertEquals` parameters which were reversed. Also, there are some other test improvements.


agent/src/test/java/com/microsoft/applicationinsights/agent/internal/agent/EnterExitClassVisitorTest.java
```diff
@@ -21,10 +21,19 @@
...
+
+import com.microsoft.applicationinsights.agent.internal.coresync.InstrumentedClassType;
+import org.junit.*;
+import org.mockito.Mockito;
+
+
 public final class EnterExitClassVisitorTest {
+
+    // TODO fix these tests
+//    @Ignore
```
littleaj: I made an attempt to fix this test, but it was a time sink. Instead, I added a TODO comment for later.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilderTest.java
```diff
@@ -46,7 +46,7 @@ public void testForbiddenSection() throws IOException {
         Set<String> excludedPrefixes = configuration.getExcludedPrefixes();
 
         assertNotNull(excludedPrefixes);
-        assertEquals(excludedPrefixes.size(), 2);
+        assertEquals(2, excludedPrefixes.size());
```
littleaj: oops, this is not correct.

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilderTest.java
```diff
@@ -62,18 +62,18 @@ public void testClassesConfiguration() throws IOException {
         AgentConfiguration configuration = testConfiguration("InstrumentationTest.xml");
         Map<String, ClassInstrumentationData> classes = configuration.getRequestedClassesToInstrument();
         assertNotNull(classes);
-        assertEquals(classes.size(), 2);
+        assertEquals(2, classes.size());
```
littleaj: also, incorrect

agent/src/test/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/RuntimeExceptionDeciderTest.java
```diff
@@ -96,6 +96,6 @@ public void testConfDataEnabledWithSuppressedDataNotRelevant() {
 
         assertNotNull(result);
         assertTrue(result.valid);
-        assertEquals(result.stackSize, 1);
+        assertEquals(1, result.stackSize);
```
littleaj: and again...incorrect

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
```diff
@@ -107,13 +107,13 @@ private static void performTest(
...
         Collection<Object> objects = result.get(key);
         assertNotNull(objects);
-        assertEquals(objects.size(), 1);
+        assertEquals(1, objects.size());
         double value = 0.0;
         for (Object obj : objects) {
             try {
                 value += Double.parseDouble(String.valueOf(obj));
             } catch (Exception e) {
-                assertFalse(true);
```
littleaj: Please, do not do this...use `Assert.fail` if you intend to fail the test when it reaches this line.

PR #779, opened 2018-12-04
Implement W3C tracestate

reyang:
@dhaval24

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
@@ -1,5 +1,9 @@
 package com.microsoft.applicationinsights.web.internal.correlation.tracecontext;
 
+import java.util.LinkedHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import org.apache.http.annotation.Experimental;
 
 /**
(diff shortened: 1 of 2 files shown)


web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -11,27 +15,97 @@
...
  * Implementations can add vendor specific details here.
  */
-@Experimental
 public class Tracestate {
 
     /**
-     * Field to capture tracestate header
+     * Internal representation of the tracestate
+     */
+    private LinkedHashMap<String, String> internalList = new LinkedHashMap<String, String>(32);
```
dhaval24: A good practice to follow is use Interface. 
`Map<String, String> internalList = new LinkedHashMap<>(32)`. You also need not specify the types in diamond operator twice.  By the way, LinkedHashMap is used to maintain insertion order. Do you need to maintain that order here?

dhaval24: Is 32 the max limit according to spec?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -11,27 +15,97 @@
...
+    public Tracestate(String input) {
+        if (input == null) {
+            throw new IllegalArgumentException("input is null");
+        }
+        
+        String[] values = DELIMITER_FORMAT_RE.split(input);
+        for (String item : values) {
+            Matcher m = MEMBER_FORMAT_RE.matcher(item);
+            if (!m.find()) {
+                throw new IllegalArgumentException(String.format("invalid string %s in tracestate", item));
```
dhaval24: In general, within Application Insights the practice that was followed was to not throw only, but do try - catch and Log. So that we don't break users application. Do we want to throw here (for generic library purpose and use try catch in implementation)?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
reyang: Yes, https://github.com/w3c/trace-context/blob/master/trace_context/HTTP_HEADER_FORMAT.md#header-value
```
There can be a maximum of 32 list-members in a list.
```

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
```diff
@@ -11,27 +15,97 @@
...
+        if (internalList.size() > 32) {
+            throw new IllegalArgumentException("cannot have more than 32 key-value pairs");
+        }
+        internalString = toInternalString();
     }
 
+    /**
+     * Ctor that creates a tracestate object from a parent one
+     */
+    public Tracestate(Tracestate parent, String key, String value) {
```
dhaval24: For application insights case we need another helper method also - which can basically swap the value of the `azure key`. Should this helper reside here? If not we would need to have almost similar method in the AI implementation.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
reyang: From class implementation perspective, it makes sense to throw if invalid argument is passed in.
For integration, we need to swallow the exception and ignore the broken header according to the W3C spec.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/tracecontext/Tracestate.java
reyang: Can we use ```new Tracestate(tracestate, "key", "new value")```?

dhaval24: should work

PR #782, opened 2018-12-07
W3C Integration

### LINKED ISSUE #716: Support W3C distributed tracing format

dhaval24:
Fix #716 

This PR is super set of all previous PRs which completes the integration of W3C for both incoming and outbound calls.

Non Goals of this PR: Backward Compatibility with AI proprietary correlation protocol. 

Related #776 #775 

Enable W3C:

Incoming Side - 

J2EE Apps add the following to the `<TelemetryModules>` tag inside ApplicationInsights.xml

```<Add type="com.microsoft.applicationinsights.web.extensibility.modules.WebRequestTrackingTelemetryModule"  W3CEnabled ="true"/>```

Springboot apps add the following property:
`azure.application-insights.web.w3c=true`

Outgoing Side - 

Add the following to AI-Agent.xml

```
<Instrumentation>
        <BuiltIn enabled="true">
            <HTTP enabled="true" W3C="true"/>
        </BuiltIn>
    </Instrumentation>
```

Remember : you need both inbound and outbound settings to be the same for correlation to work properly.

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -62,6 +62,11 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
         deltaInNS = this.newLocal(Type.LONG_TYPE);
         mv.visitVarInsn(LSTORE, deltaInNS);
 
+        // This byte code instrumentation is responsible for injecting legacy AI correlation headers which include
+        // Request-Id and Request-Context and Correlation-Context. By default this headers are propagated if W3C
+        // is turned off. Please refer to generateChildDependencyId(), retrieveCorrelationContext(),
+        // retrieveApplicationCorrelationId() from TelemetryCorrelationUtils class for details on how these headers
+        // are created.
         if (!isW3CEnabled) {
             // generate child ID
             mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TelemetryCorrelationUtils", "generateChildDependencyId", "()Ljava/lang/String;", false);
@@ -95,7 +100,11 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
             mv.visitVarInsn(ALOAD, appCorrelationId);
             mv.visitMethodInsn(INVOKEINTERFACE, "org/apache/http/HttpRequest", "addHeader", "(Ljava/lang/String;Ljava/lang/String;)V", true);
         } else {
-            // generate child ID
+            // If W3C is enabled, we propagate Traceparent, Tracecontext headers
+            // to enable correlation. Please refer to generateChildDependencyTraceparent(), generateChildDependencyTraceparent()
+            // from TraceContextCorrelation class on how to generate this headers.
+
+            // generate child Traceparent
             mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation",
                 "generateChildDependencyTraceparent", "()Ljava/lang/String;", false);
             traceparent = this.newLocal(Type.getType(Object.class));
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -54,6 +54,7 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
     private int correlationContextLocal;
     private int appCorrelationId;
     private int tracestate;
+    private int traceparent;
 
     @Override
     public void onMethodEnter() {
@@ -97,6 +98,12 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
             // generate child ID
             mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation",
                 "generateChildDependencyTraceparent", "()Ljava/lang/String;", false);
+            traceparent = this.newLocal(Type.getType(Object.class));
+            mv.visitVarInsn(ASTORE, traceparent);
+
+            mv.visitVarInsn(ALOAD, traceparent);
+            mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation",
+                "createChildIdFromTraceparentString", "(Ljava/lang/String;)Ljava/lang/String;", false);
             childIdLocal = this.newLocal(Type.getType(Object.class));
             mv.visitVarInsn(ASTORE, childIdLocal);
 
@@ -110,7 +117,7 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
             // 2 because the 1 is the method being instrumented
             mv.visitVarInsn(ALOAD, 2);
             mv.visitLdcInsn("traceparent");
-            mv.visitVarInsn(ALOAD, childIdLocal);
+            mv.visitVarInsn(ALOAD, traceparent);
             mv.visitMethodInsn(INVOKEINTERFACE, "org/apache/http/HttpRequest", "addHeader", "(Ljava/lang/String;Ljava/lang/String;)V", true);
 
             mv.visitVarInsn(ALOAD, tracestate);
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -119,7 +119,7 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
 
             mv.visitVarInsn(ALOAD, 2);
             mv.visitLdcInsn("tracestate");
-            mv.visitVarInsn(ALOAD, appCorrelationId);
+            mv.visitVarInsn(ALOAD, tracestate);
 
             mv.visitMethodInsn(INVOKEINTERFACE, "org/apache/http/HttpRequest", "addHeader", "(Ljava/lang/String;Ljava/lang/String;)V", true);
 
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -34,14 +34,17 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
 
     private final static String FINISH_DETECT_METHOD_NAME = "httpMethodFinished";
     private final static String FINISH_METHOD_RETURN_SIGNATURE = "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;IJ)V";
+    private final boolean isW3CEnabled;
 
     public HttpClientMethodVisitor(int access,
                                    String desc,
                                    String owner,
                                    String methodName,
                                    MethodVisitor methodVisitor,
-                                   ClassToMethodTransformationData additionalData) {
+                                   ClassToMethodTransformationData additionalData,
+                                   boolean isW3CEnabled) {
         super(access, desc, owner, methodName, methodVisitor, additionalData);
+        this.isW3CEnabled = isW3CEnabled;
     }
 
     private int deltaInNS;
@@ -50,6 +53,7 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
     private int childIdLocal;
     private int correlationContextLocal;
     private int appCorrelationId;
+    private int tracestate;
 
     @Override
     public void onMethodEnter() {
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

**Comment by dhaval24** [comment]
Results of W3C Test harness on sample service 

image

File: azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebModuleConfiguration.java [hunk 1]
```diff
@@ -59,7 +59,9 @@
     @Bean
     @ConditionalOnProperty(value = "azure.application-insights.default-modules.WebRequestTrackingTelemetryModule.enabled", havingValue = "true", matchIfMissing = true)
     WebRequestTrackingTelemetryModule webRequestTrackingTelemetryModule() {
-        return new WebRequestTrackingTelemetryModule();
+        WebRequestTrackingTelemetryModule w = new WebRequestTrackingTelemetryModule();
+        w.isW3CEnabled = true;
```
lmolkova: does it enable w3c by default?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,414 @@
...
+
+            Traceparent incomingTraceparent = null;
+            Traceparent outGoingTraceParent = null;
+
+            // If no incoming traceparent or multiple traceparent create a  new one.
+            if (traceparentList.size() != 1) {
+                outGoingTraceParent = new Traceparent();
+
+                // represents the id of the current request.
+                requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
```
lmolkova: this format is not correct. |traceid.spanid. is the right one

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,414 @@
...
+                    InternalLogger.INSTANCE.error(String.format("Received invalid traceparent header with exception %s, "
+                        + "distributed trace might be broken", ExceptionUtils.getStackTrace(e)));
+                } finally {
+                    if (incomingTraceparent == null) {
+
+                        // Invalid incoming traceparent. Create a new outgoing traceparent
+                        outGoingTraceParent = new Traceparent();
+                    }
+                    // set id of this request
+                    requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
```
lmolkova: |traceid.spanid.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,414 @@
...
+
+                    // represents the trace-id of this distributed trace
+                    requestTelemetry.getContext().getOperation().setId(outGoingTraceParent.getTraceId());
+
+                    if (incomingTraceparent != null) {
+                        // represents the parent-id of this request which is combination of traceparent and incoming spanId
+                        requestTelemetry.getContext().getOperation().setParentId(outGoingTraceParent.getTraceId() + "-" +
+                            incomingTraceparent.getSpanId());
```
lmolkova: |traceid.spanid.

web/src/test/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelationTests.java
```diff
@@ -0,0 +1,251 @@
...
+
+        //validate we have generated proper ID's
+        Assert.assertNotNull(requestTelemetry.getId());
+
+        Assert.assertTrue(requestTelemetry.getId().startsWith(t.getTraceId()));
+
+        //validate operation context ID's
+        OperationContext operation = requestTelemetry.getContext().getOperation();
+        Assert.assertEquals(t.getTraceId(), operation.getId());
+        Assert.assertEquals(t.getTraceId() + "-" + t.getSpanId(), operation.getParentId());
```
lmolkova: |traceid.spanid.

web/src/test/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelationTests.java
```diff
@@ -0,0 +1,251 @@
...
+        TraceContextCorrelation.resolveCorrelation(request, response, requestTelemetry);
+
+        //validate operation context ID's - there is no parent, so parentId should be null, traceId
+        // is newly generated and request.Id is based on new traceId-spanId
+        OperationContext operation = requestTelemetry.getContext().getOperation();
+
+        Assert.assertNotNull(requestTelemetry.getId());
+
+        // First trace will have it's own spanId also.
+        Assert.assertTrue(requestTelemetry.getId().startsWith(operation.getId()+"-"));
```
lmolkova: |traceid.spanid.

web/src/test/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelationTests.java
```diff
@@ -0,0 +1,251 @@
...
+        //run
+        TraceContextCorrelation.resolveCorrelation(request, response, requestTelemetry);
+
+        //validate operation context ID's - there is no parent, so parentId should be null, traceId
+        // is newly generated and request.Id is based on new traceId-spanId
+        OperationContext operation = requestTelemetry.getContext().getOperation();
+
+        Assert.assertNotNull(requestTelemetry.getId());
+        // First trace will have it's own spanId also.
+        Assert.assertTrue(requestTelemetry.getId().startsWith(operation.getId()+"-"));
```
lmolkova: |traceid.spanid.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilder.java
```diff
@@ -223,7 +224,10 @@ private void setBuiltInInstrumentation(AgentConfigurationDefaultImpl agentConfig
         new ConfigRuntimeExceptionDataBuilder().setRuntimeExceptionData(builtInElement, builtInConfigurationBuilder);
 
         nodes = builtInElement.getElementsByTagName(HTTP_TAG);
-        builtInConfigurationBuilder.setHttpEnabled(XmlParserUtils.getEnabled(XmlParserUtils.getFirst(nodes), HTTP_TAG));
+        Element httpElement = XmlParserUtils.getFirst(nodes);
+        boolean isW3CEnabled = XmlParserUtils.w3cEnabled(httpElement, W3C_ENABLED);
+        System.out.println("!----------"+isW3CEnabled+"------!");
```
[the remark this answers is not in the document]
dhaval24: Yes, will remove it

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsWebModuleConfiguration.java
dhaval24: Good catch. I need to fix this for Springboot portion. Will update.

dhaval24: updated

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilder.java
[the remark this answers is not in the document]
dhaval24: Removed

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -74,10 +74,11 @@ public static void resolveCorrelation(HttpServletRequest request, HttpServletRes
                 outGoingTraceParent = new Traceparent();
 
                 // represents the id of the current request.
-                requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
+                requestTelemetry.setId("|" + outGoingTraceParent.getTraceId() + "." + outGoingTraceParent.getSpanId()
+                + ".");
 
                 // represents the trace-id of this distributed trace
-                requestTelemetry.getContext().getOperation().setId(outGoingTraceParent.getTraceId());
+                requestTelemetry.getContext().getOperation().setId("|" + outGoingTraceParent.getTraceId() +".");
```
lmolkova: operation must not have a I and dot.
it must be equal to traceId

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -101,15 +102,16 @@ public static void resolveCorrelation(HttpServletRequest request, HttpServletRes
                         outGoingTraceParent = new Traceparent();
                     }
                     // set id of this request
-                    requestTelemetry.setId(outGoingTraceParent.getTraceId() + "-" + outGoingTraceParent.getSpanId());
+                    requestTelemetry.setId("|" + outGoingTraceParent.getTraceId() + "." + outGoingTraceParent.getSpanId()
+                    + ".");
 
                     // represents the trace-id of this distributed trace
-                    requestTelemetry.getContext().getOperation().setId(outGoingTraceParent.getTraceId());
+                    requestTelemetry.getContext().getOperation().setId("|" + outGoingTraceParent.getTraceId() + ".");
```
lmolkova: outGoingTraceParent.getTraceId()

dhaval24: I see. Okay!

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -397,9 +399,12 @@ public static String generateChildDependencyTraceparent() {
             }
 
             RequestTelemetry requestTelemetry = context.getHttpRequestTelemetry();
+            String traceId = requestTelemetry.getContext().getOperation().getId();
+
+            // get the necessary part as traceId in context is of legacy AI format.
+            traceId = traceId.substring(1, traceId.length() - 1);
```
lmolkova: you don't need it

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -411,4 +416,19 @@ public static String generateChildDependencyTraceparent() {
...
         return null;
     }
+
+    /**
+     * This is helper method to convert traceparent (W3C) format to AI legacy format for supportability
+     * @param traceparent
+     * @return legacy format traceparent
+     */
+    public static String createChildIdFromTraceparentString(String traceparent) {
+        System.out.println(traceparent + "-------");
```
lmolkova: another debug output

dhaval24: Will remove

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -411,4 +416,19 @@ public static String generateChildDependencyTraceparent() {
...
+        assert traceparent != null;
+
+        String[] traceparentArr = traceparent.split("-");
+        assert traceparentArr.length == 4;
+
+        return "|" + traceparentArr[0] + "." + traceparentArr[1] + ".";
```
lmolkova: this cannot work. traceparentArr[0]  is probably version - 00

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: I needed it before as I was appending the "|". Now I don't. Will update.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Yes it should be return "|" + traceparentArr[1] + "." + traceparentArr[2] + "."; 
1 - > traceId
2 - > spanId.

web/src/test/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModuleTests.java
```diff
@@ -286,6 +286,10 @@ public void testCrossComponentCorrelationHeadersAreCapturedWhenW3CTurnedOn() {
 
     }
 
+    private String formatedID(String id) {
```
lmolkova: no need

web/src/test/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelationTests.java
```diff
@@ -248,4 +248,9 @@ public static String getRequestSourceValue(String appId) {
 
         return String.format("cid-v1:%s", appId);
     }
+
+    private String formatedID(String id) {
```
lmolkova: no need

dhaval24:
@lmolkova I refactored `resolveCorrelation()` method to be more readable and debugable. The test harness results still hold. Though, I am trying to test E2E with a .NET app and I am finding hard time to enable W3C even after following all the docs.  https://docs.microsoft.com/en-us/azure/application-insights/application-insights-correlation?toc=/azure/azure-monitor/toc.json

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
```diff
@@ -50,43 +53,87 @@ public HttpClientMethodVisitor(int access,
...
-        mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TelemetryCorrelationUtils", "retrieveApplicationCorrelationId", "()Ljava/lang/String;", false);
-        appCorrelationId = this.newLocal(Type.getType(Object.class));
-        mv.visitVarInsn(ASTORE, appCorrelationId);
-
-        // inject headers
-        mv.visitVarInsn(ALOAD, 2);
-        mv.visitLdcInsn("Request-Id");
-        mv.visitVarInsn(ALOAD, childIdLocal);
-        mv.visitMethodInsn(INVOKEINTERFACE, "org/apache/http/HttpRequest", "addHeader", "(Ljava/lang/String;Ljava/lang/String;)V", true);
+        if (!isW3CEnabled) {
```
lmolkova: nit. It would be great to comment these blocks of IL with actual code it does (e.g. like [this](https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/develop/Src/Web/Web.Shared.Net/Implementation/ExceptionHandlersInjector.cs#L85)). Simplifies life of those who come after you including yourself

dhaval24: Actually this is super important. I will add these comments.

dhaval24: By the way, I thought .NET didn't had byte code anymore, but the link you gave has IL code there :) Do we still use IL code even in .NET ?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,448 @@
...
+                return;
+            }
+
+            if (requestTelemetry == null) {
+                InternalLogger.INSTANCE.error("Failed to resolve correlation. requestTelemetry is null.");
+                return;
+            }
+
+            Traceparent incomingTraceparent = processIncomingTraceparent(request);
+            Traceparent outGoingTraceParent = createOutgoingTraceparent(incomingTraceparent);
```
lmolkova: nit. this is not outgoing request thing. It describes incoming request processing. And incomingTraceparent is actually parent thing.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,448 @@
...
+import org.apache.commons.lang3.exception.ExceptionUtils;
+
+/**
+ * A class that is responsible for performing correlation based on W3C protocol.
+ * This is a clean implementation of W3C protocol and doesn't have the backward
+ * compatibility with AI-RequestId protocol.
+ *
+ * @author Dhaval Doshi
+ */
+public class TraceContextCorrelation {
```
lmolkova: I wonder, why it is in web packge? it's needed for outgoing requests as well that have nothing to do with web.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,448 @@
...
+     * @return Outbound Traceparent
+     */
+    public static String generateChildDependencyTraceparent() {
+        try {
+
+            RequestTelemetryContext context = ThreadContext.getRequestTelemetryContext();
+
+            //check if context is null - no correlation will happen
+            if (context == null) {
+                InternalLogger.INSTANCE.warn("No Correlation will happen, Thread context is null while generating child dependency");
```
lmolkova: why? if there is no parent context, we should still generate a new traceparent and add to the headers. This would allow us to trace dependency to the request on the next service and build at least appmap.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
lmolkova: yes. we do.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: The previous correlation worked this way atleast, assuming that there is one incoming request. This is a valid scenario though. A worker role can initiate an outbound call. I will update this one.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: In general, Java SDK has tried to blur the distinction within core and web sdk and make the life simpler for users. If you look at ApplicationInsights-SpringBoot-Starter, it usually pulls down both core and web jars, and then smartly detect if the app is MVC type than only enable web modules. But classes are generally available on classpath. So this is okay I think. Also correlation has a primary relationship with requests in general and how they propogate in distributed systems.

However, I think Tracecontext classes can be well placed in core and implementation in web. However, I am aware that this classes will eventually go out of sdk and be replaced with a library so I won't worry about this atm.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Sure thing! Will update.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Done

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
dhaval24: Done

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
dhaval24: Done

dhaval24:
image


@lmolkova @littleaj with above all possible issues are fixed. I can see correlation working completely with W3C. Can you both sign off so I can merge this PR today?

dhaval24:
https://dev.azure.com/mseng/AppInsights/_build/results?buildId=7964109 (passed CI)

PR #785, opened 2018-12-13
W3C Final implementation

dhaval24:
Final W3C implementation

#775 #782 #776 #716 

@lmolkova you can skip this review if you want as you already reviewed the old PR.
@littleaj can you take a brief overview?

**Enable W3C:**

**Incoming Side -** 

**J2EE Apps add the following to the `<TelemetryModules>` tag inside ApplicationInsights.xml**

```xml
<Add type="com.microsoft.applicationinsights.web.extensibility.modules.WebRequestTrackingTelemetryModul>
   <Param name = " W3CEnabled" value ="true"/>
   <Param name ="enableW3CBackCompat" value = "true" />

</Add>
```

**Springboot apps add the following property**:
`azure.application-insights.web.enable-W3C=true`
`azure.application-insights.web.enable-W3C-backcompat-mode=true`

**Outgoing Side -** 

Add the following to AI-Agent.xml

```xml
<Instrumentation>
        <BuiltIn enabled="true">
            <HTTP enabled="true" W3C="true" enableW3CBackCompat="true"/>
        </BuiltIn>
    </Instrumentation>
```

Note: Please note that Backward compatibility mode is enabled by default and the enableW3CBackCompat parameter is optional and should be used only when you want to turn it off.

Ideally this would be the case when all your services have been updated to newer version of SDKs supporting W3C protocol. It is highly recommended to move to newer version of SDKs with W3C support as soon as possible. 

**Please ensure that both incoming and outgoing configurations are exactly same, or else it's very likely that correlation won't work properly.**
i.e W3C and bckportMode on on incoming side -> Then W3C and backport mode should on on outgoing side as well.

- [x] CHANGELOG.md updated

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -123,7 +123,7 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
             mv.visitVarInsn(ASTORE, tracestate);
 
             // inject headers
-            // 2 because the 1 is the method being instrumented
+            // load 2nd variable because the 1st is the method being instrumented
             mv.visitVarInsn(ALOAD, 2);
             mv.visitLdcInsn("traceparent");
             mv.visitVarInsn(ALOAD, traceparent);
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
@@ -34,14 +34,17 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
 
     private final static String FINISH_DETECT_METHOD_NAME = "httpMethodFinished";
     private final static String FINISH_METHOD_RETURN_SIGNATURE = "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;IJ)V";
+    private final boolean isW3CEnabled;
 
     public HttpClientMethodVisitor(int access,
                                    String desc,
                                    String owner,
                                    String methodName,
                                    MethodVisitor methodVisitor,
-                                   ClassToMethodTransformationData additionalData) {
+                                   ClassToMethodTransformationData additionalData,
+                                   boolean isW3CEnabled) {
         super(access, desc, owner, methodName, methodVisitor, additionalData);
+        this.isW3CEnabled = isW3CEnabled;
     }
 
     private int deltaInNS;
@@ -50,6 +53,8 @@ public final class HttpClientMethodVisitor extends AbstractHttpMethodVisitor {
     private int childIdLocal;
     private int correlationContextLocal;
     private int appCorrelationId;
+    private int tracestate;
+    private int traceparent;
 
     @Override
     public void onMethodEnter() {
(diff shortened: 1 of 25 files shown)


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/HttpClientMethodVisitor.java
```diff
@@ -50,43 +53,96 @@ public HttpClientMethodVisitor(int access,
...
+            correlationContextLocal = this.newLocal(Type.getType(Object.class));
+            mv.visitVarInsn(ASTORE, correlationContextLocal);
+
+            // retrieve request context
+            mv.visitMethodInsn(INVOKESTATIC, "com/microsoft/applicationinsights/web/internal/correlation/TelemetryCorrelationUtils", "retrieveApplicationCorrelationId", "()Ljava/lang/String;", false);
+            appCorrelationId = this.newLocal(Type.getType(Object.class));
+            mv.visitVarInsn(ASTORE, appCorrelationId);
+
+            // inject headers
+            // 2 because the 1 is the method being instrumented
```
Mikhail-Pranovich: 1st ...

dhaval24:
image

E2E correlation works. W3CTestDotNetApp has W3C turned off and only emits request-ids.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,502 @@
...
+    private static Traceparent processLegacyCorrelation(HttpServletRequest request) {
+
+        String requestId = request.getHeader(TelemetryCorrelationUtils.CORRELATION_HEADER_NAME);
+
+        if (requestId != null && !requestId.isEmpty()) {
+            String legacyOperationId = TelemetryCorrelationUtils.extractRootId(requestId);
+            RequestTelemetry requestTelemetry = ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry();
+            requestTelemetry.getContext().getProperties().putIfAbsent("ai_legacyRootID", legacyOperationId);
+            requestTelemetry.getContext().getOperation().setParentId(requestId);
+            return new Traceparent(0, legacyOperationId, null, 0);
```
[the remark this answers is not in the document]
lmolkova: It is not always 16 hex encoded bytes.

If it could be used as traceid (passes validation) it should be the same, otherwise it should be a brand new trace id. Is it the case?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -0,0 +1,502 @@
...
+        }
+        return null;
```
lmolkova: Whatever goes wrong, start a brand new traceparent. It should not be null

dhaval24:
```
if (processedTraceparent == null){
                processedTraceparent = new Traceparent();
            }
``` 
Checkout above. If the processLegacyTraceparentReturnsNull, a brand new traceparent is created.

dhaval24:
@lmolkova  thanks for your comments. I have address them.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
lmolkova: But not if request-id is null or empty

dhaval24: @lmolkova Well it does, look at this code block carefully.

```
private static Traceparent processLegacyCorrelation(HttpServletRequest request) {

        String requestId = request.getHeader(TelemetryCorrelationUtils.CORRELATION_HEADER_NAME);

        try {
            if (requestId != null && !requestId.isEmpty()) {
                String legacyOperationId = TelemetryCorrelationUtils.extractRootId(requestId);
                RequestTelemetry requestTelemetry = ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry();
                requestTelemetry.getContext().getProperties().putIfAbsent("ai_legacyRootID", legacyOperationId);
                requestTelemetry.getContext().getOperation().setParentId(requestId);
                return new Traceparent(0, legacyOperationId, null, 0);
            }
        } catch (Exception e) {
            InternalLogger.INSTANCE.error(String.format("unable to create traceparent from legacy request-id header"
                + " %s", ExceptionUtils.getStackTrace(e)));
        }

        return null;
    }
```
if request-id is null or empty it would return null. Then in the function which calls this, it checks if processedTraceparent == null then create a brand new. The case is covered. Anyway, I have added Tests to assert this.

dhaval24:
@lmolkova I have now added extensive tests to cover all the possible scenarios. I believe all the concerns are resolved and we are ready to merge. 
@littleaj if you don't have question would you mind taking a quick glance and approving this.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -177,6 +179,22 @@ public void httpMethodFinished(String identifier, String method, String correlat
...
+     * @param uriObject
+     * @return
+     */
+    private String createTarget(URI uriObject) {
+        assert uriObject != null;
+        String target = uriObject.getHost();
+        if (uriObject.getPort() != 80 && uriObject.getPort() != 443) {
+            target += ":" + uriObject.getPort();
+        }
+        target += " | " + target;
```
dhaval24: @lmolkova there is a bug here. I am trying to append the same target, which is local string variable instead of passing the one from the httpFinished() method. I realized when doing final E2E testing. Will fix it, update PR and merge it.

dhaval24: Fixed. Now it's production ready all good :)

dhaval24:
#788

PR #787, opened 2018-12-16
Relocate Guava

### LINKED ISSUE #784: Web in release 2.2.1 doesn't relocate guava
We have just tried upgrading to the latest version of application insights - 2.2.1.

The 2.2.0 web version had relocated the guava packages to com.microsoft.applicationinsights.core.dependencies.google.common, but this looks to have been removed in the 2.2.1 release.

Instead the applicationinsights-web-2.2.1.jar includes files under com.google.common without any relocation.  This means that they conflict with guava itself and causes our application to no longer compile.  The guava version used by insights seems to be very old and doesn't include com.google.common.util.concurrent.SimpleTimeLimiter.create, which was added in version 22 of guava, which causes us compilation problems as we are using those features.

Note that this only affects the -web version.  The -core version correctly relocates the guava classes, but for some reason the 2.2.1 release of web no longer relocates.

dhaval24:
Fix #784 

@littleaj I fixed the above. It was quick one liner :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
 # Version 2.3.0
+- Introducing Application Insights SpringBoot Starter 1.1.1 (GA VERSION).
+- Fix [#784](https://github.com/Microsoft/ApplicationInsights-Java/issues/784) Shade guava dependency. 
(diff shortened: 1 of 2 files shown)


dhaval24:
Thanks @littleaj updating in a new PR.

PR #796, opened 2018-12-27
simplified MapUtil.copy

littleaj:
Mostly simplifying the caller side since each caller checked for null/empty. Plus, copy from source to target is a nop if it's null or empty, so why not just check for that in the method.


core/src/test/java/com/microsoft/applicationinsights/internal/util/MapUtilTest.java
```diff
@@ -1,14 +1,43 @@
...
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
 
 public class MapUtilTest {
 
+    @Rule
+    public ExpectedException expected = ExpectedException.none();
+
+    @Test
+    public void targetCannotBeNullInCopy() {
```
dhaval24: Nice to add the coverage for this scenario.

PR #797, opened 2019-01-02
added system properties to enable logging.

### LINKED ISSUE #653: Agent calls logger before logger initialization. System property to enable javaagent logger?
I noticed that there are several calls to `InternalAgentLogger` before the config file is read. This is odd because enabling the logger is done using that file.

It may be a good idea to have a system property which enables logging before the config file is read.

Thoughts?

littleaj:
Fix #653 

This allows a system property to enable logging. This will be especially useful for read log messages which are printed before the config file is read (or if it's non-existant).

This also deprecates the logAlways statement. We've discussed this method before and its pitfalls. I vote to remove it. If you disagree, let's discuss and I can revert it if needed.

### DIFF
--- a/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
+++ b/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
@@ -47,7 +47,7 @@ public enum InternalLogger {
     private final static SimpleDateFormat dateFormatter = new SimpleDateFormat("dd-MM-yyyy HH:mm:ss.SSSZ");
 
     public class PropertyKeys {
-        private static final String SDKLOGGER_PREFIX = "applicationinsights.sdklogger.";
+        private static final String SDKLOGGER_PREFIX = "applicationinsights.logger.";
 
         public static final String CONSOLE_LEVEL =                  SDKLOGGER_PREFIX + "console.level";
         public static final String FILE_LEVEL =                     SDKLOGGER_PREFIX + "file.level";
--- a/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
+++ b/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
@@ -22,9 +22,11 @@
 package com.microsoft.applicationinsights.internal.logger;
 
 import com.google.common.base.Strings;
+import org.apache.commons.lang3.StringUtils;
 
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.TimeZone;
 
@@ -43,6 +45,17 @@ public enum InternalLogger {
 
     private final static String LOGGER_LEVEL = "Level";
     private final static SimpleDateFormat dateFormatter = new SimpleDateFormat("dd-MM-yyyy HH:mm:ss.SSSZ");
+
+    public class PropertyKeys {
+        private static final String SDKLOGGER_PREFIX = "applicationinsights.sdklogger.";
+
+        public static final String CONSOLE_LEVEL =                  SDKLOGGER_PREFIX + "console.level";
+        public static final String FILE_LEVEL =                     SDKLOGGER_PREFIX + "file.level";
+        public static final String FILE_UNIQUE_PREFIX =             SDKLOGGER_PREFIX + "file.uniquePrefix";
+        public static final String FILE_BASE_FOLDER_PATH =          SDKLOGGER_PREFIX + "file.baseFolderPath";
+        public static final String FILE_NUMBER_OF_FILES =           SDKLOGGER_PREFIX + "file.numberOfFiles";
+        public static final String FILE_MAX_LOGFILE_SIZE_IN_MB =    SDKLOGGER_PREFIX + "file.numberOfTotalSizeInMB";
+    }
 
     public enum LoggingLevel {
         ALL(Integer.MIN_VALUE),
@@ -68,7 +81,12 @@ public enum InternalLogger {
         FILE
     }
 
-    private boolean initialized = false;
+    private volatile boolean initialized = false;
+
+    /**
+     * Flag so system properties are not checked each time if already checked once.
+     */
+    private volatile boolean propsNotFound = false;
 
     private LoggingLevel loggingLevel = LoggingLevel.OFF;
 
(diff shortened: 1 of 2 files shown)


ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -256,7 +321,7 @@ private void setLoggerOutput(String loggerOutputType, Map<String, String> logger
                 // If the user asked for a logger type
                 type = LoggerOutputType.valueOf(loggerOutputType.toUpperCase());
             } catch (Exception e) {
-                System.out.println(e);
+                System.err.println(e);
```
littleaj: This should be output to stderr and though I could make a case for it being removed.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -292,6 +357,9 @@ private void onInitializationError(String errorMessage) {
     }
 
     private void log(LoggingLevel requestLevel, String message, Object... args) {
+        if (!initialized && !systemPropertyInitialize()) {
```
littleaj: It may not appear so, but this is uses double-checked locking since the method is `synchronized`

littleaj: TODO: short circuit when level==OFF

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -68,7 +81,12 @@ public int getValue() {
         FILE
     }
 
-    private boolean initialized = false;
+    private volatile boolean initialized = false;
```
dhaval24: I think it's good to turn this to AtomicBoolean and same for the props variable. Then we don't need the sync block.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -85,33 +103,79 @@ private InternalLogger() {
...
+                dateFormatter.setTimeZone(TimeZone.getTimeZone(utcId));
+            } catch (Exception e) {
+                new ConsoleLoggerOutput().log(String.format("Failed to find timezone with id='%s'. Using default '%s'", utcId, dateFormatter.getTimeZone().getDisplayName()));
+            }
+        } finally {
+            initialized = true;
+        }
+    }
+
+    public synchronized boolean systemPropertyInitialize() {
```
dhaval24: This `synchronized` block is a costly lock operation especially when checking on every log message. It will significantly reduce the performance of SDK when trace logging is turned on`.

Also, I am not sure why a synchronized is needed here. Are there any known scenarios where multiple threads can try to simultaneously initialize this. Atleast not that I know of.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -85,33 +103,79 @@ private InternalLogger() {
...
+        // if the logger is already initialized, don't init again
+        // if the system properties have already been examined, don't check again
+        if (propsNotFound || initialized) {
```
dhaval24: By turning these variables to Atomic we can gurantee that initialization happens only once.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
dhaval24: I would not recommend calling this method on every log call. Instead what you can do is that move this logic of initialization to, `TelemetryConfigurationFactory` class. It has the a method `setMinimumConfiguration()` if there is no config file. Also you can prioritize this initialization over config file if needed, but regardless the method should be called only once instead of calling it everytime on each log message.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
littleaj: AtomicBoolean won't work here because both `initialized` and `propsNotFound` need to be guarded by the same lock

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -85,33 +103,79 @@ private InternalLogger() {
...
+            propsNotFound = true;
+            return initialized;
+        }
+
+        String type = console == null ? LoggerOutputType.FILE.toString() : LoggerOutputType.CONSOLE.toString();
+        Map<String, String> props = new HashMap<>();
+        props.put(LOGGER_LEVEL, level);
+
+        // if file logging is configured, read additonal properties
+        if (file != null) {
```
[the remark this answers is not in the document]
littleaj: this method's purpose is to populate the logging property map

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
littleaj: It's only called if not initialized.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
dhaval24: Ok in this case wouldn't it be good to just take an explicit lock and guard this block of code? Why do we need to lock on the whole method. Would that be possible?

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
dhaval24: Do you envision any scenario for thread safety? Do you see any cases where there can be two threads do initialization at same time. In my mind there are none. Please educate me with one, than I am fine to keep this sync. Otherwise its just wise to remove it imho.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
dhaval24: See my comment above on it.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -156,28 +156,32 @@ public synchronized boolean systemPropertyInitialize() {
...
-                props.put(FileLoggerOutput.TOTAL_SIZE_OF_LOG_FILES_IN_MB_ATTRIBUTE, fileSize);
-            }
+            populateMapForFileLogging(props);
         }
 
         initialize(type, props);
         return initialized;
     }
 
+    private void populateMapForFileLogging(final Map<String, String> props) {
```
dhaval24: This is good. Thanks

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
littleaj: We can try to determine if thread safety is necessary in a different project. My intention was not to change how this class works, just to add a feature.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
littleaj: Since the initialize method was synchronized, I kept the new part synchronized as well. I'm not changing the nature of this class, just adding functionality.

littleaj:
@dhaval24 Let me know if you need more clarification

PR #806, opened 2019-01-22
XML/SB/Agent configuration for exceptions truncation

### LINKED ISSUE #804: Provide XML/SB/Agent configuration for exceptions truncation
We need to have some control on maximum exception's chain depth and trace length in order to guarantee it's acceptance by Azure backend. 
Currently if the size of encoded exception exceeds 64Kb it's being refused by backend. 
Related issue: #482. 

Each type of configuration should provide the ability to set max depth of exception's chain and max trace length per exception. Possible names for these variables: stackSize, traceLength.

We may also hardcode those values. But such approach won't be flexible enough.

**Additional context**
Agent already have option stackSize (see ConfigRuntimeExceptionDataBuilder.java) which is used in ExceptionTelemetry.convertExceptionTree as a cap on max exceptions graph length.

vlev:
Fix #804 .

Not sure if this contribution should be considerate a significant one. 
If it's required to follow some extra procedures, please let me know.

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
@@ -29,7 +29,7 @@ import java.util.HashSet;
 public final class DataOfConfigurationForException {
     private boolean enabled = false;
     private Integer maxStackSize;
-    private Integer maxTraceLength;
+    private Integer maxExceptionTraceLength;
     private HashSet<String> suppressedExceptions = new HashSet<String>();
     private HashSet<String> validPathForExceptions = new HashSet<String>();
 
@@ -41,12 +41,12 @@ public final class DataOfConfigurationForException {
         this.maxStackSize = maxStackSize;
     }
 
-    public Integer getMaxTraceLength() {
-        return maxTraceLength;
+    public Integer getMaxExceptionTraceLength() {
+        return maxExceptionTraceLength;
     }
 
-    public void setMaxTraceLength(Integer maxTraceLength) {
-        this.maxTraceLength = maxTraceLength;
+    public void setMaxExceptionTraceLength(Integer maxExceptionTraceLength) {
+        this.maxExceptionTraceLength = maxExceptionTraceLength;
     }
 
     public HashSet<String> getSuppressedExceptions() {
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
@@ -28,16 +28,25 @@ import java.util.HashSet;
  */
 public final class DataOfConfigurationForException {
     private boolean enabled = false;
-    private int stackSize = Integer.MAX_VALUE;
+    private Integer maxStackSize;
+    private Integer maxTraceLength;
     private HashSet<String> suppressedExceptions = new HashSet<String>();
     private HashSet<String> validPathForExceptions = new HashSet<String>();
 
-    public int getStackSize() {
-        return stackSize;
+    public Integer getMaxStackSize() {
+        return maxStackSize;
     }
 
-    public void setStackSize(int stackSize) {
-        this.stackSize = stackSize;
+    public void setMaxStackSize(Integer maxStackSize) {
+        this.maxStackSize = maxStackSize;
+    }
+
+    public Integer getMaxTraceLength() {
+        return maxTraceLength;
+    }
+
+    public void setMaxTraceLength(Integer maxTraceLength) {
+        this.maxTraceLength = maxTraceLength;
     }
 
     public HashSet<String> getSuppressedExceptions() {
--- a/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
+++ b/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
@@ -151,9 +151,9 @@ public class ApplicationInsightsTelemetryAutoConfiguration {
             if (maxStackSize != null) {
                 telemetryConfiguration.setMaxStackSize(maxStackSize);
             }
-            Integer maxTraceLength = exceptionsConfig.getMaxTraceLength();
-            if (maxTraceLength != null) {
-                telemetryConfiguration.setMaxExceptionTraceLength(maxTraceLength);
+            Integer maxExceptionTraceLength = exceptionsConfig.getMaxExceptionTraceLength();
+            if (maxExceptionTraceLength != null) {
+                telemetryConfiguration.setMaxExceptionTraceLength(maxExceptionTraceLength);
             }
         }
 
--- a/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
+++ b/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
@@ -43,6 +43,8 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import com.microsoft.applicationinsights.telemetry.ExceptionTelemetryOptions;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -146,16 +148,17 @@ public class ApplicationInsightsTelemetryAutoConfiguration {
         initializePerformanceCounterContainer();
 
         ApplicationInsightsProperties.Exceptions exceptionsConfig = applicationInsightsProperties.getExceptions();
+        ExceptionTelemetryOptions options;
         if (exceptionsConfig != null) {
-            Integer maxStackSize = exceptionsConfig.getMaxStackSize();
-            if (maxStackSize != null) {
-                telemetryConfiguration.setMaxStackSize(maxStackSize);
-            }
-            Integer maxExceptionTraceLength = exceptionsConfig.getMaxExceptionTraceLength();
-            if (maxExceptionTraceLength != null) {
-                telemetryConfiguration.setMaxExceptionTraceLength(maxExceptionTraceLength);
-            }
+            options = new ExceptionTelemetryOptions(
+                    exceptionsConfig.getMaxStackSize(),
+                    exceptionsConfig.getMaxExceptionTraceLength()
+            );
+        }else{
+            options = ExceptionTelemetryOptions.empty();
         }
+        telemetryConfiguration.setExceptionTelemetryOptions(options);
+        log.debug("Exception Handling Options: {}", options);
 
         return telemetryConfiguration;
     }
--- a/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
+++ b/azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
@@ -144,6 +144,19 @@ public class ApplicationInsightsTelemetryAutoConfiguration {
         telemetryConfiguration.setChannel(telemetryChannel);
         initializeComponents(telemetryConfiguration);
         initializePerformanceCounterContainer();
+
+        ApplicationInsightsProperties.Exceptions exceptionsConfig = applicationInsightsProperties.getExceptions();
+        if (exceptionsConfig != null) {
+            Integer maxStackSize = exceptionsConfig.getMaxStackSize();
+            if (maxStackSize != null) {
+                telemetryConfiguration.setMaxStackSize(maxStackSize);
+            }
+            Integer maxTraceLength = exceptionsConfig.getMaxTraceLength();
+            if (maxTraceLength != null) {
+                telemetryConfiguration.setMaxExceptionTraceLength(maxTraceLength);
+            }
+        }
+
         return telemetryConfiguration;
     }
 
--- a/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
@@ -292,12 +292,7 @@ public class TelemetryClient {
             return;
         }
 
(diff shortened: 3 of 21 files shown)


dhaval24:
@reyang @littleaj could you please help reviewing this contribution. In short, it is to improve our exception capturing ability by triming down the stack trace size to what is supported by backend. 
@vlev thanks a lot for your contribution much appreciated.

dhaval24:
@vlev a quick comment on naming first. I would rather prefer something different than traceLength. trace is common terminology in Distributed Tracing (a.k.a correlation) and also is sometimes intermixed as logs. This would create yet another confusing.

May be exceptionTraceLength or something of that sort is better alternative. @reyang your thoughts on this?

PS: Hardcoding is not what I would prefer. Instead, customer configuration is a good choice. However, it still wont solve all the problems and some packets mights still drop because customer misconfigured it. In this case, it might be wise to log a message for troubleshooting when the backend responds with partial accept.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
```diff
@@ -28,16 +28,25 @@
  */
 public final class DataOfConfigurationForException {
     private boolean enabled = false;
-    private int stackSize = Integer.MAX_VALUE;
+    private Integer maxStackSize;
```
dhaval24: why do we need Class type `Integer` instead of primitive type `int`?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
```diff
@@ -28,16 +28,25 @@
...
+    private Integer maxTraceLength;
     private HashSet<String> suppressedExceptions = new HashSet<String>();
     private HashSet<String> validPathForExceptions = new HashSet<String>();
 
-    public int getStackSize() {
-        return stackSize;
+    public Integer getMaxStackSize() {
```
dhaval24: can this method be converted to return primitive `int` instead of Boxed type `Integer`. I do not see any specific need of it.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -144,6 +144,19 @@ public TelemetryConfiguration telemetryConfiguration(TelemetryChannel telemetryC
         telemetryConfiguration.setChannel(telemetryChannel);
         initializeComponents(telemetryConfiguration);
         initializePerformanceCounterContainer();
+
+        ApplicationInsightsProperties.Exceptions exceptionsConfig = applicationInsightsProperties.getExceptions();
+        if (exceptionsConfig != null) {
+            Integer maxStackSize = exceptionsConfig.getMaxStackSize();
+            if (maxStackSize != null) {
```
dhaval24: If this is null, shouldn't we have a default?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -144,6 +144,19 @@ public TelemetryConfiguration telemetryConfiguration(TelemetryChannel telemetryC
...
+                telemetryConfiguration.setMaxStackSize(maxStackSize);
+            }
+            Integer maxTraceLength = exceptionsConfig.getMaxTraceLength();
+            if (maxTraceLength != null) {
```
dhaval24: Same here, we should have a default number here as well.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -273,25 +273,50 @@ public void trackMetric(MetricTelemetry telemetry) {
      * @param properties Named string values you can use to search and classify trace messages.
      * @param metrics Measurements associated with this exception event. Appear in "custom metrics" in Metrics Explorer.
      */
-    public void trackException(Exception exception, Map<String, String> properties, Map<String, Double> metrics) {
+    public void trackException(Throwable exception, Map<String, String> properties,
```
dhaval24: It's a good move to make this signature to accept Throwable. I think this is also not a breaking change.

core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetryOptions.java
```diff
@@ -0,0 +1,23 @@
+package com.microsoft.applicationinsights.telemetry;
+
+public class ExceptionTelemetryOptions {
+    private int maxStackSize;
+    private int maxTraceLength;
+
+    public ExceptionTelemetryOptions(Integer maxStackSize, Integer maxTraceLength) {
```
dhaval24: I think it's good to comment out that default values would be Integer Max for future readers.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java
```diff
@@ -507,8 +506,7 @@ private void sendSQLTelemetry(MethodData methodData, Throwable throwable) {
             telemetryClient.track(telemetry);
             if (throwable != null) {
                 InternalLogger.INSTANCE.trace("Sending Sql exception");
-                ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(throwable);
```
dhaval24: Nice move to consolidate this logic.

core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
```diff
@@ -57,22 +58,20 @@ private ExceptionTelemetry() {
         initialize(data.getProperties());
     }
 
-    /**
-     * Initializes a new instance.
-     * @param stackSize The max stack size to report.
-     * @param exception The exception to track.
-     */
-    public ExceptionTelemetry(Throwable exception, int stackSize) {
```
dhaval24: Are you removing this public constructor? In that case it might be a breaking change, though rarely anyone is relying on it.

core/src/test/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetryTest.java
```diff
@@ -34,17 +34,6 @@
 
 public final class ExceptionTelemetryTest {
 
-    @Test
-    public void testSetException() {
-        NullPointerException exception = new NullPointerException("mock");
-        ExceptionTelemetry exceptionTelemetry = new ExceptionTelemetry(exception);
-
-        NullPointerException exception1 = new NullPointerException("mock");
-        exceptionTelemetry.setException(exception1);
```
dhaval24: It seems that you have removed this setter in ExceptionTelemetry class. It was a public method and therefore it means it could be a breaking change. I would like to avoid any breaking change here.

dhaval24:
@vlev to give you some more context, we need to think a bit about this design overall. It might be a good idea to have these properties configurable, however it still leaves a gap for users to make this call and it can well end up that they might misconfigure these properties and hence the telemetry doesn't end up in backend.

We also need to ensure that the problem is solved ubiquitously across languages. For example I am not sure there is a similar approach taken in C# or other Application Insights SDK. @SergeyKanzhelev what are your thoughts and how does C# handle this?

We need to understand a bit deeply wether we needed adjustments in our backend to handle large exceptions stack traces properly, wether we need a backend adjustment. or if we need to do it on SDK side we need to ensure design is consistent across SDKs. 

I would put the PR on a small hold so we can have a better design discussion here on Github. @vlev much appreciate that you brought light to this topic again.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
vlev: The idea is to avoid the duplication of decision logic for default values (in xml, sb and agent config). It seems more convenient to keep it in one place (see ExceptionTelemetryOptions.java). 
Null value, in case of the absence of specific configuration, seems like a natural choice.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/DataOfConfigurationForException.java
vlev: The idea is to avoid the duplication of decision logic for default values (in xml, sb and agent config). It seems more convenient to keep it in one place (see ExceptionTelemetryOptions.java).
Null value, in case of the absence of specific configuration, seems like a natural choice.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
vlev: Done

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
vlev: Done

core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetryOptions.java
vlev: I've added the description of default values into class level javadoc.

core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
vlev: I've reverted this change.

core/src/test/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetryTest.java
vlev: I've reverted this change.

vlev:
> 
> 
> The design seems to be good, however naming, and some breaking changes needs to be addressed at the very least I think.

I've reverted all breaking changes and renamed traceLength to exceptionTraceLength.

vlev:
> 
> 
> @vlev to give you some more context, we need to think a bit about this design overall. It might be a good idea to have these properties configurable, however it still leaves a gap for users to make this call and it can well end up that they might misconfigure these properties and hence the telemetry doesn't end up in backend.

Well, I agree - this extra configuration can not be treated as a final solution to #482. 
Nevertheless, it could provide some control and confidence in AI tracing capabilities. Also, this configuration might be useful for those who want to apply some constrains on the amount of data being sent and stored in AI backend. For instance, some legacy applications might generate quite a big amount of white noise errors. Which could be reflected in storage account bills.

dhaval24:
> Also, this configuration might be useful for those who want to apply some constrains on the amount of data being sent and stored in AI backend.

Regarding this, yes it is a perfectly valid scenario. However, we generally recommend users to implement their own class which implements `TelemetryProcessor` interface and you can then write your custom logic to trim the stacktrace. It might provide more flexibility for users to choose their own way of trimming exceptions. May be they want to trim stacktrace from top or bottom or take parts of top and bottom and trim in the middle. 

I am fine with either decision, but to make this solution a generic solution, we need to have consistency across different AI SDKs so that we do not confuse users, to adopt to one solution in one language and another solution in another.

I know this might sound a bit stringent, but once a configuration change is introduced it remains in the users configuration file for decade and it's very hard to revert it. So I do want to proceed with caution here.

vlev:
@dhaval24 
Could you please tell, if there are any updates on this issue? 
Do you guys have some timeline in mind?

Or, maybe, you need some additional input from my side?

dhaval24:
@SergeyKanzhelev do we face similar issue in .NET SDK? If yes, how did we resolved that? If not, what are your thoughts on above approach from @vlev

vlev:
@SergeyKanzhelev, @dhaval24 Any news on the issue?

gondar:
@SergeyKanzhelev do we know if similar issue is problem for .Net SDK? 

Current behaviour in Java is very worrying and basically means AppInsights is useless as centralised logs. What's the point of having those when I cannot be sure I have all exceptions there?

We already had 2 cases in our recent production when important issue was missed (our application happens to have huge stack traces).

@dhaval24 do we have any examples of how we should implement our own TelemetryProcessor to fix the issue above?

dhaval24:
@trask FYI - this issue persist in our SDK ( where huge stack traces are not sent due to schema limits). This is a design question in general and @reyang (If I remember correctly) also had some thoughts about it.

It depends on the type of application. Assuming it is a springboot type application you can easily wire up a processor like this:

```
@Bean
    public TelemetryProcessor myProcessor() {
        return (Telemetry t) -> {
            if (t instanceof ExceptionTelemetry) {
                // do some manipulation of stacktrace ...
            }
            // always return true, as there is no filtering.
            return true;
        };
    }
```

Springboot application will then wire up this processor. If it's java ee app similar logic exist. 

@littleaj / @trask can assist more on it.

gondar:
Thanks,
I sort of know the TelemetryProcessor syntax. But getting telemetry smaller was a bit trickier. Below my implementation:
`
    public static final int MAX_STACK_LENGTH = 100;
    public static final int MAX_INNER_EXCEPTIONS = 10;
    public static final int MAX_INNER_EXCEPTION_STACK_LENGTH = 10;
    @Override
    public boolean process(Telemetry telemetry) {
        if (telemetry instanceof ExceptionTelemetry) {
            Throwable throwable = ((ExceptionTelemetry) telemetry).getThrowable();
            cutStackTraceIfNeeded(throwable, MAX_STACK_LENGTH);
            ((ExceptionTelemetry) telemetry).setException(throwable, MAX_INNER_EXCEPTIONS);
        }
        return true;
    }
    private void cutStackTraceIfNeeded(Throwable throwable, int maxStackLength) {
        if (throwable == null)
            return;
        StackTraceElement[] stack = throwable.getStackTrace();
        if (stack.length > maxStackLength) {
            StackTraceElement[] throttledStack = Arrays.copyOf(stack, maxStackLength);
            throwable.setStackTrace(throttledStack);
        }
        cutStackTraceIfNeeded(throwable.getCause(), MAX_INNER_EXCEPTION_STACK_LENGTH);
    }
`
Let me know if you have some ideas of what I could possibly miss. What I dislike about this implementation the most is that transforming of throwable stackTrace happens twice: first in logsappender and than in my processor. That's going to be performance hit ;-/.

vlev:
@dhaval24 
Should I cancel this pull request? 

Looks like you've decided to stick with the workaround. 
Could you please include this into the documentation? Each of the projects using AI would need to include it into their source code.

trask:
closing, addressed by #1702

PR #813, opened 2019-02-18
Making constants in WebSyntheticRequestTelemetryInitializer public to…

adusumillipraveen:
… be able to reuse the header values in the applications using this.

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/initializers/WebSyntheticRequestTelemetryInitializer.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/initializers/WebSyntheticRequestTelemetryInitializer.java
@@ -35,14 +35,14 @@ import com.microsoft.applicationinsights.common.CommonUtils;
  * Created by gupele on 8/17/2015.
  */
 public class WebSyntheticRequestTelemetryInitializer extends WebTelemetryInitializerBase {
-    public final static String SYNTHETIC_SOURCE_NAME = "Application Insights Availability Monitoring";
-    public final static String SYNTHETIC_TEST_RUN_ID = "SyntheticTest-RunId";
-    public final static String SYNTHETIC_TEST_LOCATION = "SyntheticTest-Location";
-    public final static String SYNTHETIC_TEST_SOURCE = "SyntheticTest-Source";
-    public final static String SYNTHETIC_TEST_TEST_NAME = "SyntheticTest-TestName";
-    public final static String SYNTHETIC_TEST_SESSION_ID = "SyntheticTest-SessionId";
-    public final static String SYNTHETIC_TEST_USER_ID = "SyntheticTest-UserId";
-    public final static String SYNTHETIC_TEST_OPERATION_ID = "SyntheticTest-OperationId";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_SOURCE_NAME = "Application Insights Availability Monitoring";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_RUN_ID = "SyntheticTest-RunId";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_LOCATION = "SyntheticTest-Location";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_SOURCE = "SyntheticTest-Source";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_TEST_NAME = "SyntheticTest-TestName";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_SESSION_ID = "SyntheticTest-SessionId";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_USER_ID = "SyntheticTest-UserId";
+    public final static String APPINSIGHTS_AVAILABILITY_SYNTHETIC_TEST_OPERATION_ID = "SyntheticTest-OperationId";
 
     @Override
     protected void onInitializeTelemetry(Telemetry telemetry) {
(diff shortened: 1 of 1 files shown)


dhaval24:
@adusumillipraveen thank you for the contribution. I am bit unclear about the proposed use case. can you please elaborate the purpose of the change so it's better understood.

dhaval24:
In general it isn't problematic to make this constants public as these are header values that should remain the same always (unless backend service modifies it, which I don't see likely happening).

adusumillipraveen:
@dhaval24 , When an application is having to form a request with these headers (example - test cases) , application is having to maintain a duplicate of these headers . We are auto configuring the WebSyntheticRequestTelemetryInitializer through a bean , but having to copy the headers across all apps where we use AppInsights . Making these constants public will  remove this duplication.

dhaval24:
> @dhaval24 , When an application is having to form a request with these headers (example - test cases) , application is having to maintain a duplicate of these headers . We are auto configuring the WebSyntheticRequestTelemetryInitializer through a bean , but having to copy the headers across all apps where we use AppInsights . Making these constants public will remove this duplication.

I still did not get it very clearly. Why would your app need to have these headers (in test-case as you said). Are you testing your app against this headers and if so why it's needed?

Apologies, but I am still confused!!

adusumillipraveen:
@dhaval24 yes, we set these headers to the requests in our test cases both functional and smoke tests ,  which run on the running app integrated end to end. This will help filtering the telemetry especially while running negative tests . Does it help?

dhaval24:
> This will help filtering the telemetry especially while running negative tests

If I understand this correctly @adusumillipraveen you are running these tests to validate app insights behavior correct?

reyang:
This is by design, these headers are not part of the public API interface of the SDK.

Please refer to:
https://github.com/Microsoft/ApplicationInsights-aspnetcore/blob/master/src/Microsoft.ApplicationInsights.AspNetCore/TelemetryInitializers/SyntheticTelemetryInitializer.cs#L18


Regards,
Reiley

adusumillipraveen:
@reyang can this be modified to have them as public ? Do you see any issue in making constants public, these being static and final ?

adusumillipraveen:
@dhaval24 we are not exactly testing app insights behaviour, but willing to have a different behaviour based on the request having these headers for some of our negative test cases.

The behaviour of WebSyntheticRequestTelemetryInitializer is totally dependant  on these headers, so  one has to set these headers to be able to fully utilise the use cases covered by the class. In this case, it's for our functional test cases. But, any code setting these headers in application willing to modify the behaviour of the class needs to duplicate these constants. Making them public constants will avoid this .

timja:
Hi @dhaval24 

Just chiming in I work with @adusumillipraveen 
What we've been using these for are for testing in production and ignoring the telemetry generated in our alerts by adding these headers in our requests and excluding them from our queries.

Should we be using these headers for that?
Or should we do something else?

dhaval24:
@timja for sure these synthetic telemetry should be ignored (i.e when you receive request from our backend regarding availability testing likely they will have these headers) and this filter is explicitly designed to do that.

Just by configuring this initializer this telemetry gets filtered out.

What I still don't quite get is, "testing in production and ignoring the telemetry generated in our alerts"

can you explain this? My understanding is that by configuring this filter these request telemetry do get filtered out from App Insights and doesn't get reported to backend.

Do you generate any other telemetry as well which is not AI Telemetry per say?

dhaval24:
My understanding is different telemetry items should not be treated in very specific way as per the internal details. This gives away the abstraction.

SergeyKanzhelev:
If the reason to open those constants is to make additional enrichment of telemetry - it is a reasonable request. Perhaps rename them to make it clear that those header names are coming from Application Insights availability tests, not generic synthetic headers.

dhaval24:
@timja @adusumillipraveen can you comment on SergeyK's point. These headers' are meant to only filter out the synthetic telemetry that comes from Application insights availability service. it's not recommended to use these headers for generic synthetic request filtering. 

Would be recommended to rename this headers with ApplicationInsights Availability prefix if that is your usecase. 

If you are using for all purpose generic synthetic telemetry, it is recommended that you have this headers defined in your code base.

adusumillipraveen:
@SergeyKanzhelev @dhaval24 I have renamed the constants to be more specific.

dhaval24:
The new constant names are much cleaner. However one should note, that excessively while processing the request can have detrimental affect. One should be careful when and where this headers are accessed. 

With that in mind, @adusumillipraveen @timja as Sergey mentioned the use of these headers should not be to filter out general purpose synthetic telemetry. It should be only restricted for AI telemetry filtering. If this resonates your usecase, I would be wiling to open this up for you.

timja:
I'm not sure that it is our usecase, we saw the constants "SyntheticTest-RunId" and assumed they were general use.

While we do use the availability test, this is more general testing of the application in production.
i.e. one usecase:
We have alerts triggered on errors being recorded in our logs, if these are generated by automated tests then we ignore those.

dhaval24:
> I'm not sure that it is our usecase, we saw the constants "SyntheticTest-RunId" and assumed they were general use.
> 
> While we do use the availability test, this is more general testing of the application in production.
> i.e. one usecase:
> We have alerts triggered on errors being recorded in our logs, if these are generated by automated tests then we ignore those.

Got it! Thanks for clarification @timja. In this case my recommendation would be really to define these header constants in your code base in a separate class and possibly reuse it everywhere. The headers defined in this class are particularly for filtering out telemetry that comes as part of synthetic availability monitoring tests and should be restricted for that purpose itself.

dhaval24:
If you are okay with the recommendation, I would probably abandon this PR. Exposing the header constant for general usage would mean that these header names are used in several different codebases and it is possible that in future if availability monitoring service in Azure expect a different header name. In that case we would need to change this value breaking several users who unwantedly took dependency on this constants.

dhaval24:
abandoning this PR, as per the reasons described above. Pleas define these the generic synthetic filtering header constants in your code base as a recommendation from our side.

PR #815, opened 2019-02-25
Refactor http request instrumentation methods (Async need)

dhaval24:
This change abstracts out onBegin(), onEnd() and oException() methods into generic HttpServerHandler which is necessary for attaching async lister to Web Filter and handling asynchronous requests that execute outside the filter thread.

As a part of this PR, I have enhanced WebTelemetryModule interface to be generic and consolidated data extraction from request and response inside a common extractor class.

Deprecated `ApplicationInsightsServletResponseWrapper` in favor of `ApplicationInsightsServletExtractor` which handles extracting data for both request and response.

Note: couple of tests, have been disabled as they were flaky on mac os.

- [ ] CHANGELOG.md updated

cc: @lmolkova (optional review)

To help understand this PR better here are some details:

Core new files in this PR:

1. Interface - HttpExtractor.java
2. Class - ApplicationInsightsServletExtractor.java
3. Class - HttpServerHandler.java

Important files touched in this PR:

1. WebRequestTrackingFilter.java
2. WebRequestTrackingModule.java 
3.  WebTelemetryModule.java

Most of the other class/interfaces are touched as part of refactoring in above classes.

### DIFF
--- a/core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
+++ b/core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
@@ -29,6 +29,7 @@ import com.microsoft.applicationinsights.telemetry.TraceTelemetry;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -119,6 +120,7 @@ public class DockerContextInitializerTests {
         verify(contextPollerMock, times(0)).getDockerContext();
     }
 
+    @Ignore
     @Test
     public void testSDKInfoFileIsWrittenWithInstrumentationKey() throws IOException {
         // The expected instrumentation key below is taken from the ApplicationInsights.xml under the resources folder.
--- a/core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
+++ b/core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
@@ -31,6 +31,7 @@ import java.util.Collection;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -72,6 +73,7 @@ public final class TransmissionFileSystemOutputTest {
         testSuccessfulSends(12, 3, new Integer(SIZE_OF_MOCK_TRANSMISSION), null);
     }
 
+    @Ignore
     @Test
     public void testFetchOldestFiles() throws Exception {
         File folder = tmpFolder.newFolder(TEMP_TEST_FOLDER+"2");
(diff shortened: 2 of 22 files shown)


web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -0,0 +1,131 @@
...
+import java.net.MalformedURLException;
+import java.util.Date;
+
+/**
+ * This Helper Handler class provides the required methods to instrument requests.
+ * @param <P> The HttpRequest entity
+ * @param <Q> The HttpResponse entity
+ */
+@Experimental
+public final class HttpServerHandler<P /* >>> extends @NonNull Object */, Q> {
```
dhaval24: Should we provide an interface for this class to implement which contains onEnd and onBegin methods. Would it make sense to have generic Handler interface and this class implement it? However the question would be how many arguments would be passed to the generic interface methods? For http we need 2, but might need more or less for others? One approach can be to use varargs and fetch parameters from there but then methods needs to have have hard logic inside them for ordering of parameters inside varargs array.

core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
```diff
@@ -119,6 +120,7 @@ public void testDockerContextNotAccessedIfPollerNotCompleted() {
         verify(contextPollerMock, times(0)).getDockerContext();
     }
 
+    @Ignore
```
dhaval24: This test is flaky on Mac OS. Marked it ignore until we make test consistent.

core/src/test/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutputTest.java
```diff
@@ -72,6 +73,7 @@ public void testSuccessfulSendTenFilesWhereThereIsNoRoomForTheLastThree() throws
         testSuccessfulSends(12, 3, new Integer(SIZE_OF_MOCK_TRANSMISSION), null);
     }
 
+    @Ignore
```
dhaval24: This test is flaky on Mac OS. Marked it ignore until we make test consistent.

web/build.gradle
```diff
@@ -40,10 +40,10 @@ dependencies {
     provided group: 'javax.servlet', name: 'javax.servlet-api', version: '3.0.1'
     provided group: 'javax.enterprise', name: 'cdi-api', version: '1.1' // Java EE
     testCompile group: 'junit', name: 'junit', version: '4.12'
-    testCompile group: 'org.mockito', name: 'mockito-all', version: '1.8.0'
-    testCompile group: 'org.eclipse.jetty', name: 'jetty-server', version: '7.0.0.M0'
-    testCompile group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '7.0.0.M0'
-    testCompile group: 'org.eclipse.jetty', name: 'jetty-servlets', version: '7.0.0.M0'
+    testCompile group: 'org.mockito', name: 'mockito-all', version: '1.10.19'
```
dhaval24: Added newer version of Mockito which supports annotation and Dependency injection. Makes testing easier.

web/build.gradle
```diff
@@ -40,10 +40,10 @@ dependencies {
...
+    testCompile group: 'org.eclipse.jetty', name: 'jetty-server', version: '9.4.14.v20181114'
```
dhaval24: Newer version of jetty server to support async.

dhaval24:
As per @reyang's offline comments I will split this PR into 2 or possibly 3 parts:

1. New Interfaces
2. Refactor and ensure integration related changes for #1.

dhaval24:
Abandoning this in favor of #816, #817, #818 (smaller PRs)

PR #816, opened 2019-02-25
Add HttpExtractor interface, implementation and tests - Async Refactor part 1

dhaval24:
Split #815

Add a HttpExtractor to extract information from request and response. 

PS: The build will fail as integrations need to happen and due to dependency update of Jetty, Servlets, in test, some classes are no longer available like 'HttpMethods'


web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public String getMethod(HttpServletRequest request) {
+        return request.getMethod();
+    }
+
+    @Override
+    public String getHost(HttpServletRequest request) {
+        return request.getServerName() + ":" + request.getServerPort();
```
dhaval24: This is just to ensure the data consistency with Application Insights previous versions of SDK. Other systems like OpenCensus etc. just return ServerName here.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+    public int getStatusCode(HttpServletResponse response) {
+        if (response != null) {
+            return response.getStatus();
+        }
+        return 0;
+    }
+
+    @Override
+    public String getURI(HttpServletRequest request) {
+        return removeSessionIdFromUri(request.getRequestURI());
```
dhaval24: This is also for data integrity with previous versions of SDK. We do not capture session ids

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    private String removeSessionIdFromUri(String uri) {
+        Validate.notNull(uri);
+        int separatorIndex = uri.indexOf(';');
+        if (separatorIndex != -1) {
+            return uri.substring(0, separatorIndex);
+        }
+        return uri;
+    }
+}
```
reyang: minor, add a new line before EOF

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpExtractor.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Returns the value of user-agent header
+     * @param request HttpRequest entity
+     * @return user-agent header string
+     */
+    String getUserAgent(P request);
+
+    /**
+     * Returns the status code of request
+     * @param response HttpResponse entity
```
reyang: probably mention that 0 will be return if there is no response

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
dhaval24: Done

PR #817, opened 2019-02-25
Add HttpServer Handler (Async Refactor - Part 2)

dhaval24:
Merge #816 before this

Add a HttpServer Handler class that uses `HttpExtractor` defined in #816 to instrument begin() and end() of http requests.

PS: The build will still fail with this PR as final integrations are not a part of it. The next PR will follow with full integration.


web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -0,0 +1,130 @@
...
+     * @throws MalformedURLException
+     */
+    public RequestTelemetryContext handleStart(P request, Q response) throws MalformedURLException {
+        RequestTelemetryContext context = new RequestTelemetryContext(new Date().getTime(),null);
+        RequestTelemetry requestTelemetry = context.getHttpRequestTelemetry();
+        ThreadContext.setRequestTelemetryContext(context);
+
+        String method = extractor.getMethod(request);
+        String userAgent = extractor.getUserAgent(request);
+
```
reyang: Curious, why do we need a blank space here?

dhaval24: It's not needed per say. But when I added that my thought was it keeps nice separation of url related data in a block which is easy to ready. (though I can understand where you are coming from :) )

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -0,0 +1,130 @@
...
+        webModulesContainer.invokeOnEndRequest(request, response);
+    }
+
+    /**
+     * This method is used to capture runtime exceptions while processing request
+     * @param e Exception occurred
+     */
+    public void handleException(Exception e) {
+        try {
+            InternalLogger.INSTANCE.trace("Unhandled application exception: %s", ExceptionUtils.getStackTrace(e));
```
reyang: Minor, is this "application exception" or general "exception"?
The comment also mentioned "runtime exception", which confused me since Java has **RuntimeException**.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -0,0 +1,130 @@
...
+            if (telemetryClient != null) {
+                telemetryClient.trackException(e);
+            }
+        } catch (Exception ex) {
+            // swallow AI Exception
+        }
+    }
+}
```
reyang: Should be consistent on empty line before EOF across all files.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: I removed them.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: already fixed

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
```diff
@@ -0,0 +1,171 @@
...
+    private TelemetryConfiguration telemetryConfiguration = TelemetryConfiguration.getActive();
+    @Spy public WebModulesContainer<HttpServletRequest,HttpServletResponse> webModulesContainer =
+        new WebModulesContainer<>(telemetryConfiguration);
+    @Spy public TelemetryClient telemetryClient;
+    @InjectMocks HttpServerHandler<HttpServletRequest, HttpServletResponse> httpServerHandler;
+    @Mock HttpServletRequest request;
+    @Mock HttpServletResponse response;
+    @Mock HttpServletRequest requestWithQueryString;
+    private String url = "http://www.abc.com/xyz/opq";
+    private String url1 = "http://30thh.loc:8480/app/test%3F/a%3F+b;jsessionid=S%3F+ID?p+1=c+d&p+2=e+f#a";
```
reyang: Why do we want to have fragment in the URL, assuming it is removed by the client/browser? (security consideration I guess?)

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+        when(extractor.getUrl(request)).thenReturn(url);
+        when(extractor.getURI(request)).thenReturn("/xyz/opq");
```
reyang: It is confusing when I read getUrl and getURI, do we have a better name, probably need to rethink the naming here.

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
reyang: Do we want to cover IP address as the host? What about IPv6?

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: To be specific, this handles. - IOException, ServletException and RuntimeExceptions thrown while processing the request.

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
dhaval24: We do not store fragment. So the #a portion is not the part of our telemetry. The example in test is to ensure that we can handle any sort of request.

dhaval24: This is a great question. We can update this field to handle IpV6 as well but this change would be different from our current behavior. What is the recommendation here?

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
dhaval24: getUniformResourceLocator() and getUniformResourceIdentifier(). How does that sounds?

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: I updated the log message to be more clear.

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
dhaval24: renamed to above.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import java.net.MalformedURLException;
+import java.util.Date;
+
+/**
+ * This Helper Handler class provides the required methods to instrument requests.
+ * @param <P> The HttpRequest entity
+ * @param <Q> The HttpResponse entity
+ */
+@Experimental
+public final class HttpServerHandler<P /* >>> extends @NonNull Object */, Q> {
```
dhaval24: @reyang  I want to get your opinion on - should we make this class implement an interface `HttpHandler<P, Q>` with `handleStart()` and `handleEnd()` methods? Do we envision a scenario where we wrap httpClients as well and in that case this might be helpful. If not, then I think it's not needed.

dhaval24:
In the final integration : https://github.com/Microsoft/ApplicationInsights-Java/blob/IntegrateAsyncRefactorWork/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java this class only seems to have the correlation portion. Since the class is public we cannot really take it down, but in theory we might just move all things to handler. What are your thoughts @reyang . do you like the design where things are kept seperate as of today's state?

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
dhaval24: This is to keep the behavior consistent. There is no new logic added here.

dhaval24: This is  also to keep behavior consistent. No new functionality added here.

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
dhaval24: @reyang https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletRequest.html#getRequestURI-- 
As per this doc - getUri() returns part of URL and getURL() returns the full url that client used. Does this clarify your question.

reyang: Got it, thank you.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
reyang: If we have generic handler that covers both server and client, probably we should avoid the name handleStart and handleEnd, as the concept `start` has very different meaning for server and client.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
```diff
@@ -29,7 +29,7 @@
     private static final String USER_AGENT_HEADER = "User-Agent";
 
     @Override
-    public String getUrl(HttpServletRequest request) {
+    public String getUniformResourceLocator(HttpServletRequest request) {
```
reyang: Given that we want to follow https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletRequest.html#getRequestURI, we want to make the name consistent here.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: I think this is not very feasible atm. Clients use different request interfaces for example https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/HttpRequest.html and sometimes these interface do not have the methods like scheme etc.

dhaval24: Since we keep this API experimental for now, we can introduce interface when we feel confident.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/ApplicationInsightsServletExtractor.java
dhaval24: sure thing.

PR #827, opened 2019-03-04
Add Async Listener Implementation

dhaval24:
Related #816, #817, #818

Add Async Listener Implementation and tests.  Async Lister is used to complete the spans for Async Servlet Requests.

Note: There was a need to modify `handleEnd()` method's signature, as this method can be called in a different thread than that of `handleStart()`. Also it's not feasible to store the instance of `RequestTelemetryContext` as a final member variable of `HttpServerHandler` class, as it is created when `handleStart()` is invoked.

PS: The CI will fail as the integration with rest of the code is not part of this PR. I will push it as part of another PR.

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
@@ -13,10 +13,20 @@ import javax.servlet.http.HttpServletResponse;
 import org.apache.commons.lang3.Validate;
 import org.apache.http.annotation.Experimental;
 
+/**
+ * This class implements {@link AsyncListener} to handle span completion for async request handling.
+ */
 @Experimental
 public final class AIHttpServletListener implements Closeable, AsyncListener {
 
+    /**
+     * Instance of {@link RequestTelemetryContext}
+     */
     private final RequestTelemetryContext context;
+
+    /**
+     * Instance of {@link HttpServerHandler}
+     */
     private final HttpServerHandler<HttpServletRequest, HttpServletResponse> handler;
 
     public AIHttpServletListener(HttpServerHandler<HttpServletRequest, HttpServletResponse> handler,
@@ -58,7 +68,7 @@ public final class AIHttpServletListener implements Closeable, AsyncListener {
             try {
                 Throwable throwable = event.getThrowable();
                 if (throwable instanceof Exception) {
-                    // AI SDK can only track Exceptions.
+                    // AI SDK can only track Exceptions. It doesn't support tracking Throwable
                     handler.handleException((Exception) throwable);
                 }
             } finally{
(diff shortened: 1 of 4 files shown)


web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
```diff
@@ -13,14 +13,24 @@
...
     private final RequestTelemetryContext context;
+
+    /**
+     * Instance of {@link HttpServerHandler}
+     */
     private final HttpServerHandler<HttpServletRequest, HttpServletResponse> handler;
 
     public AIHttpServletListener(HttpServerHandler<HttpServletRequest, HttpServletResponse> handler,
-                                RequestTelemetryContext context) {
+                                    RequestTelemetryContext context) {
```
reyang: minor - indentation should align with HttpServerHandler?

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -94,16 +105,18 @@ public RequestTelemetryContext handleStart(P request, Q response) throws Malform
      * This method is used to indicate request end instrumentation, complete correlation and record timing, response
      * @param request HttpRequest object
      * @param response HttpResponse object
+     * @param context RequestTelemetryContext object
      */
-    public void handleEnd(P request, Q response) {
-        RequestTelemetryContext context = ThreadContext.getRequestTelemetryContext();
+    public void handleEnd(P request, Q response,
+        RequestTelemetryContext context) {
```
reyang: minor: is this the correct indentation?

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -121,4 +134,19 @@ public void handleException(Exception e) {
             // swallow AI Exception
         }
     }
+
+    /**
+     * Remove data from Threadlocal and ThreadLocalCleaners
+     */
+    private void cleanup() {
+        try {
+            ThreadContext.remove();
```
reyang: What would we normally do in these cleaners? Do we anticipate them to access the threadcontext?

dhaval24: In the cleaners we clean up the context object set in ThreadLocal. This is called when handleEnd() is triggered and is important.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
dhaval24: fixed

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: fixed

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
dhaval24: @reyang I have moved the context cleaning logic after cleaners run.

PR #830, opened 2019-03-04
Final Integration: Async Listener

dhaval24:
Merge #827 first

This PR integrates Async Listener to `WebRequestTrackingFilter`. I have introduced `setRequestTelemetryContext()` API for `WebTelemetryModule` interface as all the web modules need to use the same context which is propagated. 

Deprecated the usage of `InherritableThreadLocal` in favor of `ThreadLocal` as ITLS doesn't work when performing correlation in ThreadPools. 

- [x] CHANGELOG.md updated

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
@@ -59,17 +59,31 @@ public class WebRequestTrackingTelemetryModule
      */
     private final String W3C_BACKCOMPAT_PARAMETER = "enableW3CBackCompat";
 
+    /**
+     * The {@link RequestTelemetryContext} instance propogated from
+     * {@link com.microsoft.applicationinsights.web.internal.httputils.HttpServerHandler}
+     */
+    private RequestTelemetryContext requestTelemetryContext;
+
+    public void setRequestTelemetryContext(
+        RequestTelemetryContext requestTelemetryContext) {
+        this.requestTelemetryContext = requestTelemetryContext;
+    }
+
+    /** Used for test */
+    RequestTelemetryContext getRequestTelemetryContext() {
+        return this.requestTelemetryContext;
+    }
+
     // endregion Members
 
     // region Public
 
-    public WebRequestTrackingTelemetryModule() {
-
-    }
+    public WebRequestTrackingTelemetryModule() {}
 
     /**
      * Ctor that parses incoming configuration.
-     * @param configurationData
+     * @param configurationData SDK config Object
      */
     public WebRequestTrackingTelemetryModule(Map<String, String> configurationData) {
         assert configurationData != null;
@@ -111,8 +125,8 @@ public class WebRequestTrackingTelemetryModule
         }
 
         try {
-            RequestTelemetryContext context = ThreadContext.getRequestTelemetryContext();
-            RequestTelemetry telemetry = context.getHttpRequestTelemetry();
+            //RequestTelemetryContext context = ThreadContext.getRequestTelemetryContext();
+            RequestTelemetry telemetry = this.requestTelemetryContext.getHttpRequestTelemetry();
 
             // Look for cross-component correlation headers and resolve correlation ID's
             if (isW3CEnabled) {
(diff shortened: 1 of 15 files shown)


dhaval24:
@reyang I would request that if you can review this by tomorrow, would be great. Build on master is currently failing and needs this PR to be merged (As refactoring was split into multiple pr's)

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
```diff
@@ -59,17 +57,31 @@
...
+     * {@link com.microsoft.applicationinsights.web.internal.httputils.HttpServerHandler}
+     */
+    private RequestTelemetryContext requestTelemetryContext;
+
+    public void setRequestTelemetryContext(
+        RequestTelemetryContext requestTelemetryContext) {
+        this.requestTelemetryContext = requestTelemetryContext;
+    }
+
+    /** Used for test */
```
reyang: com.google.common.annotations.VisibleForTesting?

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
```diff
@@ -70,6 +72,9 @@ public void onError(AsyncEvent event) throws IOException {
                 if (throwable instanceof Exception) {
                     // AI SDK can only track Exceptions. It doesn't support tracking Throwable
                     handler.handleException((Exception) throwable);
+                } else {
```
reyang: curious - why do we want to log this? If we don't support Throwable, should we just keep silent?

web/src/test/java/com/microsoft/applicationinsights/web/internal/ThreadContextTests.java
```diff
@@ -78,6 +79,81 @@ public void run() {
         thread.join();
 
         Assert.assertNotNull(context[0]);
-        Assert.assertEquals(expectedRequestName, context[0].getHttpRequestTelemetry().getName());
+        Assert.assertSame(expectedRequestName, ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry()
+            .getName());
+    }
+
+    @Test
+    public void testChildThreadDoesnotGetContextUpdatedWhenrelyingonItls()
```
reyang: It is hard to read the method name! Need to figure out a better syntax for readability.

web/src/test/java/com/microsoft/applicationinsights/web/internal/ThreadContextTests.java
```diff
@@ -78,6 +79,81 @@ public void run() {
...
+        RequestTelemetryContext requestTelemetryContext1 = new RequestTelemetryContext(0);
+        requestTelemetryContext.getHttpRequestTelemetry().setName(expectedRequestName1);
+        ThreadContext.setRequestTelemetryContext(requestTelemetryContext1);
+
+        Assert.assertNotSame(expectedRequestName, ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry()
+            .getName());
+    }
+
+    @Test
+    public void testNewThreadCreatedWithWrappedRunnableGetsTheParentContext() throws InterruptedException {
```
reyang: This name is much better.

reyang:
> @reyang I would request that if you can review this by tomorrow, would be great. Build on master is currently failing and needs this PR to be merged (As refactoring was split into multiple pr's)

Sorry for the delay.

I do suggest that when we break down PRs, we should make sure each commit could pass the build.
I actually got surprised that you could merge even if build is failing.

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
dhaval24: We removed all the dependency to Guava. This is Guava annotation.

web/src/test/java/com/microsoft/applicationinsights/web/internal/ThreadContextTests.java
dhaval24: Sounds good.

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/AIHttpServletListener.java
dhaval24: I think it's good to have this log message, because when users create incidents against why exception is not getting captured, we can ask them to turn on logs and understand if their app is actually throwing a throwable. Good for diagnosis in my opinion.

web/src/test/java/com/microsoft/applicationinsights/web/internal/ThreadContextTests.java
dhaval24: renamed

PR #835, opened 2019-03-08
prevent duplicate telemetry items creation in async

dhaval24:
Fixes duplicate Telemetry Item generation, when using Async request.
Caught while writing integration tests.

Design ideas taken from: 
https://github.com/openzipkin/brave/blob/8d44c92a65b346ed9b348dd0c7103057d7455e86/instrumentation/servlet/src/main/java/brave/servlet/TracingFilter.java

Problem that we are trying to solve:  When an async request is created it enters the doFilter() method twice. The first time it hits the filter it comes via as `REQUEST` dispatch and from `ASYNC` dispatch in order. We do not want to instrument the same request twice and henceforth prevent handleStart() and handleEnd() from triggering. 

The proposed solution marks the request as processed with by setting an attribute to request object.


dhaval24:
@reyang merging this now as tested with smoke tests in #836 tests under major flavors of web servers.

PR #836, opened 2019-03-08
Async Request Tracking and Correlation Integration Tests

dhaval24:
Merge #835 first

PS: Will squash merge. Cannot merge until there is a way to disable JRE-7 for Springboot App.


test/smoke/appServers/jre.master.txt
```diff
@@ -1,2 +1 @@
-openjdk:7-jre
 openjdk:8-jre
```
dhaval24: @littleaj is there way to disable JRE-7 for Springboot smoke test? The async methods which return `CompletableFuture` does not work with JRE-7.

test/smoke/testApps/SpringBootTest/src/smokeTest/resources/appServers.txt
```diff
@@ -1 +1 @@
-tomcat8
\ No newline at end of file
+tomcat85
```
dhaval24: @littleaj mostly springboot apps are deployed on Tomcat 8.5. It makes sense to test on it rather than tomcat 8.

dhaval24: I added more web servers.

test/smoke/appServers/jre.master.txt
littleaj: Not yet. I'll work on it

test/smoke/testApps/SpringBootTest/src/smokeTest/resources/appServers.txt
littleaj: I think it'll work on all of the current app servers. if they fail after the merge we can trim the list. Let's add back the other tomcat versions, jetty, jbosseap6 and jbosseap7 for now.

test/smoke/testApps/SpringBootTest/src/main/java/com/springbootstartertest/controller/TestController.java
```diff
@@ -39,6 +52,17 @@ public String trackEventSpringBoot() {
...
 	@GetMapping("/throwsException")
 	public void resultCodeTest() throws Exception {
-		throw new Exception("This is an exception");
+		throw new ServletException("This is an exception");
+	}
+
+	@Async
+	@GetMapping("/asyncDependencyCall")
+	public CompletableFuture<Integer> asyncDependencyCall() throws IOException {
+		System.out.println("************* ASYNC CALL!!");
```
littleaj: we don't need this anymore :)

test/smoke/testApps/SpringBootTest/src/main/java/com/SpringBootApp.java
```diff
@@ -1,18 +1,72 @@
...
+			this.task = task;
+			this.rtc = rtc;
+		}
+
+		@Override
+		public void run() {
+			if (ThreadContext.getRequestTelemetryContext() != null) {
+				ThreadContext.remove();
+			}
+			System.out.println("*****Id is:" + rtc.getHttpRequestTelemetry().getId());
```
littleaj: another debug sysout. remove or edit so it gives more context if the test fails.

test/smoke/testApps/SpringBootTest/src/smokeTest/resources/appServers.txt
dhaval24: @littleaj Done. Will reduce if things fail. By the way earlier it had only tomcat8 :) so we are expanding coverage.

test/smoke/appServers/jre.master.txt
dhaval24: It's not immediately needed. Just keep in your backlog.

test/smoke/testApps/SpringBootTest/src/smokeTest/java/com/springbootstartertest/smoketest/SpringbootSmokeTest.java
```diff
@@ -40,11 +44,30 @@ public void trackEvent() throws Exception {
...
+	@TargetUri("/asyncDependencyCall")
+	public void testAsyncDependencyCall() {
+		assertEquals(1, mockedIngestion.getCountForType("RequestData"));
+		assertEquals(1, mockedIngestion.getCountForType("RemoteDependencyData"));
+		RequestData d = getTelemetryDataForType(0, "RequestData");
+		RemoteDependencyData rdd = getTelemetryDataForType(0,"RemoteDependencyData");
+		String requestOperationId = d.getId();
+		String rddId = rdd.getId();
+
+    	System.out.println("*****Request Id in smoke test is: " + requestOperationId);
```
littleaj: two more debug sysout lines: remove

PR #841, opened 2019-03-09
Fixed Rate Sampling Smoketest

littleaj:
smoke test for fixed rate sampling


test/smoke/testApps/FixedRateSampling/src/main/webapp/index.jsp
```diff
@@ -0,0 +1,11 @@
+<%@page pageEncoding="UTF-8" contentType="text/html;charset=UTF-8"%>
+
+<!DOCTYPE html>
+<html>
+<head>
+<title>Simple Telemetry</title>
```
dhaval24: Curious why do we name title Simple Telemetry? Shouldn't it reflect the test app name instead? In this case - Fixed Rate Sampling Test App?

test/smoke/testApps/FixedRateSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/FixedRateSamplingTest.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Test
+    @TargetUri("/fixedRateSampling")
+    public void testFixedRateSamplingNotInExcludedTypes() {
+        assertEquals(1, mockedIngestion.getCountForType("MessageData"));
+        assertEquals(100.0, getSampleRate("MessageData", 0), Math.ulp(50.0));
+    }
+
+    protected double getSampleRate(String type, int index) {
```
dhaval24: I think this method can be moved to base class AiSmokeTest. It can have utility elsewhere also.

test/smoke/testApps/FixedRateSampling/src/main/webapp/index.jsp
littleaj: That was there previously. We can change it, but this page isn't surfaced anywhere except in the code.

test/smoke/testApps/FixedRateSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/FixedRateSamplingTest.java
littleaj: I'm going to make some change to AiSmokeTest soon. I'll include this in those changes

dhaval24: Sounds good.

PR #842, opened 2019-03-09
disable unsafe xml features

littleaj:
https://cwe.mitre.org/data/definitions/611.html
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.md


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/XmlAgentConfigurationBuilder.java
```diff
@@ -459,4 +457,13 @@ public Element getTopTag(File configurationFile) throws ParserConfigurationExcep
         Element topElementTag = (Element)topNodeTag;
         return topElementTag;
     }
+
+    private DocumentBuilder createDocumentBuilder() throws ParserConfigurationException {
```
dhaval24: It's a good practice to write java docs for the method. I would encourage to add one here mentioning it mitigates the CWE.

Issue #749, opened 2018-09-24
Instrument newer versions of the postgres jdbc SDK


timja:
**Is your feature request related to a problem? Please describe.**
Postgres is not instrumented correctly by the agent

**Describe the solution you would like**
Update the code to support newer postgres SDK versions

**Describe alternatives you have considered**
We're currently manually instrumenting using the agent XML in some places but we have a lot of java applications that are missing out on this functionality.

**Additional context**
Need to update these lines of code:
https://github.com/Microsoft/ApplicationInsights-Java/blob/master/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java#L63
https://github.com/Microsoft/ApplicationInsights-Java/blob/master/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java#L222

To handle this commit:
https://github.com/pgjdbc/pgjdbc/commit/42c2e3b7953b2e68281c4ec269bd9749e135065b#diff-fc136162d0d2039cce76db7d26c9ba04

Specifically the rename from:
AbstractJdbc2Connection -> PgConnection

dhaval24:
@timja thank you very much for raising a feature request. PostGres, is something we don't have official auto collection support for, and hence you see its untouched for long. We will take this as a feature request and triage it to future feature releases. 

@nikmd23, here is a feature request for PostGres. 

@timja by the way, feel free to submit a PR if you would be willing to contribute to the project. We welcome all contributions!

PR #846, opened 2019-03-12
Instrument new version of postgres - Fix postgresql prepared statement instrumentation

luigibk:
Fix # https://github.com/Microsoft/ApplicationInsights-Java/issues/749

- [x] CHANGELOG.md updated


timja:
Tested on an application that uses Hibernate and postgres:
https://github.com/hmcts/ccd-user-profile-api
image

Tested on an application that uses JdbcTemplate and postgres:
https://github.com/hmcts/draft-store

image

dhaval24:
@luigibk @timja thank you for this contribution. I will review this PR and come back with comments :). 
Really appreciate your team stepping up to contribute.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
```diff
@@ -59,8 +59,10 @@ public void add() {
             factory = classFactoryForMySql();
             doAdd(factory, "com/mysql/jdbc/PreparedStatement");
 
-            factory = classFactoryForPostgreSql();
+            factory = classFactoryForOldPostgreSql();
             doAdd(factory, "org/postgresql/jdbc2/AbstractJdbc2Statement");
+            factory = classFactoryForNewPostgreSql();
```
dhaval24: Probably a better renaming here is recommended. New and Old doesn't seem to be old. Recommend being specific about the version of Postgres being instrumented.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
```diff
@@ -59,8 +59,10 @@ public void add() {
...
+            doAdd(factory, "org/postgresql/jdbc/PgPreparedStatement");
```
dhaval24: I would prefer to have these String signatures as constants, in the class. That helps keeping all the signatures at one place in the class. 

I know this is not fully followed in this class, but I would encourage this change.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
```diff
@@ -230,8 +231,23 @@ public ClassVisitor create(ClassInstrumentationData classInstrumentationData, Cl
...
             }
         };
+    }
 
-        return classVisitorFactory;
+    private ClassVisitorFactory classFactoryForNewPostgreSql() {
+        return new ClassVisitorFactory() {
+            @Override
+            public ClassVisitor create(ClassInstrumentationData classInstrumentationData, ClassWriter classWriter) {
+                HashSet<String> ctorSignatures = new HashSet<String>();
```
dhaval24: In general using interface is preferred.  I would like to see the change:

`Set<String> ctorSignatures = new HashSet<>();`

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
```diff
@@ -230,8 +231,23 @@ public ClassVisitor create(ClassInstrumentationData classInstrumentationData, Cl
...
+                ctorSignatures.add("(Lorg/postgresql/jdbc/PgConnection;Ljava/lang/String;III)V");
+                // NOTE: this constructor cannot be added as it does not have a String in position 2.
```
dhaval24: If the instrumentation of this constructor is missing, what functionalities would be disabled?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
luigibk: Ok, updated to jdbc2 and jdbc4

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
luigibk: From what I can see this constructor is never directly called by clients as it is not a client responsibility to provide a CachedQuery (which is internal to the driver). The constructor which takes the sql statement (String) is the one clients call and that in turn calls the overloaded (CachedQuery) version of the constructor which deals with cached queries.  

https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgPreparedStatement.java

I have removed that comment as it is not useful, actually it's probably confusing.

luigibk:
Thanks for your comments @dhaval24. Regarding your comments at the end:

1. This PR does not relate to statement (as it should be clear from the PR title). I don't have a use case for that currently. I might address it in a following PR should the need arise.
2. I used Jdbc4. That should be specific enough while reasonably future proof. Otherwise please feel free to give suggestions.
3. I could not find any tests for jdbc drivers in the test suite. From what I can see, to start with the postgresql ones, would imply adding a (dockerized) postgresql server and connect the SpringBootTest application to it. That way sql statements could be issued and related appinsights ingestions verified. Is that what you have in mind or you have a different idea?
4. Updated changelog and also bumped application version to 2.5.0 as this is a new feature even if not a big one.

gradle.properties
```diff
@@ -1,4 +1,4 @@
 // Project properties
-version=2.4.0
+version=2.5.0
```
dhaval24: I don't think we need to go 2.5.0 for this. I would combine it in 2.4.0 Please update if possible. If not, I can take care of this.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassDataProvider.java
dhaval24: This look much better :)

dhaval24:
> Thanks for your comments @dhaval24 and @littleaj. Regarding your comments at the end:
> 
> 1. This PR does not relate to statement (as it should be clear from the PR title). I don't have a use case for that currently. I might address it in a following PR should the need arise.
[Dhaval] : Yes, I understand that. So the instrumentation for statements will be missing. Wanted to clarify. 

> 2. I used Jdbc4. That should be specific enough while reasonably future proof. Otherwise please feel free to give suggestions.
[Dhaval]: I see that the most recent jdbc driver is 4.2. Though 4.0 is still the major version so it would be reasonably long lasting. Do you have clue on what changes were introduced in https://docs.oracle.com/javase/8/docs/technotes/guides/jdbc/jdbc_42.html and if it affects instrumentation. 

> 3. I could not find any tests for jdbc drivers in the test suite. From what I can see, to start with the postgresql ones, would imply adding a (dockerized) postgresql server and connect the SpringBootTest application to it. That way sql statements could be issued and related appinsights ingestions verified. Is that what you have in mind or you have a different idea?
[Dhaval]: Yes this was the original intent. I understand that this can be a big thing and might be a bit time consuming. If that is not possible for you, it's ok. for beta support we can ensure that end to end works and test manually. However, this automated testing would be nice to have.

> 4. Updated changelog and also bumped application version to 2.5.0 as this is a new feature even if not a big one.
[Dhaval]: Sure, this is a new feature. We are planning to release a newer version (2.4.0 which is packing a lot of new features - async support etc.) so I would like to bundle this feature in 2.4.0

gradle.properties
luigibk: Sorry, I had not noticed 2.4.0 had not been released yet. That's done.

dhaval24:
@luigibk @timja thank you for your contribution.

PR #848, opened 2019-03-13
fix agent sdk connection

### LINKED ISSUE #831: HTTPClient Dependency Telemetry not working when using a Scheduled Bean
### Expected behavior
Whenever we make an RestTemplate using HTTPClient call, we expect to see the dependency telemetry within Application Insights.

### Actual behavior
No Telemetry is recorded.  

### To Reproduce
Using a SampleBean class, with the Scheduled annotation: 
@Scheduled(initialDelay=5000, fixedDelay=100000)
public void init() {...}

Within the method, make a simple RestTemplate call to a second WebService (a separate Spring Boot application, with Application Insights enabled).  On the scheduled time, second WebService sees the request, reports its own Telemetry, but, a Dependency telemetry is not recorded from the first WebService.

#### Sample Application
I included a run.cmd file to auto build/run the two webservices (attaching the javaagent).  If you watch the log output, you'll see that the Demo application outputs the response from the Dummy application, however, no Telemetry is generated.  If you navigate to 'http://localhost:9090/test' - you'll see the dependency telemetry generated as expected.

### System information
 - SDK Version: 2.3.0
 - OS type and version: Windows
 - Using spring-boot? Yes
 - Additional relevant libraries (with version, if applicable):

### Observations
What I'm seeing, after jumping into the app insights java SDK is that when an HTTP request is sent via a Scheduled Bean, it never actually gets to the 'httpMethodFinished' method that the agent is supposed to execute after retrieving all the appropriate telemetry data.  When I execute a controller from the browser (which also makes a RestTemplate call), I do see that the 'httpMethodFinished' method is executed (via setting a breakpoint within that specific code base). 

[applicationInsightsDependencyScheduled.zip](https://github.com/Microsoft/ApplicationInsights-Java/files/2934958/applicationInsightsDependencyScheduled.zip)

dhaval24:
Fix #831 

This connects the agent to Filter for the entire life span of the web filter.


littleaj:
Running smoketests on this branch

dhaval24:
doCalcSendsRequestDataAndMetricData - this failed on jboss 7 and tomcat85. not sure why

dhaval24:
On testing it is discovered that this solution doesn't work for multiple applications running in a tomcat service instrumented by a single jvm agent. 
It does work for standalone springboot app. need more investigation here.

dhaval24:
@littleaj it would be good to cherry pick the commits made for the smoke test enhancements in this branch and create a seperate PR. I would like to close this and open another one with better design if possible.

dhaval24:
closing in favor of #860

PR #847, opened 2019-03-13
Fixed closable resources

littleaj:
Several Closable resources were either left open or incorrectly closed.

PR #849, opened 2019-03-13
Fix throw/return statments in finally blocks

littleaj:
This conflicts with #847. I'll need to address that before merge.


littleaj:
This includes a bug fix found by failing unit tests after correcting the finally block return statement

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -256,7 +256,7 @@ public int compare(File file1, File file2) {
             InternalLogger.INSTANCE.error("Failed to load transmission, non transmission, exception: %s", e.toString());
         } catch (IOException e) {
             InternalLogger.INSTANCE.error("Failed to load transmission, io exception: %s", e.toString());
-        }
+                }
```
dhaval24: Is this the right spacing?

PR #850, opened 2019-03-13
Created constant to reuse Random instance

littleaj:
Recreating the random object each time does not guarantee sufficient randomness.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/util/LocalStringsUtils.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/util/LocalStringsUtils.java
@@ -23,9 +23,12 @@ package com.microsoft.applicationinsights.internal.util;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 
 public class LocalStringsUtils {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/util/LocalStringsUtils.java
```diff
@@ -53,8 +53,15 @@ public static String generateRandomId(boolean removeDashes) {
         return uuid;
     }
 
+    /**
+     * Lazy init pattern.
+     */
+    private static class RandomHolder {
+        static final Random RANDOM = new Random();
```
dhaval24: recommend to replace this with ThreadLocalRandom. Has much better performance.

dhaval24: Precisely you can use this: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadLocalRandom.html#nextLong(long).

For your knowledge : here is why ThreadLocalRandom is better than Random. In nutshell it prevents contention in multithreaded environment where Random synchronizes over the seed.

http://thoughtfuljava.blogspot.com/2012/09/prefer-threadlocalrandom-over-random.html

littleaj: makes sense

PR #852, opened 2019-03-13
addressed bug where key/value pair was incorrectly overwritten

littleaj:
There are two signatures for `setTimestamp` so this needs to be a "multimap"


littleaj:
@dhaval24 static analysis tool flagged it. :)

ElliotZhang:
@littleaj there is a bug here, please fix the NPE thrown on
"foundDesc.contains(desc)" line 115.

littleaj:
@ElliotZhang good catch. thanks.

PR #853, opened 2019-03-13
Simplify jmx data fetcher

littleaj:
This loop was very hard to read. I think this is better.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
@@ -29,9 +29,14 @@ import java.util.Set;
 import java.util.Map;
 import java.util.HashMap;
 
+import javax.management.AttributeNotFoundException;
+import javax.management.InstanceNotFoundException;
+import javax.management.MBeanException;
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
+import javax.management.ReflectionException;
 import javax.management.openmbean.CompositeDataSupport;
+import javax.management.openmbean.TabularDataSupport;
 import java.lang.management.ManagementFactory;
 
 /**
@@ -83,9 +88,6 @@ public class JmxDataFetcher {
     private static Collection<Object> fetch(MBeanServer server, Set<ObjectName> objects, String attributeName, String attributeType) throws Exception {
         ArrayList<Object> result = new ArrayList<Object>();
 
-        String attr = attributeName;
-        String[] inners = null;
-
         AttributeType innerAttributeType = AttributeType.REGULAR;
         if (COMPOSITE_ATTRIBUTE_TYPE.equals(attributeType)) {
             innerAttributeType = AttributeType.COMPOSITE;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -83,45 +88,85 @@
...
+                break;
+            case REGULAR:
+                fetchRegularObjects(server, objects, attributeName, result);
+                break;
         }
 
+        return result;
+    }
+
+    static void fetchRegularObjects(MBeanServer server, Set<ObjectName> objects, String attributeName, ArrayList<Object> result)
```
dhaval24: are these methods accessed outside of this class? Looks like no. If that's the case, please make these methods private.

littleaj: I'm writing some tests for these explicitly

dhaval24: great please push those changes and let me know.

littleaj:
When writing the tests there were some cases where exceptions were thrown that I want to cover and test for. We should discuss what the correct behavior is here.

If a field returns the wrong type, there is an exception and since it's inside a loop, the loop exits and the remainder of the fetch is lost. I plan to catch the exception, log the attempted fetch and the error, then continue through the loop. I think we should make the best effort to fetch all existing fields based on the data given.

@dhaval24 @reyang Let me know if you agree with this.

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
```diff
@@ -3,26 +3,41 @@
...
 import java.util.Map;
+import java.util.Set;
 
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
 
 public class JmxDataFetcherTest {
+
+
+    protected static final int POINT_X = 111;
```
littleaj: I'm going to move these with the other constants

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -83,45 +89,85 @@
...
+
+        String[] inners = attributeName.split("\\.");
+        for (ObjectName object : objects) {
+            TabularDataSupport tabularData = (TabularDataSupport) server.getAttribute(object, inners[0]);
+            if (tabularData == null) {
+                InternalLogger.INSTANCE.warn("Could not find tabular attribute named '%s' for '%s' in object '%s'", inners[0], attributeName, object);
+                continue;
             }
+
+            CompositeDataSupport compositeData = (CompositeDataSupport) tabularData.get(inners[1]);
```
littleaj: There's a bug here: [`TabularDataSupport.get(Object)`](https://docs.oracle.com/javase/7/docs/api/javax/management/openmbean/TabularDataSupport.html#get(java.lang.Object)) calls [`TabularDataSupport.get(Object[])`](https://docs.oracle.com/javase/7/docs/api/javax/management/openmbean/TabularDataSupport.html#get(java.lang.Object[])) just by casting `Object` to `Object[]`. I don't think we've seen issues here because it appears `TabularData` in MXBeans is pretty rare. I wasn't able to find an example and only followed the documentation.

This needs to be fixed before merging. Since `get` takes and array, this also probably means the current way to reference tabular data is insufficient. I'm going to fix the cast issue and try to find a bean in JConsole with TabularData to test against.

littleaj:
Addressing the table doesn't seem to work in the way designed. I'm abandoning this effort until we know how that will work.

PR #854, opened 2019-03-13
Clarify changelog status

timja:
This confused us and we assumed everything in here was released, maybe add something to clarify at a glance what is released in the latest version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-# Version 2.4.0
+# Version 2.4.0 (Unreleased)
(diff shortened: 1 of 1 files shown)


dhaval24:
@timja in general whatever is in master is to be released :) 

For latest version and changelog you should always look at the releases on GitHub. Here it goes : https://github.com/Microsoft/ApplicationInsights-Java/releases/latest

timja:
I get that but when looking or updating the changelog it can be nice to know at a glance,
happy to close if you don't think this is useful though

dhaval24:
@timja merged. Thanks

PR #855, opened 2019-03-14
fixed possible npe

littleaj:
Fixes possible NPE from #852

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassVisitor.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassVisitor.java
@@ -111,9 +111,11 @@ public final class PreparedStatementClassVisitor extends ClassVisitor {
             return new PreparedStatementCtorMethodVisitor(access, desc, instrumentationData.getClassName(), name, originalMV, metaData);
         }
 
-        Set<String> foundDesc = nameAndSignatures.get(name);
-        if (desc != null && foundDesc.contains(desc)) {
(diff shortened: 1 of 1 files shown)


agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/sql/PreparedStatementClassVisitor.java
```diff
@@ -111,9 +111,11 @@ public MethodVisitor visitMethod(int access, String name, String desc, String si
             return new PreparedStatementCtorMethodVisitor(access, desc, instrumentationData.getClassName(), name, originalMV, metaData);
         }
 
-        Set<String> foundDesc = nameAndSignatures.get(name);
-        if (desc != null && foundDesc.contains(desc)) {
-            return new PreparedStatementSetMethod(access, desc, instrumentationData.getClassName(), name, originalMV, metaData);
+        if (desc != null) {
```
dhaval24: Curious why would this ever be null? This is basically the parent class signature as far as I know and in order to do instrumentation this signature needs to be passed.

littleaj: that check was already there. so, i can't really speak to that.

littleaj: the point of this PR is the null check on `foundDesc`

dhaval24: got it. Looks good.

PR #860, opened 2019-03-17
Breaking Change : Deprecate different agent connection modes

dhaval24:
**Breaking Change**: This PR Deprecates different Agent connection modes. Agent would only support instrumentation of 1 Application at a time. 
Note: In the current SDK version the app domain scenario (multi apps single agent) works only in single threaded environment. With this change, single app: single agent scenario will work for multi thread and async scenarios as well.

This PR also has some related code clean-up in `WebRequestTrackingFilter`. It is expected that there will be some performance improvement with this change as well.

This PR also has some changes made by @littleaj for smoketests.

#848 related.

Smoke Tests on different web servers have been disabled temporary to verify the behavior. Changes to be reverted before final merge.


littleaj:
@dhaval24 We will still need the `SelfMode` configuration option to support certain scenarios. Make sure that option is tested when making these changes. I can assist writing a smoke test which verifies this configuration.

dhaval24:
@littleaj Makes sense to write a smoke test to verify this. would need some help to write smoke tests against self mode, as  I am not sure how to use Smoke test for non webapps. I can test manually though.

dhaval24:
@littleaj tested in agent-self mode. This works fine as long as web is also available on classpath. `TelemetryCorrelationUtils` resides in web. We need to refactor this into core sdk which is out of scope for this PR.

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # CHANGELOG
 
 # Version 2.4.0 (Unreleased)
-- Fixed [#749](https://github.com/Microsoft/ApplicationInsights-Java/issues/749) introduce support for postgresql jdbc4 prepared statements.
+- Deprecated support for multiple apps instrumented with single JVM Agent. Instrumentation will only work for single apps
```
littleaj: Change "Deprecated" to "Removed".

I would also remove the second sentence. I think what will happen is the first application to register itself with the agent will be the one sending the telemetry collected from the agent; e.g. **app1** and **app2** register with the agent, **app1** registers first, all telemetry from dependency instrumentation in **app2** will appear to be sent from **app1**.

I'm concerned this may be a change for a major version.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
```diff
@@ -49,17 +49,16 @@
     private volatile long maxSqlMaxQueryThresholdInMS = 10000L;
     private volatile long redisThresholdInNS = 10000L * 1000000;
 
-    private static RuntimeExceptionDecider runtimeExceptionDecider;
+    private final RuntimeExceptionDecider runtimeExceptionDecider = new RuntimeExceptionDecider();
 
     private static ConcurrentHashMap<String, RegistrationData> notificationHandlersData = new ConcurrentHashMap<String, RegistrationData>();
```
littleaj: It appears this can be a `final` instance variable also.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/AgentConnector.java
```diff
@@ -43,6 +43,9 @@
     private String agentKey;
     private RegistrationType registrationType = RegistrationType.NONE;
```
littleaj: `@Deprecated` to RegistrationType also

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -195,55 +197,31 @@ private void configureWebAppNameContextInitializer(String appName, TelemetryConf
...
+            return false;
         }
 
         try {
-            ServletContext context = filterConfig.getServletContext();
-            String name = extractAppName(context);
-            String key = registerWebApp(appName);
-            setKey(key);
-            InternalLogger.INSTANCE.info("Successfully registered the filter '%s' in %.3fms. appName=%s",
-                this.filterName, sw.getNanoTime()/1_000_000.0, name);
```
littleaj: I want to keep the time measurements.

CHANGELOG.md
dhaval24: This is correct. This is what will happen. 
Re change for version: we plan to release beta wit this change. If there would be more votes in community to bring back the old behavior we will work on that. Currently, the multi-app scenario is already broken and doesn't work in multi-threaded environment. Modern apps are multi-threaded and we need to support those needes.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
dhaval24: `public void setExceptionData (DataOfConfigurationForException exceptionData)` this method is called to set the exceptiondata on `runtimeExceptionDecider` from `AgentImplementation.java` hence this cannot be instance variable. We need to keep that as class variable.

core/src/main/java/com/microsoft/applicationinsights/internal/agent/AgentConnector.java
dhaval24: Done

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: @littleaj I added timing to whole init method. This tells how long the filter takes to initialize. Do you want timing for `initializeAgentIfAvailable()` as well?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
littleaj: I was referring to `notificationHandlersData`. 

Since this is already a static singleton, the result is the same if it's declared as static or not.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -175,6 +175,8 @@ public void init(FilterConfig config) {
             if (StringUtils.isNotEmpty(config.getFilterName())) {
                 this.filterName = config.getFilterName();
             }
+            long end = System.currentTimeMillis();
+            InternalLogger.INSTANCE.trace("Initialized Application Insights Filter in " + (end - start) + " ms");
```
littleaj: Should use format string.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: Ah, I didn't see that. That should be fine.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
```diff
@@ -195,55 +197,31 @@ private void configureWebAppNameContextInitializer(String appName, TelemetryConf
...
         try {
-            ServletContext context = filterConfig.getServletContext();
-            String name = extractAppName(context);
-            String key = registerWebApp(appName);
-            setKey(key);
-            InternalLogger.INSTANCE.info("Successfully registered the filter '%s' in %.3fms. appName=%s",
-                this.filterName, sw.getNanoTime()/1_000_000.0, name);
+            RegistrationResult registrationResult = AgentConnector.INSTANCE.universalAgentRegisterer();
+            cleaners.add(registrationResult.getCleaner());
+            InternalLogger.INSTANCE.trace("Successfully registered the filter with appName=%s", this.appName);
```
littleaj: Let's keep it at info level

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
dhaval24: @littleaj Apologies for the confusion. This one is not being used at all in the new code path. This was being used by `public String register(ClassLoader classLoader, AgentNotificationsHandler handler)` which is now deprecated. Moot point to change this at the moment.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: String concatenation is much faster over String.format. Take a look at the stack over flow article here specifically the second answer and comments on it.

https://stackoverflow.com/questions/925423/is-it-better-practice-to-use-string-format-over-string-concatenation-in-java

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: Done

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/coresync/impl/ImplementationsCoordinator.java
littleaj: ah, yes. that is true

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
littleaj: It's generally practiced to use format strings for log messages so the JVM doesn't perform concatenation when the logger is disabled.

CHANGELOG.md
littleaj: Cool. Let's definitely release this as beta first. If someone is deploying multiple applications to a single application server, I want to know ahead of time.

Since we are doing that, I'm going to cherry-pick out changes unrelated to this PR and merge them separately. If we do end up having to revert, it'll be nicer to have one commit to revert.

web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java
dhaval24: Makes sense. Done.

CHANGELOG.md
dhaval24: Sounds great! Thanks @littleaj

PR #864, opened 2019-03-18
fixed throw inside finally

littleaj:


littleaj:
@dhaval24  I'll just deprecate the method and note that it shouldn't be removed but visibility reduced.

Issue #416, opened 2017-08-27
NoClassDefFoundError thrown from application insights


xscript:
When used in a Maven plugin, AI java SDK throws `NoClassDefFoundError` after the maven build is finished successfully. Please refer to below snapshot.
![image](https://user-images.githubusercontent.com/2300784/29747564-2cebfc54-8b31-11e7-9627-12e73ec21170.png)

It doesn't always happen. But for certain scenarios (such as unexpected exception in Maven plugin), this error will also be thrown.

Not sure what's the root cause here. Suggestion is to add `UncaughtExceptionHandler` for DataSender and Coordinator threads at [here](https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulse.java#L81-L87), so that such exception is properly handled (for my case, exception message should not be displayed to user).

dhaval24:
@xscript Can you please clarify on this issue. I believe you are getting the error when you are including AI-Java SDK in your maven project and when you try to build your maven project maven throws this exception? 

Or you are referring to building the Java-SDK? Also please tell us what all steps you took in either case to have a better understanding.

dhaval24:
@xscript as far as I remember I have very recently added AI (1.0.9) as a maven dependency in my maven project and it build successfully. I assume you are referring to the same of adding Application Insights as a maven dependency by following the steps mentioned on the documentation page.

If you are doing something in a different way can you please expand on how you are including AI in Maven Plugin so that I can replicate the bug and see what the fix can likely be. Thanks a lot in advance.

xscript:
@dhaval24 
This is not a build issue of AI SDK. This is a runtime issue when using AI SDK in a maven plugin.

### Maven Plugin project
In this project, AI SDK is referenced as the documentation. Version 1.0.9 is used. Building is successful in this project.
You can find the usage as listed below:
- https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-maven-plugin-lib/pom.xml#L139-L149
- https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-functions-maven-plugin/src/main/resources/ApplicationInsights.xml

### Normal Maven project
In this project, above Maven Plugin is referenced in `pom.xml`.
`mvn package` will trigger the execution of above Maven Plugin, hence AI SDK will be called to report telemetry.

[...]

PR #866, opened 2019-03-21
Improve lazy init; remove unnecessary executor

### LINKED ISSUE #721: Connection Pool Shutdown Issue
### Expected behavior
Application should shutdown gracefully when /shutdown endpoint is hit in Spring / SpringBoot Apps.

### Actual behavior
When shutting down SpringBoot Application via /shutdown endpoint, I see the following error and application exists with exit code 1

```
AI: ERROR 28-08-2018 19:18:29.785+0000, 60(ActiveTransmissionNetworkOutput_1-0): Failed to send, unexpected exception.
Stack Trace:
java.lang.IllegalStateException: Connection pool shut down
        at com.microsoft.applicationinsights.core.dependencies.http.util.Asserts.check(Asserts.java:34)
        at com.microsoft.applicationinsights.core.dependencies.http.pool.AbstractConnPool.lease(AbstractConnPool.java:189)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.conn.PoolingHttpClientConnectionManager.requestConnection(PoolingHttpClientConnectionManager.java:257)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.execchain.MainClientExec.execute(MainClientExec.java:176)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:185)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.execchain.RetryExec.execute(RetryExec.java:89)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.execchain.RedirectExec.execute(RedirectExec.java:111)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:108)
        at com.microsoft.applicationinsights.core.dependencies.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:56)
        at com.microsoft.applicationinsights.internal.channel.common.ApacheSender43.sendPostRequest(ApacheSender43.java:75)
        at com.microsoft.applicationinsights.internal.channel.common.TransmissionNetworkOutput.send(TransmissionNetworkOutput.java:185)
        at com.microsoft.applicationinsights.internal.channel.common.ActiveTransmissionNetworkOutput$1.run(ActiveTransmissionNetworkOutput.java:79)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

### To Reproduce
1. Create sample springboot app with a rest controller.
2. Use ApplicationInsights Springboot starter to instrument the app.
3.  Add following properties to application.properties file:
endpoints.shutdown.enabled=true
management.security.enabled=false
4. start the application with mvn spring-boot:run
5. send post request to /shutdown endpoint.

littleaj:
Fix #721.

I updated the lazy initialization procedure. The extra thread was overkill and added overhead. This may slow the first send a bit, but I think this will resolve the issue.

Take a look and make sure I didn't miss something in the refactor. I still need to run a proper test.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
@@ -35,29 +35,36 @@ import org.apache.http.params.HttpParams;
 /**
  * Created by gupele on 6/4/2015.
  */
+@SuppressWarnings("deprecation") // PoolingClientConnectionManager, DefaultHttpClient, HttpClient.getParams
 final class ApacheSender42 implements ApacheSender {
 
-    private final PoolingClientConnectionManager cm;
-    private final HttpClient httpClient;
+    private final Object lock = new Object();
+    private volatile PoolingClientConnectionManager cm;
+    private volatile HttpClient httpClient;
 
-    public ApacheSender42() {
-        cm = new PoolingClientConnectionManager();
-        cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
-        cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
-        
-        httpClient = new DefaultHttpClient(cm);
+    ApacheSender42() {
+        InternalLogger.INSTANCE.info("Using Apache HttpClient 4.2");
+    }
 
-        HttpParams params = httpClient.getParams();
+    private PoolingClientConnectionManager initializeConnectionManager() {
+        PoolingClientConnectionManager pccm = new PoolingClientConnectionManager();
+        pccm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
+        pccm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+        return pccm;
+    }
+
+    private HttpClient initializeHttpClient(PoolingClientConnectionManager pccm) {
+        HttpClient client = new DefaultHttpClient(pccm);
+
+        HttpParams params = client.getParams();
         HttpConnectionParams.setConnectionTimeout(params, REQUEST_TIMEOUT_IN_MILLIS);
         HttpConnectionParams.setSoTimeout(params, REQUEST_TIMEOUT_IN_MILLIS);
-
-        InternalLogger.INSTANCE.info("Using Apache HttpClient 4.2");
+        return client;
     }
 
     @Override
     public HttpResponse sendPostRequest(HttpPost post) throws IOException {
-        HttpResponse response = httpClient.execute(post);
-        return response;
+        return httpClient.execute(post);
     }
 
     @Override
@@ -66,16 +73,26 @@ final class ApacheSender42 implements ApacheSender {
 
     @Override
     public void close() {
-        try {
-            cm.shutdown();
-        } catch (Exception e) {
-            // chomp
+        synchronized (lock) {
+            if (cm != null) {
+                cm.shutdown();
+            }
         }
     }
 
     @Override
     public HttpClient getHttpClient() {
-        return httpClient;
+        HttpClient result = httpClient;
+        if (result == null) {
+            synchronized (lock) {
+                result = httpClient;
+                if (result == null) {
+                    cm = initializeConnectionManager();
+                    result = httpClient = initializeHttpClient(cm);
+                }
+            }
+        }
+        return result;
     }
 
     @Override
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
@@ -46,14 +46,14 @@ final class ApacheSender42 implements ApacheSender {
         InternalLogger.INSTANCE.info("Using Apache HttpClient 4.2");
     }
 
-    private PoolingClientConnectionManager initializeConnectionManager() {
+    private static PoolingClientConnectionManager initializeConnectionManager() {
         PoolingClientConnectionManager pccm = new PoolingClientConnectionManager();
         pccm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
         pccm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
         return pccm;
     }
 
-    private HttpClient initializeHttpClient(PoolingClientConnectionManager pccm) {
+    private static HttpClient initializeHttpClient(PoolingClientConnectionManager pccm) {
         HttpClient client = new DefaultHttpClient(pccm);
 
         HttpParams params = client.getParams();
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -93,14 +93,14 @@ final class ApacheSender43 implements ApacheSender {
         return result;
     }
 
-    private PoolingHttpClientConnectionManager initializeConnectionManager() {
+    private static PoolingHttpClientConnectionManager initializeConnectionManager() {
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -91,27 +68,44 @@ public void close() {
...
+            synchronized (lock) {
+                if (connectionManager != null) {
+                    connectionManager.shutdown();
+                }
+            }
         }
     }
 
     @Override
     public HttpClient getHttpClient() {
```
dhaval24: why not just return CloseableHttpClient here and prevent the cast in Line 69?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -91,27 +68,44 @@ public void close() {
...
-                            TimeUnit.MILLISECONDS.sleep(3);
-                        } catch (InterruptedException e){
-                        }
-                    }
-                    isClientInitialized = true;
+        CloseableHttpClient result = httpClient;
+        if (result == null) {
+            synchronized (lock) {
+                connectionManager = initializeConnectionManager();
+                result = httpClient;
```
dhaval24: Why do we need this line here? In the flow I didn't see that httpClient / result variable is updated anywhere. This seems redundant.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -91,27 +68,44 @@ public void close() {
...
+                if (result == null) {
+                    httpClient = result = initializeClient(connectionManager);
```
dhaval24: This seems very confusing. Please simplify these assignments.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -91,27 +68,44 @@ public void close() {
         try {
             ((CloseableHttpClient)getHttpClient()).close();
         } catch (IOException e) {
-            InternalLogger.INSTANCE.error("Failed to close http client, exception: %s", e.toString());
+            InternalLogger.INSTANCE.error("Failed to close http httpClient, exception: %s", e.toString());
+        } finally {
```
dhaval24: It seems this is a redundant locking. The methods that call this are synchronized already. My recommendation would be to prevent locks wherever possible.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: This is the interface implementation. The interface has to supports 4.2 and 4.3. `CloseableHttpClient` does not exist in 4.2

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: It's a standard pattern. A good explanation here: https://stackoverflow.com/questions/3578604/how-to-solve-the-double-checked-locking-is-broken-declaration-in-java

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: Assignments evaluate from right to left. It's equivalent to `httpClient = (result = initializeClient(connectionManager))`

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: It's good practice to keep a class thread-safe within itself. I don't want to rely on the current call stack to ensure thread safety. Since there won't be any other threads calling this, there won't be any contention. It's simply a precautionary statement. It won't affect performance.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: Yes I understand that. Taking a look at the practice you mentioned it looks reasonable. I would recommend adding the link to the stackoverflow link in the doc so future developers get aware.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: I see makes sense.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: Thanks makes sense

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: @littleaj yes I have no problems myself reading this. I am looking from the eyes of future inherriter of the code base  and it looks invasive :)

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
```diff
@@ -66,16 +73,26 @@ public void dispose(HttpResponse response) {
 
     @Override
     public void close() {
-        try {
-            cm.shutdown();
-        } catch (Exception e) {
-            // chomp
+        synchronized (lock) {
```
dhaval24: curious why we capture exception in the ApacheSender43 class but not here. Would like to learn more.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
dhaval24: Sounds good.

dhaval24:
@littleaj I am also worried that lazy initialization technique really impacts the users who are using our core sdk for custom telemetry tracking in standalone apps (in a tight loop) or maven jobs. VS China team uses the sdk to send telemetry for maven job executions and also had recently raised the problem of telemetry being dropped on the floor via email and previously via github.
https://github.com/Microsoft/ApplicationInsights-Java/issues/416#issuecomment-411640213 

Unless there is really noticeable performance gain by employing lazy initialization I would like to avoid it. 
As a reference, effective java also recommends to use this pattern only when absolutely needed. Take a look at this informative article
http://jtechies.blogspot.com/2012/07/item-71-use-lazy-initialization.html

littleaj:
@dhaval24 I'm still working through testing this. I'll post the performance numbers here when I have them

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
littleaj: The implementation of the shutdown method cannot throw

littleaj: but `close()` can

littleaj:
@dhaval24 I checked where the sender is actually being used and `getHttpClient` is either called from the QuickPulse thread or a sender thread, e.g. in ActiveTransmissionOutput. So, performance impact from lazy init will be hidden from the client.

littleaj:
superseded by #917

PR #870, opened 2019-03-21
In SmokeTests, always set container name

### LINKED ISSUE #861: JedisConnectionException while running smoke tests
Redis dependency container throws `JedisConnectionException` while running Caching Calculator Smoke Tests.

littleaj:
For some reason, the generated container names don't always end up in the docker DNS. Setting the container name explicitly addressed the issue.

Fix #861


littleaj:
Smoke test failed because the target test application name conflicts with the subsequent tests.

The idea was to let containers with failed tests remain available for future investigation, but i'm not sure if this is realistic. I have never actually done this. I usually run the test locally for debugging.

Given that, I'm fine with removing the containers after they've run. I'm going to add `--rm` to `docker run` invocations.

littleaj:
Addressed in #959

PR #883, opened 2019-03-28
Add implementation of dependency metric collection for Livemetrics

### LINKED ISSUE #882: Dependency Metrics do not show up on LiveMetrics
Quickpulse data sender currently has an empty implementation of `addDependency()` method and hence no dependency metrics are visible on Quickpulse UI. They stay flat.

dhaval24:
Fix #882


dhaval24:
> If there's a test for the other measurements, let's include this one too. Other than that, it looks good.

Nope there is apparently no test for QuickPulse codebase. This is far from ideal solution. Currently, this is to just unblock a major bug. When we get to phase for live metrics v2, we will invest in testability of this area (If done in SDK).

PR #887, opened 2019-03-28
remove ThreadLocal when it's no longer used

littleaj:
I found this during testing. We should include this in the next release.


littleaj:
FYI, backoffManager extends ThreadLocal... 

If we need ThreadLocal in the future, let's opt for field instead of a subclass.

dhaval24:
curious why is TLS needed in backoff manager? was this intended or it seems more of bug?

littleaj:
@dhaval24 I think it is needed, but I can't speak to why. When I was running tests in tomcat, there were 5 of these flagged by the leak scanner. So, I assume it's from multiple threads. Regardless, as long as we are using it; it needs to be removed as part of cleanup.

PR #888, opened 2019-03-29
the shutdownhook should not be a daemon

littleaj:
I noticed this also when shutting down tomcat: the threads weren't stopping as expected, including the shutdownhook thread. It turns out this was the problem.


littleaj:
@dhaval24 @reyang This also needs to be in the next release

dhaval24:
Can we put this as part of the 2.4.0-BETA? Seems important to me if we have issues in shutting down threads.

littleaj:
@dhaval24 absolutely.

PR #896, opened 2019-04-01
Cache sdk version string

littleaj:
We read the values in `sdk-version.properties` and `starter-version.properties` multiple times during the life of the SDK. This eliminates the need for opening/closing the stream each time the value is read and consolidates some common code.

@dhaval24 @reyang Please comment on the FIXME/TODOs. I hope to remove those before merging.

Also, @dhaval24, make sure the `break` in `SpringBootHeartbeatProvider` is correct. It appeared to be missing.


azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/HeartBeatProvider/SpringBootHeartBeatProvider.java
```diff
@@ -81,6 +80,7 @@ public Boolean call() {
                 break;
               case SPRING_BOOT_STARTER_VERSION:
                 provider.addHeartBeatProperty(fieldName, getSpringBootStarterVersionNumber(), true);
+                break;
```
dhaval24: Thanks for adding this. Appreciated. A bug on my side.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -212,21 +211,11 @@ private static void extractToLocalFolder(File dllOnDisk, String libraryToLoad) t
...
-        if (version == null) {
-            throw new RuntimeException("Failed to find SDK version.");
-        }
-
         File dllPath = LocalFileSystemUtils.getTempDir();
 
         dllPath = new File(dllPath.toString(), AI_BASE_FOLDER);
         dllPath = new File(dllPath.toString(), AI_NATIVE_FOLDER);
-        dllPath = new File(dllPath.toString(), version);
+        dllPath = new File(dllPath.toString(), PropertyHelper.getSdkVersionNumber()); // FIXME throw if version is unknown?
```
dhaval24: I would prefer marking as "unknown" rather than throwing. SDK should still send valuable telemetry

core/src/main/java/com/microsoft/applicationinsights/internal/util/DeviceInfo.java
```diff
@@ -65,13 +63,15 @@ public static String getLocale()
         final String languageTagMethodName = "toLanguageTag";
         Locale defaultLocale = Locale.getDefault();
         try {
+            // FIXME toLanguageTag is since 1.7; reflection isn't needed anymore.
```
dhaval24: lets remove reflection if there is lang support already. I have no objections if you make this change as part of this PR.

core/src/main/java/com/microsoft/applicationinsights/internal/util/PropertyHelper.java
```diff
@@ -78,18 +84,75 @@ public static Properties getSdkVersionProperties() {
...
     }
 
+    /**
+     * Returns the SDK version string, the <i>version-number</i> prefixed with "java:".
+     * The <i>version-number</i> is the value of the {@value #VERSION_PROPERTY_NAME} property in {@value #SDK_VERSION_FILE_NAME}.
+     * If the properties file cannot be read, {@value #UNKNOWN_VERSION_VALUE} is used for the <i>version-number</i>.
+     *
+     * @return "java:<i>version-number</i>" or "java:unknown"
+     */
+    public static String getSdkVersionString() {
```
dhaval24: would it make sense to reaname method as `getSdkVersionAsString()` to be a bit more clear?

littleaj: sure

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
littleaj: The change here is that it will be marked as `"java:unknown"`, so we can tell that it came from Java SDK. Before it was `"unknown"`, without the qualifier.

dhaval24: ah got it makes sense. Thanks for clarification @littleaj

littleaj: actually, here it will just be "unknown" because `:` is not a valid file name character...

PR #901, opened 2019-04-03
Reset CDS Profile fetch counters + Introduce generic PeriodTaskExecutor

dhaval24:
Currently CDS Profiler fetcher retries instantly 3 times. If the retry fails, it never retries for the entire application lifecycle. This leads to broken app-map experience as correlation needs ikey-appId resolution. 

This PR fixes this by resetting the retry counter in `CdsProfileFetcher.java` every 4 hours by default. This configuration can be over-riden. 

This PR also introduces a Generic Abstraction for creating, managing and running PeriodicTask called `PeriodicTaskManager`

In upcoming PR I will make this configurable via XML and Springboot starter.


core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        resetPeriodInMinutes = 240;
+    }
+
+    /**
+     * Returns an instance of CdsProfileFetcherPolicy
+     * @return instance of CdsProfileFetcherPolicy
+     */
+    public static CdsProfileFetcherPolicy getInstance() {
+        if (instance == null) {
+            instance = new CdsProfileFetcherPolicy();
```
reyang: Would this introduce race condition?

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * The interval in minutes for retry counters and pending tasks to be cleaned.
+     */
+    public long resetPeriodInMinutes;
+
+    /**
+     * Cached instance to be reused across SDK for CDS Profile fetch calls.
+     */
+    private static CdsProfileFetcherPolicy instance;
```
reyang: Curious, why do we want a static instance (are you looking for singleton?) versus having a per-client policy?

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        if (instance == null) {
+            instance = new CdsProfileFetcherPolicy();
+        }
+        return instance;
+    }
+
+    /**
+     * Resets the CDS configuration policy to default.
+     */
+    public void resetConfiguration() {
```
reyang: Why do we have this? Is this for test purpose only?

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        if (instance != null) {
+            InternalLogger.INSTANCE.warn(String.format("Resetting instance of CdsProfileFetcherRetryConfiguration - maxInstantRetries, " +
```
reyang: Why would we want to log?

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public int getMaxInstantRetries() {
+        return maxInstantRetries;
+    }
+
+    public long getResetPeriodInMinutes() {
+        return resetPeriodInMinutes;
+    }
+
+    public void setMaxInstantRetries(int maxInstantRetries) {
```
reyang: Needd input validation.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+        this.maxInstantRetries = maxInstantRetries;
+    }
+
+    public void setResetPeriodInMinutes(long resetPeriodInMinutes) {
+        this.resetPeriodInMinutes = resetPeriodInMinutes;
```
reyang: Same here, need validation.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -152,17 +174,29 @@ public void setEndpointAddress(String endpoint) throws MalformedURLException {
...
+        if (!this.failureCounters.containsKey(instrumentationKey)) {
+            this.failureCounters.put(instrumentationKey, 0);
         }
+        this.failureCounters.put(instrumentationKey, this.failureCounters.get(instrumentationKey) + 1);
+
     }
 
 	@Override
 	public void close() throws IOException {
         this.httpClient.close();
```
reyang: minor: indent

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Yes it would and I am fixing this.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Yes, updating and removing logs as well.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Will remove. this is only for test.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Yes. Thanks for the catch.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Yes, I think singleton is apt here. I will use enum instead of class to make singleton. Enums are threadsafe and best way to create singleton. 
speaking about the need, I think this policy should apply to all clients universally. I don't see a need of per client policy here.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Yes

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -86,9 +107,11 @@ public ProfileFetcherResult fetchAppProfile(String instrumentationKey) throws In
         ProfileFetcherResult result = new ProfileFetcherResult(null, ProfileFetcherResultTaskStatus.PENDING);
 
         // check if we have tried resolving this ikey too many times. If so, quit to save on perf.
-        Integer failureCounter = this.failureCounters.get(instrumentationKey);
-        if (failureCounter != null && failureCounter.intValue() >= MAX_RETRIES) {
-            InternalLogger.INSTANCE.warn("The profile fetch task will not execute. Max number of retries reached.");
+        if (failureCounters.containsKey(instrumentationKey) && failureCounters.get(instrumentationKey) >=
+                policyConfiguration.getMaxInstantRetries()) {
+            InternalLogger.INSTANCE.warn(String.format(
+                    "The profile fetch task will not execute for next %d hours. Max number of retries reached.",
```
littleaj: typo? `%d hours` ... `.getResetPeriodInMinutes()`

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -67,13 +80,21 @@ public CdsProfileFetcher() {
             .setDefaultRequestConfig(requestConfig)
             .useSystemProperties()
             .build());
-        
+
+        this.policyConfiguration = CdsProfileFetcherPolicy.getInstance();
+        resetService = Executors.newSingleThreadScheduledExecutor(
+                ThreadPoolUtils.createDaemonThreadFactory(CdsProfileFetcher.class, "CdsProfilePurgeService"));
+        long cachePurgeInterval = policyConfiguration.getResetPeriodInMinutes();
+        resetService.scheduleAtFixedRate(new CachePurgingRunnable(), cachePurgeInterval, cachePurgeInterval,
```
littleaj: @dhaval24 I wonder if we need an additional executor for purging the cache. If `fetchProfile` is used frequently, you could store the last time it was called and purge the cache after elapsed time is over the threshold.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Class Responsible for configuration of Cds Profile Fetch.
+ */
+public final class CdsProfileFetcherPolicy {
+
+    /**
+     * Maximum number of instant retries to CDS to resolve ikey to AppId.
+     */
+    public int maxInstantRetries;
```
littleaj: these should be private

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
dhaval24: yup :D Thanks

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
dhaval24: @littleaj there is a slight confusion due to naming. This should more be resetInterval. I do not purge cache and that is not needed. We only reset the counter. If cache part of `InstrumentationKeyResolver` already has the appId resolved there is no need to retry.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsProfileFetcherPolicy.java
dhaval24: Done

web/src/test/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcherTests.java
```diff
@@ -158,20 +165,60 @@ public void testFetchApplicationIdFailureWithNon200StatusCode() throws Interrupt
...
+        clientWrapper.setTaskAsComplete();
+        clientWrapper.setStatusCode(500);
+        result = fetcher.fetchAppProfile("ikey");
+        Assert.assertEquals(ProfileFetcherResultTaskStatus.FAILED, result.getStatus());
+
+        assertThat(fetcher.failureCounters.size(), not(0));
+
+        TimeUnit.MINUTES.sleep(2);
+
+        assertThat(fetcher.failureCounters.size(), equalTo(0));
```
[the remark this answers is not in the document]
dhaval24: Done.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
dhaval24: Done.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsRetryPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+        return maxInstantRetries;
+    }
+
+    public long getResetPeriodInMinutes() {
+        return resetPeriodInMinutes;
+    }
+
+    public void setMaxInstantRetries(int maxInstantRetries) {
+        if (maxInstantRetries < 1) {
+            throw new IllegalArgumentException("Number of CDS instant retries should be atleast 1");
```
reyang: atleast -> at least

normally people would put "maxInstantRetries must be greater than or equal to 1"

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsRetryPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+        this.maxInstantRetries = maxInstantRetries;
+    }
+
+    public void setResetPeriodInMinutes(long resetPeriodInMinutes) {
+        if (resetPeriodInMinutes < 1) {
+            throw new IllegalArgumentException("Reset interval for CDS retries should be atelast 1 minute");
```
reyang: same here, fix the typo and put a simple wording

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsRetryPolicy.java
```diff
@@ -0,0 +1,62 @@
+package com.microsoft.applicationinsights.internal.profile;
+
+/**
+ * Responsible for Cds Retry Policy configuration.
+ */
+public enum CdsRetryPolicy {
+    /**
+     * Cached instance to be reused across SDK for CDS Profile fetch calls.
+     */
+    INSTANCE;
```
reyang: This is a good learning to me, never heard of it before :)

core/src/main/java/com/microsoft/applicationinsights/internal/util/TimerTaskUtil.java
```diff
@@ -0,0 +1,76 @@
...
+        executorServiceMap.put(taskId, service);
+        service.scheduleAtFixedRate(command, initialDelay, period, unit);
+        SDKShutdownActivity.INSTANCE.register(service);
+    }
+
+    /**
+     * USED ONLY FOR TESTING
+     * Clear the map.
+     */
+    public static void reset() {
```
reyang: I don't know if it is a common practice to have public methods just for testing purpose.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsRetryPolicy.java
```diff
@@ -0,0 +1,62 @@
+package com.microsoft.applicationinsights.internal.profile;
+
+/**
+ * Responsible for Cds Retry Policy configuration.
```
reyang: fix the inconsistency in capitalization.

core/src/main/java/com/microsoft/applicationinsights/internal/profile/CdsRetryPolicy.java
dhaval24: Do you mean the use of enum? :)

core/src/main/java/com/microsoft/applicationinsights/internal/util/TimerTaskUtil.java
dhaval24: @reyang It is not a common practice at all! Just because this method is in core package and `CdsProfileFetcherTests` stays in web (legacy reasons) I need to keep this public.

dhaval24: resolved.

dhaval24:
@reyang I have redesigned the `TimerTaskUtil` and named it `PeriodicTaskManager` instead as it is mostly a management class and manages the lifecycle of `PeriodicTask`. Let me know your thoughts.

Additionally, I have also tried to resolve your previous comments.

core/src/main/java/com/microsoft/applicationinsights/internal/util/PeriodicTaskManager.java
```diff
@@ -0,0 +1,260 @@
...
+     * @param task PeriodicTask
+     * @return ScheduledFuture associated with the scheduled task.
+     */
+    public ScheduledFuture<?> executePeriodicTask(PeriodicTask task) {
+
+        if (task == null) {
+            throw new NullPointerException(" Task cannot be null");
+        }
+
+        task.validate();
```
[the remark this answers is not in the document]
dhaval24: @littleaj I agree with this point. Though, it limits the design of `validate()` method. I expect more parameters to be there in the PeriodicTask in future. Having the `validate()` method bound to parameter will force us to change the signature of the method or write overloaded method. Anyway since this is public, and the the point of creating objects which cannot be used is more costly I have taken your point :)

PR #903, opened 2019-04-08
Increase persistent http connection validation period

### LINKED ISSUE #898: Apache HTTP Client generates lot of  "http-outgoing-0 << "[read] I/O error: Read timed out"
Using apache logs:

```
 DEBUG wire                           - http-outgoing-11 << "[read] I/O error: Read timed out"
```

You can see it in idea when enable breakpoint on exception:

image

It seems to be caused by this issue:
https://stackoverflow.com/questions/34513826/apache-http-client-prints-read-i-o-error-read-timed-out

dhaval24:
Fix #898  .

Increase the persistent http connection validation period to be equal to the socket timeout period.

This prevents generation of messages like following:
` DEBUG wire                           - http-outgoing-11 << "[read] I/O error: Read timed out"` 
on the client side. 

Suggestion was borrowed based on stackoverflow article: https://stackoverflow.com/questions/34513826/apache-http-client-prints-read-i-o-error-read-timed-out

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -60,6 +60,7 @@ final class ApacheSender43 implements ApacheSender {
                 PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
                 cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
                 cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+                cm.setValidateAfterInactivity(REQUEST_TIMEOUT_IN_MILLIS);
 
                 httpClientRef.compareAndSet(null, HttpClients.custom()
                         .setConnectionManager(cm)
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -60,6 +60,7 @@ public void run() {
                 PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
                 cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
                 cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+                cm.setValidateAfterInactivity(REQUEST_TIMEOUT_IN_MILLIS);
```
littleaj: We should consider a smaller value for this.

dhaval24: Based on what was mentioned in Stackoverflow article, the value for this should be 
`min(socketTime,  keep-alive time, server connection timeout, idleConnectionMonitor)` . The connections on backend as far as I know are always keepAlive. The server doesn't respond with this value. Neither I am aware of other two. Therefore, let's keep the value similar to SocketTimeOut for now. We will iterate if we find newer details

littleaj:
@dhaval24 Here's another good reference: https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html

SanjayaKumarSahoo:
@dhaval24 is this fix is available in v2.3.1 , if not are we expecting a new release?

PR #904, opened 2019-04-08
Fix qps invariant version from 2 -> 1

dhaval24:
Fix QPS (Livemetrics) Invariant version number. Currently Java SDK only supports QPS v1.


littleaj:
@dhaval24, out of curiosity, what is different in version 2? Would it require a lot of changes to migrate?

dhaval24:
> @dhaval24, out of curiosity, what is different in version 2? Would it require a lot of changes to migrate?

@littleaj version 2 is the full Live Metrics support which .NET SDK has. It has capability of filtering and sample telemetry on the right pane.

PR #905, opened 2019-04-09
smoketest update

littleaj:
Adjust timeout for failing test.
Disable QuickPulse in smoketests since it continually pings the actual QuickPulse endpoint.
Parameterized the "slow response time" smoketest app.

### DIFF
--- a/test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
+++ b/test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
@@ -252,7 +252,7 @@ public class CoreAndFilterTests extends AiSmokeTest {
     }
 
     @Test
-    @TargetUri(value="/requestSlow", timeout=25_000) // the servlet sleeps for 20 seconds
+    @TargetUri(value="/requestSlow", timeout=35_000) // the servlet sleeps for 20 seconds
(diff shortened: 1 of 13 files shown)


test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
```diff
@@ -252,15 +252,20 @@ public void testAutoFailedRequestWithResultCode() {
...
         RequestData rd1 = getTelemetryDataForType(0, "RequestData");
         long actual = rd1.getDuration().getTotalMilliseconds();
         long expected = (new Duration(0, 0, 0, 20, 0).getTotalMilliseconds());
         long tolerance = 2 * 1000; // 2 seconds
-        assertThat(actual, both(greaterThanOrEqualTo(expected - tolerance)).and(lessThan(expected + tolerance)));
+
+        final long min = expected - tolerance;
+        final long max = expected + tolerance;
+
+        System.out.printf("Slow response time: expected=%d, actual=%d%n", expected, actual);
```
dhaval24: curious why you prefer format string over concatenation. In general concatenation is much faster :)

littleaj: If I  had to guess, I think it's from using loggers in the past. There you want to use parameters so it doesn't concat when that level is off.

PR #908, opened 2019-04-14
Upgrade gradle to 5.3.1

timja:
I'm looking at java 11 compatibility, saw this version is quite old which makes it harder to compile with newer java gradle 5.x is required for full compatibility

For significant contributions please make sure you have completed the following items:

- [na] Design discussion issue #
- [na] Changes in public surface reviewed
- [x] CHANGELOG.md updated
I would suggest using something like https://github.com/apps/release-drafter to maintain the CHANGELOG, much easier...

PR #909, opened 2019-04-14
Fix agent startup on Java 11

timja:
On java 11 this check fails:
`if ((systemClassLoader instanceof URLClassLoader)) {`
and it was falling through to a block that was returning the jar file rather than the directory that the agent is located in.

I get the following warning, not sure if its an issue:
```
```

- [x] CHANGELOG.md updated


timja:
I merged the gradle update into this because the shadow task was failing on new ASM
https://github.com/Microsoft/ApplicationInsights-Java/pull/908

timja:
I haven't tested this anymore than starting up on an app currently, but it starts and the only warning added is: 
```
```

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
```diff
@@ -167,7 +167,7 @@ public static String getAgentJarLocation() throws UnsupportedEncodingException {
         }
 
         String stringPath = AgentImplementation.class.getProtectionDomain().getCodeSource().getLocation().getPath();
-        return URLDecoder.decode(stringPath, "UTF-8");
+        return new File(URLDecoder.decode(stringPath, "UTF-8")).getParent();
```
dhaval24: Does this change work for JDK 8 as well? Or it will break JDK 8?

timja: JDK 8 will use the URLClassLoader a few lines up.

Not sure what this fall through case was for before, any idea?

dhaval24: Ah, I see. Yes, JDK8 will use the if-block above with URL class loader.

timja:
I've merged master into this now the gradle change is merged

timja:
could the tests be run against this PR @littleaj ?

littleaj:
@timja, I expect the unit tests should to be working, I've queued a build. 

I need to make some changes to the integration tests. Some of the app server images are using older versions and I just need to bump those to the java 11 supported versions.
For now, you can add `openjdk:11-jre` to https://github.com/Microsoft/ApplicationInsights-Java/blob/master/test/smoke/appServers/jre.master.txt, which will be needed regardless. Some appservers may need to exclude this. Right now, I only know JBoss EAP 6 for sure doesn't support 11.

timja:
Added java 11

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AgentImplementation.java
littleaj: I'm not really sure either. I think the CodeSource method would work in either case. This looks fine for now.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/redis/JedisMethodVisitor.java
```diff
@@ -301,7 +301,7 @@ protected int numberOfArguments() {
         return argumentTypes.length;
     }
 
-    protected Type[] getArgumentTypes() {
+    public Type[] getArgumentTypes() {
```
littleaj: Why did the visibility change here?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
```diff
@@ -153,7 +153,7 @@ protected String getMethodName() {
         return methodName;
     }
 
-    protected Type getReturnType() {
+    public Type getReturnType() {
```
littleaj: Why did the visibility change here?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/redis/JedisMethodVisitor.java
timja: can't remember right now, so something to do with the asm upgrade, will check

timja: image

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
timja: image

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/redis/JedisMethodVisitor.java
littleaj: Will you add `@Override`?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/agent/AdvancedAdviceAdapter.java
littleaj: `@Override`

timja:
@littleaj masters merged now if you want to kick off the tests

timja:
Anything holding this up @littleaj ?

littleaj:
@timja, just some unrelated build issues. I'm making sure the integration tests are in a clean state before merging this since it's going to add Java 11 to them. I had them passing in a branch today, so they should be back to normal tomorrow. Then I'm going to queue the smoke test against this PR.

@maxgolov, FYI since you've been looking into Java 9/10/11 support.

littleaj:
@timja, the smoke tests showed a few failing tests:
```
com.microsoft.applicationinsights.smoketest.PerfCountersDataTest > testPerformanceCounterData[9: jetty9, linux, openjdk_11-jre] FAILED
com.microsoft.applicationinsights.smoketest.PerfCountersDataTest > [7: tomcat85, linux, openjdk_11-jre].classMethod FAILED
com.microsoft.applicationinsights.smoketest.PerfCountersDataTest > [4: tomcat7, linux, openjdk_11-jre].classMethod FAILED
```

`classMethod` is part of the environment setup. In both cases, the test app was deployed, but the app did not respond to requests after deployment; so deployment likely failed.

`testPerformanceCounterData` did not respond with any data within the timeout period, but the application did start successfully.

Try to run the smoke tests locally (readme here: https://github.com/microsoft/ApplicationInsights-Java/tree/master/test/smoke). You can isolate the tests/appservers you need by commenting out projects in `settings.gradle`.

Some of these may just need to be excluded from running on java 11. Part of the java 9/10/11 effort is figuring out what environments we should claim to support. If there's something in the application that's incompatible, currently there is no way to exclude an app from running on a specific jre version. If this is the case, do let us know and we can add this capability.

@maxgolov, do take a look at this. I was thinking we could merge this into your branch as part of your efforts and then merge into master from there.

timja:
I probably won't have time to test this for now,
I run on Mac and running this on Windows is time consuming for me.

I had a look at @maxgolov 's branch (https://github.com/microsoft/ApplicationInsights-Java/tree/maxgolov/jdk11_support) and it looks like this would be a good fit there.

PR #912, opened 2019-04-26
Fix smokeTests in Gradle 5

littleaj:
After #908, the upgrade broke one thing in the smokeTests. Apparently, `processSmokeTestResources` no longer ran after the `assemble` task. This adds that dependency and adds an error condition if the test app is missing from its libs directory.


littleaj:
force-push was to fix the commit graph.

PR #914, opened 2019-04-27
Fix slow request integration test

littleaj:
This is addressing the broken smoke test.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -38,6 +38,7 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
+import org.apache.http.impl.NoConnectionReuseStrategy;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -60,10 +61,10 @@ public void run() {
                 PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
                 cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
                 cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
-                cm.setValidateAfterInactivity(REQUEST_TIMEOUT_IN_MILLIS);
 
                 httpClientRef.compareAndSet(null, HttpClients.custom()
                         .setConnectionManager(cm)
+                        .setConnectionReuseStrategy(new NoConnectionReuseStrategy())
```
dhaval24: This seems to be bit bizzard to me on high level it appears that connection will be recreated every time and this might impact performance. I haven't dig deep though.

littleaj: I was letting the tests run with this option to see if it helps, but it does not.

littleaj:
this was the wrong branch :D See #917 instead

PR #917, opened 2019-05-02
Fixes failing integration test

littleaj:
Includes same changes from #866, this is what end up fixing the test. Review this instead of #866.

Reworked PeriodicTaskPool (formerly PeriodicTaskManager). It didn't seem like a good idea to try to force this to be a singleton since it has a required initialization parameter. There's no way to guarantee thread safety with the instance variable exposed.


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender42.java
```diff
@@ -66,16 +73,26 @@ public void dispose(HttpResponse response) {
 
     @Override
     public void close() {
-        try {
-            cm.shutdown();
-        } catch (Exception e) {
-            // chomp
+        synchronized (lock) {
+            if (cm != null) {
+                cm.shutdown();
```
littleaj: This may not be necessary since `HttpClientBuilder.setConnectionManagerShared` is not set, i.e. the connection manager to http client relationship is 1-1 and the client should shutdown the CM when it's done.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -98,20 +103,15 @@ public void close() {
...
-                        }
-                    }
-                    isClientInitialized = true;
+        synchronized (httpClientRef) {
+            try {
+                while (httpClientRef.get() == null) {
+                    httpClientRef.wait();
                 }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
```
trask: 👍

PR #925, opened 2019-05-08
Add debug/status logging to log4j2 and logback smoketests

littleaj:
The title says it all.


test/smoke/testApps/TraceLog4j2/src/main/resources/log4j2.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<Configuration packages="com.microsoft.applicationinsights.log4j">
+<Configuration packages="com.microsoft.applicationinsights.log4j.v2" status="trace">
```
littleaj: I changed the package here because this is what we say to do in the docs (but I think it'll work either way)

https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-trace-logs#add-the-appender-to-your-logging-framework

PR #940, opened 2019-05-29
command name string comparison case insensitive

### LINKED ISSUE #907: JDBC Telemetry on explain plan is filter on uppercase "SELECT" only
### Expected behavior
JDBC telemetry should filter on query with mixed case "SELECT" cmd

### Actual behavior
Filtering only on uppercase "SELECT"

### To Reproduce
enable jdbc telemetry in AI-Agent.xml. Then create a slow jdbc query with lowercase "select" to activate the explain plan.

#### Sample Application
See the code from application insights.

https://github.com/Microsoft/ApplicationInsights-Java/blob/master/core/src/main/java/com/microsoft/applicationinsights/internal/agent/CoreAgentNotificationsHandler.java

image

philoushka:
Make the Explain query string comparison continue regardless of the casing (upper, lower, mixed) for a statement starting with `SELECT`.

Fix #907 .

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix [#898](https://github.com/Microsoft/ApplicationInsights-Java/issues/898) Increase persistent http connection validation
 timeout period.
 - Upgrade gradle to 5.3.1
+- Fix [#907](https://github.com/microsoft/ApplicationInsights-Java/issues/907) - ensure string compare is case insensitive when running a SQL explain on a select statement. 
 
 # Version 2.4.0-BETA
 - Removed support for multiple apps instrumented with single JVM Agent. Instrumentation will only work for single apps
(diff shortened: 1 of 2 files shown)


littleaj:
@philoushka Please update CHANGELOG.md in the root of the repo following the current convention (link to issue + short description of fix)

littleaj:
but other than that, it looks pretty straightforward. It should be good to merge after a successful build.

philoushka:
Thanks @littleaj , just wasn't sure which section within changelog.md to put an entry into... my initial guess would be the section for Version 2.4.0? Please give this a thumbs up if that's correct, or let me know if incorrect.

littleaj:
@trask, the build trigger worked!

littleaj:
Thanks for your contribution, @philoushka!

PR #941, opened 2019-06-04
Fix concurrency issue

trask:
This issue was introduced in 2.4.0-BETA (ea9ee4b735f96e12b4dab888741d9a524580c131).


littleaj:
I remember now that this commit had to do with this scenario: https://github.com/microsoft/ApplicationInsights-Java/wiki/Distributed-Tracing-in-Asynchronous-Java-Applications

We'll need to make sure there's a test which validates this scenario.

PR #944, opened 2019-06-12
simple perf improvements for smoke tests

littleaj:
This shaves off about 20 minutes with the current matrix in master (~19% faster).

Highlights:
* Introduces `@RequestCapturing(enabled=true)`. If this is enabled, then it will wait for the request telemetry from the test app health check and discard it. It is assumed enabled, so only add this if the test application does not have automatic request capturing (with `enabled=false`).
* Test app health check now uses a timeout instead of arbitrary delay. Currently it will retry 3 times if the timeout is reached. You may see warning logs from this change.
* When using `@TargetUri` there is now a default timeout and no delay.
* Updated flow of tests container "registration" during start up. This should resolve the issue when the shared network cannot be cleaned up because the container is still using it.
* Removed the "ExpectSomeTelemetry" annotation. It was useless.


littleaj:
@maxgolov, I think we can, but I think this should minimize those occurrences with the change to the handling of the `allContainers` collection. The only case that doesn't cover is cancelling a test run before it completes. For now, I can push the cleanup script to scm so it makes it a bit easier to work with.

PR #946, opened 2019-06-12
Internallogger findbugs fixes

littleaj:

### DIFF
--- a/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
+++ b/ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
@@ -278,14 +278,18 @@ public enum InternalLogger {
      * @param requestLevel - The level of the message
      * @param message - The message to print
      * @param args - The arguments that are part of the message
+     *
+     * @deprecated This should be avoided at all costs.
      */
     @Deprecated
     public void logAlways(LoggingLevel requestLevel, String message, Object... args) {
-        String logMessage = createMessage(requestLevel.toString(), message, args);
-        if (!initialized || loggerOutput == null) {
-            new ConsoleLoggerOutput().log(logMessage);
-        } else {
-            loggerOutput.log(logMessage);
+        synchronized (INSTANCE) {
+            String logMessage = createMessage(requestLevel.toString(), message, args);
+            if (!initialized || loggerOutput == null) {
+                new ConsoleLoggerOutput().log(logMessage);
+            } else {
+                loggerOutput.log(logMessage);
+            }
         }
     }
     /**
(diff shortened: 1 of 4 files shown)


ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -278,14 +278,18 @@ public void trace(String message, Object... args) {
...
-        if (!initialized || loggerOutput == null) {
-            new ConsoleLoggerOutput().log(logMessage);
-        } else {
-            loggerOutput.log(logMessage);
+        synchronized (INSTANCE) {
+            String logMessage = createMessage(requestLevel.toString(), message, args);
+            if (!initialized || loggerOutput == null) {
+                new ConsoleLoggerOutput().log(logMessage);
+            } else {
+                loggerOutput.log(logMessage);
```
trask: is synchronized block to prevent possible NPE on this line? if so, we could store loggerOutput into a local var and use that without needing synchronized block?

littleaj: Here's the findbugs output:

> Unsynchronized access at InternalLogger.java:[line 285]
> Field com.microsoft.applicationinsights.internal.logger.InternalLogger.loggerOutput
> Synchronized 60% of the time
> Unsynchronized access at InternalLogger.java:[line 288]
> Unsynchronized access at InternalLogger.java:[line 367]
> Synchronized access at InternalLogger.java:[line 195] (method sync, stop)
> Synchronized access at InternalLogger.java:[line 196] (stop)
> Synchronized access at InternalLogger.java:[line 353] (implicit method sync, onInitializationError <- initialize)
> Synchronized access at InternalLogger.java:[line 317] (implicit method sync, setLoggerOutput <- initialize)
> Synchronized access at InternalLogger.java:[line 335] (setLoggerOutput)
> Synchronized access at InternalLogger.java:[line 340] (setLoggerOutput)

I reviewed all these spots and it seems fine to me because of the volatile `isInitialized`. I'm updating the sync blocks to sync on `INSTANCE` instead of `this`. This should have the same effect but looks more consistent.

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
```diff
@@ -91,7 +91,7 @@ public int getValue() {
 
     private LoggerOutput loggerOutput = null;
 
-    private InternalLogger() {
+    InternalLogger() {
```
trask: does it need to be package accessible?

littleaj: private is redundant; enum constructors are always private. (another reason why I prefer to avoid the enum singleton pattern)

trask: 👍 how about removing the constructor?

ApplicationInsightsInternalLogger/src/main/java/com/microsoft/applicationinsights/internal/logger/InternalLogger.java
trask: 👍

PR #949, opened 2019-06-13
Remove printstacktrace and rework resource handling

littleaj:

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/config/JaxbAppInsightsConfigurationBuilder.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/config/JaxbAppInsightsConfigurationBuilder.java
@@ -54,24 +54,14 @@ class JaxbAppInsightsConfigurationBuilder implements AppInsightsConfigurationBui
                 return null;
             }
 
-            ApplicationInsightsXmlConfiguration applicationInsights = (ApplicationInsightsXmlConfiguration)unmarshaller.unmarshal(resourceFileReader);
-
-            return applicationInsights;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/config/JaxbAppInsightsConfigurationBuilder.java
```diff
@@ -54,25 +58,10 @@ public ApplicationInsightsXmlConfiguration build(InputStream resourceFile) {
...
-            }
-        } catch (NullPointerException e) {
-            e.printStackTrace();
-        } finally {
-            try {
-                resourceFile.close();
-            } catch (IOException e) {
-                e.printStackTrace();
-            }
+            InternalLogger.INSTANCE.error("Failed to parse configuration file: '%s'", ExceptionUtils.getStackTrace(e));
```
trask: I think this is before InternalLogger is initialized, so won't log?

littleaj: yes, but there you can turn on the logger with system properties instead of config. There are a few places where this occurs.

trask: should we enable logger via system properties during smoke tests?

littleaj: yeah, good call.

PR #950, opened 2019-06-13
Still more new agent work

trask:


web/src/main/java/com/microsoft/applicationinsights/web/internal/config/WebReflectionUtils.java
```diff
@@ -1,21 +1,23 @@
-package com.microsoft.applicationinsights.internal.config;
```
littleaj: If we're releasing 2.5, I don't think we can move this class since it's public. Unless there's a technical reason why it needs to be in this package.

agent/build.gradle
```diff
@@ -73,16 +82,29 @@ shadowJar {
...
     transform(com.github.jengelman.gradle.plugins.shadow.transformers.AppendingTransformer) {
         resource = 'META-INF/xyzzy.instrumentation.list'
     }
-}
-archivesBaseName = 'applicationinsights-agent'
 
-jar {
-    dependsOn shadowJar
-    enabled = false
     manifest {
```
littleaj: Why did the manifest config move from the jar task?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/model/IncomingSpanImpl.java
```diff
@@ -209,8 +211,8 @@ private void finishBuildingTelemetry(long endTimeMillis) {
         try {
             requestTelemetry.setUrl(removeSessionIdFromUri(getUrl(detail)));
         } catch (MalformedURLException e) {
-            InternalLogger.INSTANCE.error("%s", e.toString());
-            InternalLogger.INSTANCE.trace("Stack trace is%n%s", ExceptionUtils.getStackTrace(e));
+            logger.error(e.getMessage());
```
littleaj: I want to stop using the `"Stack trace is"` log message pattern. I don't think it's useful. Here, I'd rather see `logger.error(e.getMessage(), e)`

agent/build.gradle
```diff
@@ -51,7 +51,10 @@ shadowJar {
     relocate('org.slf4j', "${shadowPrefix}.org.slf4j")
     // FIXME need to relocate logback also, but currently causes instrumentation to be relocated
     // relocate('ch.qos.logback', "${shadowPrefix}.ch.qos.logback")
-    relocate('com.fasterxml.jackson', "${shadowPrefix}.com.fasterxml.jackson")
+
+    // this shades internal use of java util logging which is needed to prevent java util logging from being initialized
+    // early (e.g. in the case of jboss which wants to do some setup before jul is initialized)
+    relocate('java.util.logging', 'org.glowroot.xyzzy.engine.jul')
```
littleaj: @trask, Do you think this will fix the issue with JBoss in master?

test/smoke/appServers/build.gradle
```diff
@@ -56,6 +56,7 @@ subprojects {
 		doLast {
 			def jresFile = file("$resourcesOutputDir/${appServerShortName}.jre.txt")
 			jresFile.getParentFile().mkdirs()
+			jresFile.delete()
```
littleaj: Is this necessary or should it be part of the `clean` task?

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/CommonsLogFactoryClassFileTransformer.java
```diff
@@ -33,7 +33,8 @@
 
 import static org.objectweb.asm.Opcodes.ASM7;
 
-// this is needed because gradle shadow doesn't shade META-INF/services/org.apache.commons.logging.LogFactory
+// this is needed because gradle shadow doesn't shade the constant
+// "META-INF/services/org.apache.commons.logging.LogFactory" that is in commons-logging code
```
littleaj: shadowJar does have a method for shading SPI files. Did it not work in this case?

agent/build.gradle
trask: now there are 2 tasks that extend jar

agent/build.gradle
trask: It should fix the issue with java.util.logging

test/smoke/appServers/build.gradle
trask: i think i tried in clean, but `$resourcesOutputDir/${appServerShortName}` did not resolve, i think clean was too early, probably there's a way, but my gradle isn't good enough yet

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/CommonsLogFactoryClassFileTransformer.java
trask: it shaded the SPI file, but it didn't shade the constant that was in the .class file, so the .class file still pointed to `META-INF/services/org.apache.commons.logging.LogFactory`

web/src/main/java/com/microsoft/applicationinsights/web/internal/config/WebReflectionUtils.java
trask: this is a good question, the reason for the move may not be applicable anymore, i'll look into it

agent/build.gradle
littleaj: makes sense

agent/build.gradle
littleaj: I'm going to give it a try

test/smoke/appServers/build.gradle
littleaj: I think clean might already be deleting these files...I'll have to check.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/CommonsLogFactoryClassFileTransformer.java
littleaj: ah, I see. I didn't think there was an explicit reference to this file.

trask:
Closing, this is outdated now after some design changes, I'll submit new PR in a few days

PR #954, opened 2019-06-22
removed type parameters from public interface.

littleaj:
The interface also did not have type parameters in 2.3.1.


karolz-ms:
Thank you for making this change, Arthur. I assume this will restore `WebTelemetryModule` backward compatibility with previous SDK releases?

@xiaomi7732 FYI

littleaj:
@karolz-ms , yes, it's reverted to the original version. You can see [here](https://github.com/microsoft/ApplicationInsights-Java/blob/2.4.0-BETA/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebTelemetryModule.java) the type parameters and additional method was added in the 2.4.0-BETA, but did not exist in [2.3.1](https://github.com/microsoft/ApplicationInsights-Java/blob/2.3.1/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebTelemetryModule.java) or [2.3.0](https://github.com/microsoft/ApplicationInsights-Java/blob/2.3.0/web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebTelemetryModule.java).

PR #974, opened 2019-07-11
Clean up smoketest testApps

littleaj:
There were a few apps that were either unused or not smoketest apps in the smoketest directory. We need to be more strict about what we include under /test/smoke/ for the changes I want to make in the future.


test/webapps/PerfTestApp/build.gradle
```diff
@@ -1,7 +1,9 @@
 apply plugin: 'war'
 
+def aiSdkVersion = version
+
 dependencies {
-    compile aiCoreJar
+    compile "com.microsoft.azure:applicationinsights-core:$aiSdkVersion"
```
trask: Can we use `$version` here directly? I think it's understood when working in multi-module project that all versions are the same.

test/webapps/PerfTestAppSpringBoot/build.gradle
```diff
@@ -17,8 +17,10 @@ compileJava.targetCompatibility = 1.7
 compileSmokeTestJava.sourceCompatibility = 1.8
 compileSmokeTestJava.targetCompatibility = 1.8
 
+def aiSdkVersion = version
+
 dependencies {
-    compile springBootStarterJar
+    compile "com.microsoft.azure:applicationinsights-spring-boot-starter:$aiSdkVersion"
```
trask: Same question above about using `$version` directly here

test/webapps/PerfTestApp/build.gradle
littleaj: I wanted to have the option to use a different version than the one in the repo.

littleaj:
@trask, I think my autocrlf=true (checkout `\r\n`, check-in `\n`). I can change it to as-is if you prefer.

trask:
It looks like we had some files with bad line endings stored in git from long ago and the line endings got fixed when you moved them 👍

trask:
@littleaj it looks this has merge conflicts due to #975

PR #975, opened 2019-07-12
Build script updates

littleaj:
* remove usages of deprecated apis
* remove/upgrade deprecated plugins
* upgrade gradle wrapper to 5.5.1


littleaj:
I'm going to revert the gradle upgrade. There were a lot of changes around that. It'll be better to address it separately.

PR #978, opened 2019-07-15
DateFormatter is not null-safe

feigi:
Fix #.
If the timestamp property is not set, a call to BaseTelemetry#toString will result in an NPE. This is irritating as for example IntelliJ's debugger and the Mockito mocking framework call #toString.

- [-] Design discussion issue #
- [-] Changes in public surface reviewed
- [-] CHANGELOG.md updated

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
@@ -160,7 +160,7 @@ public abstract class BaseTelemetry<T extends Domain> implements Telemetry {
         tmp.setBaseData(getData());
         tmp.setBaseType(this.getBaseTypeName());
         envelope.setData(tmp);
-        envelope.setTime(getTimestamp() != null ? LocalStringsUtils.getDateFormatter().format(getTimestamp() : "null"));
+        if (getTimestamp() != null) envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp());
         envelope.setTags(context.getTags());
 
         envelope.serialize(writer);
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
```diff
@@ -160,7 +160,7 @@ public void serialize(JsonTelemetryDataSerializer writer) throws IOException {
         tmp.setBaseData(getData());
         tmp.setBaseType(this.getBaseTypeName());
         envelope.setData(tmp);
-        envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp()));
+        envelope.setTime(getTimestamp() != null ? LocalStringsUtils.getDateFormatter().format(getTimestamp() : "null"));
```
littleaj:
```suggestion
        if (getTimestamp() != null) envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp());
```
Then https://github.com/microsoft/ApplicationInsights-Java/blob/e29b3d9118ca1ab59ff54ddedb6c53e35c278095/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/Envelope.java#L219-L231
and https://github.com/microsoft/ApplicationInsights-Java/blob/e29b3d9118ca1ab59ff54ddedb6c53e35c278095/core/src/main/java/com/microsoft/applicationinsights/telemetry/JsonTelemetryDataSerializer.java#L210-L216
take care of a default value.

core/src/main/java/com/microsoft/applicationinsights/telemetry/BaseTelemetry.java
```diff
@@ -160,7 +160,7 @@ public void serialize(JsonTelemetryDataSerializer writer) throws IOException {
         tmp.setBaseData(getData());
         tmp.setBaseType(this.getBaseTypeName());
         envelope.setData(tmp);
-        envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp()));
+        if (getTimestamp() != null) envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp());
```
littleaj: missing closing parenthesis
```suggestion
        if (getTimestamp() != null) envelope.setTime(LocalStringsUtils.getDateFormatter().format(getTimestamp()));
```

Issue #979, opened 2019-07-15
Bring back custom agent instrumentation


nitinprabhu:
- We are currently trying to update our java version 11 which is supported by https://github.com/microsoft/ApplicationInsights-Java/blob/2.5.0-BETA/CHANGELOG.md
but the problem is it has removed support for <Class> custom instrumentation in AI-Agent.xml.

- We do use Class custom instrumentation so we would like to know how we can achieve the same in 2.50-BETA release or future releases.
```
<?xml version="1.0" encoding="utf-8"?>
<ApplicationInsightsAgent>
    <AgentLogger>INFO</AgentLogger>
    <Instrumentation>
        <BuiltIn enabled="true">
            <HIBERNATE enabled="true"/>
            <HTTP enabled="true"/>
            <JDBC enabled="true"/>
        </BuiltIn>

        <Class name="uk.gov.hmcts.reform.sendletter.services.encryption.PgpEncryptionUtil" type="Encryption">
            <Method name="encryptFile"/>
        </Class>
        <Class name="uk.gov.hmcts.reform.sendletter.services.pdf.PdfCreator" type="PdfCreator">
            <Method name="createFromTemplates"/>
            <Method name="createFromBase64Pdfs"/>
        </Class>
    </Instrumentation>
</ApplicationInsightsAgent>

```

littleaj:
We're adding this back in the next release. :)

nitinprabhu:
Thanks @littleaj for the update

PR #980, opened 2019-07-16
Bring back custom agent instrumentation

trask:
See #979

### DIFF
--- a/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/builder/XmlAgentConfigurationBuilder.java
+++ b/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/builder/XmlAgentConfigurationBuilder.java
@@ -19,15 +19,20 @@
  * DEALINGS IN THE SOFTWARE.
  */
 
-package com.microsoft.applicationinsights.agent.internal.config;
+package com.microsoft.applicationinsights.agent.internal.config.builder;
 
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
+import com.google.common.base.Strings;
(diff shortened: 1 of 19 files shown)


test/smoke/testApps/SpringBootTest/src/main/java/com/springbootstartertest/controller/TargetObject.java
```diff
@@ -0,0 +1,26 @@
...
+    String three() throws Exception {
+        throw new Exception("Three");
+    }
+
+    static class NestedObject {
+
+        void four(boolean x, int[] y, String[][] z) {
+        }
+    }
+}
```
littleaj: Could we add a case like this:
```java
String five() {
    sb.append(one());
    sb.append(two("call_two"));
    four(false, new int[0], new String[0][0]);
    return sb.append("_Five").toString();
}
```
I've seen this pattern before where a service is monitoring the high-level operations as well as the low-level parts of those operations.

agent/src/main/java/com/microsoft/applicationinsights/agent/internal/config/builder/XmlAgentConfigurationBuilder.java
```diff
@@ -142,7 +186,54 @@ private BuiltInInstrumentation getBuiltInInstrumentation(Element instrumentation
...
+    private Element getClassDataElement(Node item) {
+        if (item.getNodeType() != Node.ELEMENT_NODE) {
+            return null;
+        }
+
+        Element classNode = (Element) item;
+
+        String strValue = classNode.getAttribute(ENABLED_ATTRIBUTE);
+        if (!Strings.isNullOrEmpty(strValue)) {
+            boolean isEnabled = Boolean.valueOf(strValue);
```
littleaj:
```suggestion
            boolean isEnabled = Boolean.parseBoolean(strValue);
```

test/smoke/testApps/SpringBootTest/src/main/java/com/springbootstartertest/controller/TargetObject.java
trask: Done, and this uncovered a bug that prior tests missed!

PR #989, opened 2019-07-23
Enable dependency-check build plugin

littleaj:
This will fail the build if it finds any CVEs in dependencies. You can find the reports in `<project>/build/reports/dependency-check-report.html`. The html has descriptions of CVEs and also has buttons which produce snippets for the suppressions file, if needed.

Most of our use of the dependencies flagged are for their interfaces and the CVEs were generally related to runtime issues. I upgraded libraries where I could. I'll roll-back and evaluate again if the smoke tests fail because of the upgrades.

### DIFF
--- a/agent/dependency-check-suppressions.xml
+++ b/agent/dependency-check-suppressions.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+    <suppress>
+        <notes><![CDATA[
+      file name: instrumentation-logback-0.14.2.jar
+      ]]></notes>
+        <packageUrl regex="true">^pkg:maven/org\.glowroot\.instrumentation/instrumentation\-logback@.*$</packageUrl>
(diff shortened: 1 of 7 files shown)


agent/dependency-check-suppressions.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+    <suppress>
+        <notes><![CDATA[
+      file name: instrumentation-logback-0.14.2.jar
```
trask: Nice to add reason, like in others, e.g.
```
This is logback instrumentation, not the logback library itself.
```

littleaj:
@trask, one FYI: this used to have `check.dependsOn dependencyCheckAnalyze`, but the build adds some plugins dynamically at build time (findbugs/PMD) which were being flagged. So instead I'm going to add a separate task to the build which runs `dependencyCheckAnalyze`. 

This also allows us to grab the junit results files so the failures show in the test results tab. The build server doesn't allow multiple patterns for junit test results and I didn't see a way to change the output file name in the dependencyCheck plugin.

PR #993, opened 2019-07-25
Use tls1.2 by default

littleaj:
You can use `applicationinsights.ssl.protocols` to override. Empty value means <strike>let the JVM/Http client decide (i.e. the current behavior).</strike> use defaults: TLSv1.2 or v1.3.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -44,8 +44,6 @@ import org.apache.http.impl.client.HttpClients;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
 import org.apache.http.ssl.SSLContexts;
 
-import javax.net.ssl.SSLContext;
-
 /**
  * Created by gupele on 6/4/2015.
  */
@@ -59,13 +57,14 @@ final class ApacheSender43 implements ApacheSender {
                 new Runnable() {
                     @Override
                     public void run() {
-                        PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
                         final String[] allowedProtocols = SSLOptionsUtil.getAllowedProtocols();
-                        if (allowedProtocols != null) {
-                            SSLConnectionSocketFactory sf = new SSLConnectionSocketFactory(SSLContexts.createDefault(), allowedProtocols, null, SSLConnectionSocketFactory.getDefaultHostnameVerifier());
+                        final PoolingHttpClientConnectionManager cm;
+                        if (allowedProtocols.length == 0) {
+                            cm = new PoolingHttpClientConnectionManager();
+                        } else {
                             cm = new PoolingHttpClientConnectionManager(RegistryBuilder.<ConnectionSocketFactory>create()
+                                    .register("https", new SSLConnectionSocketFactory(SSLContexts.createDefault(), allowedProtocols, null, SSLConnectionSocketFactory.getDefaultHostnameVerifier()))
                                     .register("http", PlainConnectionSocketFactory.getSocketFactory())
-                                    .register("https", sf)
                                     .build());
                         }
                         cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -62,16 +62,7 @@ final class ApacheSender43 implements ApacheSender {
                         PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
                         final String[] allowedProtocols = SSLOptionsUtil.getAllowedProtocols();
                         if (allowedProtocols != null) {
-                            SSLContext sslContext;
-                            try {
-                                sslContext = SSLContexts.custom().useProtocol("TLS").build();
-                            } catch (NoSuchAlgorithmException | KeyManagementException e) {
-                                if (InternalLogger.INSTANCE.isWarnEnabled()) {
-                                    InternalLogger.INSTANCE.warn("Could not configure custom SSL context. Using system default: %s", ExceptionUtils.getStackTrace(e));
-                                }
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -53,14 +57,29 @@ static ApacheSender43 create() {
...
                 new Runnable() {
-
                     @Override
                     public void run() {
                         PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
+                        final String[] allowedProtocols = SSLOptionsUtil.getAllowedProtocols();
+                        if (allowedProtocols != null) {
+                            SSLContext sslContext;
+                            try {
+                                sslContext = SSLContexts.custom().useProtocol("TLS").build();
```
trask: This looks like same as `SSLContexts.createDefault()`

core/src/main/java/com/microsoft/applicationinsights/internal/util/SSLOptionsUtil.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Arrays;
+
+public class SSLOptionsUtil {
+
+    public static final String APPLICATION_INSIGHTS_SSL_PROTOCOLS_PROPERTY = "applicationinsights.ssl.protocols";
+
+    @Nullable
+    public static String[] getAllowedProtocols() {
+        final String rawProp = System.getProperty(APPLICATION_INSIGHTS_SSL_PROTOCOLS_PROPERTY);
+        final String[] defaultValue = {"TLSv1.2"};
```
trask: Java 11 uses TLSv1.3 by default
Java 8 uses TLSv1.2 by default
Maybe better to add special handling for Java 7 only?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -53,14 +57,29 @@ static ApacheSender43 create() {
...
+                            } catch (NoSuchAlgorithmException | KeyManagementException e) {
+                                if (InternalLogger.INSTANCE.isWarnEnabled()) {
```
trask: I'm thinking error level, or maybe even fatal(?)

core/src/main/java/com/microsoft/applicationinsights/internal/util/SSLOptionsUtil.java
littleaj: Two things here:
1. I wonder if we can add TLSv1.3 as another allowed protocol. As long as it doesn't fail on java 7, that should be fine.
2. Based on the guidance provided, we need to be explicit and allow to roll back to TLSv1 or TLSv1.1 if the customer needs that.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
littleaj: If i'm using the default SSLContext, this no longer applies.

core/src/main/java/com/microsoft/applicationinsights/internal/util/SSLOptionsUtil.java
littleaj: I added TLSv1.3 and TLSv1.2 as possible options and now it validates each value to be sure there are no errors at runtime trying to parse a protocol name.

core/src/main/java/com/microsoft/applicationinsights/internal/util/SSLOptionsUtil.java
```diff
@@ -0,0 +1,71 @@
...
+     * <p>Finds the list of supported SSL/TLS protocols. If custom protocols are specified with the system property {@value #APPLICATION_INSIGHTS_SSL_PROTOCOLS_PROPERTY}, this overrides the defaults.
+     * An error will be logged if the property contains no supported protocols.</p>
+     *
+     * <p>If no supported protocols are specified, the defaults are used (see static constructor). If no default protocols are available on this JVM, an error is logged.</p>
+     *
+     * @return An array of supported protocols. If there are none found, an empty array.
+     */
+    public static String[] getAllowedProtocols() {
+        final String rawProp = System.getProperty(APPLICATION_INSIGHTS_SSL_PROTOCOLS_PROPERTY);
+        if (rawProp == null) { // empty means defer to JVM defaults
```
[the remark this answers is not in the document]
littleaj: The result is the same, but a message is logged when the property is set to empty. I'm going to make one tweak so a warning is logged when the value is set to empty string.

I think it's an odd choice to set this property to empty and therefore warrants a warning message.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -72,8 +75,12 @@ public CdsProfileFetcher(CdsRetryPolicy retryPolicy) {
                 .setConnectionRequestTimeout(5000)
                 .build();
 
-        setHttpClient(HttpAsyncClients.custom()
-                .setDefaultRequestConfig(requestConfig)
+        final HttpAsyncClientBuilder httpAsyncClientBuilder = HttpAsyncClients.custom().setDefaultRequestConfig(requestConfig);
+        final String[] allowedProtocols = SSLOptionsUtil.getAllowedProtocols();
+        if (allowedProtocols.length > 0) {
```
trask: this is always `true` now?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -51,16 +50,23 @@
...
         final ApacheSender43 sender = new ApacheSender43();
+        final String[] allowedProtocols = SSLOptionsUtil.getAllowedProtocols();
         Thread initThread = new Thread(
                 new Runnable() {
-
                     @Override
                     public void run() {
-                        PoolingHttpClientConnectionManager cm = new PoolingHttpClientConnectionManager();
+                        final PoolingHttpClientConnectionManager cm;
+                        if (allowedProtocols.length == 0) {
```
trask: this is always `false` now?

PR #994, opened 2019-07-25
Correct the source/target compatibility property usage

littleaj:
The new version of gradle had some issue with the `tasks.withType` method. This seemed to work just fine.

### DIFF
--- a/test/fakeIngestion/servlet/build.gradle
+++ b/test/fakeIngestion/servlet/build.gradle
@@ -1,7 +1,7 @@
+apply plugin: 'java'
 apply plugin: 'war'
 
 repositories {
-    jcenter()
     mavenCentral()
 }
 
@@ -14,4 +14,4 @@ dependencies {
 
 }
 
-compileJava.sourceCompatibility = 1.8
+sourceCompatibility = 1.7
--- a/test/webapps/PerfTestApp/build.gradle
+++ b/test/webapps/PerfTestApp/build.gradle
@@ -16,7 +16,7 @@ configurations.all {
 }
 
 dependencies {
-    compile "com.microsoft.azure:applicationinsights-core:$aiSdkVersion"
+    compile project(':core')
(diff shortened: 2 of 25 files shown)


settings.gradle
```diff
@@ -36,9 +36,9 @@ if (System.env.'COLLECTD_HOME') {
 //include 'distributions'
 
 
+include ':test:webapps:PerfTestApp'
```
trask: i thought this was sample app?

test/fakeIngestion/servlet/build.gradle
```diff
@@ -1,7 +1,7 @@
+apply plugin: 'java'
 apply plugin: 'war'
 
 repositories {
-    jcenter()
```
trask: 👍

test/webapps/PerfTestApp/build.gradle
```diff
@@ -3,25 +3,18 @@ apply plugin: 'war'
 def aiSdkVersion = version
 
 dependencies {
-    compile "com.microsoft.azure:applicationinsights-core:$aiSdkVersion"
+    compile project(':core')
```
trask: 👍

trask:
Also looks like some issues with build

settings.gradle
littleaj: oops, I didn't mean to include that...will revert. I was using that application for testing.

littleaj:
@trask, build issues were because of some compatibility issues between `:test:smoke:framework:testCases` and `:test:webapps:PerfTestApp`. #995 fixes this. The inclusion of the PerfTestApp project in settings.xml was unintentional and since it's reverted, that should address the build issues.

littleaj:
I think #995 needs to be merged first. That's how I was running locally.

littleaj:
ok. Fixed a few more compatibility issues. Just waiting on builds to pass

PR #1003, opened 2019-08-02
ConnectionString Implementation

### LINKED ISSUE #935: QuickPulse needs ability to change LiveMetrics endpoint
https://github.com/microsoft/ApplicationInsights-Java/blob/5ef196e19e67f5691b5af404fe2ec0dd52049862/core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java#L38

This needs to be configurable so Live Metrics works in all environments. This will also make it more testable.

littleaj:
Skip reviewing `KeyValuePairScanner` for now (it's not fully implemented and not currently being used). It relates to point 2 below.

Things I'm still considering:
1. Validate values after parsing:
If they're invalid they will fail when the respective component tries to use them. This is basically the current behavior (component validates input before use).
2. Use `MapSplitter` instead of our own custom KVP scanner:
Pros: saves time
Cons: it's `@Beta` and could change drastically in a future version. Less control/flexibility
3. Since you can specify authorization type with `Authorization` and it could have different values in the future, perhaps `ConnectionConfiguration.ikey` should instead be an abstract object currently having one implementation, e.g. InstrumentationKeyAuthorization.
I haven't already done this because we have no information on what the next Authorization type could be.

@trask , let me know your thoughts on the current implementation. Let me know if there's a test you think should be covered.

Fix #935


core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    public static ConnectionConfiguration parse(String connectionString) throws ConnectionStringParseException {
+        // parse key value pairs
+        final Map<String, String> kvps;
+        try {
+            kvps = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            kvps.putAll(Splitter.on(';').trimResults().omitEmptyStrings().withKeyValueSeparator('=').split(connectionString));
+        } catch (IllegalArgumentException e) {
+            throw new InvalidConnectionStringException(e);
```
trask: I think add message, e.g. "Error parsing connection string: ...", and include problematic connection string (in case it's picking up connection string that user didn't expect.

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/UnsupportedAuthorizationTypeException.java
```diff
@@ -0,0 +1,22 @@
+package com.microsoft.applicationinsights.internal.config.connection;
+
+public class UnsupportedAuthorizationTypeException extends ConnectionStringParseException {
```
trask: remove unused constructors

core/src/test/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionStringParserTests.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.junit.*;
+import org.junit.rules.ExpectedException;
+
+import java.util.UUID;
+
+import static com.microsoft.applicationinsights.internal.config.connection.ConnectionString.parse;
+import static org.junit.Assert.*;
+
+public class ConnectionStringParserTests {
```
trask: test for case insensitivity would be good

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -0,0 +1,192 @@
...
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    public static ConnectionConfiguration parse(String connectionString) throws ConnectionStringParseException {
+        // parse key value pairs
+        final Map<String, String> kvps;
+        try {
+            kvps = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            kvps.putAll(Splitter.on(';').trimResults().omitEmptyStrings().withKeyValueSeparator('=').split(connectionString));
```
trask: 👍 for MapSplitter

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.AbstractMap.SimpleEntry;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class ConnectionString {
+    private ConnectionString(){}
+
+    @VisibleForTesting
+    static class KeyValuePairScanner implements Iterator<Map.Entry<String, String>> {
```
trask: I like the `MapSplitter` approach below

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -0,0 +1,192 @@
...
+        // TODO validate values?
+
+        // check for authorization
+        String authorizationType = kvps.get(ConnectionStringKeys.AUTHORIZATION);
+        if (authorizationType != null && !"ikey".equalsIgnoreCase(authorizationType)) {
+            throw new UnsupportedAuthorizationTypeException(authorizationType + " is not a supported Authorization value. Supported values: [\"ikey\"].");
+        }
+
+        // get ikey
+        String instrumentationKey = kvps.remove(ConnectionStringKeys.INSTRUMENTATION_KEY);
```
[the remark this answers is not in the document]
littleaj: that was from a previous impl where I was logging the extra values

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
littleaj: Sounds good:

```
com.microsoft.applicationinsights.internal.config.connection.InvalidConnectionStringException: Error parsing connection string: "Authorization=ikey;InstrumentationKey==fake-ikey"
	at com.microsoft.applicationinsights.internal.config.connection.ConnectionString.parseInto(ConnectionString.java:21)
	at com.microsoft.applicationinsights.internal.config.connection.ConnectionStringTests.multipleKeySeparatorsIsInvalid(ConnectionStringTests.java:194)
	...
Caused by: java.lang.IllegalArgumentException: Chunk [InstrumentationKey==fake-ikey] is not a valid entry
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:217)
	at com.google.common.base.Splitter$MapSplitter.split(Splitter.java:513)
	at com.microsoft.applicationinsights.internal.config.connection.ConnectionString.parseInto(ConnectionString.java:19)
```

core/src/main/java/com/microsoft/applicationinsights/TelemetryConfiguration.java
```diff
@@ -220,6 +226,23 @@ public void setRoleName(String roleName) {
...
 
+    public String getConnectionString() {
+        return connectionString;
+    }
+
+    public void setConnectionString(String connectionString) {
+        try {
+            ConnectionString.parseInto(connectionString, this);
+        } catch (ConnectionStringParseException e) {
+            throw new IllegalArgumentException("Invalid connection string: "+connectionString, e);
```
MS-TimothyMothra: If we assume that a connection string contains Secrets, we shouldn't emit its contents in an Exception.

core/src/test/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionStringParsingTests.java
```diff
@@ -0,0 +1,254 @@
...
+    public void nonKeyValueStringIsInvalid() throws ConnectionStringParseException {
+        exception.expect(InvalidConnectionStringException.class);
+        parseInto(UUID.randomUUID().toString(), config);
+    }
+
+    @Test
+    public void missingAuthorizationIsInvalid() throws ConnectionStringParseException {
+        exception.expect(InvalidConnectionStringException.class);
+        parseInto("LiveEndpoint=https://live.example.com", config);
+    }
```
MS-TimothyMothra: nit: Spec states that Authorization is optional because we currently only support the "ikey" value.
You're technically not wrong to require it so not a deal-breaker.

littleaj:
I have another branch where I'm adding integration tests. There's going to be a good amount of test framework changes since we can now alter all endpoints, so I'm going to create a separate PR for those changes.

The integration tests will be done before this feature is released.

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/EndpointProvider.java
```diff
@@ -0,0 +1,95 @@
...
+    @VisibleForTesting static final String INGESTION_URI_PATH = "v2/track";
+    @VisibleForTesting static final String LIVE_URI_PATH = "QuickPulseService.svc";
+    @VisibleForTesting static final String API_PROFILES_APP_ID_URI_PREFIX = "api/profiles/"; // <base uri, with host>/api/profiles/<ikey>/appid
+    @VisibleForTesting static final String API_PROFILES_APP_ID_URI_SUFFIX = "/appId";
+
+    private final AtomicReference<URI> ingestionEndpoint = new AtomicReference<>();
+    private final AtomicReference<URI> ingestionEndpointURL = new AtomicReference<>();
+    private final AtomicReference<URI> liveEndpointURL = new AtomicReference<>();
+    private final AtomicReference<URI> profilerEndpoint = new AtomicReference<>();
+    private final AtomicReference<URI> snapshotEndpoint = new AtomicReference<>(); // TODO is this one needed?
```
trask: i think use `volatile` instead of `AtomicReference`, since not doing any atomic operations on them

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -50,19 +52,20 @@
...
+    public static final String DEFAULT_SERVER_URI = "https://dc.services.visualstudio.com/v2/track";
 
     // For future use: re-send a failed transmission back to the dispatcher
     private TransmissionDispatcher transmissionDispatcher;
 
-    private final String serverUri;
+    private String serverUri;
 
     private volatile boolean stopped;
+    private final AtomicReference<TelemetryConfiguration> configuration = new AtomicReference<>();
```
trask: i think use `volatile` instead of `AtomicReference`, since not doing any atomic operations on it

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionStringParseException.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.applicationinsights.internal.config.connection;
+
+public abstract class ConnectionStringParseException extends Exception {
+
+    ConnectionStringParseException() {
```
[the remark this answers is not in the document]
littleaj: It's used in catch clauses and throws declarations. I suppose it's not needed. I could refactor to use multi-catch, if you prefer.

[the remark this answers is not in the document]
trask: oh, i just meant is the no-arg constructor used

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -61,10 +61,15 @@
     private final CdsRetryPolicy retryPolicy;
 
     public CdsProfileFetcher() {
-        this(new CdsRetryPolicy());
+        this(new CdsRetryPolicy(), TelemetryConfiguration.getActive());
```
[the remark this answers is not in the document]
littleaj: I meant to deprecate the other two constructors and only use the one with the configuration argument, but still wanted them to work until removed.

core/src/test/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactoryTest.java
```diff
@@ -44,21 +43,26 @@
...
 import static org.hamcrest.Matchers.*;
 import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
 
 import static org.junit.Assert.*;
 
 public final class TelemetryConfigurationFactoryTest {
 
+    @Rule
+    public EnvironmentVariables envVars = new EnvironmentVariables();
```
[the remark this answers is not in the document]
littleaj: I know! I wish I had this in my toolbox years ago...

littleaj:
Also, I'm going to remove the strict parsing of `Authorization`. The current version of connection string will only support `InstrumentationKey`. The value of `Authorization` can be safely ignored until properly supported.

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -0,0 +1,143 @@
...
+            throw new InvalidConnectionStringException("Could not parse connection string.");
+        }
+
+        mapToConnectionConfiguration(kvps, targetConfig);
+    }
+
+    private static void mapToConnectionConfiguration(Map<String, String> kvps, TelemetryConfiguration config) throws ConnectionStringParseException {
+        // check for authorization
+        String authorizationType = kvps.get(Keywords.AUTHORIZATION);
+        if (!(Strings.isNullOrEmpty(authorizationType) || "ikey".equalsIgnoreCase(authorizationType))) {
```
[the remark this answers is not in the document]
littleaj: It's in my nature to use the minimal number of logic operations 😸. This is being removed anyway because it relates to the `Authorization` kvp

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/EndpointProvider.java
littleaj: In this case, I think it's equivalent since URI is immutable.

web/src/main/java/com/microsoft/applicationinsights/web/extensibility/modules/WebRequestTrackingTelemetryModule.java
```diff
@@ -70,7 +70,9 @@ public WebRequestTrackingTelemetryModule() {}
      * @param configurationData SDK config Object
      */
     public WebRequestTrackingTelemetryModule(Map<String, String> configurationData) {
-        assert configurationData != null;
```
littleaj: @trask, I removed the all the assert statements.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
```diff
@@ -94,17 +97,18 @@ public void setEnabled(boolean enabled) {
   }
 
   public String getInstrumentationKey() {
-    // First try getting ikey from application.properties
```
littleaj: These were being evaluated in the wrong order.

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
```diff
@@ -205,22 +224,17 @@ public void setInProcess(InProcess inProcess) {
        */
       private boolean developerMode = false;
       /** Endpoint address. */
-      private String endpointAddress = TransmissionNetworkOutput.DEFAULT_SERVER_URI;
```
littleaj: We want to allow `null` so we can tell if a custom endpoint is set in application.properties

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
```diff
@@ -50,8 +50,7 @@ private TraceContextCorrelation() {}
      * @param response
      * @param requestTelemetry
      */
-    public static void resolveCorrelation(HttpServletRequest request, HttpServletResponse response,
-        RequestTelemetry requestTelemetry) {
+    public static void resolveCorrelation(HttpServletRequest request, HttpServletResponse response, RequestTelemetry requestTelemetry) {
```
trask: can you minimize formatting changes in this file to help with merge conflicts since it has several changes also in the codeless branch?

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -99,11 +95,30 @@ public CdsProfileFetcher(CdsRetryPolicy retryPolicy, TelemetryConfiguration conf
...
+        try {
+            return internal_fetchAppProfile(instrumentationKey, configuration);
+        } catch (ExecutionException | IOException e) {
+            throw new ApplicationIdResolutionException(e);
+        }
+    }
 
-        if (instrumentationKey == null || instrumentationKey.isEmpty()) {
-            throw new IllegalArgumentException("instrumentationKey must be not null or empty");
+    private ProfileFetcherResult internal_fetchAppProfile(String instrumentationKey, TelemetryConfiguration configuration) throws InterruptedException, ExecutionException, IOException {
```
trask: `internal_fetchAppProfile` ==> `internalFetchAppProfile`

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -37,16 +38,16 @@
...
+import java.net.URI;
 import java.net.URL;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
-public class CdsProfileFetcher implements AppProfileFetcher {
+public class CdsProfileFetcher implements AppProfileFetcher, ApplicationIdResolver {
```
trask: what was the reason for not passing in config to constructor, and instead passing it in fetch method?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/helpers/IkeyResolver.java
```diff
@@ -33,22 +33,22 @@
   private static final String EXTERNAL_PROPERTY_IKEY_NAME_SECONDARY = "APPINSIGHTS_INSTRUMENTATIONKEY";
 
     public static String getIkeyFromEnvironmentVariables() {
-    String v = System.getProperty(EXTERNAL_PROPERTY_IKEY_NAME);
+    String v = System.getenv(EXTERNAL_PROPERTY_IKEY_NAME);
```
trask: 👍

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
```diff
@@ -94,17 +97,18 @@ public void setEnabled(boolean enabled) {
   }
 
   public String getInstrumentationKey() {
-    // First try getting ikey from application.properties
-    if (StringUtils.isNotBlank(instrumentationKey)) {
-      return instrumentationKey;
-    }
 
     // If above fails try getting ikey from environment variables or system properties
```
trask: remove `If above fails`

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsProperties.java
```diff
@@ -94,17 +97,18 @@ public void setEnabled(boolean enabled) {
...
+    if (v != null) {
+      return v;
     }
-    return v;
+
+    // Try getting ikey from application.properties
+    if (StringUtils.isNotBlank(instrumentationKey)) {
+      return instrumentationKey;
+    }
+    throw new IllegalStateException("Instrumentation Key must be set to report telemetry");
```
trask: throwing exception?

azure-application-insights-spring-boot-starter/src/main/java/com/microsoft/applicationinsights/autoconfigure/ApplicationInsightsTelemetryAutoConfiguration.java
```diff
@@ -125,8 +125,11 @@ public void setTelemetryProcessors(
...
     @Bean
     @DependsOn("internalLogger")
-    public TelemetryConfiguration telemetryConfiguration(TelemetryChannel telemetryChannel) {
+    public TelemetryConfiguration telemetryConfiguration() {
         TelemetryConfiguration telemetryConfiguration = TelemetryConfiguration.getActiveWithoutInitializingConfig();
+        if (StringUtils.isNotEmpty(applicationInsightsProperties.getConnectionString())) {
+            telemetryConfiguration.setConnectionString(applicationInsightsProperties.getConnectionString());
+        }
         telemetryConfiguration.setTrackingIsDisabled(!applicationInsightsProperties.isEnabled());
         telemetryConfiguration.setInstrumentationKey(applicationInsightsProperties.getInstrumentationKey());
```
trask: will this throw exception if using connection string and not instrumentation key directly?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/ConfiguredTransmitterFactory.java
```diff
@@ -2,6 +2,17 @@
...
+    /**
+     * Either {@code configuration} or {@code endpoint} could be null, but one must be non-null.
+     * @param configuration The configuration for the current TelemetryClient
+     * @param endpoint The endpoint to use.
+     * @param maxTransmissionStorageCapacity
+     * @param throttlingIsEnabled
+     * @param maxInstantRetries
+     * @return
+     */
+    TelemetriesTransmitter<T> create(@Nullable TelemetryConfiguration configuration, @Nullable String endpoint, String maxTransmissionStorageCapacity, boolean throttlingIsEnabled, int maxInstantRetries);
```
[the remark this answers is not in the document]
littleaj: This is what I had at first, but it seemed easier to maintain backwards compatibility if we included it here. I'll see if I can refactor it out because I would prefer it not be there (we will have to remove it eventually anyway)

test/smoke/appServers/Jetty.9/jetty9.linux.partial.dockerfile
```diff
@@ -30,7 +30,6 @@ ENV JETTY_STOP_KEY stopitplease
 WORKDIR /root/docker-stage
 
 RUN java -jar $JETTY_HOME/start.jar jetty.base=$JETTY_BASE --add-to-start=http,jsp,deploy,jstl --update-ini jetty.http.port=8080 --update-ini jetty.deploy.extractWars=true
-RUN java -jar $JETTY_HOME/start.jar jetty.base=$JETTY_BASE --add-to-start=console-capture --update-ini jetty.console-capture.timezone=PST
```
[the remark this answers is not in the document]
littleaj: Yes, this is what was causing stdout to be sent to file and not printed to console.

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/AppProfileFetcher.java
```diff
@@ -29,16 +29,18 @@
...
+public interface AppProfileFetcher extends Closeable {
+    /**
+     * Fetches the application profile and returns the appId corresponding to the
+     * instrumentation key provided.
+     * @param instrumentationKey The instrumentation key for which to fetch the appId.
+     * @throws ExecutionException
+     * @throws InterruptedException
+     * @throws IOException
+     * @throws ParseException
+     */
```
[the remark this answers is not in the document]
littleaj: I'm not sure why the diff is showing up like this. I'm seeing LFs locally and I have core.autocrlf set to input

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
littleaj: This class is only used as a member of InstrumentationKeyResolver which is an enum singleton and therefore can't take parameters to its constructor.

IMO we should try to avoid singletons as much as possible.

trask: make sense

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/TraceContextCorrelation.java
littleaj: makes sense

I'd like to set a max width in checkstyle that makes sense for modern monitors. I found this file hard to read.

trask: Yeah, i'm not opposed to reformatting in general, just not in this file for now :-)

web/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/AppProfileFetcher.java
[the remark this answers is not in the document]
trask: Ah, your commit removed a space at the beginning of all of the lines (which is good) 😅

PR #985, opened 2019-07-19
Normalize white space

trask:
Newlines:

  git add --renormalize .

Tabs:

  git grep -l -P '\t' -- '*.java' '*.gradle' '*.xml' '*.jsp' '*.sh' '*.dockerfile' '*.proto' '*.md' | xargs sed -i 's/\t/    /g'

  manually updated NOTICE

  ignored .cpp/.h

I will rebase and re-do above steps (and force push) this PR once other pending PRs are merged to avoid causing merge conflicts for them.

PR #1006, opened 2019-08-05
Normalize white space

trask:
Remove trailing spaces

```
git grep -l -P ' $' -- '*.java' '*.gradle' '*.xml' '*.jsp' '*.sh' '*.dockerfile' '*.proto' '*.md' | xargs sed -i 's/ \+$//g'
```

@littleaj I'll wait until you merge the 2.4.1 fix into master so it doesn't cause conflicts for you, then I'll re-run / force push if looks good.


littleaj:
@trask, is this going to also convert tabs to spaces?

trask:
@littleaj, this only removes trailing spaces, i think we got rid of all tabs in #985

PR #1010, opened 2019-08-07
Remove commons-logging.properties

trask:
It does not appear to be doing anything currently since it is applying
OFF to unshaded apache http client package instead of shaded package
name.

Also, if monitored application is using commons logging, it could pick
up this file and cause different logging behavior in the monitored
application.


trask:
@littleaj 👍 smoke tests look good (no http logging)

PR #1011, opened 2019-08-07
add asserts to verify response code.

### LINKED ISSUE #810: Application Insights showing 200 for failed requests
Even from the latest application insights version **2.3.0**, the Response code issue which is described in the below fix is not resolved,

https://github.com/Microsoft/ApplicationInsights-Java/pull/603

Did we find any better alternatives or solution for the issue?

With Regards,
Dinesh Ayyapillai.

littleaj:
This should Fix #810

### DIFF
--- a/web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
+++ b/web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
@@ -14,6 +14,8 @@ import java.util.Date;
 import java.util.List;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
+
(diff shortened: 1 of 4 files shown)


littleaj:
latest commit should fix #600 (pending review)

web/src/main/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandler.java
```diff
@@ -112,6 +114,11 @@ public void handleEnd(ServletRequest request, ServletResponse response,
         int resultCode = extractor.getStatusCode(response);
         requestTelemetry.setSuccess(resultCode < 400);
         requestTelemetry.setResponseCode(Integer.toString(resultCode));
```
trask: I think these three lines above can go in a conditional `... == null` and then the `if` below can be `else`

PR #1012, opened 2019-08-09
Better agent diagnostic logging

trask:
@littleaj thoughts on system property name `applicationInsights.logging.level`? It currently only applies to agent logging, so maybe should be more scoped, e.g. `applicationInsights.agent.logging.level`?


littleaj:
@trask, I like the `agent` qualifier.

PR #1016, opened 2019-08-09
Minor reworkings

littleaj:
Some style changes, some log message clarification, removed some redundancy

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
@@ -14,8 +14,7 @@ public class HeartbeatDefaultPayload {
   /**
    * List of default payloads which would be added
    */
-  private static final List<HeartBeatPayloadProviderInterface> defaultPayloadProviders =
-      new ArrayList<>();
+  private static final List<HeartBeatPayloadProviderInterface> defaultPayloadProviders = new ArrayList<>();
 
   static {
     defaultPayloadProviders.add(new DefaultHeartBeatPropertyProvider());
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
```diff
@@ -59,19 +58,17 @@ public static boolean addDefaultPayLoadProvider(HeartBeatPayloadProviderInterfac
...
         for (HeartBeatPayloadProviderInterface payloadProvider : defaultPayloadProviders) {
           if (disabledProviders != null && disabledProviders.contains(payloadProvider.getName())) {
 
             // skip any azure specific modules here
             continue;
           }
 
-          boolean fieldsAreSet = payloadProvider.setDefaultPayload(disabledFields, provider).call();
-          populatedFields = populatedFields || fieldsAreSet;
+          populatedFields |= payloadProvider.setDefaultPayload(disabledFields, provider).call();
```
trask: I think this makes it a bit harder to read, since `x |= y` for booleans is easily confused with `x = x || y`. I see you are making very nice usage of the non-short circuiting behavior of `|=`, but it took me a while to parse it correctly in my head 😭.

core/src/main/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatDefaultPayload.java
```diff
@@ -59,19 +58,17 @@ public static boolean addDefaultPayLoadProvider(HeartBeatPayloadProviderInterfac
...
         for (HeartBeatPayloadProviderInterface payloadProvider : defaultPayloadProviders) {
           if (disabledProviders != null && disabledProviders.contains(payloadProvider.getName())) {
 
             // skip any azure specific modules here
             continue;
           }
 
-          boolean fieldsAreSet = payloadProvider.setDefaultPayload(disabledFields, provider).call();
-          populatedFields = populatedFields || fieldsAreSet;
+          populatedFields = populatedFields || payloadProvider.setDefaultPayload(disabledFields, provider).call();
```
trask: I don't know if this was intentional, but before it was still executing `setDefaultPayload` on every loop, but now it will not execute `setDefaultPayload` once `populatedFields` is `true`

littleaj: Let me see if there are side effects of setDefaultPayload. It may not need to run every loop

littleaj:
```suggestion
          boolean fieldsAreSet = payloadProvider.setDefaultPayload(disabledFields, provider).call();
          populatedFields = populatedFields || fieldsAreSet;
```

PR #1017, opened 2019-08-09
Fix GCPerformanceCounter not working

### LINKED ISSUE #929: It's not possible to turn on GCPerformanceCounter using <JvmPC name="GC" enabled="true" />
It's not possible to turn on GCPerformanceCounter with such configuration:

```
 <PerformanceCounters collectionFrequencyInSec="60">
        <UseBuiltIn>true</UseBuiltIn>
        <Jvm enabled="true">
            <JvmPC name="GC" enabled="true" />
        </Jvm>
    </PerformanceCounters>
```

Is it possible to turn it on somehow?

trask:
Fix #929

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/jvm/GCPerformanceCounter.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/jvm/GCPerformanceCounter.java
@@ -80,7 +80,7 @@ public final class GCPerformanceCounter implements PerformanceCounter {
             MetricTelemetry mtTotalTime = new MetricTelemetry(GC_TOTAL_TIME, timeToReport);
 
             mtTotalCount.markAsCustomPerfCounter();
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/jvm/GCPerformanceCounter.java
```diff
@@ -80,7 +80,7 @@ public void report(TelemetryClient telemetryClient) {
             MetricTelemetry mtTotalTime = new MetricTelemetry(GC_TOTAL_TIME, timeToReport);
 
             mtTotalCount.markAsCustomPerfCounter();
-            mtTotalCount.markAsCustomPerfCounter();
+            mtTotalTime.markAsCustomPerfCounter();
```
littleaj: :) good catch

littleaj:
@trask, is this going in beta3 or later?

PR #1018, opened 2019-08-12
Add App Services sample

trask:


trask:
@littleaj, I renamed the existing sample to `misc`, open to suggestions on better naming

trask:
Added `README.md` and fixed `groupId`

PR #1024, opened 2019-08-20
Test is being flaky

littleaj:
I think this is why the test is failing

### DIFF
--- a/web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
+++ b/web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
@@ -13,8 +13,6 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
-import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
 import org.mockito.InOrder;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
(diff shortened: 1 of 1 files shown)


web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
```diff
@@ -161,7 +157,11 @@ public void timeStatusCodeAreSetWhenHandleEndIsInvoked() throws MalformedURLExce
         assertThat(rt.getUrl().toString(), equalTo(url));
         assertThat(rt.getContext().getUser().getUserAgent(), equalTo("User-Agent"));
         assertThat(rt.getTimestamp(), is(CoreMatchers.<Date>notNullValue()));
-
+        try {
+            TimeUnit.MILLISECONDS.sleep(100); // pause for a moment
```
trask: Can you add comment why `sleep` is needed?

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
```diff
@@ -161,7 +157,11 @@ public void timeStatusCodeAreSetWhenHandleEndIsInvoked() throws MalformedURLExce
...
+        } catch (InterruptedException e) {
```
trask: No need to catch this

littleaj: yup. will rethrow

web/src/test/java/com/microsoft/applicationinsights/web/internal/httputils/HttpServerHandlerTest.java
littleaj: will do

PR #1028, opened 2019-08-22
Reduce the logging output for "Backoff has been reset"

### LINKED ISSUE #894: Remove/suppress "Backoff has been reset" log message
https://github.com/Microsoft/ApplicationInsights-Java/blob/71d58c4a45afe8b8df286a61db49de36e4405012/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionPolicyManager.java#L129-L133

This log message should, at the very least, be a `TRACE`, but at this point I'm voting to remove. During testing, logs get flooded with this. I don't find it useful.

Perhaps it could be more state-based: if backoff has been enabled, **only then** log when it's reset. I'd find that more useful, but as long as there's a message for when backoff time is increased as well. The state changes are the most interesting parts of this procedure.

jdubois:
As mentioned in #894 this message floods the console during development. I don't think it should be totally removed (anyway it shouldn't be commented out!), as it can be interesting to know everything went well, when you are debugging.

Fix #894


littleaj:
@jdubois, the issue with the message was that it was logged unconditionally, but I do agree that this message isn't always useless and that it should be a trace message (not commented out :) ). I created #1030 from your changes which logs only when the state has changed.

PR #1045, opened 2019-08-30
Reduce sporadic test failures

trask:


test/smoke/testApps/CustomInstrumentation/src/smokeTest/java/com/microsoft/applicationinsights/smoketestapp/CustomInstrumentationTest.java
```diff
@@ -47,12 +44,9 @@ public void customInstrumentationOne() {
...
     @TargetUri("/customInstrumentationTwo")
-    public void customInstrumentationTwo() {
-        List<Envelope> rdList = mockedIngestion.getItemsEnvelopeDataType("RequestData");
-        List<Envelope> rddList = mockedIngestion.getItemsEnvelopeDataType("RemoteDependencyData");
-
-        assertThat(rdList, hasSize(1));
-        assertThat(rddList, hasSize(1));
+    public void customInstrumentationTwo() throws Exception {
+        List<Envelope> rdList = mockedIngestion.waitForItems("RequestData", 1);
+        List<Envelope> rddList = mockedIngestion.waitForItems("RemoteDependencyData", 1);
```
littleaj: Before this change, if !=1 items were received it would throw AssertionException causing a test failure. Now this will throw IllegalStateException and test fails with an error. Perhaps waitForItems should fail with some type of AssertionException instead. The tearDown flow seems to be more stable when it follows that path.

Perhaps in the future we could have something like: `mockedIngestion.assertReceived(numberOfItems(1), ofType("RequestData"))`, but we don't have to do this now.

test/fakeIngestion/standalone/src/main/java/com/microsoft/applicationinsights/test/fakeingestion/MockedAppInsightsIngestionServer.java
```diff
@@ -104,6 +106,17 @@ public Envelope waitForItem(Predicate<Envelope> condition, int timeout, TimeUnit
...
     }
 
+    public List<Envelope> waitForItems(final String type, final int numItems) throws Exception {
+        List<Envelope> items = waitForItems(new Predicate<Envelope>() {
+            @Override public boolean apply(Envelope input) {
+                return input.getData().getBaseType().equals(type);
+            }
+        }, numItems, 10, TimeUnit.SECONDS);
+        if (items.size() > numItems) {
+            throw new IllegalStateException("Received too many items of type " + type);
```
trask: @littleaj how about this?
```suggestion
            throw new AssertionError("Expecting " + numItems + " of type " + type + ", but received " + items.size());
```

test/smoke/testApps/CustomInstrumentation/src/smokeTest/java/com/microsoft/applicationinsights/smoketestapp/CustomInstrumentationTest.java
trask: 👍 I made changed from `IllegalStateException` to `AssertionError`, and improved message in the standard form of "expected X but received Y".

PR #1059, opened 2019-09-06
smoketest updates for connection strings

littleaj:
<strike>Waiting on #1003 to merge (this is a subbranch)</strike>

Adds connections string to smoketest app configs

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/ConfiguredTransmissionOutput.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/ConfiguredTransmissionOutput.java
@@ -1,7 +1,0 @@
-package com.microsoft.applicationinsights.internal.channel;
-
-import com.microsoft.applicationinsights.TelemetryConfiguration;
-
-public interface ConfiguredTransmissionOutput extends TransmissionOutput {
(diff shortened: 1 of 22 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/ConfiguredTransmissionOutput.java
```diff
@@ -1,7 +0,0 @@
-package com.microsoft.applicationinsights.internal.channel;
```
littleaj: I had intended to remove this in the last pr, but this is fine too

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -263,10 +263,10 @@ private HttpPost createTransmissionPostRequest(Transmission transmission) {
         return request;
     }
 
-    @Override
-    public void setConfiguration(TelemetryConfiguration configuration) {
-        this.configuration = configuration;
-    }
+//    @Override
```
littleaj: will delete

Issue #790, opened 2018-12-18
WebFlux and Netty APIs are not supported in Springboot.


dhaval24:
Update this https://docs.microsoft.com/en-us/java/azure/spring-framework/configure-spring-boot-java-applicationinsights?view=azure-java-stable document - Web Flux and Netty APIs are not supported in SpringBoot.

mercer:
Any plans to offer support?

dhaval24:
@mercer we definitely want to better support webflux and netty. I do not have a timeline for this yet, but this is definitely in our backlogs.
We are very open to contributions as well. Please feel free to contribute/raise a PR if you could :)

marceloverdijk:
Is there an update on this?
I’m using Micronaut (based on Netty) and would like to setup trace logging using Application Insights.
As Netty has no notion of a `java.servlet.Filter` I’m not able to setup the integration unfortunately - as from the docs the integration is based on `Java’s.servlet.*`.

dhaval24:
@marceloverdijk unfortunately, there is still no update on this. However, this feature is definitely on our backlog to support Netty.  If you are open to submit a PR, we would welcome your contribution and I can help review the code and ensure a successful contribution. 

Currently the web integration is based on servelet stack. However, you ca definitely track the requests manually by using `TelemetryClient.trackRequeust()` if you would like to do.

marceloverdijk:
@dhaval24 thx for the feedback.
I will look into the `TelemetryClient.trackRequeust()`

Does the trace logging via Logback as described here https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-trace-logs work without the http filter?

That links refers to the http filter, so it's not clear what is needed to setup this trace logging.

dhaval24:
@marceloverdijk yes log4j and logback would work without http filter. They have no dependency on it.

marceloverdijk:
@dhaval24 that sounds promising! I have some further questions related to the setup in that case, but I don't think it's best to hijack this issue for that. Should I create a new issue with the question, or is there another channel that I can use best?

dhaval24:
@mercer yes please create a new issue for it. This issue is mostly to track Netty and WebFlux support :)

dsibilio:
I've noticed the Pull Request which added Netty/Webflux support has been merged into the **codeless** branch (https://github.com/microsoft/ApplicationInsights-Java/pull/1074).

[...]

PR #1074, opened 2019-09-25
Add Netty support

trask:
This also brings WebFlux support.


settings.gradle
```diff
@@ -55,6 +55,7 @@ include ':test:smoke:appServers:Tomcat.8.5'
 include ':test:smoke:appServers:JBossEAP.7'
 include ':test:smoke:appServers:JBossEAP.6'
 include ':test:smoke:appServers:Jetty.9'
+include ':test:smoke:appServers:JavaSE'
```
littleaj: nice!

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -553,26 +563,30 @@ private static void startTestApplicationContainer() throws Exception {
 
         final ContainerInfo containerInfo = new ContainerInfo(containerId, currentImageName);
         currentContainerInfo.set(containerInfo);
-        try {
-            String url = String.format("http://localhost:%s/", String.valueOf(appServerPort));
-            System.out.printf("Verifying appserver has started (%s)...%n", url);
+        if (currentImageName.startsWith("javase_")) {
```
littleaj: what's the rationale here?

littleaj:
@trask, this PR resolves #790. We need to remove that line from the docs also :)

test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
trask: 👍 Good point, I was thinking with Java SE app there would not be an app server to handle the url, but that's not the case.

trask: I forgot why 😭. There was a reason. I added comment to explain.

trask:
Since this is not going into master, I'm not sure if we should mark it as resolving #790.

PR #1085, opened 2019-10-01
Diagnostic logs in JSON

littleaj:
This writes diagnostic logs to file in json for use in azure app services.

### DIFF
--- a/agent-codeless/build.gradle
+++ b/agent-codeless/build.gradle
@@ -143,11 +143,13 @@ jar {
     dependsOn shadowJar2
     enabled = false
 }
-
+def logbackContribVersion = '0.1.5'
 dependencies {
     compile 'org.slf4j:slf4j-api:1.7.26'
     compile 'org.slf4j:jcl-over-slf4j:1.7.26' // this is needed to pick up logging from apache http client
     compile 'ch.qos.logback:logback-classic:1.2.3'
+    compile "ch.qos.logback.contrib:logback-json-classic:$logbackContribVersion"
+
     // not using gson because it has dependency on java.sql.*, which is not available in Java 9+ bootstrap class loader
     // only complaint so far about moshi is that it doesn't give line numbers when there are json formatting errors
     compile 'com.squareup.moshi:moshi:1.8.0'
@@ -174,6 +176,9 @@ dependencies {
     compile(project(path: ':core', configuration: 'shadow'))
 
     testCompile 'junit:junit:4.12'
+    testCompile 'org.hamcrest:hamcrest-core:1.3'
+    testCompile 'org.hamcrest:hamcrest-library:1.3'
+    testCompile 'com.github.stefanbirkner:system-rules:1.19.0'
 }
 
 // region Publishing properties
--- a/agent-codeless/build.gradle
+++ b/agent-codeless/build.gradle
@@ -179,6 +179,7 @@ dependencies {
     testCompile 'org.hamcrest:hamcrest-core:1.3'
     testCompile 'org.hamcrest:hamcrest-library:1.3'
     testCompile 'com.github.stefanbirkner:system-rules:1.19.0'
+    testCompile 'org.mockito:mockito-core:1.10.19'
 }
 
 // region Publishing properties
--- a/agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/InstrumentationKeyFinder.java
+++ b/agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/InstrumentationKeyFinder.java
@@ -1,4 +1,4 @@
-package com.microsoft.applicationinsights.agentc.internal.diagnostics.log;
+package com.microsoft.applicationinsights.agentc.internal.diagnostics;
 
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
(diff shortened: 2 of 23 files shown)


agent-codeless/build.gradle
```diff
@@ -143,11 +143,13 @@ jar {
     dependsOn shadowJar2
     enabled = false
 }
-
+def logbackContribVersion = '0.1.5'
 dependencies {
     compile 'org.slf4j:slf4j-api:1.7.26'
     compile 'org.slf4j:jcl-over-slf4j:1.7.26' // this is needed to pick up logging from apache http client
     compile 'ch.qos.logback:logback-classic:1.2.3'
+    compile "ch.qos.logback.contrib:logback-json-classic:$logbackContribVersion"
```
littleaj: there used to be more dependencies which is why i declared a variable


...and I wanted to try out multi-line comments

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/InstrumentationKeyFinder.java
```diff
@@ -0,0 +1,41 @@
...
+    public String getName() {
+        return "ikey";
+    }
+
+    @Nullable
+    @Override
+    public String getValue() {
+        try {
+            return Global.getTelemetryClient().getContext().getInstrumentationKey();
+        } catch (NullPointerException npe) {
```
trask: catching `NullPointerException` looks odd, can we null check somehow instead?

also, are we checking this just for performance vs just always using getting from env vars below?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/ResourceIdFinder.java
```diff
@@ -0,0 +1,43 @@
...
+        String diagLogMountPath = System.getenv(ResourceIdFinder.DIAGNOSTIC_LOGS_MOUNT_PATH_ENV_VAR);
+
+        if (!Strings.isNullOrEmpty(diagLogMountPath)) {
+            try {
+                String resId = new String(Files.readAllBytes(Paths.get(diagLogMountPath, ResourceIdFinder.RESOURCE_ID_FILE_NAME)), StandardCharsets.UTF_8);
+                if (!resId.isEmpty()) {
+                    return resId;
+                }
+            } catch (Exception e) {
+                // nop
```
trask: Log this over normal logback channel?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/SiteNameFinder.java
```diff
@@ -0,0 +1,25 @@
...
+        return SITE_NAME_FIELD_NAME;
+    }
+
+    @Override
+    protected String populateValue() {
+        String fromEnv = System.getenv(SiteNameFinder.WEBSITE_SITE_NAME_ENV_VAR);
+        if (Strings.isNullOrEmpty(fromEnv)) {
+            return null;
+        }
+        return fromEnv;
```
trask: Check out `Strings.emptyToNull()` 👍

agent-codeless/src/test/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/log/MoshiJsonFormatterTests.java
```diff
@@ -0,0 +1,46 @@
...
+    public void formatterSerializesSimpleMap() throws Exception {
+        Map<String, Object> m = new HashMap<>();
+        m.put("s1", "v1");
+        m.put("int1", 123);
+        m.put("b", true);
+        assertEquals("{\"b\":true,\"int1\":123,\"s1\":\"v1\"}", formatter.toJsonString(m));
+    }
+
+    @Test
+    public void formatterWithPrettyPrintPrintsPretty() throws Exception {
```
trask: Do we need pretty print option?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/InstrumentationKeyFinder.java
```diff
@@ -0,0 +1,41 @@
...
+            // ignore. TelemetryClient not initialized yet.
+        }
+        final String connStr = System.getenv(TelemetryConfigurationFactory.CONNECTION_STRING_ENV_VAR_NAME);
+        if (!Strings.isNullOrEmpty(connStr)) {
+            try {
+                // see ConnectionString.parseInto
+                Map<String, String> kvps = new HashMap<>(Splitter.on(';').trimResults().omitEmptyStrings().withKeyValueSeparator('=').split(connStr));
+                return kvps.get(Keywords.INSTRUMENTATION_KEY);
+            } catch (Exception e) {
+                return null;
```
trask: log exception via normal logback channel?

agent-codeless/src/test/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/log/MoshiJsonFormatterTests.java
littleaj: no, but I've been using it for development. The prod config won't have it on.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/InstrumentationKeyFinder.java
littleaj: `Global.getTelemetryClient()` throws npe if client is not set. getContext is never null.

I was thinking this would be the most reliable, but we can defer to env vars if you think that is sufficient.

PR #1088, opened 2019-10-07
status file

littleaj:
Diagnostics file for app services and possibly others in the future.


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/PidFinder.java
```diff
@@ -0,0 +1,39 @@
...
+import java.lang.management.RuntimeMXBean;
+
+public class PidFinder extends CachedDiagnosticsValueFinder {
+    public static final String PROPERTY_NAME = "PID";
+
+    @Nullable
+    @Override
+    protected String populateValue() {
+        // will only work with sun based jvm
+        final RuntimeMXBean rb = ManagementFactory.getRuntimeMXBean();
```
Mikhail-msft: This looks fragile.
Can we use Java 9 API for this? ProcessHandle.current().pid();
and as a fallback (in case we need to support older jvms) - use the random string.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,138 @@
...
+            public void run() {
+                Map<String, Object> map = getJsonMap();
+
+                String fileName = constructFileName(map);
+
+                final File file = new File(directory, fileName);
+                boolean dirsWereCreated = file.getParentFile().mkdirs();
+                if (dirsWereCreated || file.getParentFile().exists()) {
+                    try (BufferedSink buffer = Okio.buffer(Okio.sink(file))) {
+                        new Builder().build().adapter(Map.class).indent(" ").nullSafe().toJson(buffer, map);
```
Mikhail-msft: that's all magic to me, we need Trask to review those.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,138 @@
...
+                    } catch (IOException e) {
```
Mikhail-msft: we want to catch all - no matter what happens during logging the status, the attach should not be broken

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,138 @@
...
+            @Override
+            public void run() {
+                Map<String, Object> map = getJsonMap();
+
+                String fileName = constructFileName(map);
+
+                final File file = new File(directory, fileName);
+                boolean dirsWereCreated = file.getParentFile().mkdirs();
+                if (dirsWereCreated || file.getParentFile().exists()) {
+                    try (BufferedSink buffer = Okio.buffer(Okio.sink(file))) {
```
Mikhail-msft: please make sure that proper flags are used for createfile: FileMode.Create, FileAccess.Write, FileShare.Read, FileOptions.DeleteOnClose | FileOptions.WriteThrough

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/MachineNameFinder.java
```diff
@@ -0,0 +1,28 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class MachineNameFinder extends CachedDiagnosticsValueFinder {
+    public static final String PROPERTY_NAME = "MachineName";
+
+    @Nullable
+    @Override
+    protected String populateValue() {
```
Mikhail-msft: is WEBSITES_SITENAME is covered somewhere else?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
```diff
@@ -123,9 +129,7 @@ private static void addLibJars(Instrumentation instrumentation, File agentJarFil
 
     private static void start(Instrumentation instrumentation, File agentJarFile) throws Exception {
 
-        Path agentJarPath = agentJarFile.toPath();
-        Path sdkNamePrefixPath = agentJarPath.resolveSibling("appsvc.codeless");
-        if (Files.exists(sdkNamePrefixPath)) {
+        if (DiagnosticsHelper.isAppService()) {
```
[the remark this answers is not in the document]
littleaj: Perhaps. I can rename.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/PidFinder.java
littleaj: Since we need to maintain Java 7/8 compatibility, I can't use it directly. I could call it via reflection, but this will work for oracle/openjdk based jvms and in app services, this will always be the case.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/MachineNameFinder.java
littleaj: Yes, here: https://github.com/microsoft/ApplicationInsights-Java/blob/codeless/agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/SiteNameFinder.java

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/PidFinder.java
Mikhail-msft: We plan to re-use this code outside of AppSvcs, where any jvm could be present. Let's have a following logic: java9+ - reflection, else - the current code, else - random string?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
littleaj: The Okio library creates the file. Does it need to delete on close?

I don't think WriteThrough is possible with java.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/PidFinder.java
littleaj: random string?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
Mikhail-msft: yes, we want the file to be deleted on process termination

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,197 @@
...
+        return buffer;
+    }
+
+    @VisibleForTesting
+    static Map<String, Object> getJsonMap() {
+        Map<String, Object> map = new LinkedHashMap<>(CONSTANT_VALUES);
+        for (DiagnosticsValueFinder finder : VALUE_FINDERS) {
+            final String value = finder.getValue();
+            if (!Strings.isNullOrEmpty(value)) {
+                map.put(capitalize(finder.getName()), value);
```
trask: Is `capitalize()` needed? it looks like names are already capitalized?

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @VisibleForTesting
+    static final String FILE_EXTENSION = ".json";
+
+    private static final String DEFAULT_STATUS_FILE_DIRECTORY = "/home/LogFiles/ApplicationInsights/status";
+
+    @VisibleForTesting
+    static String directory;
+
+    @GuardedBy("StatusFile.class")
```
trask: I think better to `@Guard` and synchronize below on an internal field that isn't visible externally, so you don't have to worry about anyone else using the same lock. (even if you have to create the field, e.g. `private final Object lock = new Object()`

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
littleaj: Not all are. I'm keeping the same style as existing status files.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,197 @@
...
+    private static String uniqueId;
+
+    @GuardedBy("StatusFile.class")
+    private static BufferedSink buffer;
+
+    private static final ThreadPoolExecutor WRITER_THREAD = new ThreadPoolExecutor(1, 1, 750L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<Runnable>(), ThreadPoolUtils.createNamedDaemonThreadFactory("StatusFileWriter"));
```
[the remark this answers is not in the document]
littleaj: The first argument is how many threads to create initially and how many threads to keep active by default; it defines initial and min threads. Line 66 (`WRITER_THREAD.allowCoreThreadTimeOut(true);`) lets idle threads die after keepAliveTime expires; so first arg only defines initial threads.

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public static <T> void putValueAndWrite(String key, T value) {
+        if (shouldNotWrite()) {
+            return;
+        }
+        CONSTANT_VALUES.put(key, value);
+        write();
+    }
+
+    public static void write() {
```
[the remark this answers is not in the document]
littleaj: It's used outside this package

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
```diff
@@ -84,12 +86,18 @@ public static void premain(Instrumentation instrumentation, File agentJarFile) {
...
             instrumentation.addTransformer(new LegacyPerformanceCounterClassFileTransformer());
+            success = true;
         } catch (ThreadDeath td) {
             throw td;
         } catch (Throwable t) {
             startupLogger.error("Agent failed to start.", t);
             t.printStackTrace();
         } finally {
+            try {
+                StatusFile.putValueAndWrite("RegisteredClassFileTransformers", success);
```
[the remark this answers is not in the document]
littleaj: `AgentInitializedSuccessfully=true/false`

PR #1090, opened 2019-10-07
fix quickpulse url construction.

### LINKED ISSUE #1089: QuickPulse regression in 2.5.1-beta
The QuickPulse urls are malformed. From the http logs:
```http
POST /QuickPulseService.svcping?ikey=00000000-0000-0000-0000-000000000000 HTTP/1.1
```
(missing slash before 'ping')

littleaj:
add tests to verify.

Fix #1089

### DIFF
--- a/core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
+++ b/core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
@@ -28,6 +28,7 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import com.microsoft.applicationinsights.TelemetryConfigurationTestHelper;
 import com.microsoft.applicationinsights.extensibility.initializer.docker.internal.Constants;
 import com.microsoft.applicationinsights.extensibility.initializer.docker.internal.DockerContext;
 import com.microsoft.applicationinsights.extensibility.initializer.docker.internal.DockerContextPoller;
@@ -38,10 +39,8 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.Test;
+
+import org.junit.*;
 
 /**
  * Created by yonisha on 7/29/2015.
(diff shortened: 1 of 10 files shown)


littleaj:
@trask, I had to make quite a few changes to prevent static variable values from bleeding into other tests which were causing failures. Ping me if you want me to go through the details.

core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
```diff
@@ -38,10 +39,8 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.Test;
+
+import org.junit.*;
```
trask: Did you change Intellij settings? I think default is 99 imports before it uses '*'

core/src/test/java/com/microsoft/applicationinsights/internal/heartbeat/HeartbeatTests.java
```diff
@@ -16,19 +17,22 @@
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import org.junit.AfterClass;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
+
+import org.junit.*;
```
trask: Same here re: Intellij settings

core/src/test/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcherTests.java
```diff
@@ -1,11 +1,15 @@
 package com.microsoft.applicationinsights.internal.quickpulse;
 
 import com.microsoft.applicationinsights.TelemetryConfiguration;
-import org.junit.Assert;
-import org.junit.Test;
+import org.junit.*;
```
trask: Same here re: Intellij settings

core/src/test/java/com/microsoft/applicationinsights/extensibility/initializer/docker/DockerContextInitializerTests.java
littleaj: Yes. For tests, I allow * for `org.junit`, `org.hamcrest` and `org.mockito` and there are a few static ones too, e.g. `org.junit.Assert.*`. For prod code, I prefer to be explicit. For tests I like the splats for the test framework classes. Intellisense is usually quicker/better with its suggestions.

I think the checkstyle settings reflect this, but maybe not.

trask: How do you configure this differently in Intellij for test vs prod?

littleaj: I don't think you can. I have the count settings set to 500 (default is 99 but that didn't always work for me), plus rules for org.junit, Mockito, Matchers and Assert which always use `*`. These settings are under Editor/Code Style/Java/Imports

PR #1103, opened 2019-10-17
Improve startup overhead

trask:


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -46,33 +44,13 @@
  */
 final class ApacheSender43 implements ApacheSender {
 
-    private final AtomicReference<CloseableHttpClient> httpClientRef = new AtomicReference<>();
+    private final Object lock = new Object();
+
+    private RuntimeException initException; // guarded by lock
```
littleaj: `@GuardedBy`?

trask:
These improvements were not too big. More significant improvements coming with Glowroot Instrumentation 0.14.7

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
trask: 👍

PR #1106, opened 2019-10-17
A couple of environment variable improvements

trask:


agent-codeless/src/test/java/com/microsoft/applicationinsights/agentc/internal/ConfigurationBuilderTest.java
```diff
@@ -1,13 +1,20 @@
...
+import org.junit.contrib.java.lang.system.EnvironmentVariables;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public class ConfigurationBuilderTest {
 
+    @Rule
+    public EnvironmentVariables envVars = new EnvironmentVariables();
```
littleaj: 👍

core/src/main/java/com/microsoft/applicationinsights/internal/config/TelemetryConfigurationFactory.java
```diff
@@ -388,6 +390,30 @@ private void setRoleName(ApplicationInsightsXmlConfiguration userConfiguration,
...
+                if (roleInstance.length() == 0) {
+                    return;
+                }
+
+                configuration.setRoleInstance(roleInstance);
+            }
+        } catch (Exception e) {
+            InternalLogger.INSTANCE.error("Failed to set role instance: '%s'", e.toString());
+        }
+    }
```
littleaj: Are the environment variables a "codeless agent only" mechanism? If not, perhaps they should be read in this class instead/also.

PR #1107, opened 2019-10-19
Improve startup time

trask:
In some environments `getCanonicalHostName()` can take several seconds.

I think get `getHostName()` should be fine, since we are calling it on `localhost` anyways (this is what Glowroot uses to capture local host name, see https://github.com/glowroot/glowroot/blob/v0.13.5/agent/core/src/main/java/org/glowroot/agent/init/EnvironmentCreator.java#L66).


littleaj:
@trask, this seems easy enough to move also include in master

PR #1108, opened 2019-10-19
Add instrumentation for java.util.logging

trask:


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
```diff
@@ -227,59 +227,93 @@ private static void start(Instrumentation instrumentation, File agentJarFile) th
         jdbcConfig.put("captureResultSetNavigate", false);
         jdbcConfig.put("captureGetConnection", false);
 
-        jdbcConfig.put("explainPlanThresholdMillis", getExplainPlanThresholdInMS(configuration, 10000));
+        Number explainPlanThresholdInMS = getExplainPlanThresholdInMS(configuration);
+        if (explainPlanThresholdInMS == null) {
+            jdbcConfig.put("explainPlanThresholdMillis", 10000);
+        } else {
+            jdbcConfig.put("explainPlanThresholdMillis", explainPlanThresholdInMS);
+        }
```
littleaj: why the pattern change? It seemed more compact in the previous version.

trask: i think i did that to match the pattern for `getLoggingThreshold(configuration)`

PR #1120, opened 2019-10-25
Update sampling config

trask:


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/Configuration.java
```diff
@@ -25,24 +25,42 @@
...
     public Map<String, Map<String, Object>> instrumentation = Collections.emptyMap();
     public List<CustomInstrumentation> customInstrumentation = Collections.emptyList();
-    public @Nullable FixedRateSampling fixedRateSampling;
 
     public boolean debug;
     public boolean developerMode;
 
+    public static class Sampling {
+
+        public @Nullable FixedRateSampling fixedRate;
```
littleaj: (I guess this PR is better suited for a rename, if needed)
Are we keeping the fixedRate terminology?

trask: I realized that we need to sync config with other languages, so didn't change for now, but moving under experimental to give time to sort it out.

PR #1125, opened 2019-10-28
Codeless azure functions

trask:


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/model/TraceContextCorrelationCore.java
```diff
@@ -65,6 +65,38 @@ private TraceContextCorrelationCore() {
...
+            requestTelemetry.getContext().getOperation().setParentId("|" + incomingTraceparent.getTraceId() + "." +
+                    incomingTraceparent.getSpanId() + ".");
+
+            Tracestate tracestate = getTracestate(request, requestHeaderGetter, incomingTraceparent, getAppId());
+            int traceflag = incomingTraceparent.getTraceFlags();
+
+            return new DistributedTraceContext(requestTelemetry, tracestate, traceflag);
+
+        } catch (Exception e) {
+            logger.debug("unable to perform correlation", e);
```
littleaj: Is this a common occurrence? My first instinct is to log this as a warning.

trask: I think the reason it's logged as `debug` (copied from method above `resolveCorrelationForRequest`) is that it can be caused by user sending bad (e.g. malicious) http header and flooding the log

PR #1136, opened 2019-11-03
Remove modules

trask:


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/model/DistributedTraceContext.java
```diff
@@ -38,28 +35,11 @@
     private final Tracestate tracestate;
     private final int traceflag;
 
-    // ApplicationInsights format
```
littleaj: I thought we still needed this for backwards compatibility. Is that handled somewhere else?

trask: `DistributedTraceContext` just represents internal state, where everything is W3C now, but (I double-checked 😀) we still propagate `Request-Id` externally to maintain `backCompat`

PR #1143, opened 2019-11-06
update status file to use shutdown hook on linux

littleaj:


agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/status/StatusFile.java
```diff
@@ -140,8 +141,13 @@ private static BufferedSink getBuffer(File file) throws IOException {
             if (buffer != null) {
                 buffer.close();
             }
-            buffer = Okio.buffer(Okio.sink(file.toPath(), StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE,
-                    StandardOpenOption.WRITE, StandardOpenOption.TRUNCATE_EXISTING));
+            if (SystemInformation.INSTANCE.isWindows()) {
+                buffer = Okio.buffer(Okio.sink(file.toPath(), StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE,
```
trask: can you add a comment explaining why we need the different behavior on windows vs linux, since it was complicated and we'll probably forget the details in a few months :-)

PR #1147, opened 2019-11-07
Disable SDK quick pulse and heartbeat

trask:


littleaj:
@trask, we're disabling heartbeat? Maybe we should discuss this in more detail.

PR #1153, opened 2019-11-11
Add azure pipeline yaml for smoke tests

trask:


gradle/common-java.gradle
```diff
@@ -126,7 +126,7 @@ if (java7JreDir) {
 } else {
     def msg = "IMPORTANT: Environment variable 'JAVA_JRE_7' is not defined - Install JRE 7 and set 'JAVA_JRE_7' to prevent runtime compatibility issues!"
     if (!isJava7) {
-        if (isBuildServer || isRelease) {
+        if ((isBuildServer && !System.env.SMOKE_APP) || isRelease) {
```
littleaj: I'm not sure I agree with this. If this is still building the artifacts, I still think we need the java 7 jars on the bootstrapclasspath

settings.gradle
```diff
@@ -49,11 +49,15 @@ if (System.properties.AI_SAMPLES) {
...
-include ':test:smoke:appServers:Jetty.9'
+if (System.env.SMOKE_APP_SERVER) {
+    include ":test:smoke:appServers:$System.env.SMOKE_APP_SERVER"
+} else {
+    include ':test:smoke:appServers:Tomcat.7'
+    include ':test:smoke:appServers:Tomcat.8.5'
+    include ':test:smoke:appServers:Wildfly.8'
+    include ':test:smoke:appServers:Wildfly.14'
+    include ':test:smoke:appServers:Jetty.9'
+}
```
littleaj: 👍

test/smoke/appServers/JBossEAP.gradle
```diff
@@ -1,41 +0,0 @@
-if (!project.hasProperty('jbossEapVersion')) {
```
littleaj: Why are we removing JBossEAP?

gradle/common-java.gradle
trask: This pipeline is not building the artifacts, just running smoke tests

test/smoke/appServers/JBossEAP.gradle
trask: Switching from testing JBossEAP to Wildfly so we can download without username/password, and seems better to use Wildfly anyways since that is what App Services supports

azure-pipelines.yml
```diff
@@ -0,0 +1,296 @@
...
+
+variables:
+  # smoke tests are run inside of docker linux containers so don't use windows performance counters anyways
+  GradleParams: --info -DskipWinNative=true
+  isRelease: false
+
+jobs:
+- job: smoke
+  strategy:
+    matrix:
```
littleaj: This will get unwieldy as we add dimensions. I'd like to generate the matrix.

A simple example: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-configuration (2nd one)

Since it just needs json, a gradle task might be easier.

trask: That's cool that you can generate the matrix, I was wondering about that! Let's come back to it later? (also, I tried splitting out JRE as a 3rd dimension but it ended up exceeding 256 matrix slice limit 😭)

littleaj: Sounds good.

Maybe we can split this into two builds. I saw you can specify the yaml file to use in the build definition settings.

gradle/common-java.gradle
littleaj: Where does it get the artifacts? I didn't see that part

trask: Sorry, what I mean is, this is not replacing the `AppInsights-Java-Gradle-CI` build, only the `AppInsights-Java-SmokeTest-MultiConfig` build.

I don't think we need to compile against Java 7 bootstrap classes when running smoke tests, since any issue will be caught during `AppInsights-Java-Gradle-CI`.

trask:
That's a good question! Let's find out :-).

PR #1157, opened 2019-11-12
Add nuspec to project

littleaj:


trask:
Since this is App Services specific, and we need same for AKS/Functions, what about `package/appsvc/Microsoft.ApplicationInsights.JavaCodelessAgent.AppSvc.nuspec`?

littleaj:
@trask, there will be a subfolder for each variant; all included in one package.

PR #1171, opened 2019-12-04
Fix springboot dependencies

littleaj:
This does a few things:
1. Fixes the unnecessary inclusion of xmlpull and xstream in spring boot shadow
2. Uses shadow configuration for core and web so we don't have to manage their dependencies here (related to 1)
3. Removes an unnecessary relocation.

### DIFF
--- a/azure-application-insights-spring-boot-starter/build.gradle
+++ b/azure-application-insights-spring-boot-starter/build.gradle
@@ -47,8 +47,7 @@ sourceSets {
 
 def springBootVersion = '1.5.21.RELEASE'
 dependencies {
-    compile(project(path: ':core', configuration: 'shadow'))
-    compile(project(path: ':web', configuration: 'shadow'))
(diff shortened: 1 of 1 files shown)


azure-application-insights-spring-boot-starter/build.gradle
```diff
@@ -47,14 +47,16 @@ sourceSets {
 
 def springBootVersion = '1.5.21.RELEASE'
 dependencies {
-    compile(project(':core'))
-    compile(project(':web'))
+    compile(project(path: ':core', configuration: 'shadow'))
```
trask: Do we need `:core` since it's included in `:web` shadow configuration?

littleaj: Maybe not. I'll give it a shot and if the tests fail, i'll revert.

littleaj: Looks like it worked

PR #1177, opened 2019-12-12
Fix bugs from static analysis

littleaj:
Bugs found by lgtm.com

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
@@ -83,8 +83,7 @@ final class SenderThreadsBackOffManager extends ThreadLocal<SenderThreadLocalBac
 
     @Override
     protected SenderThreadLocalBackOffData initialValue() {
-        long addSeconds = threadsSecondsDifference.incrementAndGet();
-        senderThreadLocalData = new SenderThreadLocalBackOffData(backOffTimeoutsInMilliseconds, addSeconds * 1000L);
+        senderThreadLocalData = new SenderThreadLocalBackOffData(backOffTimeoutsInMilliseconds, threadsSecondsDifference.incrementAndGet() * 1000L);
         registerSenderData(senderThreadLocalData);
         return senderThreadLocalData;
     }
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
@@ -83,8 +83,8 @@ final class SenderThreadsBackOffManager extends ThreadLocal<SenderThreadLocalBac
 
     @Override
     protected SenderThreadLocalBackOffData initialValue() {
-        int addSeconds = threadsSecondsDifference.incrementAndGet();
-        senderThreadLocalData = new SenderThreadLocalBackOffData(backOffTimeoutsInMilliseconds, addSeconds * 1000);
+        long addSeconds = threadsSecondsDifference.incrementAndGet();
+        senderThreadLocalData = new SenderThreadLocalBackOffData(backOffTimeoutsInMilliseconds, addSeconds * 1000L);
         registerSenderData(senderThreadLocalData);
         return senderThreadLocalData;
     }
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
```diff
@@ -83,8 +83,8 @@ public synchronized void stopAllSendersBackOffActivities() {
 
     @Override
     protected SenderThreadLocalBackOffData initialValue() {
-        int addSeconds = threadsSecondsDifference.incrementAndGet();
-        senderThreadLocalData = new SenderThreadLocalBackOffData(backOffTimeoutsInMilliseconds, addSeconds * 1000);
+        long addSeconds = threadsSecondsDifference.incrementAndGet();
```
trask: this cast to long is not needed, is it for clarity?
```suggestion
        int addSeconds = threadsSecondsDifference.incrementAndGet();
```

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -109,7 +109,7 @@ public TransmissionFileSystemOutput(String folderPath, String maxTransmissionSto
                                                                         DEFAULT_CAPACITY_MEGABYTES,
                                                                         MAX_TRANSMISSION_STORAGE_CAPACITY_NAME,
                                                                         maxTransmissionStorageCapacity);
-        capacityInBytes = capacityEnforcer.getCurrentValue() * 1024 * 1024;
+        capacityInBytes = ((long)capacityEnforcer.getCurrentValue()) * 1024L * 1024L;
```
trask: similar to above, this cast is not needed, or is it for clarity?
```suggestion
        capacityInBytes = capacityEnforcer.getCurrentValue() * 1024L * 1024L;
```

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
```diff
@@ -212,7 +212,7 @@ public Transmission fetchOldestFile() {
     }
 
     public void setCapacity(int suggestedCapacity) {
-        this.capacityInBytes = capacityEnforcer.normalizeValue(suggestedCapacity) * 1024 * 1024;
+        this.capacityInBytes = ((long)capacityEnforcer.normalizeValue(suggestedCapacity)) * 1024L * 1024L;
```
trask: same as above
```suggestion
        this.capacityInBytes = capacityEnforcer.normalizeValue(suggestedCapacity) * 1024L * 1024L;
```

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
littleaj: The constructor is as shown:
https://github.com/microsoft/ApplicationInsights-Java/blob/a4437aa0817797e9e19e0f4ed7156290c9b22644/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadLocalBackOffData.java#L55

The multiplication of `addSeconds * 1000` was flagged because it's `int * int` could overflow if `addSeconds` is sufficiently large (I don't think it's possible in this case, but...). Changing the type seemed like the most straightforward way to address it.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
littleaj: I couldn't recall if the cast was needed or not. Since the other operands are `long`, this cast is redundant.

For reference: [Java Language Spec, Section 5.6.2](https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.6.2)

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java
littleaj: Same as above

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/SenderThreadsBackOffManager.java
littleaj: I suppose `addSeconds` could be inlined and the literal changed to `long`. That's probably cleaner

littleaj: I see what you meant now. I forgot the literal was already changed to `long` 😄

PR #1186, opened 2020-01-08
Fix ChannelFetcher memory leak

trask:


core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -51,10 +53,16 @@
...
+        @Deprecated
         private final List<ChannelFetcher> fetchers = new ArrayList<ChannelFetcher>();
         private final List<Stoppable> stoppables = new ArrayList<Stoppable>();
         private final List<Closeable> closeables = new ArrayList<Closeable>();
 
+        public synchronized void register(TelemetryChannel channel) {
+            channels.put(channel, true);
+        }
+
         public synchronized void register(ChannelFetcher fetcher) {
```
littleaj: if the `fetchers` list is removed we can make this `@Deprecated` and a nop

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -51,10 +53,16 @@
     private static class SDKShutdownAction implements Runnable {
         private boolean stopped = false;
 
+        private final Map<TelemetryChannel, Boolean> channels = new WeakHashMap<>();
+        @Deprecated
         private final List<ChannelFetcher> fetchers = new ArrayList<ChannelFetcher>();
```
littleaj: We can remove this since it's private.

core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -507,6 +497,7 @@ public void flush() {
     TelemetryChannel getChannel() {
         if (this.channel == null) {
             this.channel = configuration.getChannel();
+            SDKShutdownActivity.INSTANCE.register(this.channel);
```
littleaj: I think this should still be in the constructor, `getChannel()` since it's called in `track()`.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -51,10 +53,16 @@
     private static class SDKShutdownAction implements Runnable {
         private boolean stopped = false;
 
+        private final Map<TelemetryChannel, Boolean> channels = new WeakHashMap<>();
```
littleaj: I like this, but doesn't this mean when a `TelemetryClient` instance is GC'ed and its `TelemetryChannel` key is dropped that the channel's threads will still exist? If the `TelemetryClient` instance is being disposed, its `SDKShutdownAction` should be run.

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
```diff
@@ -126,23 +134,29 @@ private void stopInternalLogger() {
...
+            } catch (ThreadDeath td) {
+                throw td;
+            } catch (Throwable t) {
                 try {
-                    TelemetryChannel channelToStop = fetcher.fetch();
-                    if (channelToStop != null) {
-                        channelToStop.stop(getPerThreadTimeout(), getPerThreadTimeUnit());
-                    }
+                    InternalLogger.INSTANCE.error("Failed to stop channel: '%s'", t.toString());
+                    InternalLogger.INSTANCE.trace("Stack trace generated is %s", ExceptionUtils.getStackTrace(t));
```
littleaj: Since we're changing things, could you please consolidate these :) Also, I think this can be a warning. It's just to hint that there may be some threads hanging around.

trask: 👍 Updated

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
trask: 👍 Updated to non-weak

core/src/main/java/com/microsoft/applicationinsights/internal/shutdown/SDKShutdownActivity.java
trask: Let's keep `fetchers` list for now just in the (very unlikely) case someone is using it.

PR #1196, opened 2020-02-03
Add http proxy configuration

trask:
Using environment variable `APPLICATIONINSIGHTS_HTTP_PROXY`.

Format is `[scheme://]host[:port]` where scheme defaults to `http` and port defaults to `80` for http and `443` for https.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
@@ -29,6 +29,7 @@ import javax.annotation.concurrent.GuardedBy;
 
 import com.microsoft.applicationinsights.internal.logger.InternalLogger;
 import com.microsoft.applicationinsights.internal.util.SSLOptionsUtil;
+import org.apache.http.HttpHost;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -143,9 +146,11 @@ private static CloseableHttpClient init() {
...
-        return HttpClients.custom()
-                .setConnectionManager(cm)
-                .useSystemProperties()
-                .build();
+        HttpClientBuilder builder = HttpClients.custom()
+                .setConnectionManager(cm);
+        if (proxy != null) {
+            builder.setProxy(proxy);
+        }
+        return builder.build();
```
littleaj: I think that if a proxy is set with our custom property/variable, then that would override the system properties. If it was not set, then the system properties' settings would still be used. I'm not sure what would happen if `useSytemProperties` and `setProxy` are used together. I hope it would work as I described, but if not, I think we can add `useSystemProperties` in an else clause on line 153.

Also, we have used the debug logging properties for the http client in the past. I'm not sure if the absence of `useSystemProperties` would prevent that.

What do you think?

trask: Yes, this makes sense to me 👍
(and I checked and setProxy takes precedence over useSystemProperties so we are good there)

PR #1201, opened 2020-02-14
Fix appsvc windows log config and env var config 

littleaj:
<strike>APPLICATIONINSIGHTS_LOGDIR overrides user accessible log directory, which also overrides the 'site.logdir' value.</strike>

APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY is linux only and overrides the internal log directory.

Updated package to contain different configs for windows and linux.

- [x] CHANGELOG.md updated

### DIFF
--- a/agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
+++ b/agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
@@ -11,7 +11,13 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>Application Insights Java Codeless Agent.</description>
     <copyright>© Microsoft Corporation. All rights reserved.</copyright>
-    <releaseNotes />
+    <releaseNotes>
+        * Fixed IPA log directory defaults for Windows.
+        * Introduced environment variable `APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY` for internal self-diagnostic logs.
+        * Updated Status File location to be relative to the log directory (controlled by same variables/properties).
+        * Added env var `APPLICATIONINSIGHTS_EXTENSION_LOG_FILE_ENABLED`. $HOME/LogFiles/ApplicationInsights/*.log is disabled when set to "false".
+        * Added env var `APPLICATIONINSIGHTS_EXTENSION_STATUS_FILE_ENABLED`. $HOME/LogFile/ApplicationInsights/status/status_*.json is disabled when set to "false".
+    </releaseNotes>
     <tags>Microsoft ApplicationInsights Java Codeless</tags>
   </metadata>
   <files>
--- a/agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
+++ b/agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>Microsoft.ApplicationInsights.JavaCodelessAgent</id>
-    <version>2.5.1</version>
+    <version>2.5.2</version>
     <title>Application Insights Java Codeless Agent Package</title>
     <authors>Microsoft</authors>
     <licenseUrl>http://go.microsoft.com/fwlink/?LinkID=391182</licenseUrl>
--- a/agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
+++ b/agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
@@ -23,7 +23,6 @@ package com.microsoft.applicationinsights.agentc.internal;
 
 import java.io.File;
 import java.lang.instrument.Instrumentation;
-import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -64,10 +63,10 @@ import com.microsoft.applicationinsights.telemetry.PageViewTelemetry;
 import com.microsoft.applicationinsights.telemetry.RemoteDependencyTelemetry;
 import com.microsoft.applicationinsights.telemetry.RequestTelemetry;
 import com.microsoft.applicationinsights.telemetry.TraceTelemetry;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpHost;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.glowroot.instrumentation.engine.config.InstrumentationDescriptor;
-import org.glowroot.instrumentation.engine.config.InstrumentationDescriptors;
 import org.glowroot.instrumentation.engine.impl.InstrumentationServiceImpl.ConfigServiceFactory;
 import org.glowroot.instrumentation.engine.impl.SimpleConfigServiceFactory;
 import org.glowroot.instrumentation.engine.init.EngineModule;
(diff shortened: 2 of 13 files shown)


Mikhail-msft:
what env should be used for rolling file telemetry - "APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY" ? and if so, will it work on windows as well?

littleaj:
`APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY` is for internal logs on linux (possibly more platforms in the future). To set the log directory on windows use the system property `site.logdir`. Be sure to include the drive letter.

agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
```diff
@@ -11,7 +11,12 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>Application Insights Java Codeless Agent.</description>
     <copyright>© Microsoft Corporation. All rights reserved.</copyright>
-    <releaseNotes />
+    <releaseNotes>
+        * Fixed IPA log directory defaults for Windows.
+        * Introduced environment variable `APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY` for internal self-diagnostic logs.
+        * Split appsvc folder into `windows` and `linux` since the logging configuration differs for each platform.
+        * Updated Status File location to be relative to the log directory (controlled by same variables/properties).
+    </releaseNotes>
```
littleaj: I wasn't sure how to format this.

agent-codeless/package/Microsoft.ApplicationInsights.JavaCodelessAgent.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>Microsoft.ApplicationInsights.JavaCodelessAgent</id>
-    <version>2.5.1</version>
+    <version>2.5.2</version>
```
littleaj: @trask 2.5.2 or 2.6.0?

trask: somewhere in between? 😄

i'm good with `2.5.2`

agent-codeless/package/appsvc/linux/ai.logback.xml
```diff
@@ -11,7 +11,7 @@
     <appender name="LOGDIR" class="com.microsoft.applicationinsights.agentc.shadow.ch.qos.logback.core.rolling.RollingFileAppender">
         <file>${site.logdir:-/home/LogFiles}/ApplicationInsights/application-insights-extension.log</file>
         <rollingPolicy class="com.microsoft.applicationinsights.agentc.shadow.ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-            <fileNamePattern>${site.logdir:-/home/LogFiles}/ApplicationInsights/application-insights-extension-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
+            <fileNamePattern>${site.logdir:-/home/LogFiles}/ApplicationInsights}/application-insights-extension-%d{yyyy-MM-dd}.%i.log</fileNamePattern>
```
trask: looks like an extra `}` snuck in

agent-codeless/src/main/java/com/microsoft/applicationinsights/agentc/internal/MainEntryPoint.java
```diff
@@ -120,24 +117,9 @@ public static void premain(Instrumentation instrumentation, File agentJarFile) {
...
-                startupLogger.error("Could not load appsvc logging config", t);
-            }
-        } else {
-            File logbackXmlOverride = new File(agentJarFile.getParentFile(), "ai.logback.xml");
-            if (logbackXmlOverride.exists()) {
-                System.setProperty("ai.logback.configurationFile", logbackXmlOverride.getAbsolutePath());
-            }
+        File logbackXmlOverride = new File(agentJarFile.getParentFile(), "ai.logback.xml");
+        if (logbackXmlOverride.exists()) {
+            System.setProperty("ai.logback.configurationFile", logbackXmlOverride.getAbsolutePath());
```
trask: this is nice change 👍

Mikhail-msft:
> `APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY` is for internal logs on linux (possibly more platforms in the future). To set the log directory on windows use the system property `site.logdir`. Be sure to include the drive letter.

Let's not differentiate if possible - we would need the same logic for windows containers

Mikhail-msft:
> I don't think we need to specify `D:` drive everywhere for windows since that will be default, @Mikhail-msft what do u think?

we should use following (not sure about precedence): %HOME%\LogFiles or site.logdir. I don't think we should hardcode d:\ anywhere - that's quite atypical for windows.

littleaj:
> > I don't think we need to specify `D:` drive everywhere for windows since that will be default, @Mikhail-msft what do u think?
> 
> we should use following (not sure about precedence): %HOME%\LogFiles or site.logdir. I don't think we should hardcode d:\ anywhere - that's quite atypical for windows.
@Mikhail-msft @trask 

This makes sense to me. I actually had this but removed it because I wasn't sure if we could rely on `HOME`. I still think there should be a default value for Windows. Should I just use the inferred drive (i.e. no drive specified)? e.g. `\home\LogFiles`?

Also, I think `site.logdir` would take precedence and `%HOME%\LogFiles` would be the default value if `site.logdir` was missing.

littleaj:
> > `APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY` is for internal logs on linux (possibly more platforms in the future). To set the log directory on windows use the system property `site.logdir`. Be sure to include the drive letter.
> 
> Let's not differentiate if possible - we would need the same logic for windows containers

We have to split these. Ping me for details.

littleaj:
@trask @Mikhail-msft this has been updated with changes from our last conversation. I've tested it on both windows and linux azure web apps.

agent-codeless/src/main/resources/appsvc.ai.logback.xml
```diff
@@ -3,58 +3,21 @@
...
-        <appender-ref ref="LOGDIR"/>
-        <appender-ref ref="APPSVCTBS"/>
+        <appender-ref ref="CONSOLE" />
+        <appender-ref ref="USER_LOG_FILE" /><!-- optional; platform dependent -->
+        <appender-ref ref="DIAGNOSTICS_FILE" /><!-- optional; platform dependent -->
     </root>
 
-    <logger name="applicationinsights.diagnostics" level="info" />
-</configuration>
+    <logger name="applicationinsights.extension.diagnostics" level="${applciationinsights.extension.diagnostics.level:-info}" />
```
trask:
```suggestion
    <logger name="applicationinsights.extension.diagnostics" level="${applicationinsights.extension.diagnostics.level:-info}" />
```

agent-codeless/src/main/resources/rp-logger-config/diagnostics.appender.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<included>
+    <property scope="system" name="ai.appsvc.logfile.name" value="${ai.logfile.name:-application-insights-extension}" />
+    <property scope="system" name="ai.appsvc.logfile.ext" value="${ai.logfile.ext:-.log}" />
+    <property scope="system" name="ai.appsvc.logfile.old.ext" value="${ai.appsvc.logfile.ext:-.log}.old" />
```
trask: I think I'd prefer if these didn't end up in the global System properties.

Would it work to use default (local) scope properties with a bit of duplication between the two xml config files instead?

agent-codeless/src/main/resources/rp-logger-config/user-logfile.appender.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<included>
+    <property scope="system" name="ai.logdir" value="${site.logdir:-${HOME:-.}/LogFiles}/ApplicationInsights" />
+    <property scope="system" name="ai.logfile.name" value="application-insights-extension" />
+    <property scope="system" name="ai.logfile.ext" value=".log" />
```
trask: Same comment as above

agent-codeless/src/main/resources/appsvc.ai.logback.xml
littleaj: good catch. ty

agent-codeless/src/main/resources/rp-logger-config/diagnostics.appender.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<included>
+    <property scope="local" name="ai.appsvc.logfile.name" value="${ai.logfile.name:-application-insights-extension}" />
+    <property scope="local" name="ai.appsvc.logfile.ext" value="${ai.logfile.ext:-.log}" />
+    <property scope="local" name="ai.appsvc.logfile.old.ext" value="${ai.appsvc.logfile.ext:-.log}.old" />
```
trask:
```suggestion
    <property scope="local" name="ai.appsvc.logfile.name" value="application-insights-extension" />
    <property scope="local" name="ai.appsvc.logfile.ext" value=".log" />
    <property scope="local" name="ai.appsvc.logfile.old.ext" value=".log.old" />
```
This seems less confusing to me, compared to referencing properties defined in the other file.

PR #1207, opened 2020-02-15
remove files from master that shouldn't have been merged over

littleaj:


littleaj:
Depends on test fix in #1210

PR #1225, opened 2020-04-02
Add ETW provider/appender

littleaj:


etw/java/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/events/model/IpaEtwEventErrorBase.java
```diff
@@ -0,0 +1,47 @@
...
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+package com.microsoft.applicationinsights.agentc.internal.diagnostics.etw.events.model;
+
+public abstract class IpaEtwEventErrorBase extends IpaEtwEventBase {
+    private StacktraceProvider stacktrace;
```
trask, line 24: Is it important that this has no dependencies? Wondering if we need the `StacktraceProvider` abstraction, or can we depend directly on `IThrowableProxy` here?

etw/java/src/main/jni/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/EtwProvider.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private static File loadLibrary() throws IOException {
+        final String fileName = getDllFilenameForArch();
+
+        final File targetDir = DllFileUtils.buildDllLocalPath();
+        final File dllPath = new File(targetDir, fileName);
+
+        if (!dllPath.exists()) {
+            DllFileUtils.extractToLocalFolder(dllPath, fileName);
+        }
```
trask, line 74: If we release new version with updated DLL, it won't overwrite the old DLL until the temp folder gets cleaned? Do you know how often temp folder is cleaned on App Service Windows?

etw/java/build.gradle
```diff
@@ -0,0 +1,122 @@
...
+    jni32 {
+        transitive = false
+    }
+    provided {
+        transitive = false
+    }
+    testRuntime.extendsFrom provided
+}
+
+def jniSrcDir = 'src/main/jni'
```
trask, line 45: I don't see this directory (or anything else in this module)?

agent-codeless/build.gradle
```diff
@@ -30,6 +30,23 @@ def instrumentationVersion = '0.14.9'
 
 def shadowPrefix = 'com.microsoft.applicationinsights.agentc.shadow'
 
+sourceSets {
+    devtest {
+        java {
+            srcDir file('src/devtest/java')
```
trask, line 36: I don't see this directory?

agent-codeless/src/devtest/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/DiagnosticsLoggerProxy.java
```diff
@@ -0,0 +1,316 @@
+package com.microsoft.applicationinsights.agentc.internal.diagnostics.etw;
+
+import com.microsoft.applicationinsights.agentc.internal.diagnostics.DiagnosticsHelper;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Marker;
+
+public class DiagnosticsLoggerProxy implements Logger {
```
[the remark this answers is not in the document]
littleaj: I made a "devtest" artifact which includes this class for use in the etw-testapp. Just grabbing the logger with slf4j from the test app didn't work. See the build file for `:agent-codeless` and `:etw:etw-testapp`

etw/java/src/main/jni/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/EtwProvider.java
littleaj: it won't overwrite, but it's in a versioned folder. It should pick up the correct one. However, it may be better to version the DLL files themselves.

etw/java/build.gradle
littleaj: `etw/java/src/main/jni`. which module?

etw/java/src/main/jni/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/EtwProvider.java
trask: oh, that's good and addresses my concern, i missed the versioning of the containing folder

agent-codeless/build.gradle
littleaj: hmm, i'm not sure why these aren't showing up for you. `agent-codeless/src/devtest/java`

etw/java/build.gradle
trask: Sorry, github UI failing me (and I'm failing it).

agent-codeless/build.gradle
trask: Same, github UI failing me (and I'm failing it).

agent-codeless/src/devtest/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/DiagnosticsLoggerProxy.java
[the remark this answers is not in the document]
trask: 👍

etw/java/src/main/java/com/microsoft/applicationinsights/agentc/internal/diagnostics/etw/events/model/IpaEtwEventErrorBase.java
littleaj: This was the only dependency on logback. I used to use `IThrowableProxy` and I changed it in commit addb62921ba337b49e076781ae56483fbf9bcbbf, but I don't recall exactly what broke when I fixed the circular dependency. I can backtrack if this becomes an issue.

trask: Oh ya, I had missed before that this class is in `etw:java` 👍

etw/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+Provider Name: **`Microsoft-ApplicationInsights-Java-IPA`**
+Provider GUID: **`1f0dc33f-30ae-5ff3-8b01-8ca9b8509233`**
+
+# Building
+
+## Build Requirements
+To build the native resources, you'll need
+* Visual Studio Build Tools, C++
+* Windows 10 SDK (developed with `10.0.18362.0`)
```
maxgolov, line 11: nit: it may be possible to include a script that installs the required deps using chocolatey. Chocolatey is available in GitHub actions, as well as it's easy to install it with a PowerShell script.
Sharing some thoughts here:

Install choco if it's not installed:
```
@"%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "[System.Net.ServicePointManager]::SecurityProtocol = 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))" && SET "PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin"
```

Install Tools and SDK using choco:
```
choco install -y --includeOptional visualstudio2017-workload-netcoretools
choco install -y windows-sdk-10-version-1809-all
```

Note: this installation step takes about 6-10 minutes on clean machine.

etw/native/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+def javaIncludeDir = new File(System.env.'JAVA_HOME', "include")
+def programFilesX86 = System.env.'ProgramFiles(x86)' ?: 'C:/Program Files (x86)'
+
+// TraceLoggingProvider.h in Windows 10 SDK
+def winSdkDir = "$programFilesX86/Windows Kits/10"
+if (System.env.APPINSIGHTS_WIN_SDK_PATH) {
+    winSdkDir = System.env.APPINSIGHTS_WIN_SDK_PATH
+    logger.info "Windows SDK path set with environment variable, APPINSIGHTS_WIN_SDK_PATH"
+}
+def vsToolsDir = "$programFilesX86/Microsoft Visual Studio 14.0"
```
maxgolov, line 38: nit: 14.0 is a bit old'ish....

etw/native/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+ext {
+    headersDir = "$projectDir/src/main/headers"
+}
+library {
+    baseName = "applicationinsights-java-etw-provider"
+    linkage = [ Linkage.SHARED ]
+    targetMachines = [ machines.windows.x86, machines.windows.x86_64 ]
+    privateHeaders.from file(headersDir)
+    dependencies {
+        implementation files(file("$winSdkLibPath/x64/advapi32.lib"))
```
maxgolov, line 71: nit: you can also specify the lib files to link using msvc compiler-specific 'pragma comment lib': https://docs.microsoft.com/en-us/cpp/preprocessor/comment-c-cpp?view=vs-2019
That way if you need to manage other dependencies for your native code, you could add more to source rather than maintaining it in the outside gradle script. You'd also need only one line instead of 4 (you can then omit the dependencies section entirely).

etw/native/build.gradle
```diff
@@ -0,0 +1,135 @@
...
+    includes file("${System.env.JAVA_HOME}/include")
+    includes file("${System.env.JAVA_HOME}/include/win32")
+
+    compilerArgs.add '/W4'
+    compilerArgs.add '/std:c++14'
+    compilerArgs.add '/EHsc'
+    compilerArgs.add '/sdl'
+
+    if (project.hasProperty(NATIVE_BUILD_VARIANT_PROPERTY) && 'release'.equalsIgnoreCase(project.properties[NATIVE_BUILD_VARIANT_PROPERTY])) {
+        logger.info "Configuring ${it} for release build"
```
maxgolov, line 91: That's cool........ you can also use .vcxproj and **msbuild** to prescribe default flags for Release and Debug, without manually passing individual compiler args and flags. It'd be more natural for C++ dev dealing with Visual Studio to just edit the .vcxproj file for that.

etw/native/src/main/cpp/etw_provider.cpp
```diff
@@ -0,0 +1,435 @@
...
+void javaThrowUnknownError(JNIEnv * env, std::string message) noexcept {
+    jclass cls = env->FindClass("java/lang/RuntimeException");
+    if (cls != NULL) {
+        std::string m = "Unknown error from " DLL_FILENAME_STR + message;
+        env->ThrowNew(cls, m.c_str());
+    }
+    env->DeleteLocalRef(cls);
+}
+
+char * jstring2cstr(JNIEnv * env, jstring &jstr_input, char * cstr_output, aijnierr_t field_id) throw(aijnierr_t) {
```
maxgolov, line 397: nit: since you're already using C++ **string** elsewhere - I would've just returned **std::string** from these helper methods. That will help you to get rid of a bunch of **delete[]** at the end of **writeEvent_IpaEtwEvent** function. Basically let the platform to manage the lifecycle of **std::string** for you, don't manage lifecycle of C-style const char * strings. Also your code would be smaller.

etw/README.md
littleaj: This is a good idea. I might do this in the future.

etw/native/build.gradle
littleaj: The other JNI library is using this as a minimum. Otherwise, I would use the latest.

etw/native/build.gradle
littleaj: Nice! I'll definitely do this.

etw/native/src/main/cpp/etw_provider.cpp
littleaj: Makes sense. I'll make this change.

etw/native/build.gradle
littleaj: I did try that, but it was a bit difficult working with it and the gradle plugin. I'm more familiar with gradle than vcxproj :/ You can still generate the vcxproj with `.\gradlew :etw:native:visualStudio` if one prefers to work in Visual Studio.

PR #1250, opened 2020-06-01
Merge codeless to opentelemetry

littleaj:
Contains ETW changes


littleaj:
I'm not really sure how to handle submodules, @trask. should it not be updated? Maybe that's why LGTM is failing.

PR #1253, opened 2020-06-04
Fix backoff retries

trask:


trask:
> Seems straightforward, though I wonder if we are now missing tests for the backoff code.

Unfortunately we were already missing tests for the backoff code. The tests I removed were testing code that has not been used for a long time 😕.

PR #1254, opened 2020-06-04
Improve logging/exception handling loading JNI perf counter lib

littleaj:
I made the logs more specific and added info messages when we find the PID from the MXBean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 2.5.2
+* Added additional error logging for troubleshooting issues loading JNI performance counter library. 
+
 # Version 2.6.1
 * Fix shutdown issue due to non-daemon thread
   ([#1245](https://github.com/microsoft/ApplicationInsights-Java/pull/1245))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # CHANGELOG
 
-# Version 2.5.2
+# Version 2.6.2
 * Added additional error logging for troubleshooting issues loading JNI performance counter library. 
 
 # Version 2.6.1
@@ -362,4 +362,3 @@ Livemetrics UX.
 - Support collecting performance counters from 32-bit Windows machines.
 - Support manual tracking of dependencies using a new ```trackDependency``` method API.
 - Ability to tag a telemetry item as synthetic, by adding a ```SyntheticSource``` property to the reported item.
-
--- a/core/native.gradle
+++ b/core/native.gradle
@@ -49,7 +49,7 @@ if (System.env.APPINSIGHTS_VS_PATH) {
 
 idea {
     module {
-        sourceDirs += file("$projectDir/src/windows/cpp")\
+        sourceDirs += file("$projectDir/src/windows/cpp")
     }
 }
 
--- a/core/native.gradle
+++ b/core/native.gradle
@@ -20,6 +20,7 @@
  */
 
 apply plugin: 'cpp'
+apply plugin: 'idea'
 
 buildscript {
     repositories {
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 2.5.2
```
trask, line 3:
```suggestion
# Version 2.6.2
```

core/native.gradle
```diff
@@ -46,6 +47,12 @@ if (System.env.APPINSIGHTS_VS_PATH) {
     println "Using custom Visual Studio Tools path: $vsToolsDir"
 }
 
+idea {
+    module {
+        sourceDirs += file("$projectDir/src/windows/cpp")\
```
trask, line 52: is the trailing `\` intentional?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -75,11 +76,15 @@ public static boolean initialize() {
...
+            if (InternalLogger.INSTANCE.isErrorEnabled()) {
+                InternalLogger.INSTANCE.error("Error initializing JNI Performance Counter library. Windows performance counters will not be used.: "+ ExceptionUtils.getStackTrace(e));
+            }
         } catch (Throwable e) {
             try {
-                InternalLogger.INSTANCE.error(
-                    "Failed to load native dll, Windows performance counters will not be used. " +
-                "Please make sure that Visual C++ Redistributable is properly installed: %s.", e.toString());
+                if (InternalLogger.INSTANCE.isErrorEnabled()) {
+                    InternalLogger.INSTANCE.error("Unexpected error initialiing JNI Performance Counter library. Windows performance counters will not be used: "+ExceptionUtils.getStackTrace(e));
```
trask, line 86:
```suggestion
                    InternalLogger.INSTANCE.error("Unexpected error initializing JNI Performance Counter library. Windows performance counters will not be used: "+ExceptionUtils.getStackTrace(e));
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
```diff
@@ -150,14 +155,26 @@ public static double getValueOfPerformanceCounter(String name) {
...
 
+    /**
+     * Performance Counters identify a process by "Instance Name."
+     * This will be the executable name without the extension, e.g. a process running java.exe will have an instance name "java".
+     * If there are multiple instances of the same executable, an additional identifier is appended. By default this looks like "java#1", "java#2".
+     * For some reason, the instance name can change after the process starts with the default naming scheme.
+     *
+     * To workaround this, add a DWORD registry value named 'ProcessNameFormat' set to the value '2' to the key
+     * 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\PerfProc\Performance'.
+     * This changes the naming scheme from "java#2" to "java_PID" where PID is the current process id. It also makes the name constant for the life of the process.
```
trask, line 166: I'm really hoping this isn't the problem. Having to instruct customers to modify a registry settings to make basic monitoring work would be unfortunate.

core/native.gradle
littleaj: oops, no. I'll fix that

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JniPCConnector.java
littleaj: Seems like a rare occurrence, but I wanted it documented somewhere. It took me a long time to find because all the SO articles linked to the same KB article which is 404. I finally found a reply with the instructions.

PR #1259, opened 2020-06-11
Update README with sampling.enabled property

smgn:
Add azure.application-insights.telemetry-processor.sampling.enabled property to the spring-boot-starter's readme.


smgn:
@littleaj , @dhaval24 , would you mind reviewing it? It took me half a day to find this property in the sources and I don't want others to have the same experience.

PR #1284, opened 2020-07-17
Update ETW Event names

littleaj:


heyams:
@trask do we still need this change? I can change the PR to be based off master..

trask:
@heyams it looks like a reasonable change, but I don't really understand the implications, maybe try it out when you are testing etw to understand it better?

heyams:
i double checked.. i don't much value having this change. We do have Level in the etw event. there is no need to use taskName to track logging level again. Let's chat offline if needed.


image

PR #1293, opened 2020-07-29
Resolve NullPointer exception when running under test with MockMvc

### LINKED ISSUE #1281: MockMvc-based tests fail with App Insights enabled
This the same issue as #895 . I was able to implement the recommended solution to do a test setup that stubbed out the App Insights initialization but this isn't a great solution for my project. 

Our approach is to build our APIs so we can enable App Insights through configuration where/as needed (by adding it in via our pom.xml), so we are trying to limit our app code to the Spring logging and instrumentation stack (SLF4J, Micrometer). This is working except in this case I need to add App Insights code to my tests need AI-specific code to work around this issue, breaking our approach.

Is it possible to put a null check [before in this line](https://github.com/microsoft/ApplicationInsights-Java/blob/master/web/src/main/java/com/microsoft/applicationinsights/web/internal/WebRequestTrackingFilter.java#L131) to avoid this issue entirely? That would avoid going through the extra plumbing to avoid this issue in my tests.

I'm happy to make a PR on this if the team agrees with this approach. 

The other approach would be a simple way to disable App Insights when running under test, but I don't see a way to do that.

andrejpk:
Fix #1281 by adding a simple null check in WebRquestTrackingFilter

If needed, here's a line for CHANGELOG.md (I don't know what version to tag it under. I'm happy to add this to my PR if I know where to put it):

* Fixed NullPointer when testing with MockMvc ([#](https://github.com/microsoft/ApplicationInsights-Java/issues/1281))

- [x] Changes in public surface reviewed (n/a)
- [x] CHANGELOG.md updated (see note above)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 2.6.2-BETA.3
+* Fixed NullPointer when testing with MockMvc ([#1281](https://github.com/microsoft/ApplicationInsights-Java/issues/1281))
+
 # Version 2.6.2-BETA.2
 * Fixed RequestTelemetryContext initialization ([#1247](https://github.com/microsoft/ApplicationInsights-Java/issues/1247)). Thanks, librucha!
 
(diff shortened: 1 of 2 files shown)


littleaj:
@andrejpk , Thanks for this! You can make a new section at the top of CHANGELOG.md for 2.6.2-BETA.3. We'll update it later if needed.

andrejpk:
Thanks, @littleaj , added.

PR #1311, opened 2020-09-11
Improve JMX performance counter

heyams:
Fix [#8242793](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=8242793), [#8087964](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=8087964), [#7965248](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=7965248)

boolean JMX counter will look like the following in Application Insights:
image

Testing was done manually using the spring-petclinic application.

### DIFF
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
@@ -32,6 +32,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 
+import com.microsoft.applicationinsights.agent.bootstrap.configuration.InstrumentationSettings.JmxMetric;
 import com.microsoft.applicationinsights.agent.bootstrap.configuration.InstrumentationSettings.PreviewConfiguration;
 import com.microsoft.applicationinsights.agent.bootstrap.diagnostics.DiagnosticsHelper;
 import com.squareup.moshi.JsonAdapter;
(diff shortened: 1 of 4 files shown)


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -64,6 +65,18 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+        threadCountJmxMetric.attribute = "ThreadCount";
+        threadCountJmxMetric.display = "Current Thread Count";
+        preview.jmxMetrics.add(threadCountJmxMetric);
+
+        JmxMetric classCountJmxMetric = new JmxMetric();
+        classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
+        classCountJmxMetric.attribute = "LoadedClassCount";
+        classCountJmxMetric.display = "Loaded Class Count";
+        preview.jmxMetrics.add(classCountJmxMetric);
+
```
trask, line 79: what happens if user has configured one of these in their ApplicationInsights.json file?

heyams: Good catch.  It will get sent twice.  I will remove the duplicate.

heyams: With further testing, duplicate only happens if the display name is different.  If they're the same, it only sent once.

trask (approved):

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
heyams: With this change, it will use customer's display name since that is what they prefer.

PR #1317, opened 2020-09-17
Add 3 env variables

heyams:
Fix [#7963772](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=7963772), [#8261291](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=8261291), [#8206362](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=8206362).

### DIFF
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
@@ -80,10 +80,9 @@ public class ConfigurationBuilder {
 
     private static void loadSamplingRateEnvVar(PreviewConfiguration preview) {
         final String samplingRate = overlayWithEnvVar(APPLICATIONINSIGHTS_SAMPLING_RATE, "100");
-        if (samplingRate != "100") {
-            configurationMessages.add(new ConfigurationMessage("Update default sampling rate from 100 to {}", samplingRate));
+        if (!samplingRate.equals("100")) {
             FixedRateSampling fixedRateSampling = new FixedRateSampling();
-            fixedRateSampling.percentage = new Double(samplingRate);
+            fixedRateSampling.percentage = Double.parseDouble(samplingRate);
             Sampling sampling = new Sampling();
             sampling.fixedRate = fixedRateSampling;
             preview.sampling = sampling;
@@ -92,14 +91,12 @@ public class ConfigurationBuilder {
 
     private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
         final String logCaptureThreshold = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, "INFO");
-        if (logCaptureThreshold != "INFO") {
-            configurationMessages.add(new ConfigurationMessage("Update default log capture threshold from INFO to {}", logCaptureThreshold));
-            final Map<String, Object> threshold = new HashMap<String, Object>() {{
-                put("threshold", logCaptureThreshold);
-            }};
-            preview.instrumentation = new HashMap<String, Map<String, Object>>() {{
-                put("logging", threshold);
-            }};
+        if (!logCaptureThreshold.equals("INFO")) {
+            Map<String, Object> logging = preview.instrumentation.get("logging");
+            if (logging == null) {
+                logging = new HashMap<>();
+            }
+            logging.put("threshold", logCaptureThreshold);
         }
     }
 
@@ -108,28 +105,26 @@ public class ConfigurationBuilder {
 
         // JmxMetrics env variable has higher precedence over jmxMetrics config from ApplicationInsights.json
         if (jmxMetricsEnvVarJson != null && !jmxMetricsEnvVarJson.isEmpty()) {
-            configurationMessages.add(new ConfigurationMessage("Update JMX metrics to {}", jmxMetricsEnvVarJson));
             Moshi moshi = new Moshi.Builder().build();
             Type listOfJmxMetrics = Types.newParameterizedType(List.class, JmxMetric.class);
             JsonReader reader = JsonReader.of(new Buffer().writeUtf8(jmxMetricsEnvVarJson));
             reader.setLenient(true);
             JsonAdapter<List<JmxMetric>> jsonAdapter = moshi.adapter(listOfJmxMetrics);
             preview.jmxMetrics = jsonAdapter.fromJson(reader);
-        } else {
-            if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=Threading", "ThreadCount")) {
-                JmxMetric threadCountJmxMetric = new JmxMetric();
-                threadCountJmxMetric.objectName = "java.lang:type=Threading";
-                threadCountJmxMetric.attribute = "ThreadCount";
-                threadCountJmxMetric.display = "Current Thread Count";
-                preview.jmxMetrics.add(threadCountJmxMetric);
-            }
-            if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
-                JmxMetric classCountJmxMetric = new JmxMetric();
-                classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
-                classCountJmxMetric.attribute = "LoadedClassCount";
-                classCountJmxMetric.display = "Loaded Class Count";
-                preview.jmxMetrics.add(classCountJmxMetric);
-            }
+        }
+        if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=Threading", "ThreadCount")) {
+            JmxMetric threadCountJmxMetric = new JmxMetric();
+            threadCountJmxMetric.objectName = "java.lang:type=Threading";
+            threadCountJmxMetric.attribute = "ThreadCount";
+            threadCountJmxMetric.display = "Current Thread Count";
+            preview.jmxMetrics.add(threadCountJmxMetric);
+        }
+        if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
+            JmxMetric classCountJmxMetric = new JmxMetric();
+            classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
+            classCountJmxMetric.attribute = "LoadedClassCount";
+            classCountJmxMetric.display = "Loaded Class Count";
+            preview.jmxMetrics.add(classCountJmxMetric);
         }
     }
 
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
@@ -95,6 +95,7 @@ public class ConfigurationBuilder {
             Map<String, Object> logging = preview.instrumentation.get("logging");
             if (logging == null) {
                 logging = new HashMap<>();
+                preview.instrumentation.put("logging", logging);
             }
             logging.put("threshold", logCaptureThreshold);
         }
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
@@ -70,9 +70,12 @@ public class ConfigurationBuilder {
     }
 
     private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
-        if (preview.instrumentation.containsKey("logging")) {
-            Map<String, Object> logging = preview.instrumentation.get("logging");
-            logging.put("threshold", overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, (String) logging.get("threshold")));
+        final String loggingEnvVar = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, (String)null);
+        if (loggingEnvVar != null) {
+            Map<String, Object> threshold = new HashMap<String, Object>() {{
+                put("threshold", loggingEnvVar);
+            }};
+            preview.instrumentation.put("logging", threshold);
         }
     }
 
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
@@ -48,7 +48,7 @@ public class ConfigurationBuilder {
     private static final String APPLICATIONINSIGHTS_CONFIGURATION_CONTENT = "APPLICATIONINSIGHTS_CONFIGURATION_CONTENT";
     private static final String APPLICATIONINSIGHTS_CONFIGURATION_FILE = "APPLICATIONINSIGHTS_CONFIGURATION_FILE";
     private static final String APPLICATIONINSIGHTS_JMX_METRICS = "APPLICATIONINSIGHTS_JMX_METRICS";
-    private static final String APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD = "APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD";
+    private static final String APPLICATIONINSIGHTS_LOGGING_THRESHOLD = "APPLICATIONINSIGHTS_LOGGING_THRESHOLD";
     private static final String APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE = "APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE";
 
     private static final String APPLICATIONINSIGHTS_ROLE_NAME = "APPLICATIONINSIGHTS_ROLE_NAME";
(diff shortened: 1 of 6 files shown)


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+        loadLogCaptureEnvVar(preview);
+        loadJmxMetrics(preview);
+
+        return config;
+    }
+
+    private static void loadSamplingRateEnvVar(PreviewConfiguration preview) {
+        final String samplingRate = overlayWithEnvVar(APPLICATIONINSIGHTS_SAMPLING_RATE, "100");
+        if (samplingRate != "100") {
+            configurationMessages.add(new ConfigurationMessage("Update default sampling rate from 100 to {}", samplingRate));
```
trask, line 84: `configurationMessages` are logged at `WARN` level, i think ok to not log this
```suggestion
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
-        if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
-            JmxMetric classCountJmxMetric = new JmxMetric();
-            classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
-            classCountJmxMetric.attribute = "LoadedClassCount";
-            classCountJmxMetric.display = "Loaded Class Count";
-            preview.jmxMetrics.add(classCountJmxMetric);
+    private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
+        final String logCaptureThreshold = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, "INFO");
+        if (logCaptureThreshold != "INFO") {
+            configurationMessages.add(new ConfigurationMessage("Update default log capture threshold from INFO to {}", logCaptureThreshold));
```
trask, line 96: same as above
```suggestion
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+        loadSamplingRateEnvVar(preview);
+        loadLogCaptureEnvVar(preview);
+        loadJmxMetrics(preview);
+
+        return config;
+    }
+
+    private static void loadSamplingRateEnvVar(PreviewConfiguration preview) {
+        final String samplingRate = overlayWithEnvVar(APPLICATIONINSIGHTS_SAMPLING_RATE, "100");
+        if (samplingRate != "100") {
```
trask, line 83:
```suggestion
        if (!samplingRate.equals("100")) {
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
 
-        if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
-            JmxMetric classCountJmxMetric = new JmxMetric();
-            classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
-            classCountJmxMetric.attribute = "LoadedClassCount";
-            classCountJmxMetric.display = "Loaded Class Count";
-            preview.jmxMetrics.add(classCountJmxMetric);
+    private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
+        final String logCaptureThreshold = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, "INFO");
+        if (logCaptureThreshold != "INFO") {
```
trask, line 95: .equals()

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
         }
+    }
 
-        return config;
+    private static void loadJmxMetrics(PreviewConfiguration preview) throws IOException {
+        String jmxMetricsEnvVarJson = overlayWithEnvVar(APPLICATIONINSIGHTS_JMX_METRICS, null);
+
+        // JmxMetrics env variable has higher precedence over jmxMetrics config from ApplicationInsights.json
+        if (jmxMetricsEnvVarJson != null && !jmxMetricsEnvVarJson.isEmpty()) {
+            configurationMessages.add(new ConfigurationMessage("Update JMX metrics to {}", jmxMetricsEnvVarJson));
```
trask, line 111: same as above
```suggestion
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+    private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
+        final String logCaptureThreshold = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, "INFO");
+        if (logCaptureThreshold != "INFO") {
+            configurationMessages.add(new ConfigurationMessage("Update default log capture threshold from INFO to {}", logCaptureThreshold));
+            final Map<String, Object> threshold = new HashMap<String, Object>() {{
+                put("threshold", logCaptureThreshold);
+            }};
+            preview.instrumentation = new HashMap<String, Map<String, Object>>() {{
+                put("logging", threshold);
+            }};
```
trask, line 102: currently this will overwrite existing instrumentation settings map provided from json file

if u change:
`public Map<String, Map<String, Object>> instrumentation = Collections.emptyMap()`
to
`public Map<String, Map<String, Object>> instrumentation = new HashMap<>()`

then can simplify:

```suggestion
            Map<String, Object> logging = preview.instrumentation.get("logging");
            if (logging == null) {
                logging = new HashMap<>();
            }
            logging.put("threshold", logCaptureThreshold);
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,70 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+                threadCountJmxMetric.display = "Current Thread Count";
+                preview.jmxMetrics.add(threadCountJmxMetric);
+            }
+            if (!jmxMetricExisted(preview.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
+                JmxMetric classCountJmxMetric = new JmxMetric();
+                classCountJmxMetric.objectName = "java.lang:type=ClassLoading";
+                classCountJmxMetric.attribute = "LoadedClassCount";
+                classCountJmxMetric.display = "Loaded Class Count";
+                preview.jmxMetrics.add(classCountJmxMetric);
+            }
```
trask, line 132: we want to add these additional default metrics whether or not env var was set, similar to how cpu and mem metrics are always captured

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,66 @@ public static Configuration create(Path agentJarPath) throws IOException {
         Configuration config = loadConfigurationFile(agentJarPath);
         PreviewConfiguration preview = config.instrumentationSettings.preview;
 
-        preview.roleName = overlayWithEnvVar(APPLICATIONINSIGHTS_ROLE_NAME, WEBSITE_SITE_NAME, preview.roleName);
+        preview.roleName = overlayWithEnvVars(APPLICATIONINSIGHTS_ROLE_NAME, WEBSITE_SITE_NAME, preview.roleName);
         preview.roleInstance =
-                overlayWithEnvVar(APPLICATIONINSIGHTS_ROLE_INSTANCE, WEBSITE_INSTANCE_ID, preview.roleInstance);
+                overlayWithEnvVars(APPLICATIONINSIGHTS_ROLE_INSTANCE, WEBSITE_INSTANCE_ID, preview.roleInstance);
 
+        loadSamplingRateEnvVar(preview);
```
trask, line 74: i think if we change
`public FixedRateSampling fixedRate;`
to
`public FixedRateSampling fixedRate = new FixedRateSampling();`

this this can be simpler:

```suggestion
        preview.sampling.fixedRate.percentage = overlayWithEnvVar(APPLICATIONINSIGHTS_SAMPLING_RATE, preview.sampling.fixedRate.percentage);
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,66 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+        loadLogCaptureEnvVar(preview);
```
trask, line 75: i think we could do the same here, by initializing `preview.instrumentation` map in PreviewConfiguration constructor, e.g.

```
        public PreviewConfiguration() {
            instrumentation.put("logging", new HashMap<>());
        }
```

then this could be simpler also:
```suggestion
        Map<String, Object> logging = preview.instrumentation.get("logging");
        logging.put("threshold", overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, logging.get("threshold")));
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,66 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+        loadJmxMetrics(preview);
+
+        return config;
+    }
+
+    private static void loadSamplingRateEnvVar(PreviewConfiguration preview) {
+        final String samplingRate = overlayWithEnvVar(APPLICATIONINSIGHTS_SAMPLING_RATE, "100");
+        if (!samplingRate.equals("100")) {
```
trask, line 83: i don't think this will handle the case correctly where json file has "50", and env var has "100"

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -61,27 +67,66 @@ public static Configuration create(Path agentJarPath) throws IOException {
...
+            fixedRateSampling.percentage = Double.parseDouble(samplingRate);
+            Sampling sampling = new Sampling();
+            sampling.fixedRate = fixedRateSampling;
+            preview.sampling = sampling;
+        }
+    }
+
+    private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
+        final String logCaptureThreshold = overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, "INFO");
+        if (!logCaptureThreshold.equals("INFO")) {
```
trask, line 94: same here, i don't think it will handle correctly json value "WARN" and env var "INFO"

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
heyams: This is not clear to me.  percentage might be null? percentage is Double, this overlay method returns string.

trask: oh yes, maybe overload `overlayWithEnvVar`:

```
    static Double overlayWithEnvVar(String name, Double defaultValue) {
        String value = getEnv(name);
        if (value != null && !value.isEmpty()) {
            return Double.parseDouble(value);
        }
        return defaultValue;
    }
```

trask:
hey, i was thinking, we should add some tests for the env var configuration, i pushed an example of how this can be done, checkout out https://github.com/microsoft/ApplicationInsights-Java/commit/342dd18af1bd9c82618e45a17987b19bee13e16f

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -57,31 +63,45 @@
...
-        preview.roleInstance =
-                overlayWithEnvVar(APPLICATIONINSIGHTS_ROLE_INSTANCE, WEBSITE_INSTANCE_ID, preview.roleInstance);
+        return config;
+    }
+
+    private static void loadLogCaptureEnvVar(PreviewConfiguration preview) {
+        if (preview.instrumentation.containsKey("logging")) {
+            Map<String, Object> logging = preview.instrumentation.get("logging");
+            logging.put("threshold", overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, (String) logging.get("threshold")));
+        }
```
trask, line 76: i don't think the condition is needed now with the new initialization? (also we always want to overlay env var)

```suggestion
        Map<String, Object> logging = preview.instrumentation.get("logging");
        logging.put("threshold", overlayWithEnvVar(APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD, (String) logging.get("threshold")));
```

heyams: The new initialization caused the micrometer to fail.  micrometer is using the same preview.instrumentation.. we can't initialize it to logging only.  it also supports for any kind of instrumentation names.  I need to delete the initialization and revert back to what i had originally.

trask: ok, makes sense

can you add unit tests that cover micrometer also? will be a good way to verify configuration behavior

also, need to overlay env var whether or not `preview.instrumentation.containsKey("logging")`, can you add unit test for that also?

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -32,20 +32,26 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ConfigurationBuilder {
 
     private static final String APPLICATIONINSIGHTS_CONFIGURATION_CONTENT = "APPLICATIONINSIGHTS_CONFIGURATION_CONTENT";
     private static final String APPLICATIONINSIGHTS_CONFIGURATION_FILE = "APPLICATIONINSIGHTS_CONFIGURATION_FILE";
+    private static final String APPLICATIONINSIGHTS_JMX_METRICS = "APPLICATIONINSIGHTS_JMX_METRICS";
+    private static final String APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD = "APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD";
+    private static final String APPLICATIONINSIGHTS_SAMPLING_RATE = "APPLICATIONINSIGHTS_SAMPLING_RATE";
```
trask, line 54: sorry, i think i picked the wrong name here, can you change this to `APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE`? that will match the json attributes better, and match the terminology better from other places as well, e.g. https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling#fixed-rate-sampling

agent/agent-bootstrap/src/test/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationTest.java
```diff
@@ -47,19 +52,71 @@ public void shouldParse() throws IOException {
...
+        assertEquals(60, preview.heartbeat.intervalSeconds);
+        assertEquals(3, preview.jmxMetrics.size());
+        assertEquals("error", preview.instrumentation.get("logging").get("threshold"));
+        assertEquals("true", preview.instrumentation.get("micrometer").get("enabled"));
+        assertEquals("true", preview.instrumentation.get("logging").get("enabled"));
+    }
+
+    @Test
+    public void shouldOverrideSamplingRate() throws IOException {
+        envVars.set("APPLICATIONINSIGHTS_SAMPLING_RATE", "25");
```
trask, line 71:
```suggestion
        envVars.set("APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE", "25");
```

agent/agent-bootstrap/src/test/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationTest.java
```diff
@@ -47,19 +52,71 @@ public void shouldParse() throws IOException {
...
+
+        Configuration configuration = loadConfiguration();
+        ConfigurationBuilder.overlayEnvVars(configuration);
+
+        assertTrue(configuration.instrumentationSettings.preview.sampling.fixedRate.percentage == 25);
+    }
+
+    @Test
+    public void shouldOverrideLogCaptureThreshold() throws IOException {
+        envVars.set("APPLICATIONINSIGHTS_LOG_CAPTURE_THRESHOLD", "TRACE");
```
trask, line 81:
```suggestion
        envVars.set("APPLICATIONINSIGHTS_LOGGING_THRESHOLD", "TRACE");
```

trask (approved):
👍

PR #1319, opened 2020-09-18
Drop java 7 

heyams:


trask (commented):
Try search also for "Java 7", there are some README.md files under smoke apps that can be deleted now, maybe other places also

trask (approved):

PR #1327, opened 2020-09-24
Fix for url with spaces

trask:
This was fixed in upstream otel repo recently, and will be available in the otel 0.9.0 release, but patching locally for now.


heyams:
👍

heyams (approved):

PR #1332, opened 2020-09-29
update sql dependency name

xiao-lix:
This PR is to finish [workitem](https://msazure.visualstudio.com/One/_workitems/edit/7963836)


trask (approved):
👍

PR #1339, opened 2020-10-05
Use the correct metric weight for Live Metrics

heyams:
This will address this issue [#206218273](https://portal.microsofticm.com/imp/v3/incidents/details/206218273/home).


trask (approved):
Great find 👍

PR #1341, opened 2020-10-08
Lazily set connection string in the Azure Function instrumentation

heyams:


instrumentation/azure-functions/src/main/java/io/opentelemetry/instrumentation/auto/azure/functions/AzureFunctionsInstrumentation.java
```diff
@@ -86,5 +91,25 @@ public static Scope methodEnter(@Advice.Argument(0) final Object request)
     public static void methodExit(@Advice.Enter final Scope scope) {
       scope.close();
     }
+
+    /**
+     * Azure Function Linux with consumption plan doesn't come with 'APPLICATIONINSIGHTS_CONNECTION_STRING' env variable during the initialization.
+     * Need to set this connection string lazily when it becomes available.
+     */
+    private static void lazilySetConnectionString() {
+      if (!AiConnectionString.hasConnectionString()) {
```
trask, line 100: just a comment here
```suggestion
      // race condition (two initial requests happening at the same time) is not a worry here
      // because at worst they both enter the condition below and update the connection string
      if (!AiConnectionString.hasConnectionString()) {
```

instrumentation/azure-functions/src/main/java/io/opentelemetry/instrumentation/auto/azure/functions/AzureFunctionsInstrumentation.java
```diff
@@ -86,5 +91,25 @@ public static Scope methodEnter(@Advice.Argument(0) final Object request)
...
+        String connectionString = System.getenv("APPLICATIONINSIGHTS_CONNECTION_STRING");
+        if (connectionString != null && !connectionString.isEmpty()) {
+          AiConnectionString.setConnectionString(connectionString);
+          // TODO to be deleted later once the testing is completed
+          log.debug("Lazily set the connection string for Azure Function Linux Consumption Plan" + connectionString);
+        } else {
```
trask, line 106: fall back to `APPINSIGHTS_INSTRUMENTATIONKEY` to support azure functions that were created prior to `APPLICATIONINSIGHTS_CONNECTION_STRING`

instrumentation/jdbc/src/main/java/io/opentelemetry/instrumentation/auto/jdbc/JdbcTracer.java
```diff
@@ -84,7 +84,6 @@ public Span startSpan(Statement statement, String query) {
     }
 
     DBInfo dbInfo = extractDbInfo(connection);
-
```
trask, line 87: can u undo this formatting change? only because we are merging into here from otel, so will help avoid merge conflicts

instrumentation/jdbc/src/main/java/io/opentelemetry/instrumentation/auto/jdbc/StatementInstrumentation.java
```diff
@@ -86,7 +86,6 @@ public static void onEnter(
         @Advice.Local("otelSpan") Span span,
         @Advice.Local("otelScope") Scope scope,
         @Advice.Local("otelCallDepth") Depth callDepth) {
-
```
trask, line 89: same here, thx

instrumentation/azure-functions/src/main/java/io/opentelemetry/instrumentation/auto/azure/functions/AzureFunctionsInstrumentation.java
```diff
@@ -69,6 +71,20 @@ public AzureFunctionsInstrumentation() {
...
     public static Scope methodEnter(@Advice.Argument(0) final Object request)
         throws ReflectiveOperationException {
+      // race condition (two initial requests happening at the same time) is not a worry here
+      // because at worst they both enter the condition below and update the connection string
+      if (!AiConnectionString.hasConnectionString()) {
+        String connectionString = System.getenv("APPLICATIONINSIGHTS_CONNECTION_STRING");
+        if (!Strings.isNullOrEmpty(connectionString)) {
+          AiConnectionString.setConnectionString(connectionString);
+        } else {
+          // if the instrumentation key is neither null nor empty , we will create a default connection string based on the instrumentation key.
```
trask, line 81:
```suggestion
          // if the instrumentation key is neither null nor empty , we will create a default connection string based on the instrumentation key.
          // this is to support Azure Functions that were created prior to the introduction of connection strings
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/aiconnectionstring/AiConnectionString.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.api.aiconnectionstring;
+
+public class AiConnectionString {
+
+  private static Accessor accessor;
```
trask, line 21: just to be safe, since set and get are from different threads
```suggestion
  private static volatile Accessor accessor;
```

trask (approved):
👍

PR #1342, opened 2020-10-08
Add azure function linux consumption plan support on the agent side

heyams:
This is pending on the otel-fork changes #1341.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
@@ -40,7 +40,7 @@ public class ConnectionStringAccessor implements AiConnectionString.Accessor {
     @Override public void setValue(String value) {
         if (!Strings.isNullOrEmpty(value)) {
             TelemetryConfiguration.getActive().setConnectionString(value);
-            logger.info("Set connection string lazily for the Azure Function Consumption Plan.");
+            logger.debug("Set connection string lazily for the Azure Function Consumption Plan.");
         }
     }
 }
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
@@ -31,17 +31,23 @@ public class ConnectionStringAccessor implements AiConnectionString.Accessor {
 
     private static final Logger logger = LoggerFactory.getLogger(ConnectionStringAccessor.class);
 
+    public ConnectionStringAccessor() {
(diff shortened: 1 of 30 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override public boolean hasValue() {
+        //check for instrumentation key value is sufficient here because it's updated each time when the connection string is updated.
+        String instrumentationKey = TelemetryConfiguration.getActive().getInstrumentationKey();
+        return !Strings.isNullOrEmpty(instrumentationKey);
+    }
+
+    @Override public void setValue(String value) {
+        if (!Strings.isNullOrEmpty(value)) {
+            TelemetryConfiguration.getActive().setConnectionString(value);
+            logger.info("Set connection string lazily for the Azure Function Consumption Plan.");
```
trask, line 43:
```suggestion
            logger.debug("Set connection string lazily for the Azure Function Consumption Plan.");
```

agent/agent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -29,6 +30,14 @@ public static void installAgentTracer() {
                     TraceConfig.getDefault().toBuilder()
                             .setSampler(new FixedRateSampler(fixedRateSamplingPercentage))
                             .build());
+        } else {
+            // TODO will revisit to delete DefaultSpan
+            // For the case of AzureFunctionInstrumentation, TraceParent comes in this format: 00-ea5a92ef3cf03649acf1b89c2b5d4211-b70350d129319a4b-00
+            // Sampling is always off.  We need to use AlwaysOnSampler.
```
trask, line 36:
```suggestion
            // OpenTelemetry default sampling is "parent based", which means don't sample if remote traceparent sampled flag was not set,
            // and Azure Functions is not setting the sampled flag on traceparent currently, so we can't use the default currently, and instead default to "always on" in this case
            // TODO revisit using "parent based" both for 100% and fixed-rate sampler above
```

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -82,6 +83,11 @@ public Exporter(TelemetryClient telemetryClient) {
 
     @Override
     public CompletableResultCode export(Collection<SpanData> spans) {
+        if (Strings.isNullOrEmpty(TelemetryConfiguration.getActive().getInstrumentationKey())) {
+            logger.info("Instrumentation key is null or empty.");
```
trask, line 87:
```suggestion
            logger.debug("Instrumentation key is null or empty.");
```

trask (approved):
👍

PR #1348, opened 2020-10-18
Support resource attributes

xiao-lix:
Support resource attributes. 
[workitem](https://msazure.visualstudio.com/One/_workitems/edit/7013136)


trask (approved):
👍

PR #1350, opened 2020-10-19
Add smoke test for SpringWebClient

heyams:
Fix [#8406278](https://msazure.visualstudio.com/One/_boards/board/t/Azure%20Attach%20(Java%20NET)/Features/?workitem=8406278)


trask (approved):
👍

PR #1355, opened 2020-10-22
Update exporter mappings

trask:
Improvements to the exporter mappings.

Also a commit that fixes `CdsProfileFetcher` to use configured http proxy.

### DIFF
--- a/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
+++ b/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
@@ -78,9 +78,6 @@ public class Exporter implements SpanExporter {
     private static final AttributeKey<String> SPAN_SOURCE_ATTRIBUTE_NAME = AttributeKey.stringKey(AiAppId.SPAN_SOURCE_ATTRIBUTE_NAME);
     private static final AttributeKey<String> SPAN_TARGET_ATTRIBUTE_NAME = AttributeKey.stringKey(AiAppId.SPAN_TARGET_ATTRIBUTE_NAME);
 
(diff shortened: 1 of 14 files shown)


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -412,34 +402,41 @@ private static void setProperties(Map<String, String> properties, long timeEpoch
...
+            // if you use just HTTP, Breeze will remove appid from the target
+            telemetry.setType("Http (tracked component)");
+            telemetry.setTarget(target + " | " + targetAppId);
+        }
+
+        String url = removeAttributeString(attributes, SemanticAttributes.HTTP_URL);
         if (url != null) {
             try {
                 URI uriObject = new URI(url);
-                String target = createTarget(uriObject);
```
heyams, line 430: createTarget appends port to be part of the target.. i don't see that get added back.. are we dropping port entirely?

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -553,14 +561,6 @@ private static boolean removeAttributeBoolean(Map<AttributeKey<?>, Object> attri
         }
     }
 
-    private static String createTarget(URI uriObject) {
```
heyams, line 556: i was referring to this method.

heyams (approved):
👍

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: good catch! yes, we want to keep the port, i will fix 👍

PR #1360, opened 2020-10-26
Support live update of sampling rate and connection string without a restart

heyams:
Fixed [#8534376](https://msazure.visualstudio.com/One/_workitems/edit/8534376)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -187,6 +196,44 @@ public void run() {
                 }
             }
         });
+
+        pollJsonConfigEveryMinute();
+    }
+
+    private static volatile Long lastModifiedTime;
+    private static void pollJsonConfigEveryMinute() {
+        Executors.newSingleThreadScheduledExecutor().scheduleWithFixedDelay(new Runnable() {
```
trask, line 205: need to make this a daemon thread so it doesn't prevent JVM shutdown

also, let's give the thread a name, for when it shows up in thread dump

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -187,6 +196,44 @@ public void run() {
...
+            @Override public void run() {
+                Path path = MainEntryPoint.getConfigPath();
+                if (path != null) {
+                    if (Files.exists(path)) {
```
trask, line 209: to avoid some of the nesting:

```
if (path == null) {
    return;
}
if (!Files.exist(path)) {
    return;
}
try {
    ...
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -69,10 +76,11 @@
 import org.slf4j.LoggerFactory;
 
 import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 public class BeforeAgentInstaller {
 
-    private static Logger startupLogger = LoggerFactory.getLogger("com.microsoft.applicationinsights.agent");
+    private static final Logger startupLogger = LoggerFactory.getLogger("com.microsoft.applicationinsights.agent");
```
heyams, line 83: when i tested this change with spring-petclinic app, startupLogger didn't output to console even though i set it to info level... it does when I used "system.out.println".  Do you know what might be the cause?  Everywhere else, we pass Class name the LoggerFactor.getLogger.. here we use part of the package name... is there a reason for that? don't want to disclose class name in the log?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -187,6 +195,54 @@ public void run() {
...
         });
+
+        pollJsonConfigEveryMinute();
+    }
+
+    private static volatile Long lastModifiedTime;
+    private static void pollJsonConfigEveryMinute() {
+        Thread pollingThread = new Thread(new JsonConfigPolling());
+        pollingThread.setDaemon(true);
+        Executors.newSingleThreadScheduledExecutor().scheduleWithFixedDelay(pollingThread, 60, 60, SECONDS);
```
trask, line 206: this is a bit more consistent with other places in code:

Executors.newSingleThreadScheduledExecutor(ThreadPoolUtils.createDaemonThreadFactory(JsonConfigPolling.class))
.scheduleWithFixedDelay(...)

and then don't need to create `Thread`

heyams: scheduleWithFixedDelay requires a runnable.  I also looked into the implementation of ThreadPoolUtils.. it created a new thread and set daemon to true.  it's essentially the same..

heyams: it's mostly for threadFactory and threadPool.

trask: `ThreadPoolUtils` creates a `ThreadFactory` (not a `Thread`), which is a more typical way to configure a thread pool

heyams: yup, spotBug complained too: DMI_THREAD_PASSED_WHERE_RUNNABLE_EXPECTED.  😢

heyams: Updated.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
```diff
@@ -21,7 +21,11 @@
 
 package com.microsoft.applicationinsights.agent.bootstrap.configuration;
 
+import java.nio.file.Path;
+
 public class Configuration {
 
     public InstrumentationSettings instrumentationSettings = new InstrumentationSettings();
+
+    public transient Path configPath;
```
trask, line 30:
```suggestion
    // transient so that Moshi will ignore when binding from json
    public transient Path configPath;
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -292,4 +281,21 @@ private void log(Logger logger) {
             logger.warn(message, args);
         }
     }
+
+    public static Configuration loadJsonConfigFile(Path configPath) throws IOException{
+        assert(Files.exists(configPath));
```
trask, line 286: assert keyword is not used much (since not active by default without `-ea` flag), better to throw exception, e.g.

```suggestion
        if (!Files.exists(configPath)) {
            throw new IllegalStateException("config file does not exist: " + configPath);
        }
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -187,6 +196,53 @@ public void run() {
...
+        pollJsonConfigEveryMinute();
+    }
+
+    private static volatile Long lastModifiedTime;
+    private static void pollJsonConfigEveryMinute() {
+        Executors.newSingleThreadScheduledExecutor(ThreadPoolUtils.createDaemonThreadFactory(JsonConfigPolling.class))
+                .scheduleWithFixedDelay(new JsonConfigPolling(), 60, 60, SECONDS);
+    }
+
+    private static class JsonConfigPolling implements Runnable {
```
trask, line 209: i think this would be good class to extract out into top-level, to avoid cluttering `BeforeAgentInstaller`

you can track `lastModifiedTime` as instance field

and would be great if we can get `lastModifiedTime` from before the initial loading (add another field to `Configuration`?), that way we avoid a potential race condition of the config file being updated between when we read it the first time, and when this thread starts checking for updates

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
```diff
@@ -64,6 +71,8 @@ public static void start(Instrumentation instrumentation, URL bootstrapURL) {
             Configuration config = ConfigurationBuilder.create(agentJarFile.toPath());
             configuration = config.instrumentationSettings;
             configPath = config.configPath;
+            BasicFileAttributes attributes = Files.readAttributes(configPath, BasicFileAttributes.class);
+            lastModifiedTime = attributes.lastModifiedTime().toMillis();
```
trask, line 75: let's capture `lastModifiedTime` before reading the file, just to completely eliminate race condition of the file being updated in between reading it and capturing the `lastModifiedTime`

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -292,4 +282,27 @@ private void log(Logger logger) {
...
         }
     }
+
+    public static Configuration loadJsonConfigFile(Path configPath) throws IOException{
+        if (!Files.exists(configPath)) {
+            throw new IllegalStateException("config file does not exist: " + configPath);
+        }
+        
+        BasicFileAttributes attributes = Files.readAttributes(configPath, BasicFileAttributes.class);
+        Long lastModifiedTime = attributes.lastModifiedTime().toMillis();
```
trask, line 292:
```suggestion
        // important to read last modified before reading the file, to prevent possible race condition
        // where file is updated after reading it but before reading last modified, and then since
        // last modified doesn't change after that, the new updated file will not be read afterwards
        Long lastModifiedTime = attributes.lastModifiedTime().toMillis();
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+import com.microsoft.applicationinsights.agent.bootstrap.configuration.Configuration;
+import com.microsoft.applicationinsights.agent.bootstrap.configuration.ConfigurationBuilder;
+import com.microsoft.applicationinsights.agent.bootstrap.configuration.InstrumentationSettings.FixedRateSampling;
+import com.microsoft.applicationinsights.internal.util.ThreadPoolUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+public final class JsonConfigWatcher {
```
trask, line 42: no need for nested class now, can do something like this:
```suggestion
private static class JsonConfigPolling implements Runnable {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private static volatile Long lastModifiedTime = MainEntryPoint.getLastModifiedTime();
```
trask, line 44: and then this can become instance field
```suggestion
    private volatile Long lastModifiedTime = MainEntryPoint.getLastModifiedTime();
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigPolling.java
```diff
@@ -31,58 +31,56 @@
...
 import org.slf4j.LoggerFactory;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
-public final class JsonConfigWatcher {
+public class JsonConfigPolling implements Runnable {
 
-    private static volatile Long lastModifiedTime = MainEntryPoint.getLastModifiedTime();
-    private static final Logger logger = LoggerFactory.getLogger(JsonConfigWatcher.class);
+    private volatile Long lastModifiedTime = MainEntryPoint.getLastModifiedTime();
```
trask, line 44: instead of pulling this from MainEntryPoint, i think cleaner to push lastModifiedTime into `pollJsonConfigEveryMinute(long)` method, and then push from there into `JsonConfigPolling` constructor

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigPolling.java
```diff
@@ -31,58 +31,56 @@
...
+                    logger.debug("Override fixed rate sampling percentage from " + Global.getSamplingPercentage() + " to " + sampling.percentage + " ");
+                    Global.setSamplingPercentage(sampling.percentage);
                 }
-            } catch (IOException e) {
-                e.printStackTrace();
-                logger.error("Error occurred when polling json config file: " + e.toString());
             }
+        } catch (IOException e) {
+            e.printStackTrace();
+            logger.error("Error occurred when polling json config file: " + e.toString());
```
trask, line 83:
```suggestion
             logger.error("Error occurred when polling json config file: {}", e.getMessage(), e);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigPolling.java
trask: the idea is that makes JsonConfigPolling self-contained, easier to test, and understand on it's own

trask (approved):
👍

PR #1361, opened 2020-10-28
Exporter mapping improvements

trask:
A few things:
* by adding special case for "spring-scheduling" INTERNAL span, it will prevent misc INTERNAL spans from showing up in request table
* fixed mapping of CONSUMER spans (due to new instrumentation in 0.10.0 that hasn't been released yet)
* fixed dependency type for messaging spans that are not PRODUCER

### DIFF
--- a/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
+++ b/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
@@ -53,6 +53,7 @@ import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.data.SpanData.Event;
 import io.opentelemetry.sdk.trace.data.SpanData.Link;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
+import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.Span.Kind;
 import io.opentelemetry.trace.SpanId;
 import io.opentelemetry.trace.attributes.SemanticAttributes;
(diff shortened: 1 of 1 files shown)


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -118,16 +119,19 @@ private void export(SpanData span) {
...
+            } else if ("spring-scheduling".equals(stdComponent) && !SpanId.isValid(span.getParentSpanId())) {
+                // TODO need semantic convention for determining whether to map INTERNAL to request or dependency
+                //  (or need clarification to use SERVER for this)
                 exportRequest(span, attributes);
             } else {
                 exportRemoteDependency(span, attributes, true);
             }
         } else if (kind == Kind.CLIENT || kind == Kind.PRODUCER) {
             exportRemoteDependency(span, attributes, false);
+        } else if (kind == Kind.CONSUMER && !span.getHasRemoteParent()) {
```
heyams, line 131: getHasRemoteParent, can we fix it in the opentelemetry-api? it doesn't seem to follow the naming convention for boolean getter.

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -118,16 +119,19 @@ private void export(SpanData span) {
             boolean isLog = removeAttributeBoolean(attributes, AI_INTERNAL_LOG);
             if (isLog) {
                 exportLogSpan(span, attributes);
-            } else if (!SpanId.isValid(span.getParentSpanId())) {
-                // TODO revisit this decision
-                // maybe user-generated telemetry?
-                // otherwise this top-level span won't show up in Performance blade
+            } else if ("spring-scheduling".equals(stdComponent) && !SpanId.isValid(span.getParentSpanId())) {
```
heyams, line 122: Should i rename SpringScheduler to SpringScheduling for smoke test?  It's SpringSchedulingInstrumentation on the opentelemetry as well.

heyams (approved):
👍

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: sure, that seems more consistent 👍

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: good point! I just opened https://github.com/open-telemetry/opentelemetry-java/issues/1910

PR #1364, opened 2020-10-28
More configuration changes

trask:
* Round sampling percentage to the nearest 100/N, and log warning if the result is "significantly" different from what the user configured
* Make OpenTelemetry API support opt-in prior to OTel GA


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -126,6 +126,9 @@ private static void start(Instrumentation instrumentation) throws Exception {
             properties.put("ota.integration.java-util-logging.enabled", "false");
             properties.put("ota.integration.logback.enabled", "false");
         }
+        if (!config.preview.openTelemetryApiSupport) {
+            properties.put("ota.integration.opentelemetry-api.enabled", "false");
```
heyams, line 130: // TODO update doc?

heyams (approved):
👍

PR #1365, opened 2020-10-29
Send JMX metrics to CustomMetrics instead of PerformanceCounter

heyams:
CustomMetrics:
"Private Bytes"
“Heap Memory Used (MB)”
“GC Total Time”
“GC Total Count”
"Loaded Class Count"
"Current Thread Count"
"Suspected Deadlocked Threads"

Keep the following in PerformanceCounter (used by the Performance blade on the portal):
“Available Bytes”
“% Processor Time”
“IO Data Bytes/sec”

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
@@ -29,6 +29,8 @@ import com.microsoft.applicationinsights.telemetry.MetricTelemetry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.microsoft.applicationinsights.internal.perfcounter.Constants.TOTAL_MEMORY_PC_COUNTER_NAME;
+
 /**
  * A performance counter that sends {@link com.microsoft.applicationinsights.telemetry.MetricTelemetry}
  *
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
```diff
@@ -46,11 +48,15 @@ public JmxMetricPerformanceCounter(String id, String objectName, Collection<JmxA
...
-    telemetry.markAsCustomPerfCounter();
+        MetricTelemetry telemetry = new MetricTelemetry();
         telemetry.setName(displayName);
         telemetry.setValue(value);
-        telemetry.getProperties().put("CustomPerfCounter", "true");
+
+        // keep "Available Bytes" under PerformanceCounter to prevent breaking performance blade on the ApplicationInsights Portal
+        if (displayName.equals(TOTAL_MEMORY_PC_COUNTER_NAME)) {
+            telemetry.markAsCustomPerfCounter();
+        }
```
trask, line 58: `TOTAL_MEMORY_PC_COUNTER_NAME` looks like it's captured in `FreeMemoryPerformanceCounter`, not here?

trask:
It looks like one valid smoke test failure

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/JmxMetricPerformanceCounter.java
heyams: TelemetryConfigurationFactory.java: line 357.  it's eventually JmxMetricPerformanceCounter.

if (PerformanceCounterContainer.INSTANCE.register(new JmxMetricPerformanceCounter(entry.getKey(), entry.getKey(), entry.getValue()))) {

trask (approved):

PR #1366, opened 2020-10-29
Exporter mapping updates for target/source

trask:


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -405,19 +397,38 @@ private static void setProperties(Map<String, String> properties, String level,
...
+                    logger.debug(e.getMessage(), e);
+                }
+            }
         }
-        Long port = removeAttributeLong(attributes, SemanticAttributes.NET_PEER_PORT);
-        if (target != null && port != null && port != 443 && port != 80) {
-            target += ":" + port;
+        if (target == null) {
+            // this should not happen, just a failsafe
+            target = "http";
```
heyams, line 429: Http?  type is 'Http', should we make it consistent with type here?

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -459,14 +469,53 @@ private static void applyDatabaseClientSpan(Map<AttributeKey<?>, Object> attribu
...
+            return null;
+        }
+        // append net.peer.port to target
+        Long port = removeAttributeLong(attributes, SemanticAttributes.NET_PEER_PORT);
+        if (port != null && port != 443 && port != 80) {
+            return target + ":" + port;
         } else {
-            // TODO fall back to PEER_SERVICE/NET_PEER_NAME/NET_PEER_IP (e.g. for Redis)?
-            telemetry.setTarget(dbSystem);
+            return target;
```
heyams, line 518: Can you remove else block here?  simply "return target;"

trask:
@heyams thanks! I added a GA work item to apply your feedback, so we can merge this without waiting for another CI smoke test build

heyams (approved):

PR #1384, opened 2020-11-14
Remove guava usage from Azure Functions instrumentation

trask:
cc: @TsuyoshiUshio


instrumentation/azure-functions/src/main/java/io/opentelemetry/instrumentation/auto/azure/functions/AzureFunctionsInstrumentationHelper.java
```diff
@@ -1,14 +1,13 @@
 package io.opentelemetry.instrumentation.auto.azure.functions;
 
-import com.google.common.base.Strings;
```
trask, line 3: this is the fix, the other changes are just small cleanups

TsuyoshiUshio: Great :)

PR #1387, opened 2020-11-19
Enable Awesome Logging

kryalama:
Feature request

Includes friendly error messages for
* Missing connection string
* Invalid SSL certificate issues when not able to connect to IngestionEndPoint Url, Live endpoint Url and CdsProfiler endpoint url.
* Invalid Telemetry Processor Configuration

The following are forwarding urls used in the log messages
https://go.microsoft.com/fwlink/?linkid=2151450  -> Troubleshooting Docs
https://go.microsoft.com/fwlink/?linkid=2151557  -> Telemetry Processors
https://go.microsoft.com/fwlink/?linkid=2151451  -> Java IPA Overview

### DIFF
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
@@ -103,7 +103,8 @@ public class MainEntryPoint {
         }
     }
 
-    private static FriendlyException getFriendlyException(Throwable t) {
+    // visible for testing
+    static FriendlyException getFriendlyException(Throwable t) {
         if (t instanceof FriendlyException) {
             return (FriendlyException) t;
         }
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
@@ -122,7 +122,11 @@ public class MainEntryPoint {
                 startupLogger.error(message, t);
             }
         } else {
-            System.err.println(message);
+            if (isFriendlyException) {
+                System.err.println(message);
+            } else {
+                t.printStackTrace();
+            }
         }
     }
 
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
@@ -21,17 +21,15 @@
 package com.microsoft.applicationinsights.agent.bootstrap;
 
 import java.io.File;
-import java.lang.instrument.ClassFileTransformer;
-import java.lang.instrument.IllegalClassFormatException;
 import java.lang.instrument.Instrumentation;
 import java.net.URL;
 import java.nio.file.Path;
-import java.security.ProtectionDomain;
 
 import ch.qos.logback.classic.Level;
 import com.microsoft.applicationinsights.agent.bootstrap.configuration.Configuration;
 import com.microsoft.applicationinsights.agent.bootstrap.configuration.Configuration.SelfDiagnostics;
 import com.microsoft.applicationinsights.agent.bootstrap.configuration.ConfigurationBuilder;
+import com.microsoft.applicationinsights.agent.bootstrap.customExceptions.FriendlyException;
 import com.microsoft.applicationinsights.agent.bootstrap.diagnostics.DiagnosticsHelper;
 import com.microsoft.applicationinsights.agent.bootstrap.diagnostics.status.StatusFile;
 import io.opentelemetry.javaagent.bootstrap.AgentInitializer;
@@ -83,11 +81,14 @@ public class MainEntryPoint {
         } catch (ThreadDeath td) {
             throw td;
         } catch (Throwable t) {
-            if (startupLogger != null) {
-                startupLogger.error("ApplicationInsights Java Agent failed to start", t);
+
+            FriendlyException friendlyException = getFriendlyException(t);
+            if (friendlyException != null) {
+                logErrorMessage(startupLogger, friendlyException.getMessage(), true, t);
             } else {
-                t.printStackTrace();
+                logErrorMessage(startupLogger, "ApplicationInsights Java Agent failed to start", false, t);
             }
+
         } finally {
             try {
                 StatusFile.putValueAndWrite("AgentInitializedSuccessfully", success, startupLogger != null);
@@ -99,6 +100,29 @@ public class MainEntryPoint {
                 }
             }
             MDC.clear();
+        }
+    }
+
+    private static FriendlyException getFriendlyException(Throwable t) {
+        if (t instanceof FriendlyException) {
+            return (FriendlyException) t;
+        }
+        Throwable cause = t.getCause();
+        if (cause == null) {
+            return null;
+        }
+        return getFriendlyException(cause);
+    }
+
+    private static void logErrorMessage(Logger startupLogger, String message, boolean isFriendlyException, Throwable t) {
+        if (startupLogger != null) {
+            if (isFriendlyException) {
+                startupLogger.error(message);
+            } else {
+                startupLogger.error(message, t);
+            }
+        } else {
+            System.err.println(message);
         }
     }
 
--- a/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
+++ b/agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
@@ -29,7 +29,7 @@ import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
-import com.microsoft.applicationinsights.agent.bootstrap.configuration.ConfigurationBuilder.ConfigurationException;
+import com.microsoft.applicationinsights.agent.bootstrap.customExceptions.FriendlyException;
 
 import static java.util.concurrent.TimeUnit.MINUTES;
 
@@ -122,17 +122,23 @@ public class Configuration {
         public List<ProcessorAction> actions; // specific for processor type "attributes"
         public NameConfig name; // specific for processor types "log" and "span"
 
-        private static void isValidRegex(String value) {
+        private static void isValidRegex(String value) throws FriendlyException {
             try {
                 Pattern.compile(value);
             } catch (PatternSyntaxException exception) {
-                throw new ConfigurationException("User provided processor config do not have valid regex:" + value);
+                throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                "User provided processor configuration do not have valid regex:" + value,
+                                "Please provide a valid regex in the telemetry processors configuration. " +
+                                "Learn more about telemetry processors here: https://go.microsoft.com/fwlink/?linkid=2151557");
             }
         }
 
-        public void validate() {
+        public void validate() throws FriendlyException {
             if (type == null) {
-                throw new ConfigurationException("User provided config has a processor with no type!!!");
+                throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                "User provided processor configuration has a processor with no type!!!",
+                                "Please provide a type in the telemetry processors configuration. " +
+                                "Learn more about telemetry processors here: https://go.microsoft.com/fwlink/?linkid=2151557");
             }
             if (include != null ) {
                 include.validate(type);
@@ -144,10 +150,13 @@ public class Configuration {
             validateLogOrSpanProcessorConfig();
         }
 
-        public void validateAttributeProcessorConfig() {
+        public void validateAttributeProcessorConfig() throws FriendlyException {
             if (type == ProcessorType.attribute) {
                 if (actions == null || actions.isEmpty()) {
-                    throw new ConfigurationException("User provided config has invalid attribute processor configuration with empty actions!!!");
+                    throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                    "User provided configuration has invalid attribute processor configuration with empty actions!!!",
+                                    "Please provide atleast one action in the attribute processors configuration. " +
+                                    "Learn more about telemetry processors here: https://go.microsoft.com/fwlink/?linkid=2151557");
                 }
                 for (ProcessorAction action : actions) {
                     action.validate();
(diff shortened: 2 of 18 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -99,6 +100,8 @@ public QuickPulseStatus ping() {
                         break;
                 }
             }
+        } catch (FriendlyException e) {
+            // chomp
```
trask, line 104: this seems worth logging, e.g. if SSL cert issue

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -185,6 +187,12 @@ public boolean sendSync(Transmission transmission) {
             } catch (IOException ioe) {
                 ex = ioe;
                 logger.error("Failed to send", ioe);
+            } catch (FriendlyException e) {
+                ex = e;
+                if(!friendlyExceptionThrown) {
+                    logger.error(e.getMessage());
```
trask, line 193: let's log at the bottom of the message something like

`(this message is only logged the first time it occurs after startup)`

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -71,8 +73,14 @@ private ApacheSender43() {
...
     @Override
-    public HttpResponse sendPostRequest(HttpPost post) throws IOException {
-        return getHttpClient().execute(post);
+    public HttpResponse sendPostRequest(HttpPost post) throws IOException, FriendlyException {
+        try {
+            return getHttpClient().execute(post);
+        } catch (SSLHandshakeException e) {
+            throw new FriendlyException("ApplicationInsights Java Agent failed to send telemetry data.",
+                    "Unable to find valid certification path to requested target",
+                    "Please import the Breeze certificate into the java key store.");
```
trask, line 82: can we log the hostname here, so they know which host they need to import the SSL cert for?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -78,8 +78,9 @@ public HttpResponse sendPostRequest(HttpPost post) throws IOException, FriendlyE
             return getHttpClient().execute(post);
         } catch (SSLHandshakeException e) {
             throw new FriendlyException("ApplicationInsights Java Agent failed to send telemetry data.",
-                    "Unable to find valid certification path to requested target",
-                    "Please import the Breeze certificate into the java key store.");
+                    "Unable to find valid certification path to requested target.",
+                    "Please import the Breeze certificate from "+post.getURI().getHost()+", into the java key store.",
```
trask, line 82:
```suggestion
                    "Please import the SSL certificate from "+post.getURI().getHost()+", into the java key store.",
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
```diff
@@ -102,6 +103,29 @@ public static void start(Instrumentation instrumentation, URL bootstrapURL) {
...
+
+    private static void logErrorMessage(Logger startupLogger, String message, boolean isFriendlyException, Throwable t) {
+        if (startupLogger != null) {
+            if (isFriendlyException) {
+                startupLogger.error(message);
+            } else {
+                startupLogger.error(message, t);
+            }
+        } else {
+            System.err.println(message);
```
trask, line 125:
```suggestion
            if (isFriendlyException) {
                System.err.println(message);
            } else {
                t.printStackTrace();
            }
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -108,7 +108,10 @@ private static void start(Instrumentation instrumentation) throws Exception {
         Configuration config = MainEntryPoint.getConfiguration();
         if (!hasConnectionStringOrInstrumentationKey(config)) {
             if (!("java".equals(System.getenv("FUNCTIONS_WORKER_RUNTIME")))) {
-                throw new ConfigurationException("No connection string or instrumentation key provided");
+                throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                            "No connection string or instrumentation key provided",
+                                            "Please provide connection string or instrumentation key.","");
+                //throw new ConfigurationException("No connection string or instrumentation key provided");
```
trask, line 114:
```suggestion
```

trask (commented):
also try to repro and catch SSL cert issue in `CdsProfileFetcher`

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
```diff
@@ -102,6 +103,29 @@ public static void start(Instrumentation instrumentation, URL bootstrapURL) {
...
 
+    private static FriendlyException getFriendlyException(Throwable t) {
+        if (t instanceof FriendlyException) {
+            return (FriendlyException) t;
+        }
+        Throwable cause = t.getCause();
+        if (cause == null) {
+            return null;
+        }
+        return getFriendlyException(cause);
```
heyams, line 114: what is the point of this recursion here?

core/src/main/java/com/microsoft/applicationinsights/customExceptions/FriendlyException.java
```diff
@@ -0,0 +1,38 @@
+package com.microsoft.applicationinsights.customExceptions;
+
+public class FriendlyException extends Exception {
```
heyams, line 3: can we try to remove this duplicate class?

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/customExceptions/FriendlyException.java
```diff
@@ -0,0 +1,40 @@
...
+        messageBuilder.append(banner).append(System.lineSeparator());
+        messageBuilder.append("*************************").append(System.lineSeparator());
+        messageBuilder.append(System.lineSeparator());
+        messageBuilder.append("Description:").append(System.lineSeparator());
+        messageBuilder.append(description).append(System.lineSeparator());
+        messageBuilder.append(System.lineSeparator());
+        messageBuilder.append("Action:").append(System.lineSeparator());
+        messageBuilder.append(action).append(System.lineSeparator());
+        if (!note.isEmpty()) {
+            messageBuilder.append(action).append(System.lineSeparator());
```
heyams, line 34: look like this line is the only difference?

heyams: line 32 takes care of it.

heyams (commented):
Can you add some unit tests?

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
kryalama: In the MainEntryPoint class there are different kind of exceptions thrown and FriendlyException is chained as part of these exceptions. We have to look for Exception where the cause if FriendlyException.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/customExceptions/FriendlyException.java
kryalama: To make the message more clear I added one more line

core/src/main/java/com/microsoft/applicationinsights/customExceptions/FriendlyException.java
kryalama: As of now we duplicated this class because we are not able to import core packages to agent-bootstrap. Added a note in one of the classes that the class is duplicated.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
heyams: my concern is will this lead to an infinite loop when throws is not FriendlyException?

kryalama: If there is no FriendlyException, the cause will eventually be null and the function returns null. Will try to add a unit test for this case.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
```diff
@@ -181,9 +196,12 @@ public void validate() {
     public static class ToAttributeConfig {
         public List<String> rules;
 
-        public void validate() {
+        public void validate() throws FriendlyException {
             if(rules==null || rules.isEmpty()) {
-                throw new ConfigurationException("User provided config has invalid toAttribute value with no rules!!!");
+                throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                "User provided config has invalid toAttribute value with no rules!!!",
+                                "Please provide atleast one rule under the toAttribute section of processor configuration. " +
```
trask, line 203:
```suggestion
                                "Please provide at least one rule under the toAttribute section of processor configuration. " +
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/processors/ExporterWithAttributeProcessorTest.java
```diff
@@ -443,7 +443,7 @@ public void complexActionTest() {
     }
 
     @Test
-    public void simpleIncludeTest() {
+    public void simpleIncludeTest() throws FriendlyException {
```
trask, line 446: let's change `FriendlyException` to extend `RuntimeException` so we don't need to clutter up with all the throws clauses

core/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -122,7 +122,8 @@ public ProfileFetcherResult fetchApplicationId(String instrumentationKey, Teleme
             if (cause != null && cause instanceof SSLHandshakeException) {
                 throw new FriendlyException("ApplicationInsights Java Agent failed to connect to CdsProfile end point.",
                         "Unable to find valid certification path to requested target.",
-                        "Please import the SSL certificate from "+ Defaults.PROFILER_ENDPOINT +", into the java key store.",
+                        "Please import the SSL certificate from "+ Defaults.PROFILER_ENDPOINT +", into the java key store. "+
+                        "Learn more about importing the certificate here: https://go.microsoft.com/fwlink/?linkid=2151450",
```
trask, line 126: Can we link directly to the anchor: https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-troubleshoot#ssl-certificate-issues?

core/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
```diff
@@ -122,7 +122,8 @@ public ProfileFetcherResult fetchApplicationId(String instrumentationKey, Teleme
             if (cause != null && cause instanceof SSLHandshakeException) {
                 throw new FriendlyException("ApplicationInsights Java Agent failed to connect to CdsProfile end point.",
                         "Unable to find valid certification path to requested target.",
-                        "Please import the SSL certificate from "+ Defaults.PROFILER_ENDPOINT +", into the java key store.",
+                        "Please import the SSL certificate from "+ Defaults.PROFILER_ENDPOINT +", into the java key store. "+
```
trask, line 125: I don't think CdsProfileFetcher reads from PROFILER_ENDPOINT

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
```diff
@@ -122,17 +122,23 @@
...
         public NameConfig name; // specific for processor types "log" and "span"
 
-        private static void isValidRegex(String value) {
+        private static void isValidRegex(String value) throws FriendlyException {
             try {
                 Pattern.compile(value);
             } catch (PatternSyntaxException exception) {
-                throw new ConfigurationException("User provided processor config do not have valid regex:" + value);
+                throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
+                                "User provided processor configuration do not have valid regex:" + value,
```
trask, line 130: I don't think "User provided" is needed, how about: (and applies below too)
```suggestion
                                "Telemetry processor configuration does not have valid regex:" + value,
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/Configuration.java
```diff
@@ -269,17 +302,26 @@ private void validateSpanProcessorIncludeExclude() {
         public String value;
         public String fromAttribute;
 
-        public void validate() {
+        public void validate() throws FriendlyException {
 
                 if (this.key == null || this.key.isEmpty()) {
-                    throw new ConfigurationException("User provided config has invalid action with empty key!!!");
+                    throw new FriendlyException("ApplicationInsights Java Agent failed to start.",
```
trask, line 308: just a thought, instead of passing "ApplicationInsights Java Agent failed to start" everywhere, which requires knowledge of how this code is used, it may be nicer to only provide that part of the message when performing the log (e.g. from the catch block where we know that we are failing the agent to start)

core/src/main/java/com/microsoft/applicationinsights/web/internal/correlation/CdsProfileFetcher.java
kryalama: updated the url to point to ssl certificate issues section

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/MainEntryPoint.java
kryalama: added unit test for getFriendlyException method

trask (approved):
Good stuff 👍

PR #1400, opened 2020-12-02
fix for issue where null pointer exception is thrown when attribute value is not provided and matchType is regexp

kryalama:
fix for issue where null pointer exception is thrown when attribute value is not provided and matchType is regexp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 3.0.1 GA
+* fix for an issue where null pointer exception is being thrown when attribute value is not provided and matchType is regexp
+
 # Version 3.0.0 GA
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+# Version 3.0.1 GA
+* fix for an issue where null pointer exception is being thrown when attribute value is not provided and matchType is regexp
```
trask, line 4:
```suggestion
* Telemetry processor config throws null pointer exception when attribute value is not provided and matchType is regexp
```

trask (commented):
👍

heyams (approved):

trask (approved):

PR #1408, opened 2020-12-09
Fix apache http request not captured through Azure Function

heyams:
Fixed [#8695413](https://msazure.visualstudio.com/One/_boards/board/t/TelReachSDKTeam/Backlog%20items/?workitem=8695413)

Azure function injects apache http client class into class loader, and our agent uses the cached version instead of the class from the class loader. Use a system property to disable ClassLoaderOptimization for Azure Function.  This change will fix all instrumentation for Azure Function.

Java worker will need to add this argument to the jvm command: 
-DClassLoaderOptimizationOptOut=true

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
@@ -137,7 +137,7 @@ public abstract class InstrumentationModule {
     InstrumentationContextProvider contextProvider = getContextProvider();
 
     AgentBuilder agentBuilder = parentAgentBuilder;
-    log.debug("##################### classLoaderOptimizationOptOutFlag: {}", classLoaderOptimizationOptOutFlag);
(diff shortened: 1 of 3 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -50,6 +50,8 @@
   private static final Logger log = LoggerFactory.getLogger(InstrumentationModule.class);
   private static final Logger muzzleLog = LoggerFactory.getLogger("muzzleMatcher");
 
+  private static boolean classLoaderOptimizationOptOutFlag = Boolean.getBoolean("ClassLoaderOptimizationOptOut");
```
trask, line 53: it's traditional for Java system properties to start with lowercase
```suggestion
  private static boolean classLoaderOptimizationOptOutFlag = Boolean.getBoolean("classLoaderOptimizationOptOut");
```

trask (approved):
👍

PR #1410, opened 2020-12-12
Update to otel-java-instrumentation 0.12.0

trask:
This also brings interop with 2.x SDK:

* ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry().getProperties().put("key1", "val1")
* ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry().setName("new name")

TODO after merging need to update CI with new smoke test


gradle.properties
```diff
@@ -1,3 +1,3 @@
 // Project properties
-version=3.0.0
+version=3.0.1-BETA
```
heyams, line 2: We used to call Preview for older agent versions before GA, and then increment.  what would be the version name for next preview version? 3.0.2-BETA?

trask: 3.0.1-BETA.2

heyams (approved):

PR #1416, opened 2020-12-18
Fix logging configuration

### LINKED ISSUE #1415: Instrumentation logging level ignored
### Expected behavior
When set the following in applicationinsights.json, I expect to turn off auto log collection (JUL in my case):
```
{
  "instrumentation": {
    "logging": {
      "level": "OFF"
    }
  }
}
```

### Actual behavior
I am still seeing log traces appear in azure monitor

### To Reproduce
Set config as described above and check if trace logs are being sent

trask:
The code was using `level` in one place and `threshold` in another.

`level` is correct (and what's in documentation).

This PR also improves the tests so that the would have caught this issue.

Resolves #1415

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
@@ -243,15 +243,15 @@ public class BeforeAgentInstaller {
         if (logging == null) {
             return defaultValue;
         }
-        Object thresholdObj = logging.get("level");
-        if (thresholdObj == null) {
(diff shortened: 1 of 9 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -243,12 +243,12 @@ private static String getLoggingFrameworksThreshold(Configuration config, String
         if (logging == null) {
             return defaultValue;
         }
-        Object thresholdObj = logging.get("threshold");
+        Object thresholdObj = logging.get("level");
```
heyams, line 246: levelObj?

heyams (approved):

PR #1438, opened 2021-01-12
Remove duplicate configuration setting for suppressing logging

trask:
Simpler when there is just one way to configure a particular behavior.

In this case, we can already suppress log capture using APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL=OFF (and similar in json).


heyams (commented):
what about APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE or APPLICATIONINSIGHTS_ROLE_NAME, or connection string? users can set them through both env var or json config.. this is inconsistent with the rest of the configs though.

trask:
> what about APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE or APPLICATIONINSIGHTS_ROLE_NAME, or connection string? users can set them through both env var or json config.. this is inconsistent with the rest of the configs though.

yeah, this isabout being having two different settings:

```
{
  "instrumentation": {
    "logging": {
      "enabled": false,
      "level": "OFF"
    }
  }
}
```

we don't need `"enabled": false` since you can accomplish the same thing with `"level": "OFF"`

heyams (approved):
👍

PR #1440, opened 2021-01-15
Fix App Services logback parser failure 

heyams:
Fix [9130498](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/AEP/Cobalt/2101?workitem=9130498).


trask (approved):
👍

PR #1473, opened 2021-02-05
Log invalid json attributes as WARN only

kryalama:


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -279,42 +279,79 @@ private void log(Logger logger) {
...
+                throw new FriendlyException(getJsonEncodingExceptionMessage(configPath.toAbsolutePath().toString(), ex.getMessage()),
                         "Learn more about configuration options here: https://go.microsoft.com/fwlink/?linkid=2153358");
-
             } catch(Exception e) {
                 throw new ConfigurationException("Error parsing configuration file: " + configPath.toAbsolutePath().toString(), e);
             }
         }
     }
+
+    static String getJsonEncodingExceptionMessage(String configPath, String message) {
```
kryalama, line 311: I added this to make the error messages more clear to the customers. Moshi's exception messages are not very relevant to our usecases.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -279,42 +279,79 @@ private void log(Logger logger) {
...
-                }
-                configuration.configPath = configPath;
-                configuration.lastModifiedTime = lastModifiedTime;
                 return configuration;
             } catch(JsonDataException ex) {
-                throw new FriendlyException("Application Insights Java agent's configuration file "+configPath.toAbsolutePath().toString()+" has the following issue:\n"+ex.getMessage(),
-                        "Learn more about configuration options here: https://go.microsoft.com/fwlink/?linkid=2153358");
-
+                Logger logger = LoggerFactory.getLogger(ConfigurationBuilder.class);
+                logger.warn(ex.getMessage());
```
trask, line 294: I think this should be inside the "if strict" condition below

Also, good to make Logger `private static final` at top of class

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -279,42 +279,79 @@ private void log(Logger logger) {
...
-
             } catch(Exception e) {
                 throw new ConfigurationException("Error parsing configuration file: " + configPath.toAbsolutePath().toString(), e);
             }
         }
     }
+
+    static String getJsonEncodingExceptionMessage(String configPath, String message) {
+        String DEFAULT_MESSAGE = "Application Insights Java agent's configuration file "+ configPath +
+                " has a malformed JSON\n";
```
trask, line 313: can move this up to `private static final`

trask (commented):
nice tests 👍

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/configuration/ConfigurationBuilder.java
kryalama: I think its not worth it as the DEFAULT_MESSAGE is built on string 'configPath'

trask: oh good point, I missed that! typically UPPER_CASE names are reserved for java constants

kryalama: Ya I changed it to lower case :)

trask (approved):

trask:
@kryalama I was wrong about making Logger static! see the update I pushed to the PR.

also, I think we should parse the warning message a bit to remove the "Cannot skip" part below, the rest seems good:

```
2021-02-06 15:20:13.722-08 WARN  c.m.a.a.b.c.ConfigurationBuilder - Cannot skip unexpected NAME at $.httpProxy
```

trask (approved):

PR #1477, opened 2021-02-07
Fix sampling rate recorded for dependencies

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigPolling.java
```diff
@@ -46,22 +43,22 @@
...
-        this.path = path;
-        this.lastModifiedTime = lastModifiedTime;
-        this.lastReadSamplingPercentage = lastReadSamplingPercentage;
-    }
-
     // passing in lastReadSamplingPercentage instead of using the real samplingPercentage, because the real
     // samplingPercentage is rounded to nearest 100/N, and we want to know specifically when the underlying config value changes
     // which is lastReadSamplingPercentage
-    public static void pollJsonConfigEveryMinute(Path path, long lastModifiedTime, double lastReadSamplingPercentage) {
+    public static void pollJsonConfigEveryMinute(Path path, long lastModifiedTime, Configuration lastReadConfig) {
```
heyams, line 52: is it necessary to expose whole config? i don't find anything else is used except the sampling rate.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/JsonConfigPolling.java
```diff
@@ -81,22 +78,23 @@ public void run() {
...
                 lastModifiedTime = fileTime.toMillis();
-                Configuration configuration = ConfigurationBuilder.loadJsonConfigFile(path);
+                Configuration config = ConfigurationBuilder.loadJsonConfigFile(path);
+
                 // important to overlay env vars here, so that we don't overwrite the value set by env var
-                ConfigurationBuilder.overlayEnvVars(configuration);
+                ConfigurationBuilder.overlayEnvVars(config);
 
-                if (!configuration.connectionString.equals(TelemetryConfiguration.getActive().getConnectionString())) {
+                if (!config.connectionString.equals(lastReadConfig.connectionString)) {
```
heyams, line 86: i see why you need config object instead of sampling rate here. what is wrong with the old way retrieving the connection string?

trask: I don't think any functional difference, but this way the comparison against prior value is the same for across sampling percentage and connection string (and any future fields)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/ConnectionStringAccessor.java
```diff
@@ -48,7 +47,7 @@ public void setValue(String value) {
             // now that we know the user has opted in to tracing, we need to init the propagator and sampler
             DelegatingPropagator.getInstance().setUpStandardDelegate();
             // TODO handle APPLICATIONINSIGHTS_SAMPLING_PERCENTAGE
-            DelegatingSampler.getInstance().setAlwaysOnDelegate();
+            DelegatingSampler.getInstance().setDelegate(100);
```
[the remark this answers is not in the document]
trask: for Azure Functions consumption users who don't opt in to monitoring, it's more efficient if we use `Sampler.alwaysOff()` (it will return noop spans, etc which will be more efficient than full spans, etc)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/Global.java
```diff
@@ -8,23 +8,13 @@
...
 
-    private static volatile double samplingPercentage = 100;
-
     // this can be null if agent failed during startup
     @Nullable
     public static TelemetryClient getTelemetryClient() {
         return telemetryClient;
     }
 
-    public static double getSamplingPercentage() {
```
heyams, line 19: I still think sampling rate inside TraceState is a bit awkward since you will need to strip it out before sending out the payload.  Can we keep it in Global? AiSampler can access Global, but then you may need to setDelegate inside AiSampler again.  Can we chat offline about this?

trask:
Closing, will open a fresh PR for this

PR #1478, opened 2021-02-08
Fix disabling micrometer

### LINKED ISSUE #1445: How to disable Micrometer metrics (with SpringBoot Actuator) and JMX auto-collection
Hi,
in SB Actuator, there are 2 ways to expose micrometer metrics per docs - Web (HTTP) or JMX.
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-endpoints

In docs for Codeless monitoring, it is said, that Micrometer and JMX Metrics are auto-collected and in configuration, you can disable collection of micrometer metrics.
https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config

I did disable collection of micrometer metrics, however they still seem to be collected (version 3.0). 
```
"micrometer": {
    "enabled": false
}
```
I suspect it is because Actuator by default exposes them via JMX too (in my config, I expose them via Web and JMX).

If that is true, my question is - is it possible to disable auto-collection of micrometer metrics exposed via JMX? Or possibly disable all auto-collection of JMX metrics? And maybe this distinction of exposure of actuator metrics could be mentioned in docs?

I am going to disable exposure of micrometer metrics by JMX in actuator config, I just wanted to know if it was possible to disable auto-collection of JMX metrics.

Thank you.

trask:
Resolves #1445

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
@@ -65,6 +65,7 @@ import com.microsoft.applicationinsights.internal.util.PropertyHelper;
 import com.microsoft.applicationinsights.web.internal.correlation.CdsProfileFetcher;
 import io.opentelemetry.instrumentation.api.aiconnectionstring.AiConnectionString;
 import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.instrumentation.api.config.ConfigBuilder;
 import org.apache.http.HttpHost;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


kryalama:
@trask Is it possible to add some tests later?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
```diff
@@ -125,6 +126,7 @@ private static void start(Instrumentation instrumentation) throws Exception {
         // TODO need some kind of test for these configuration properties
         if (!isInstrumentationEnabled(config, "micrometer")) {
             properties.put("otel.instrumentation.micrometer.enabled", "false");
```
heyams, line 128: look like micrometer is disabled already... this is for disabling actuator-metrics.  is actuator-metrics part of micrometer?

kryalama (approved):

heyams (approved):

trask:
> @trask Is it possible to add some tests later?

👍 we have todo marker:

// TODO need some kind of test for these configuration properties

and a work item for that TODO: https://msazure.visualstudio.com/One/_workitems/edit/9022308

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java
trask: they are both for capturing micrometer metrics, but in different ways:

"micrometer" instrumentation captures micrometer metrics captured via the global micrometer api

"actuator-metrics" instrumentation captures micrometer metrics captured via spring boot actuator metrics

PR #1482, opened 2021-02-09
Collect perf counters using Oshi 

heyams:
will convert the rest on a separate PR.


core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,83 @@
...
+import oshi.hardware.HWDiskStore;
+import oshi.hardware.HardwareAbstractionLayer;
+import oshi.software.os.OSProcess;
+import oshi.software.os.OperatingSystem;
+
+public class OshiPerformanceCounter implements PerformanceCounter {
+
+    private static final Logger logger = LoggerFactory.getLogger(OshiPerformanceCounter.class);
+    private static final String ID = Constants.PERFORMANCE_COUNTER_PREFIX + "OshiPerformanceCounter";
+    private final static double NANOS_IN_SECOND = 1000000000.0;
```
trask, line 38:
```suggestion
    private final static double NANOS_IN_SECOND = 1_000_000_000;
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,83 @@
...
+        return ID;
+    }
+
+    @Override public void report(TelemetryClient telemetryClient) {
+        long currentCollectionInNanos = System.nanoTime();
+
+        SystemInfo systemInfo = new SystemInfo();
+        HardwareAbstractionLayer hal = systemInfo.getHardware();
+        OperatingSystem osInfo = systemInfo.getOperatingSystem();
+        OSProcess processInfo = osInfo.getProcess(osInfo.getProcessId());
```
trask, line 52: I think this code can be run once during constructor, and re-use same `processInfo` in this method (calling `updateAttributes` first each time the method is run)

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // system.disk.io
+        for (HWDiskStore diskStore : hal.getDiskStores()) {
+            currentProcessIO += (double) (diskStore.getReadBytes() + diskStore.getWriteBytes());
+        }
+
+        if (lastCollectionInNanos != -1) {
+            double timeElapsedInSeconds = ((double)(currentCollectionInNanos - lastCollectionInNanos)) / NANOS_IN_SECOND;
+            double value = (currentProcessIO - prevProcessIO) / timeElapsedInSeconds;
+            prevProcessIO = currentProcessIO;
```
trask, line 64: this happens below also
```suggestion
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,83 @@
...
+            send(telemetryClient, value, Constants.PROCESS_IO_PC_METRIC_NAME);
+            logger.trace("Sent performance counter for '{}': '{}'", Constants.PROCESS_IO_PC_METRIC_NAME, value);
+        }
+
+        prevProcessIO = currentProcessIO;
+        lastCollectionInNanos = currentCollectionInNanos;
+
+        // runtime.java.cpu_time
+        processInfo.updateAttributes();
+        long totalProcessorTime = (long) ((processInfo.getUserTime() + processInfo.getKernelTime()) * 0.001);
```
trask, line 74: can you add comment here about why multiplying by 0.001?

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+import oshi.software.os.OperatingSystem;
+
+public class OshiPerformanceCounter implements PerformanceCounter {
+
+    private static final Logger logger = LoggerFactory.getLogger(OshiPerformanceCounter.class);
+    private static final String ID = Constants.PERFORMANCE_COUNTER_PREFIX + "OshiPerformanceCounter";
+    private final static double NANOS_IN_SECOND = 1_000_000_000;
+    private long lastCollectionInNanos = -1;
+    private double prevProcessIO;
+    private HardwareAbstractionLayer hal;
```
trask, line 41: doesn't look like need a field for this, can be local var
```suggestion
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+        processInfo = osInfo.getProcess(osInfo.getProcessId());
+    }
+
+    @Override public String getId() {
+        return ID;
+    }
+
+    @Override public void report(TelemetryClient telemetryClient) {
+        long currentCollectionInNanos = System.nanoTime();
+        double currentProcessIO = 0.0;
```
trask, line 57:
```suggestion
        double currentProcessIO = 0;
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // process io
+        currentProcessIO += (double) (processInfo.getBytesRead() + processInfo.getBytesWritten());
+
+        if (lastCollectionInNanos != -1) {
+            double timeElapsedInSeconds = ((double)(currentCollectionInNanos - lastCollectionInNanos)) / NANOS_IN_SECOND;
```
trask, line 63: this cast isn't needed since NANOS_IN_SECONDS is a double
```suggestion
            double timeElapsedInSeconds = (currentCollectionInNanos - lastCollectionInNanos) / NANOS_IN_SECOND;
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override public void report(TelemetryClient telemetryClient) {
+        long currentCollectionInNanos = System.nanoTime();
+        double currentProcessIO = 0.0;
+
+        // process io
+        currentProcessIO += (double) (processInfo.getBytesRead() + processInfo.getBytesWritten());
+
+        if (lastCollectionInNanos != -1) {
```
trask, line 62: weirdly, System.nanoTime() can return negative values, so would probably be better to track the first time in a different way (maybe just a new field), just in the slim slim case that it returns -1

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+            double timeElapsedInSeconds = ((double)(currentCollectionInNanos - lastCollectionInNanos)) / NANOS_IN_SECOND;
+            double value = (currentProcessIO - prevProcessIO) / timeElapsedInSeconds;
+            send(telemetryClient, value, Constants.PROCESS_IO_PC_METRIC_NAME);
+            logger.trace("Sent performance counter for '{}': '{}'", Constants.PROCESS_IO_PC_METRIC_NAME, value);
+        }
+
+        prevProcessIO = currentProcessIO;
+        lastCollectionInNanos = currentCollectionInNanos;
```
trask, line 70: for consistency (or both lastX)
```suggestion
        prevProcessIO = currentProcessIO;
        prevCollectionInNanos = currentCollectionInNanos;
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // runtime.java.cpu_time
```
trask, line 72:
```suggestion
```

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -0,0 +1,84 @@
...
+        processInfo.updateAttributes();
```
trask, line 73: need to call `updateAttribute` at the top of this method. I think that may be why the values were 1 min off

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
heyams: this is required for disk.io.  now it's no longer needed. will delete.

core/src/main/java/com/microsoft/applicationinsights/internal/perfcounter/OshiPerformanceCounter.java
heyams: System.nanoTime() won't return negative values.. -1 is the initialization to make sure it skips this if block for the very 1st time

trask: check out https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--

trask (commented):
👍 can you attach SNAPSHOT agent to this PR, so we have a copy with side-by-side reporting for future comparison if needed, and then go ahead and remove old counter, and the underlying native code in this PR

heyams:
@trask can't attach jar directly here. You can download it [here](https://dev.azure.com/mseng/AppInsights/_build/results?buildId=14339660&view=artifacts&pathAsName=false&type=publishedArtifacts).

trask (commented):
@heyams a couple of things

* "\\Processor(_Total)\\% Processor Time" doesn't match between the new and old way for me
* Let's capture IO and Processor Total at startup, so that the first reporting of data can report the diff already
* I'm getting this during startup:
  ```
  2021-02-16 11:26:45.277-08 ERROR o.s.o.windows.WindowsOperatingSystem - AdjustTokenPrivileges failed. Error: 1300
  ```

trask (approved):

PR #1486, opened 2021-02-10
Removing unnecessary dependencies (otel-fork)

trask:


instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
```diff
@@ -25,17 +25,22 @@ dependencies {
   // in public API.
   library group: 'com.amazonaws', name: 'aws-lambda-java-events', version: '2.2.1'
 
-  implementation(
+  compileOnly(
     'com.fasterxml.jackson.core:jackson-databind:2.10.5.1',
     'com.fasterxml.jackson.module:jackson-module-afterburner:2.9.10',
     'commons-io:commons-io:2.2')
-  implementation deps.slf4j
+  compileOnly deps.slf4j
```
heyams, line 32: i googled the diff between compile vs. compileOnly.  Got this 'The difference is that packages your java code use from a compileOnly dependency will not be listed as Import-Package manifest entries.". is this the only diff? Just curious.

heyams (approved):

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
trask: by making it `compileOnly`, gradle (and maven) won't treat it as a transitive dependency when you use the (containing) artifact

PR #1494, opened 2021-02-16
Improved operation name logging for Netty based applications

neinoi:
Fix Operation name logging for Netty based applications

No discussion found on this subject, nor issue.

### DIFF
--- a/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
+++ b/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
@@ -46,7 +46,7 @@ public class HttpServerRequestTracingHandler extends SimpleChannelUpstreamHandle
     HttpRequest request = (HttpRequest) message;
 
     Context context =
-        tracer().startSpan(request, ctx.getChannel(), channelTraceContext, "netty.request");
+        tracer().startSpan(request, ctx.getChannel(), channelTraceContext, request.method() + " " + request.uri());
     try (Scope ignored = context.makeCurrent()) {
       ctx.sendUpstream(event);
       // the span is ended normally in HttpServerResponseTracingHandler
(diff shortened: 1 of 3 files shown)


instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
```diff
@@ -46,7 +46,7 @@ public void messageReceived(ChannelHandlerContext ctx, MessageEvent event) {
     HttpRequest request = (HttpRequest) message;
 
     Context context =
-        tracer().startSpan(request, ctx.getChannel(), channelTraceContext, "netty.request");
+        tracer().startSpan(request, ctx.getChannel(), channelTraceContext, request.getMethod() + " " + request.getUri());
```
trask, line 49: we need to follow OpenTelemetry semantic conventions (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name)

```suggestion
        tracer().startSpan(request, ctx.getChannel(), channelTraceContext, "HTTP " + request.getMethod());
```

if you want a more specific operation name including the URI, check out https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-upgrade-from-2x#operation-names

trask (commented):
Hi @neinoi! Can you submit this PR upstream to https://github.com/open-telemetry/opentelemetry-java-instrumentation? Then we will pull it in to our fork after the next upstream release.

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
neinoi: Hi @trask , I need to have more details in my insights. With my proposition I get something like this in the operation name : 
GET /api/v3/tracks

Before that, all calls were grouped by « netty.request » name.
With your proposition, I’ll have something like that : 
HTTP GET
HTTP POST

(For the upstream PR, no pb)

trask:
closing, this was merged upstream https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2317, thx @neinoi!

PR #1495, opened 2021-02-16
Add env vars for disabling instrumentation

### LINKED ISSUE #1488: Possibility to disable spring-schedule tracing
Hi

I have an azure function written in Java which is using `spring-cloud-function-adapter-azure`, part of my code containing method with `@Scheduled` annotation with `fixedDelay` set to 10 seconds. I enabled application insights agent by setting environment variable `XDT_MicrosoftApplicationInsights_Java=1` in function app configuration. Logic under annotated method doesn't give me any valuable information so I would like to exclude all spring-scheduling request from tracing in application insights. Currently after each 10 seconds I'm getting information about execution, it's making mess in logs and price of usage is rising.

image
 

Disable spring-scheduling requests in codeless way (e.g by using environment variable, configuration file etc)

I went through open telemetry [documentation](https://opentelemetry.io/docs/java/automatic_instrumentation/#specific-libraries) and application agent [source code](https://github.com/microsoft/ApplicationInsights-Java/blob/master/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/BeforeAgentInstaller.java#L99) and thought I can solve my issue by setting `JAVA_OPTS` in application insights to `-Dotel.instrumentation.spring-scheduling.enabled=false` or `-Dotel.integration.spring-scheduling.enabled=false` but it didn't help.

trask:
Also adds spring-scheduling and jms to list of instrumentation that can be disabled, resolves #1488.

Also adds smoke tests for configurations with disabled instrumentation.

### DIFF
--- a/test/smoke/testApps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
+++ b/test/smoke/testApps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
@@ -1,0 +1,19 @@
+package com.microsoft.applicationinsights.smoketest;
+
+import org.junit.*;
+
+import static org.junit.Assert.*;
+
+@UseAgent("disabled_micrometer")
+public class MicrometerDisabledTest extends AiSmokeTest {
+
+    @Test
+    @TargetUri("/test")
+    public void doMostBasicTest() throws Exception {
+        mockedIngestion.waitForItems("RequestData", 1);
+
+        // sleep a bit and make sure no micrometer metrics are reported
+        Thread.sleep(10000);
+        assertEquals(0, mockedIngestion.getCountForType("MetricData"));
+    }
+}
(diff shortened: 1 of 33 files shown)


test/smoke/testApps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
```diff
@@ -0,0 +1,19 @@
...
+@UseAgent("disabled_micrometer")
+public class MicrometerDisabledTest extends AiSmokeTest {
+
+    @Test
+    @TargetUri("/test")
+    public void doMostBasicTest() throws Exception {
+        mockedIngestion.waitForItems("RequestData", 1);
+
+        // sleep a bit and make sure no micrometer metrics are reported
+        Thread.sleep(10000);
```
heyams, line 16: why here needs to wait 2.x longer? it's 5000 elsewhere.

trask: metrics are only reported every 5 seconds, and then batching can take another 5 seconds

heyams (approved):

PR #1496, opened 2021-02-16
Run smoke tests using GitHub Actions

trask:


heyams (approved):

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,55 @@
+name: PR build
+
+on: pull_request
```
trask, line 3: TODO on merge also

PR #1498, opened 2021-02-18
Always capture operation_Id on trace/exception

trask:
Previously, we didn't capture operation_Id on trace/exception records when there was no parent, because there was nothing that it needed to be correlated with.

But it's a bit weird to explain why some trace/exception records don't have an operation_Id.

Opening this as a draft pending discussion with others.


trask:
decided against this, as we are move towards OpenTelemetry where operationId = traceId and there is no trace id for standalone logs in OpenTelemetry

PR #1510, opened 2021-02-23
Roles Support in QuickPulse Java SDK

rambhatt-msft:
Fix # .

- [ ] CHANGELOG.md updated


core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -48,28 +49,35 @@
...
+        }
+
         sb.append("[{");
         formatDocuments(sb);
         sb.append("\"Instance\": \"").append(instanceName).append("\",");
         sb.append("\"InstrumentationKey\": \"").append(ikey).append("\",");
         sb.append("\"InvariantVersion\": 1,");
-        sb.append("\"MachineName\": \"").append(instanceName).append("\",");
+        sb.append("\"MachineName\": \"").append(machineName).append("\",");
+        sb.append("\"RoleName\": ").append(roleName).append(",");
```
trask, line 80: is it same to remove above conditional:

```
if (!LocalStringsUtils.isNullOrEmpty(roleName)) {
    roleName = "\"" + roleName + "\"";
}
```

and do this (more similar to attributes)?

```suggestion
        sb.append("\"RoleName\": \"").append(roleName).append("\",");
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -86,12 +94,13 @@ private DefaultQuickPulseDataFetcher(ArrayBlockingQueue<HttpPost> sendQueue, Tel
...
     @Override
-    public void prepareQuickPulseDataForSend() {
+    public void prepareQuickPulseDataForSend(String redirectedEndpoint) {
         try {
             QuickPulseDataCollector.FinalCounters counters = QuickPulseDataCollector.INSTANCE.getAndRestart();
 
             final Date currentDate = new Date();
-            final HttpPost request = networkHelper.buildRequest(currentDate, getEndpointUrl());
+            final String endPointURL = LocalStringsUtils.isNullOrEmpty(redirectedEndpoint) ? getQuickPulseEndpoint() : redirectedEndpoint;
+            final HttpPost request = networkHelper.buildRequest(currentDate, this.getEndpointUrl(endPointURL));
```
trask, line 103:
```suggestion
            final String endpoint = LocalStringsUtils.isNullOrEmpty(redirectedEndpoint) ? getQuickPulseEndpoint() : redirectedEndpoint;
            final HttpPost request = networkHelper.buildRequest(currentDate, this.getEndpointUrl(endpoint));
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -114,11 +123,12 @@ public void prepareQuickPulseDataForSend() {
     }
 
     @VisibleForTesting
-    String getEndpointUrl() {
-        return getQuickPulseEndpoint() + "/post?ikey=" + getInstrumentationKey();
+    String getEndpointUrl(String endpointURL) {
+        return endpointURL + "/post?ikey=" + getInstrumentationKey();
```
trask, line 127: or maybe `endpointPrefix`?
```suggestion
    String getEndpointUrl(String endpoint) {
        return endpoint + "/post?ikey=" + getInstrumentationKey();
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataSender.java
```diff
@@ -90,7 +90,7 @@ public void run() {
         } catch (Throwable t) {
             try {
                 stopped = true;
-                quickPulseStatus = QuickPulseStatus.ERROR;
+                quickPulseHeaderInfo = new QuickPulseHeaderInfo(QuickPulseStatus.QP_IS_ON);
```
trask, line 93: is this a change in behavior?

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -49,39 +50,55 @@
...
+        }
 
         pingPrefix = "{" +
                 "\"Documents\": null," +
                 "\"Instance\":\"" + instanceName + "\"," +
                 "\"InstrumentationKey\": null," +
                 "\"InvariantVersion\": 2," +
-                "\"MachineName\":\"" + instanceName + "\"," +
+                "\"MachineName\":\"" + machineName + "\"," +
+                "\"RoleName\":" + roleName + "," +
```
trask, line 78: same comment as above about removing conditional and changing this, primarily to match the other attributes

```suggestion
                "\"RoleName\":\"" + roleName + "\"," +
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -49,39 +50,55 @@
...
+        this(apacheSender, null, machineName, instanceName, roleName, quickPulseId);
     }
 
     @Override
-    public QuickPulseStatus ping() {
+    public QuickPulseHeaderInfo ping(String redirectedEndpoint) {
         final Date currentDate = new Date();
-        final HttpPost request = networkHelper.buildRequest(currentDate, getQuickPulsePingUri());
+        final String endPointURL = LocalStringsUtils.isNullOrEmpty(redirectedEndpoint) ? getQuickPulseEndpoint() : redirectedEndpoint;
+        final HttpPost request = networkHelper.buildPingRequest(currentDate, getQuickPulsePingUri(endPointURL), quickPulseId, machineName, roleName, instanceName);
```
trask, line 100:
```suggestion
        final String endpoint = LocalStringsUtils.isNullOrEmpty(redirectedEndpoint) ? getQuickPulseEndpoint() : redirectedEndpoint;
        final HttpPost request = networkHelper.buildPingRequest(currentDate, getQuickPulsePingUri(endpoint), quickPulseId, machineName, roleName, instanceName);
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -117,16 +134,16 @@ public QuickPulseStatus ping() {
     }
 
     @VisibleForTesting
-    String getQuickPulsePingUri() {
-        return getQuickPulseEndpoint() + "/ping?ikey=" + getInstrumentationKey();
+    String getQuickPulsePingUri(String endpointURL) {
+        return endpointURL + "/ping?ikey=" + getInstrumentationKey();
```
trask, line 138: or `endpointPrefix`?
```suggestion
    String getQuickPulsePingUri(String endpoint) {
        return endpoint + "/ping?ikey=" + getInstrumentationKey();
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseHeaderInfo.java
```diff
@@ -0,0 +1,34 @@
+package com.microsoft.applicationinsights.internal.quickpulse;
+
+public class QuickPulseHeaderInfo {
+
+    private QuickPulseStatus quickPulseStatus;
+    private String QpsServiceEndpointRedirect;
+    private long QpsServicePollingInterval;
```
trask, line 7:
```suggestion
    private final QuickPulseStatus quickPulseStatus;
    private final String qpsServiceEndpointRedirect;
    private final long qpsServicePollingIntervalMillis;
```

trask (commented):
Thanks @rambhatt-msft, this is looking great!

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
rambhatt-msft: This is a great question 😃 The problem here is that, without this conditional, if the value of `roleName` is `null` then it will become `roleName: "null"` in the resulting json, as in a string called "null" which is incorrect, we want it to be `roleName: null` so the service interprets this correctly.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataSender.java
rambhatt-msft: Thank you for catching it! This was a mistake on my part

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
rambhatt-msft: This is a great question 😃 The problem here is that, without this conditional, if the value of `roleName` is `null` then it will become `roleName: "null"` in the resulting json, as in a string called "null" which is incorrect, we want it to be `roleName: null` so the service interprets this correctly.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
```diff
@@ -33,7 +34,27 @@
...
+
+    public HttpPost buildPingRequest(Date currentDate, String address, String quickPulseId, String machineName, String roleName, String instanceName) {
+
+        HttpPost request = buildRequest(currentDate, address);
+
+        request.addHeader(QPS_ROLE_NAME_HEADER, roleName);
+        request.addHeader(QPS_MACHINE_NAME_HEADER, machineName);
+        request.addHeader(QPS_STREAM_ID_HEADER, quickPulseId);
+        request.addHeader(QPS_INSTANCE_NAME_HEADER, instanceName);
+        request.addHeader(QPS_INVARIANT_VERSION_HEADER, "2");
```
tokaplan, line 55: Why 2?

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
```diff
@@ -48,17 +69,31 @@ public boolean isSuccess(HttpResponse response) {
...
-                return QuickPulseStatus.QP_IS_OFF;
+    public QuickPulseHeaderInfo getQuickPulseHeaderInfo(HttpResponse response) {
+        Header[] headers = response.getAllHeaders();
+        QuickPulseStatus status = QuickPulseStatus.ERROR;
+        long servicePollingIntervalHint = -1;
+        String serviceEndpointRedirect = null;
+        final QuickPulseHeaderInfo quickPulseHeaderInfo;
+
+        for (Header header: headers) {
+            if (QPS_STATUS_HEADER.equals(header.getName())) {
```
tokaplan, line 80: Are we good on case sensitivity here?

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
rambhatt-msft: > Why 2?

checkout DefaultQuickPulsePingSender.java : 76, this is to match whats in the body

`"\"InvariantVersion\": 2," +`

tokaplan: Let's bump it - here and in the body. In fact, can we please have it in a centralized location rather than in 2 separate places? We usually bump the invariant version when we add behaviors.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
trask: oh I see, and the others are never null, makes sense 👍

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
trask: 👍 for equalsIgnoreCase

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -75,7 +75,7 @@ private DefaultQuickPulseDataFetcher(ArrayBlockingQueue<HttpPost> sendQueue, Tel
         formatDocuments(sb);
         sb.append("\"Instance\": \"").append(instanceName).append("\",");
         sb.append("\"InstrumentationKey\": \"").append(ikey).append("\",");
-        sb.append("\"InvariantVersion\": 1,");
```
rambhatt-msft, line 71: @trask Do you know why this value and the one below in `DefaultQuickPulsePingSender.java ` are different? They really should not be, as far as I understand.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -73,7 +73,7 @@ public DefaultQuickPulsePingSender(ApacheSender sender, TelemetryConfiguration c
                 "\"Documents\": null," +
                 "\"Instance\":\"" + instanceName + "\"," +
                 "\"InstrumentationKey\": null," +
-                "\"InvariantVersion\": 2," +
```
rambhatt-msft, line 63: @trask and here.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
trask: hey @rambhatt-msft, I'm not sure why, it does seem like a mistake that they are different. I looked at the git history and found this: #904. Is it ok for us to go to v2/v3, even though we don't support filtering and sample telemetry?

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulse.java
```diff
@@ -44,6 +44,7 @@
 public enum QuickPulse implements Stoppable {
     INSTANCE;
 
+    public static final Integer QP_INVARIANT_VERSION = 1;
```
trask, line 47: it's more conventional to make this a primitive
```suggestion
    public static final int QP_INVARIANT_VERSION = 1;
```

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
```diff
@@ -33,7 +34,25 @@
...
+    private static final String QPS_INVARIANT_VERSION_HEADER = "x-ms-qps-invariant-version";
+
+    public HttpPost buildPingRequest(Date currentDate, String address, String quickPulseId, String machineName, String roleName, String instanceName) {
+
+        HttpPost request = buildRequest(currentDate, address);
+        request.addHeader(QPS_ROLE_NAME_HEADER, roleName);
+        request.addHeader(QPS_MACHINE_NAME_HEADER, machineName);
+        request.addHeader(QPS_STREAM_ID_HEADER, quickPulseId);
+        request.addHeader(QPS_INSTANCE_NAME_HEADER, instanceName);
+        request.addHeader(QPS_INVARIANT_VERSION_HEADER, QuickPulse.QP_INVARIANT_VERSION.toString());
```
trask, line 53: if you change above to primitive, then:
```suggestion
        request.addHeader(QPS_INVARIANT_VERSION_HEADER, Integer.toString(QuickPulse.QP_INVARIANT_VERSION));
```

trask (approved):
👍 will wait for your confirmation before merging

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseNetworkHelper.java
rambhatt-msft: thanks for the suggestion

rambhatt-msft:
> 👍 will wait for your confirmation before merging

I am ready with the change. @tokaplan when you feel comfortable with it, we can merge.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
rambhatt-msft: We should keep the version unchanged at 1. Since higher versions, like you said, support filtering and sample telemetry and the version numbers are used to distinguish between them. This change I did now is gonna be in all SDKs so its not a distinguishing feature.

tokaplan (approved):

trask:
devops please run

trask:
Thanks @rambhatt-msft!

rambhatt-msft:
> Thanks @rambhatt-msft!

Thank for approving @trask  😃

PR #1519, opened 2021-02-28
Simplify retrieval of app id

trask:
A couple of things in this PR
* Simplifies code for retrieving app id
* Removes dependency on `httpasyncclient` (retrieving app id now goes over same http client as ingestion)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
```diff
@@ -36,6 +36,7 @@
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
```
heyams, line 38: Look like this import is not needed.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -26,17 +26,19 @@
...
 public interface ApacheSender {
-    final static int DEFAULT_MAX_TOTAL_CONNECTIONS = 200;
-    final static int REQUEST_TIMEOUT_IN_MILLIS = 60000;
-    final static int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 20;
+    int DEFAULT_MAX_TOTAL_CONNECTIONS = 200;
+    int REQUEST_TIMEOUT_IN_MILLIS = 60000;
+    int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 20;
 
-    HttpResponse sendPostRequest(HttpPost post) throws IOException, FriendlyException;
+    HttpResponse sendRequest(HttpUriRequest post) throws IOException, FriendlyException;
```
heyams, line 41: (HttpUriRequest request)?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -164,7 +164,7 @@ public boolean sendSync(Transmission transmission) {
                 // POST the transmission data to the endpoint
                 request = createTransmissionPostRequest(transmission);
                 httpClient.enhanceRequest(request);
-                response = httpClient.sendPostRequest(request);
+                response = httpClient.sendRequest(request);
```
heyams, line 167: I checked HttpPost (HttpEntityEnclosingRequestBase) is not a type of HttpUriRequest, which is HttpRequest, and HttpMessage.  I don't see other changes related to it.. line 165 returns HttpPost? where is the magic?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -26,17 +26,19 @@
...
+import org.apache.http.client.methods.HttpUriRequest;
 
 /**
  * Created by gupele on 6/4/2015.
  */
 public interface ApacheSender {
-    final static int DEFAULT_MAX_TOTAL_CONNECTIONS = 200;
-    final static int REQUEST_TIMEOUT_IN_MILLIS = 60000;
-    final static int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 20;
+    int DEFAULT_MAX_TOTAL_CONNECTIONS = 200;
```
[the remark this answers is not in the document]
trask: all interface fields are static and final, so `final static` was redundant here

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender43.java
trask: all unused imports are cleaned up in #1520

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
trask: HttpPost extends HttpEntityEnclosingRequestBase extends HttpRequestBase implements HttpUriRequest 😄

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
trask: done

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -26,17 +26,19 @@
...
+    int REQUEST_TIMEOUT_IN_MILLIS = 60000;
```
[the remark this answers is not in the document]
trask: same

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/ApacheSender.java
```diff
@@ -26,17 +26,19 @@
...
+    int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 20;
```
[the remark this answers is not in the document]
trask: same

heyams (approved):

heyams (commented):
one last note: should ApacheSender handle the backoff systematically instead of AppIdSupplier handles it individually?  Use the same backoff timer?

trask:
> one last note: should ApacheSender handle the backoff systematically instead of AppIdSupplier handles it individually? Use the same backoff timer?

could be nice in the future, let's see how things look after we move most of the http client stuff to Azure SDK (where we will be using a different http client)

PR #1524, opened 2021-03-02
Re-order Exporter to match Azure SDK Exporter

trask:
This is just re-ordering of code inside of Exporter to make it easier to compare and see differences with https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/monitor/azure-monitor-opentelemetry-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java


heyams (approved):

trask:
cc: @srnagar

PR #1526, opened 2021-03-03
Fix role name on Azure Functions

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -200,7 +200,7 @@ public static void overlayEnvVars(Configuration config) throws IOException {
 
         if (isTrimEmpty(config.role.name)) {
             // only use WEBSITE_SITE_NAME as a fallback
```
heyams, line 202: comment can be deleted too?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationTest.java
```diff
@@ -190,24 +190,35 @@ public void shouldOverrideRoleName() throws IOException {
...
 
+    @Test
+    public void shouldOverrideRoleNameWithLowercaseWebsiteEnvVarOnAzFn() throws IOException {
+        envVars.set("FUNCTIONS_WORKER_RUNTIME", "java");
+        envVars.set("WEBSITE_SITE_NAME", "Role Name From Website Env");
+
+        Configuration configuration = loadConfiguration("applicationinsights_NoRole.json");
+        ConfigurationBuilder.overlayEnvVars(configuration);
+
+        assertEquals("role name from website env", configuration.role.name);
```
heyams, line 219: Look like case doesn't match.

trask: see the name of this test 😄

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
trask: I think the comment is still good, in other cases env var overrides the json config, but in this case it's only a fallback

heyams (approved):

PR #1527, opened 2021-03-03
Intellij inspections

trask:
Ran a couple of Intellij auto-fix inspections:
* Field may be final
* Field can be local


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -72,15 +72,15 @@
     // For future use: re-send a failed transmission back to the dispatcher
     private TransmissionDispatcher transmissionDispatcher;
 
-    private String serverUri;
+    private final String serverUri;
 
-    private volatile boolean stopped;
```
heyams, line 77: this is thread-safe.  did intellij inspection fail in this case?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -72,15 +72,15 @@
...
-    private volatile TelemetryConfiguration configuration;
```
heyams, line 78: same here.

core/src/main/java/com/microsoft/applicationinsights/internal/config/AddTypeXmlElement.java
```diff
@@ -32,7 +32,7 @@
 
     private String type;
 
-    private ArrayList<ParamXmlElement> paramElements = new ArrayList<>();
+    private final ArrayList<ParamXmlElement> paramElements = new ArrayList<>();
```
heyams, line 35: not sure about this.  are you sure this will be an empty list?

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
```diff
@@ -49,13 +49,13 @@
...
-    private String instanceName;
-    private String machineName;
-    private String quickPulseId;
+    private final String pingPrefix;
+    private final String roleName;
+    private final String instanceName;
+    private final String machineName;
+    private final String quickPulseId;
     private long lastValidTransmission = 0;
-    private static volatile AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
```
heyams, line 58: same concern here.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
trask: if it can be changed to local var, then that solves the thread-safety issue 😄

core/src/main/java/com/microsoft/applicationinsights/internal/config/AddTypeXmlElement.java
trask: `final` means you can't re-assign it, but doesn't mean you can't mutate the ArrayList

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
trask: same

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulsePingSender.java
trask: same

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
heyams: volatile or local var?

trask: local vars have no thread-safety problems, since they are not shared across threads

heyams: do we need stopped at all?  i don't see that it becomes true anywhere in the class...  final does not make sense to stopped, it should be modifiable after the initial value.

trask: 👍 good point, removed

heyams (approved):
👍

PR #1528, opened 2021-03-03
Change ApacheSender into a real Apache CloseableHttpClient

trask:
So it can be re-used from profiler.


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
```diff
@@ -27,34 +6,38 @@
...
+import org.apache.http.protocol.HttpContext;
 import org.apache.http.ssl.SSLContexts;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/**
- * Created by gupele on 6/4/2015.
- */
-public final class ApacheSender43 implements ApacheSender {
+public class LazyHttpClient extends CloseableHttpClient {
```
heyams, line 32: final?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
```diff
@@ -64,56 +47,43 @@
...
                 initException = e;
             }
-            return httpClient;
+            return delegate;
         }
     }
 
-    @Override
-    public void enhanceRequest(HttpPost request) {
+    public static void dispose(HttpResponse response) {
```
heyams, line 76: why change this method to be static?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
```diff
@@ -64,56 +47,43 @@
...
 
-    @Override
-    public HttpResponse sendRequest(HttpUriRequest request) throws IOException, FriendlyException {
-        try {
-            return getHttpClient().execute(request);
-        } catch (SSLHandshakeException e) {
-            String completeUrl = "https://" + request.getURI().getHost();
-            throw SSLUtil.newSSLFriendlyException(completeUrl);
-        }
+    public static CloseableHttpClient getInstance() {
```
heyams, line 52: can we return LazyHttpClient? I know LazyHttpClient is an instance of closeableHttpClient, Thought that will make it clear between delegate and instance.

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
```diff
@@ -64,56 +47,43 @@
...
+        try {
+            if (response != null) {
+                ((CloseableHttpResponse) response).close();
+            }
+        } catch (IOException e) {
+            logger.error("Failed to send or failed to close response, exception: {}", e.toString());
+        }
+    }
+
+    public static void enhanceRequest(HttpPost request) {
```
heyams, line 86: does it need to been static?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -138,8 +139,8 @@ public boolean sendSync(Transmission transmission) {
             try {
                 // POST the transmission data to the endpoint
                 request = createTransmissionPostRequest(transmission);
-                httpClient.enhanceRequest(request);
-                response = httpClient.sendRequest(request);
+                LazyHttpClient.enhanceRequest(request);
```
heyams, line 142: httpClient.enhanceRequest if it's not static.. i'm not sure why it got changed to be static. what is the benefit of doing this?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
```diff
@@ -138,8 +139,8 @@ public boolean sendSync(Transmission transmission) {
...
+                response = httpClient.execute(request);
```
heyams, line 143: is this why you make instance to be CloseablehttpClient instead of lazyhttpclient? sendRequest is removed, which is a wrapper method around calling execute?

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
trask: we don't typically mark our classes as final, that's typically important for public APIs to prevent users from subclassing

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
trask: to reflect better that it doesn't rely on any instance state

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
trask: I think this is ok, reflects that callers don't need any special details of LazyHttpClient, only the interface

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
trask: static makes it clear that the method doesn't rely on instance state

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
trask: same as above

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionNetworkOutput.java
trask: Using `CloseableHttpClient` also makes it easy to mock in tests

heyams (approved):

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
heyams: is it a good practice in java to use a static method whenever it doesn't rely on the instance state? it's just a bit confusing to use non-static and static interchangeably.  are you more concerning about performance gain in this case? Static method does occupy less space and memory allocation happens once,  binding occurs at compile time or earlier time instead of run time, and etc?

trask: there's not a perf difference, it's more readability, whether a method is just a "utility" method (input->output) or if it is really part of the object state/behavior, in this case, probably better to move these static methods to some standalone util class, but it's all going to change soon when we move to Azure SDKs

heyams: util class makes sense to me :)

PR #1536, opened 2021-03-05
lazily set website site name

heyams:


trask (commented):
Instead of having separate `AiConnectionString` and `AiWebsiteSiteName` classes (and more in the future), how about a single `LazyConfiguration` class, and then we can move all the logic of reading the env vars into the AI code, and won't need to modify the otel-fork in the future when we add support for other lazy env vars?

trask (approved):

PR #1538, opened 2021-03-05
Populate client IP

### LINKED ISSUE #1535: Agent 3.0.2 track the wrong ClientIP when sending telemetry from server
**Expected behavior**
When unmasking Client IP in Log Analytics (https://docs.microsoft.com/en-us/azure/azure-monitor/app/ip-collection?tabs=net), the Client IP should be the public IP of the end-user.  

**Actual behavior**
When unmasking Client IP in Log Analytics (https://docs.microsoft.com/en-us/azure/azure-monitor/app/ip-collection?tabs=net), the IP appearing is the server outbound IP (and not the end user Public IP).  

**ENVIRONMENT**
Web Server: JBoss Wildfly 17.0.1 Standalone
Web Application: Sample Maven Web Application Hello World

**To Reproduce**
Deploy applicationinsights-agent-3.0.2.jar and the web application on a Wildfly Standalone web server. The agent 3.0.2 send server telemetry to app insight, but the client IP is the outbound ip of the server. There is also no AI tracking cookie attached to the responses, is that expected ?

trask:
Resolves #1535

### DIFF
--- a/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
+++ b/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
@@ -528,6 +528,15 @@ public class Exporter implements SpanExporter {
         requestData.setId(span.getSpanId());
         requestData.getContext().getOperation().setId(span.getTraceId());
 
+        String locationIp = attributes.get(SemanticAttributes.HTTP_CLIENT_IP);
+        if (locationIp == null) {
+            // only use net.peer.ip if http.client_ip is not available
+            locationIp = attributes.get(SemanticAttributes.NET_PEER_IP);
+        }
+        if (locationIp != null) {
+            requestData.getContext().getLocation().setIp(locationIp);
+        }
+
         String aiLegacyParentId = span.getSpanContext().getTraceState().get("ai-legacy-parent-id");
         if (aiLegacyParentId != null) {
             // see behavior specified at https://github.com/microsoft/ApplicationInsights-Java/issues/1174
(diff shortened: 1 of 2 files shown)


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -528,6 +528,15 @@ private void exportRequest(SpanData span) {
         requestData.setId(span.getSpanId());
         requestData.getContext().getOperation().setId(span.getTraceId());
 
+        String locationIp = attributes.get(SemanticAttributes.HTTP_CLIENT_IP);
+        if (locationIp == null) {
```
heyams, line 532: empty check?

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -528,6 +528,15 @@ private void exportRequest(SpanData span) {
...
+            // only use net.peer.ip if http.client_ip is not available
+            locationIp = attributes.get(SemanticAttributes.NET_PEER_IP);
+        }
+        if (locationIp != null) {
```
heyams, line 536: i checked customMetrics table, it has the value of "0.0.0.0".  when it's empty, does Breeze assign this value to client_ip?

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: in this context, null means absent, and "" means (for some strange reason) user intentionally set "", so makes sense(-ish) to pass it on

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: Breeze masks client IP by default, you have to manually override:
https://docs.microsoft.com/en-us/azure/azure-monitor/app/ip-collection?tabs=net#storing-ip-address-data

heyams (approved):

PR #1545, opened 2021-03-10
Update Apache HttpClient version

trask:
Also set timeouts more consistently.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
@@ -15,6 +15,7 @@ import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.config.RegistryBuilder;
+import org.apache.http.config.SocketConfig;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
```diff
@@ -118,12 +109,20 @@ private static CloseableHttpClient init() {
                         .build());
         cm.setMaxTotal(DEFAULT_MAX_TOTAL_CONNECTIONS);
         cm.setDefaultMaxPerRoute(DEFAULT_MAX_CONNECTIONS_PER_ROUTE);
+        cm.setDefaultSocketConfig(SocketConfig.custom()
+                .setSoTimeout(REQUEST_TIMEOUT_IN_MILLIS)
```
heyams, line 113: what is setSoTimeout?

heyams (approved):

core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/LazyHttpClient.java
trask: socket timeout, from reading a bunch of Apache HttpClient jira tickets yesterday (😭), I think it's used for SSL connection/handshake, which occurs outside of "request" scope, so outside of the request-level timeouts we set later in this method

PR #1547, opened 2021-03-11
Update change log for 3.0.3-BETA

trask:


heyams (approved):
Can we add lazy loading env var WEBSITE_SITE_NAME for Azure Functions? or don't need to mention it since it's 1st party?

trask:
> Can we add lazy loading env var WEBSITE_SITE_NAME for Azure Functions? or don't need to mention it since it's 1st party?

added 👍

PR #1550, opened 2021-03-12
Always write to log on startup failure

trask:
Also introduces env var `APPLICATIONINSIGHTS_SELF_DIAGNOSTICS_FILE_PATH`, which can also be helpful in this case for redirecting log file to another location.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/MainEntryPoint.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/MainEntryPoint.java
@@ -21,6 +21,7 @@
 package com.microsoft.applicationinsights.agent.internal.wasbootstrap;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.lang.instrument.Instrumentation;
 import java.net.URL;
 import java.nio.file.Path;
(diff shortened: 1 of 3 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/MainEntryPoint.java
```diff
@@ -134,19 +135,34 @@ private static void logErrorMessage(Logger startupLogger, String message, boolea
...
                 if (isFriendlyException) {
                     startupLogger.error(message);
                 } else {
                     startupLogger.error(message, t);
                 }
-            } catch (Throwable e) {
-                // If the startupLogger still have some issues being initialized, just print the error stack trace
+            } catch (Throwable ignored) {
+                // this is a last resort in cases where the JVM doesn't have write permission to the directory where the agent lives
+                // and the user has not specified APPLICATIONINSIGHTS_SELF_DIAGNOSTICS_FILE_PATH
```
heyams, line 149: file.path is currently default to "applicationinsights.log". for azure function, it doesn't have the write permission in the agent jar folder. can we make azure function or other attaches to default to "/var/log/applicationinsights.log"?

heyams: of course, we can always ask them to provide a value for this env var or declare it in the json config. 
this env var will have a higher precedence over the json config, right? similar to other env vars' precedence.

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationTest.java
```diff
@@ -337,6 +337,16 @@ public void shouldOverrideSelfDiagnosticsLevel() throws IOException {
         assertEquals("DEBUG", configuration.selfDiagnostics.level);
     }
 
+    @Test
+    public void shouldOverrideSelfDiagnosticsFilePath() throws IOException {
+        envVars.set("APPLICATIONINSIGHTS_SELF_DIAGNOSTICS_FILE_PATH", "/tmp/ai.log");
```
heyams, line 342: before overriding, can we assertEquals the value from the applicationinsights.json?

trask: the `shouldParse` test above does this

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/MainEntryPoint.java
trask: this shouldn't happen for users, since they can't control the json file, and so can't mess it up accidentally and cause it to fail to start

also will still log to `/tmp/applicationinsights.log` now which is better than before (just not the more ideal `/var/log/applicationinsights.log`, trying to balance the amount of special cases we add to the code for different compute environments)

heyams (approved):

PR #1551, opened 2021-03-12
Sdk name improvements

trask:
Two sdk name improvements:

* use `f*r_java` when running in Azure Functions on App Services (instead of `awr_java`)
* use `s*r_java` when running in Spring Cloud with `springcloud.codeless` marker file

(and removed the restriction on using json files with attach, since spring cloud uses one)


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/DiagnosticsHelper.java
```diff
@@ -46,47 +44,40 @@ private DiagnosticsHelper() { }
 
     public static void setAgentJarFile(Path agentPath) {
         if (Files.exists(agentPath.resolveSibling("appsvc.codeless"))) {
-            appServiceCodeless = true;
+            appSvcAttachForLoggingPurposes = true;
+            if ("java".equals(System.getenv("FUNCTIONS_WORKER_RUNTIME"))) {
```
heyams, line 48: 👍 for the dedicated plan, azure functions is using app service's prefix. this will fix it :)

heyams (approved):
👍

PR #1554, opened 2021-03-13
Remove unneeded dependency

trask:
Found this when comparing jar file sizes between 3.0.2 and 3.0.3-BETA


heyams (approved):
how much difference does it make? 
🎉

heyams:
Found it. 2.2 MB increase from 3.0.2 to 3.0.3-BETA.

PR #1513, opened 2021-02-25
Initial proposal for providing access to JFR profiles from within Application Insights

johnoliver:
This is a draft pull request to gain feedback, as there are still outstanding issues that need to be addressed:

1. This PR includes a JFR library under `/java-flight-recorder` (Developed by David Grieve) this is currently in the process of being released at https://github.com/microsoft/jfr-streaming when this is done this library will be removed from this commit and will instead rely on the published library.
2. Currently this will compile but fail at run time due to requiring `public void addHandler(Handler handler) throws SecurityException {}` to be added to PatchLogger in otel.

This provides 4 main components:

- java-flight-recorder (to be removed in favour of project https://github.com/microsoft/jfr-streaming)
  - Invokes JFR on demand and obtains jfr profile files.
- Alert library
  - Observes a stream of telemetry data, and if its rolling average breaches a configured threshold issues an alert.
- Agent service profiler. 
  - Coordinates the process of receiving an alert, invoking the profiler and uploading the results.
    - Invokes JFR profiler when an AlertBreach happens.
    - Closes profile when complete.
    - Uploads profile.
  - Monitors for configuration changes and notifies Alert subsystem of alert threshold changes
- ProfilerService
  - Main entry point for the profiler subsystem.
  - Initialises and wires up above services

A TelemetryObserver has been created, the TelemetryClient notifies observers of telemetry data sent to the client. The alerting component uses this mechanism to observe telemetry data.

For significant contributions please make sure you have completed the following items:

- [ ] Design discussion issue #
- [ ] Changes in public surface reviewed
- [ ] CHANGELOG.md updated

PR #1556, opened 2021-03-15
Update Intellij project xml

trask:
This keeps getting changed automatically for me locally, and noticed the same is happening for others in #1513, so makes sense to make the change to the file under source control.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -27,24 +27,6 @@
       </option>
       <option name="JD_ADD_BLANK_AFTER_PARM_COMMENTS" value="true" />
     </JavaCodeStyleSettings>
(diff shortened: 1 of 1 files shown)


heyams:
weird.. it didn't happen to me. can we add it to .gitignore? We ignore lots of *.xml under .idea/ but keep codestyles.

trask:
> weird.. it didn't happen to me. can we add it to .gitignore? We ignore lots of *.xml under .idea/ but keep codestyles.

I'd like to sync with the OTel code style, which doesn't rely on `.idea`, but instead uses `spotless` gradle plugin and `.editorconfig` file. This PR is just a band-aid for now.

heyams (approved):

trask:
closing and re-opening to re-trigger CI

PR #1557, opened 2021-03-15
Use parent-based sampler mostly

trask:
This way non-remote spans (e.g. dependencies) will just re-use the parent decision instead of re-calculating based on the trace id (same end result).


heyams (approved):

trask:
closing and re-opening to re-trigger CI

PR #1558, opened 2021-03-15
More consistent rp integration logging and rp configuration story

trask:
* Unifies app svc "user-logging" with regular "self diagnostics"
* Rename "attach" to "rp integration"
* Removes a couple of env vars that we aren't using
* Sets default self diagnostic file path for Azure Functions on Linux to `"/var/log/applicationinsights/" + DEFAULT_NAME`
* Do not read `applicationinsights.json` from agent directory for rp integrations because that makes it difficult to merge applicationinsights.json from RP and user json from `APPLICATIONINSIGHTS_CONFIGURATION_CONTENT`

PR #1559, opened 2021-03-15
Add applicationinsights-rp.json

trask:
Move configuration needed by Azure Spring Cloud into separate json file, to make it explicit that these are the only configuration options we support for RP, and to make it clear the RP config in this file always takes precedence.

Also so that we can implement not reading `applicationinsights.json` from agent directory for rp integrations in #1558 because reading that makes it difficult to merge applicationinsights.json from RP and user json from `APPLICATIONINSIGHTS_CONFIGURATION_CONTENT`


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -86,6 +83,11 @@ public static Configuration create(Path agentJarPath) throws IOException {
                             " (and note that metricIntervalSeconds applies to all auto-collected metrics," +
                             " not only micrometer)"));
         }
+        overlayEnvVars(config);
+        // rp configuration should always be last (so it takes precedence)
+        if (rpConfiguration != null) {
```
heyams, line 88: can we add a note saying it's only available for spring cloud for now?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -233,6 +235,15 @@ public static void overlayEnvVars(Configuration config) throws IOException {
         loadInstrumentationEnabledEnvVars(config);
     }
 
+    private static void overlayRpConfiguration(Configuration config, RpConfiguration rpConfiguration)  {
+        if (rpConfiguration.connectionString != null) {
```
heyams, line 239: empty check? unless you want to let them disable java agent by setting an empty connection string?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -233,6 +235,15 @@ public static void overlayEnvVars(Configuration config) throws IOException {
...
+            config.connectionString = rpConfiguration.connectionString;
+        }
+        if (rpConfiguration.sampling != null) {
```
heyams, line 242: if (rpconfiguration.sampling != null && config.sampling.percentage != rpConfiguration.sampling.percentage) {
only update when it's different?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/RpConfigurationBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+        Path configPath = agentJarPath.resolveSibling("applicationinsights-rp.json");
+        if (Files.exists(configPath)) {
+            return loadJsonConfigFile(configPath);
+        }
+        return null;
+    }
+
+    public static RpConfiguration loadJsonConfigFile(Path configPath) throws IOException{
+        if (!Files.exists(configPath)) {
+            throw new IllegalStateException("config file does not exist: " + configPath);
```
heyams, line 58: "rp config file"?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/RpConfigurationPolling.java
```diff
@@ -23,79 +23,68 @@
...
 import com.microsoft.applicationinsights.agent.internal.sampling.SamplingPercentage;
+import com.microsoft.applicationinsights.agent.internal.wasbootstrap.configuration.RpConfiguration;
+import com.microsoft.applicationinsights.agent.internal.wasbootstrap.configuration.RpConfigurationBuilder;
 import com.microsoft.applicationinsights.internal.util.ThreadPoolUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
-public class JsonConfigPolling implements Runnable {
```
heyams, line 43: json live polling should support for both applicationinsights.json and applicationinsights-rp.json.
can we not make it specific for RpConfiguration?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
trask: I don't think any reason not to update the value when it's the same here

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/RpConfigurationPolling.java
trask: json live polling is only for rp configuration, it's not a documented or supported end-user feature

heyams: for non-rp scenario, do we want to support live polling of connection string and sampling rate? living polling was requested by Azure SDK (Asir specifically), right? That's not rp integrated.

heyams (approved):

PR #1573, opened 2021-03-23
Respect parent sampled unless override match

trask:
For example, if upstream span decides to sample 100% of `/login` requests, then downstream spans should respect that decision (as long as there is no specific override match for the downstream span), instead of applying the default sampling percentage to the downstream spans.


heyams (approved):
👍

PR #1581, opened 2021-03-24
Disable code signing

heyams:


.pipelines/pipeline.user.windows.yml
```diff
@@ -12,9 +12,10 @@ versioning:
     - !!defaultcommand
       name: 'Generate Version'
       command: '.scripts/version.cmd'
+# excluding our source code from being scanned for code sign validation
 signing_options:
-  profile: '170'
-  codesign_validation_glob_pattern: 'file|**/agent/agent/build/libs/applicationinsights-agent-*.jar'
+  profile: 'none'
+  codesign_validation_glob_pattern: 'f:regex|\.(jar|pom)$'  # recursively find files ending in .jar, .pom
```
trask, line 18: is the pattern still needed for profile `none`?

trask (approved):

.pipelines/pipeline.user.windows.yml
heyams: i have one of the project admins review it as well.. he offered some feedback.. it's updated.

heyams: i will testing it without regex pattern.. once i'm done testing this change.

heyams:
profile: 'none'
codesign_validation_glob_pattern: 'f:regex|\.(jar|pom)$'

@trask this skipped code signing.. not sure about the new change yet.. testing it now.

heyams:
@trask Testing Oyster's feedback and removing pattern entirely.. both worked.  but without pattern is a bit faster.. saving ~7mins.

PR #1587, opened 2021-03-29
Update submodule

### LINKED ISSUE #1563: Spring-webflux-client instrumentation module issue.
I came across an issue related to the logic of _**io.opentelemetry.instrumentation.spring.webflux.client.WebClientTracingFilter**_ class if spring reactive WebClient call is presented in the Reactor upstream chain.
Steps to reproduce:
For example - assume we have to call some external http service 1 but such a call should be preceded by another http service 2 call - in such a case last call resides in the upstream position with regarding to former one.
And parent open telemetry context for http service 1 will contain span under the following key - _**ContextKey.named("opentelemetry-traces-client-span-key")**_ which is considered as a signal to not create new child span. But according to the logic of 
_**io.opentelemetry.instrumentation.spring.webflux.client.WebClientTracingFilter.MonoWebClientTrace**_ publisher - if mentioned above condition will be met - then subscribe method will be left without any subscription - and it is resulting in hanging subscriber chain.
I have found a work around to this problem by turning off spring-webflux-client instrumentation module - otel.instrumentation.spring.webflux.client.enabled=false - and then provides custom module which will be picked up by java Service Loader API. 

```@Override
public void subscribe(CoreSubscriber<? super ClientResponse> subscriber) {
  Context parentContext = Context.current();
  if (!tracer().shouldStartSpan(parentContext)) {
    return;
  }

  ClientRequest.Builder builder = ClientRequest.from(request);
  Context context = tracer().startSpan(parentContext, request, builder);
  try (Scope ignored = context.makeCurrent()) {
    this.next
        .exchange(builder.build())
        .doOnCancel(
            () -> {
              tracer().onCancel(context);
              tracer().end(context);
            })
        .subscribe(new TraceWebClientSubscriber(subscriber, context));
  }
}
```
But anyway it will be very useful to address such an issue. Thanks!

### LINKED ISSUE #1578: Broken e2e transaction when using Spring Cloud Gateway (spring-webflux) with agent 3.0.3-BETA.*
### Expected behavior
Single e2e transaction should be built for single operation incoming into reactive server and going out to downstream services.

### Actual behavior
While using AI agent version 3.0.2 traces from reactive gateway to downstream services are correctly correlated and mapped to initial operation transaction (`tx = [request -> reactive-gateway -> downstream-service -> ...]`), but when using AI agent version 3.0.3-BETA or 3.0.3-BETA.2 e2e transaction is broke, request to reactive gateway is logged separately (different operation IDs, different transactions) from downstream service (`tx1 = [request -> reactive-gateway]`, `tx2 = [request -> downstream-service -> ...]`).
I have not tested reactive clients like `org.springframework.web.reactive.function.client.WebClient` directly.

### To Reproduce
Steps to reproduce the behavior: simple Spring MVC application and Spring Cloud Gateway application, gateway routes to spring mvc application, both applications uses AI javaagent.

#### Sample Application
If applicable, provide a sample application which reproduces the issue.

### System information
 - SDK Version: 3.0.3-BETA.*
 - OS type and version: Docker `openjdk:11.0.9-jre-slim`
 - Application Server type and version (if applicable): Netty 0.9.10.RELEASE
 - Using spring-boot? Yes
 - Additional relevant libraries (with version, if applicable): spring-boot-starter-webflux:2.3.2.RELEASE

### Logs

### Screenshots

Expected behavior:
*Single transaction*
image

Actual behavior:
*Two separate transactions*
image
image

trask:
See #1585

Resolves #1563
Resolves #1578

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/ConfigOverride.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/ConfigOverride.java
@@ -46,6 +46,9 @@ class ConfigOverride {
             properties.put("otel.instrumentation.opentelemetry-api.enabled", "false");
             properties.put("otel.instrumentation.opentelemetry-annotations.enabled", "false");
         }
+        if (!config.preview.azureSdkInstrumentation) {
+            properties.put("otel.instrumentation.azure-sdk.enabled", "false");
+        }
         properties.put("otel.propagators", DelegatingPropagatorProvider.NAME);
         // AI exporter is configured manually
         properties.put("otel.traces.exporter", "none");
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
@@ -167,6 +167,7 @@ public class Configuration {
         public SamplingPreview sampling = new SamplingPreview();
         public List<ProcessorConfig> processors = new ArrayList<>();
         public boolean openTelemetryApiSupport;
+        public boolean azureSdkInstrumentation;
         // applies to perf counters, default custom metrics, jmx metrics, and micrometer metrics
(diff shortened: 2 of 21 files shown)


heyams (approved):

heyams:
looks like azure sdk smoke test failed. @trask

PR #1593, opened 2021-03-30
Suppress redis encoding events

### LINKED ISSUE #1586: A way to suppress redis.encode.start/end custom events
Our applications currently generates a lot of custom `redis.encode.start` and `redis.encode.end` events. Is there a way to suppress these custom events without disabling redis auto-collected telemetry?

We have tried to use new sampling override functionality, but it didn't work.

We have tried rules using `db.statement` and `db.system` attributes:
```json
{
  "attributes": [
	{
	  "key": "db.statement",
	  "value": "INFO",
	  "matchType": "strict"
	},
	{
	  "key": "db.system",
	  "value": "redis",
	  "matchType": "strict"
	}
  ],
  "percentage": 0
}
```

Also we have tried using `ai.operation.name` attribute:
```json
{
  "attributes": [
	{
	  "key": "ai.operation.name",
	  "value": "redis.encode.end",
	  "matchType": "strict"
	}
  ],
  "percentage": 0
}
```

Nothing helped, these events are still being sent to AI.

Also maybe there is a way to get complete list of Span attributes that can be used for sampling overrides (to filter out such dependencies like periodic clustered Quatz job selects, etc.)?

trask:
These events are pretty noisy, and not necessary, so it's reasonable to suppress them. Plus they also don't correspond to any OTel semantic conventions.

Resolves #1586

### DIFF
--- a/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
+++ b/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
@@ -616,6 +616,12 @@ public class Exporter implements SpanExporter {
 
     private void exportEvents(SpanData span, Double samplingPercentage) {
         for (EventData event : span.getEvents()) {
+            boolean lettuce51 =
+                    span.getInstrumentationLibraryInfo().getName().equals("io.opentelemetry.javaagent.lettuce-5.1");
+            if (lettuce51 && event.getName().startsWith("redis.encode.")) {
+                // special case as these are noisy and come from the underlying library itself
+                continue;
+            }
             EventTelemetry telemetry = new EventTelemetry(event.getName());
             String operationId = span.getTraceId();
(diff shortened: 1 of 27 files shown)


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -616,6 +616,12 @@ private static String nullAwareConcat(String str1, String str2, String separator
 
     private void exportEvents(SpanData span, Double samplingPercentage) {
         for (EventData event : span.getEvents()) {
+            boolean lettuce51 =
+                    span.getInstrumentationLibraryInfo().getName().equals("io.opentelemetry.javaagent.lettuce-5.1");
```
heyams, line 620: We only suppress lettuce-5.1 and allow other versions if available? What about lettuce-5.0 and 4.0?

heyams (approved):

PR #1599, opened 2021-04-01
Configure the JMX metrics to include multiple attributes per objectName

MS-jgol:
@trask Please review

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
@@ -280,11 +280,22 @@ public class AiComponentInstaller implements ComponentInstaller {
         // configure custom jmx metrics
         ArrayList<JmxXmlElement> jmxXmls = new ArrayList<>();
         for (JmxMetric jmxMetric : config.jmxMetrics) {
-            JmxXmlElement jmxXml = new JmxXmlElement();
-            jmxXml.setName(jmxMetric.name);
-            jmxXml.setObjectName(jmxMetric.objectName);
-            jmxXml.setAttribute(jmxMetric.attribute);
-            jmxXmls.add(jmxXml);
+            if (!jmxMetric.attribute.isEmpty()) {
+                JmxXmlElement jmxXml = new JmxXmlElement();
+                jmxXml.setName(jmxMetric.name);
+                jmxXml.setObjectName(jmxMetric.objectName);
+                jmxXml.setAttribute(jmxMetric.attribute);
+                jmxXmls.add(jmxXml);
+            }
+            else if (!jmxMetric.attributes.isEmpty()) {
+                for (String attribute : jmxMetric.attributes) {
+                    JmxXmlElement jmxXml = new JmxXmlElement();
+                    jmxXml.setName(getJmxDisplayName(jmxMetric.objectName) + attribute);
+                    jmxXml.setObjectName(jmxMetric.objectName);
+                    jmxXml.setAttribute(attribute);
+                    jmxXmls.add(jmxXml);
+                }
+            }
         }
         xmlConfiguration.getPerformance().setJmxXmlElements(jmxXmls);
 
@@ -296,6 +307,19 @@ public class AiComponentInstaller implements ComponentInstaller {
             xmlConfiguration.getChannel().setDeveloperMode(true);
         }
         return xmlConfiguration;
+    }
+
+    //assuming that every other element is not needed - such as "type", "worker", "name"
+    //assuming same set of delimiters in every object name
+    private static String getJmxDisplayName(String attribute) {
+        String nameParts[] = attribute.split("=|:|,");
+        String name = "";
+        if(nameParts!=null&&nameParts.length>0) {
+            for(int i=0; i<nameParts.length;i=i+2) {
+                name+= nameParts[i] + " / ";
+            }
+        }
+        return name;
     }
 
     private static ParamXmlElement newParamXml(String name, String value) {
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
@@ -96,7 +96,8 @@ public class Configuration {
 
         public String name;
         public String objectName;
-        public String attribute;
+        public String attribute = "";
+        public List<String> attributes = new ArrayList<>();
     }
 
     public static class Instrumentation {
--- a/agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationTest.java
+++ b/agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationTest.java
@@ -53,9 +53,10 @@ public class ConfigurationTest {
         assertEquals("abc", configuration.customDimensions.get("some key"));
         assertEquals("def", configuration.customDimensions.get("another key"));
         assertEquals(10.0, configuration.sampling.percentage, 0);
-        assertEquals(3, configuration.jmxMetrics.size());
+        assertEquals(4, configuration.jmxMetrics.size());
         assertEquals("Thread Count", configuration.jmxMetrics.get(0).name);
         assertEquals("java.lang:type=Threading", configuration.jmxMetrics.get(0).objectName);
+        // java.lang/Threading/ThreadCount
         assertEquals("ThreadCount", configuration.jmxMetrics.get(0).attribute);
         assertEquals("error", configuration.instrumentation.logging.level);
         assertEquals(60, configuration.heartbeat.intervalSeconds);
@@ -201,7 +202,7 @@ public class ConfigurationTest {
         assertEquals("InstrumentationKey=00000000-0000-0000-0000-000000000000", configuration.connectionString);
         assertEquals("Something Good", configuration.role.name);
         assertEquals("xyz123", configuration.role.instance);
-        assertEquals(3, configuration.jmxMetrics.size());
+        assertEquals(4, configuration.jmxMetrics.size());
         assertEquals("error", configuration.instrumentation.logging.level);
         assertTrue(configuration.instrumentation.micrometer.enabled);
         assertEquals(60, configuration.heartbeat.intervalSeconds);
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
@@ -46,6 +46,7 @@ import org.slf4j.LoggerFactory;
 public class JmxDataFetcher {
 
     private static final Logger logger = LoggerFactory.getLogger(JmxDataFetcher.class);
+    private static ArrayList<JmxAttributeData> warningShown= new ArrayList<>();
 
     /**
      * Gets an object name and its attributes to fetch and will return the data.
@@ -72,13 +73,24 @@ public class JmxDataFetcher {
                 Collection<Object> resultForAttribute = fetch(server, objects, attribute.attribute);
                 result.put(attribute.metricName, resultForAttribute);
             } catch (Exception e) {
-                logger.warn("Failed to fetch JMX object '{}' with attribute '{}': ", objectName, attribute.attribute);
-                throw e;
+                //Is it ok if we print out the exception text instead of throwing the exception so we can continue the loop?
+                //Is it ok to print out and log the exception once per metric so that we do not flood the logs?
+                if(!warningShown.contains(attribute)) {
+                    warningShown.add(attribute);
+                    System.out.println(e.toString());
+                    logger.error(e.toString());
+                    logger.warn("Failed to fetch JMX object '{}' with attribute '{}': ", objectName, attribute.attribute);
+                }
+
+                //throw e;
             }
         }
 
         return result;
     }
+
+    // objectName = "gc=*"
+    // attribute = "collectionTime"
 
     private static Collection<Object> fetch(MBeanServer server, Set<ObjectName> objects, String attributeName)
             throws AttributeNotFoundException, MBeanException, ReflectionException, InstanceNotFoundException {
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
@@ -136,6 +136,7 @@ public class ConfigurationBuilder {
             threadCountJmxMetric.objectName = "java.lang:type=Threading";
             threadCountJmxMetric.attribute = "ThreadCount";
             config.jmxMetrics.add(threadCountJmxMetric);
+            System.out.println("ThreadCount metric did not exist");
         }
         if (!jmxMetricExisted(config.jmxMetrics, "java.lang:type=ClassLoading", "LoadedClassCount")) {
             JmxMetric classCountJmxMetric = new JmxMetric();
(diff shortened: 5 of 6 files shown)


MS-jgol:
@trask I added some comments within the code that I will need to remove - there are a few questions/assumptions for an easier review

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -280,11 +280,22 @@ private static ApplicationInsightsXmlConfiguration buildXmlConfiguration(Configu
...
+                JmxXmlElement jmxXml = new JmxXmlElement();
+                jmxXml.setName(jmxMetric.name);
+                jmxXml.setObjectName(jmxMetric.objectName);
+                jmxXml.setAttribute(jmxMetric.attribute);
+                jmxXmls.add(jmxXml);
+            }
+            else if (!jmxMetric.attributes.isEmpty()) {
+                for (String attribute : jmxMetric.attributes) {
+                    JmxXmlElement jmxXml = new JmxXmlElement();
+                    jmxXml.setName(getJmxDisplayName(jmxMetric.objectName) + attribute);
```
trask, line 293: you can calculate jmxDisplayName once outside of the for-each loop

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -298,6 +309,19 @@ private static ApplicationInsightsXmlConfiguration buildXmlConfiguration(Configu
         return xmlConfiguration;
     }
 
+    //assuming that every other element is not needed - such as "type", "worker", "name"
+    //assuming same set of delimiters in every object name
+    static String getJmxDisplayName(String attribute) {
+        String[] nameParts = attribute.split("=|:|,");
+        StringBuilder name = new StringBuilder();
+        if(nameParts.length>0) {
```
trask, line 317: I don't think you need this conditional, ok to loop 0 times below

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -96,7 +96,8 @@
 
         public String name;
         public String objectName;
-        public String attribute;
+        public String attribute = "";
+        public List<String> attributes = new ArrayList<>();
```
trask, line 100: can you add comment here explaining why there is both attribute and attributes?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstallerTest.java
```diff
@@ -0,0 +1,15 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class AiComponentInstallerTest {
+
+    @Test
+    public void jmxObjectNameCreation() {
+        assertEquals("java.lang / Threading / ", AiComponentInstaller.getJmxDisplayName("java.lang:type=Threading"));
+        assertEquals("java.nio / BufferPool / direct / ", AiComponentInstaller.getJmxDisplayName("java.nio:type=BufferPool,name=direct"));
+        assertEquals("java.lang / GarbageCollector / PS MarkSweep / ", AiComponentInstaller.getJmxDisplayName("java.lang:type=GarbageCollector,name=PS MarkSweep"));
+        assertEquals("Catalina / RequestProcessor / \"http-nio-8080\" / HttpRequest1 / ", AiComponentInstaller.getJmxDisplayName("Catalina:type=RequestProcessor,worker=\"http-nio-8080\",name=HttpRequest1"));
```
trask, line 13: 👍

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationTest.java
```diff
@@ -53,9 +53,10 @@ public void shouldParse() throws IOException {
         assertEquals("abc", configuration.customDimensions.get("some key"));
         assertEquals("def", configuration.customDimensions.get("another key"));
         assertEquals(10.0, configuration.sampling.percentage, 0);
-        assertEquals(3, configuration.jmxMetrics.size());
+        assertEquals(4, configuration.jmxMetrics.size());
         assertEquals("Thread Count", configuration.jmxMetrics.get(0).name);
         assertEquals("java.lang:type=Threading", configuration.jmxMetrics.get(0).objectName);
+        // java.lang/Threading/ThreadCount
```
trask, line 59:
```suggestion
```

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -72,14 +73,25 @@
...
         }
 
         return result;
     }
 
+    // objectName = "gc=*"
+    // attribute = "collectionTime"
+
     private static Collection<Object> fetch(MBeanServer server, Set<ObjectName> objects, String attributeName)
             throws AttributeNotFoundException, MBeanException, ReflectionException, InstanceNotFoundException {
```
trask, line 96: I like the idea of a comment here with example parameters, but it's a little unclear

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
```diff
@@ -96,15 +101,20 @@ private static void performTest(
...
-        verify(result, "Int", expectedInt);
-        verify(result, "Double", expectedDouble);
-        verify(result, "Long", expectedLong);
+            verify(result, "Int", expectedInt);
+            verify(result, "Double", expectedDouble);
+            verify(result, "Long", expectedLong);
+        }
+        catch(Exception e) {
+            Assert.fail("Exception thrown: "+e);
+        }
```
trask, line 117: it's not clear why this catch is needed?

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
```diff
@@ -47,17 +47,22 @@ public long getLongSample() {
...
+            server.registerMBean(testStub, mxbeanName);
+            List<JmxAttributeData> attributes = new ArrayList<>();
+            attributes.add(new JmxAttributeData("Int", "WrongNameIntSample"));
+            attributes.add(new JmxAttributeData("Double", "WrongNameDoubleSample"));
+            attributes.add(new JmxAttributeData("Long", "WrongNameLongSample"));
+            JmxDataFetcher.fetch("JSDKTests:type=TestStub3", attributes);
+        }
+        catch (Exception e) {
+            Assert.fail("Exception thrown: "+e);
+        }
```
trask, line 65: same here

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -46,6 +46,7 @@
 public class JmxDataFetcher {
 
     private static final Logger logger = LoggerFactory.getLogger(JmxDataFetcher.class);
+    private static ArrayList<JmxAttributeData> warningShown= new ArrayList<>();
```
trask, line 49: set will perform `contains` better, and concurrent data structure is important since accessed across multiple threads
```suggestion
    private static Set<JmxAttributeData> warningShown= Collections.newSetFromMap(new ConcurrentHashMap<>());
```

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -72,14 +73,25 @@
                 Collection<Object> resultForAttribute = fetch(server, objects, attribute.attribute);
                 result.put(attribute.metricName, resultForAttribute);
             } catch (Exception e) {
-                logger.warn("Failed to fetch JMX object '{}' with attribute '{}': ", objectName, attribute.attribute);
-                throw e;
+                //Is it ok if we print out the exception text instead of throwing the exception so we can continue the loop?
+                //Is it ok to print out and log the exception once per metric so that we do not flood the logs?
+                if(!warningShown.contains(attribute)) {
+                    warningShown.add(attribute);
```
trask, line 79:
```suggestion
                if(warningShown.add(attribute)) {
```

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -72,14 +73,25 @@
                 Collection<Object> resultForAttribute = fetch(server, objects, attribute.attribute);
                 result.put(attribute.metricName, resultForAttribute);
             } catch (Exception e) {
-                logger.warn("Failed to fetch JMX object '{}' with attribute '{}': ", objectName, attribute.attribute);
-                throw e;
+                //Is it ok if we print out the exception text instead of throwing the exception so we can continue the loop?
+                //Is it ok to print out and log the exception once per metric so that we do not flood the logs?
```
trask, line 77: yes and yes

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
```diff
@@ -72,14 +73,25 @@
...
+                if(!warningShown.contains(attribute)) {
+                    warningShown.add(attribute);
+                    System.out.println(e.toString());
+                    logger.error(e.toString());
+                    logger.warn("Failed to fetch JMX object '{}' with attribute '{}': ", objectName, attribute.attribute);
```
trask, line 82:
```suggestion
                    logger.warn("Failed to fetch JMX object '{}' with attribute '{}': {}", objectName, attribute.attribute, e.toString());
```

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
MS-jgol: Not very familiar with the required test structure. Do we need a way to handle the exception here?

core/src/test/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcherTest.java
MS-jgol: Same here - not very familiar with the required test structure. Do we need a way to handle the exception here? What is the proper way for the test purposes - remove? change to finally?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
MS-jgol: Done

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
MS-jgol: Changed

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
MS-jgol: Right... the name would be the same for all attributes...

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -298,6 +310,18 @@ private static ApplicationInsightsXmlConfiguration buildXmlConfiguration(Configu
         return xmlConfiguration;
     }
 
+    //assuming that every other element is not needed - such as "type", "worker", "name"
+    //assuming same set of delimiters in every object name
+    static String getJmxDisplayName(String attribute) {
+        String[] nameParts = attribute.split("=|:|,");
```
trask, line 316: can you add a test with one of these separators inside of double quotes? I think you may need more complex algorithm for this, see https://github.com/glowroot/glowroot/blob/master/common2/src/main/java/org/glowroot/common2/repo/util/Gauges.java#L124-L189

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
trask: just checking you saw this (github is hiding it behind "2 hidden conversations" button)

trask:
> Same here - not very familiar with the required test structure. Do we need a way to handle the exception here? What is the proper way for the test purposes - remove? change to finally?

I pushed an update to your branch to remove those two catch blocks

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
MS-jgol: @trask Did not see this... this is a placeholder for the future work, I can remove it, not needed here

core/src/main/java/com/microsoft/applicationinsights/internal/jmx/JmxDataFetcher.java
MS-jgol: Thanks :) 
Removed the comments as they are not needed

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
MS-jgol: @trask Decided to try another approach... See if you approve. Basically, I was thinking that too if we include the domain, type and name, followed by attribute it will be more readable. Using ObjectName for parsing. 
Questions:
1. Are you ok with using ObjectName class? In case the objectname format may change in the future, this is probably a safe option to go with in the longer run?
2. Is the format I described above adequate? 
3. If point 1. is ok, but point 2. is not, I can keep using ObjectName but modify how I extract the properties.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -313,11 +316,23 @@ else if (!jmxMetric.attributes.isEmpty()) {
...
+            ObjectName nameObject = new ObjectName(attribute);
+            name.append(nameObject.getDomain()).append(" / ");
+            Hashtable properties = nameObject.getKeyPropertyList();
+
+            if(properties.containsKey("type")) {
+                name.append(properties.get("type")).append(" / ");
+            }
+            if(properties.containsKey("name")) {
+                name.append(properties.get("name")).append(" / ");
+            }
```
trask, line 332: only handling `type` and `name` will lead to duplicate names for mbean names that vary only on other fields (which is common)

what do you think of simplifying, and just appending the full objectName + attributeName? it's not as pretty, but would require less documentation/explanation for users

MS-jgol: @trask 
Just like in Glowroot? 

For example:
1.
"Catalina:type=RequestProcessor,worker="http-nio-8080",name=HttpRequest1"
would be displayed as:
Catalina / RequestProcessor / "http-nio-8080" / HttpRequest1 / (plus attribute name here)

2. "tomcat.jdbc:name=\"jdbc/storage\",context=/,engine=Catalina,type=ConnectionPool,host=localhost,class=org.apache.tomcat.jdbc.pool.DataSource"

Would be displayed as:

"tomcat.jdbc / "jdbc/storage" / / / Catalina / ConnectionPool / localhost / org.apache.tomcat.jdbc.pool.DataSource / (plus attribute name here)

Please confirm

trask: no, suggesting potentially simpler:

Catalina:type=RequestProcessor,worker="http-nio-8080",name=HttpRequest1 (plus attribute name here)

MS-jgol: @trask 
I changed the naming algorithm up again. See what you think.
Will generate names as:
Catalina /RequestProcessor /"http-nio-8080" /HttpRequest1 / (plus attribute name here)

I can always revert to the simple way as you suggested earlier. Or I can implement an algorithm like you did in GlowRoot, to identify the beginning and end of the quoted property (such as "http-nio-8080" in the object name above)

trask:
@heyams can you look at the pipeline failure?

heyams:
> @heyams can you look at the pipeline failure?

onebranch experienced an outage with their service account.. `System.InvalidOperationException: Task Get Service Tree Data failed to exit with a success (0) exit code` will go away automatically.  I triggered a new build.

trask:
closing, we are going to align with OpenTelemetry custom JMX metrics configuration options once that is implemented

PR #1601, opened 2021-04-01
New preview flag

trask:


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -604,6 +607,18 @@ private void exportRequest(SpanData span) {
         exportEvents(span, samplingPercentage);
     }
 
+    private String getTelemetryName(SpanData span) {
+        String name = span.getName();
+        if (!httpMethodInOperationName || !name.startsWith("/")) {
+            return name;
+        }
+        String httpMethod = span.getAttributes().get(SemanticAttributes.HTTP_METHOD);
+        if (httpMethod == null) {
```
heyams, line 616: include empty? otherwise line 619 would return "" + " " + name.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -175,6 +175,8 @@
         // ignoreRemoteParentNotSampled is currently needed
         // because .NET SDK always propagates trace flags "00" (not sampled)
         public boolean ignoreRemoteParentNotSampled = true;
+        // TODO consider turning this on by default in 3.1.0
+        public boolean httpMethodInOperationName;
```
heyams, line 179: If i remembered correctly, we used to have http method in operation name, then it got removed.. now we add it back as a config? i wonder what was the reason to remove it at the first place.

heyams: it used to be HTTP GET + operationName

heyams (approved):

PR #1606, opened 2021-04-03
Update APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL at runtime

heyams:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -166,4 +171,14 @@ public Properties asJavaProperties() {
...
+
+  public void updateLogCaptureThreshold(String property, String value) {
+    if (LOG_CAPTURE_THRESHOLD.equals(property)) {
+      String level = getProperty(property);
+      if (value != null && !value.isEmpty() && !value.equalsIgnoreCase(level)) {
+        getAllProperties().replace(LOG_CAPTURE_THRESHOLD, level, value.toUpperCase());
+        logger.info("Update {} from {} to {}", LOG_CAPTURE_THRESHOLD, level, value);
+      }
+    }
+  }
```
trask, line 183:
```suggestion
  public void updateLogCaptureThreshold(String value) {
    // TODO add test for dynamically updating in each logging instrumentation module
    getAllProperties().put(LOG_CAPTURE_THRESHOLD, value.toUpperCase());
  }
```

heyams: i don't think it should get updated when value is empty or null.  That would turn the level OFF. Updated.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -166,4 +171,11 @@ public Properties asJavaProperties() {
   public boolean isAgentDebugEnabled() {
     return getBooleanProperty("otel.javaagent.debug", false);
   }
+
+  public void updateLogCaptureThreshold(String value) {
+    if (value != null && !value.isEmpty()) {
+      getAllProperties().put(LOG_CAPTURE_THRESHOLD, value.toUpperCase());
+      logger.info("Update {} to {}", LOG_CAPTURE_THRESHOLD, value.toUpperCase());
+    }
+  }
```
trask, line 180: I think this will have a better chance of getting upstreamed:

(config can't do logging upstream because it's used to configure logging, and this way it's not a one-off for a single property)

```suggestion
  public void updateProperty(String key, String value) {
    getAllProperties().put(key, value);
  }
```

trask (approved):

PR #1607, opened 2021-04-03
Update instrumentation logging level at runtime

heyams:
This is based on #1606


trask (approved):
let's make sure this feature is included an end-to-end test in Azure Functions environment. it feels like something that could easily break in a future release, especially since updating otel config dynamically at runtime is not really supported upstream

PR #1609, opened 2021-04-05
Fix github source permission on the release pipeline

heyams:


.pipelines/pipeline.user.windows.official.yml
```diff
@@ -18,3 +18,11 @@ build:
...
           exclude:
             - 'buildSrc/**/*'
+    - !!buildcommand
+      name: 'SetSrcAsArtifact'
+      command: 'gci.cmd' # this file is empty
+      artifacts:
+        - from: 'src/'
+          to: 'out'
+          include:
+            - '**/*'
```
trask, line 28: this is copying all `src/` folders to `out`? how does this help with the permission issue? also, did you need to add `gci.cmd` to source control?

heyams: this will make source available in the artifacts so that in the release pipeline i don't need to create a separate artifact for github.  release pipeline will need to be updated.. added gci.cmd.

trask (approved):

PR #1610, opened 2021-04-06
Bump azure-storage-blob to the latest version 12.11.0-beta.2

johnoliver:
Resolves incompatible versions of azure-core


heyams:
@johnoliver can you create a new PR? We just added you to applicationinsights-java-contributors group to see if that makes any differences.  Reopen/closing PR doesn't seem to work. Thanks.

PR #1614, opened 2021-04-07
Fix wrong comparison operator

domi-87:
This resulted in an incorrect error message when the interval was set to 1 second.


trask (approved):
thx @domi-87!

PR #1621, opened 2021-04-13
Proposal to add GC mbean monitoring and use GC events for profile triggering

johnoliver:
This adds a library that monitors and parses gc events from mbean notification events and alerts observers of GCevents.

The memory profiling alerting trigger is then modified to be triggered based on the fill level of the tenured space after a tenured collection.

- [ ] CHANGELOG.md updated


agent/agent-gc-monitor/gc-monitor-api/src/main/java/com/microsoft/gcmonitor/memorypools/MemoryPool.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Representation of a JVM memory pool
+ */
+public abstract class MemoryPool {
+
+    private final String name;
+    private final Set<GarbageCollector> garbageCollectors;
+    private final boolean isTenuredPool;
+    private final boolean isYoungPool;
+    private final boolean isHeap;
```
trask, line 18: `isTenuredPool` -> `tenuredPool`

agent/agent-gc-monitor/gc-monitor-core/src/main/java/com/microsoft/gcmonitor/JMXGCMonitorFactory.java
```diff
@@ -0,0 +1,20 @@
+package com.microsoft.gcmonitor;
+
+import javax.management.MBeanServerConnection;
+import java.lang.management.ManagementFactory;
+import java.util.concurrent.ExecutorService;
+
+public class JMXGCMonitorFactory implements GcMonitorFactory {
```
trask, line 7: I've ended up liking this style:
```suggestion
public class JmxGcMonitorFactory implements GcMonitorFactory {
```

agent/agent-gc-monitor/gc-monitor-core/src/main/java/com/microsoft/gcmonitor/collectors/JMXGarbageCollector.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.management.openmbean.CompositeData;
+import java.lang.management.GarbageCollectorMXBean;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Implementation of a GarbageCollectorStats that provides access to the stats via a MxBean
+ */
+public class JMXGarbageCollector implements GarbageCollectorStats {
```
trask, line 24: to not get this confused with `GarbageCollector`
```suggestion
public class JmxGarbageCollectorStats implements GarbageCollectorStats {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -570,6 +571,11 @@ public void validate() throws FriendlyException {
         public boolean enabled = false;
     }
 
+    public static class GcEventConfiguration {
+        public boolean reportAllGcEvents = false;
+        public boolean reportTenuredGcEvents = false;
```
trask, line 576: or maybe a single enum property, similar to logging `level`, e.g. `NONE`, `TENURED`, `ALL`?
```suggestion
        public boolean reportAll = false;
        public boolean reportTenured = false;
```

core/src/main/java/com/microsoft/applicationinsights/internal/profiler/GcEventMonitor.java
```diff
@@ -0,0 +1,138 @@
...
+                    alertingSubsystem.track(AlertMetricType.MEMORY, percentage);
+                }
+            }
+        }
+    }
+
+    /**
+     * If gc reporting is enabled, send gc event to Application Insights
+     */
+    private static void emmitGcEvent(TelemetryClient telemetryClient, GcEventMonitorConfiguration gcEventMonitorConfiguration, GCCollectionEvent event) {
```
trask, line 98:
```suggestion
    private static void emitGcEvent(TelemetryClient telemetryClient, GcEventMonitorConfiguration gcEventMonitorConfiguration, GCCollectionEvent event) {
```

agent/agent-gc-monitor/gc-monitor-api/src/main/java/com/microsoft/gcmonitor/memorypools/MemoryPools.java
```diff
@@ -0,0 +1,345 @@
...
+        Optional<MemoryPools> pool = MemoryPools.findPoolFor(id);
+
+        if (pool.isPresent()) {
+            return pool.get().factory.apply(managers);
+        } else {
+            throw new IllegalArgumentException("Cannot find memory pool " + id);
+        }
+    }
+
+    public static String getIdentifier(String name) {
```
trask, line 86: can you add a comment why we need an id and cannot use name directly?

trask (commented):
this is a really great addition too!

a few initial comments so far

agent/agent-gc-monitor/gc-monitor-api/src/main/java/com/microsoft/gcmonitor/memorypools/MemoryPools.java
johnoliver: this is a remnant of when we used this as a key inside JSON, it is not needed anymore, can remove it

agent/agent-gc-monitor/gc-monitor-api/src/main/java/com/microsoft/gcmonitor/MemoryManagers.java
```diff
@@ -0,0 +1,74 @@
...
+    private boolean isComposedOf(Set<GarbageCollector> collectors) {
+        Set<Class<? extends GarbageCollector>> collectorClasses = collectors.stream()
+                .map(GarbageCollector::getClass)
+                .collect(Collectors.toSet());
+        for (Class<? extends GarbageCollector> manager : managers) {
+            if (!collectorClasses.contains(manager)) {
+                return false;
+            }
+        }
+        return true;
```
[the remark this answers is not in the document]
johnoliver: I believe so yes

agent/agent-tooling/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     def includeProfiler = project.hasProperty("includeProfiler") ? project.includeProfiler.toBoolean() : false
     if (includeProfiler) {
         implementation project(":agent:agent-profiler:agent-service-profiler")
+        implementation project(":agent:agent-gc-monitor:gc-monitor-core")
```
trask, line 37: this seems useful outside of the profiler(?)

core/src/main/java/com/microsoft/applicationinsights/internal/profiler/GcReportingLevel.java
```diff
@@ -0,0 +1,7 @@
+package com.microsoft.applicationinsights.internal.profiler;
+
+public enum GcReportingLevel {
+    NONE,
+    TENURED_ONLY,
+    ALL
+}
```
trask, line 7: this doesn't feel too intuitive from configuration perspective, I wonder if what you had (two separate booleans) was better, or if different naming would help here...

trask (approved):
thx!

agent/agent-tooling/build.gradle
johnoliver: It would be useful outside the profiler, as you can see if the reporting level is set we stream back telemetry events that should give a high level of detail about garbage collection activity compared to our current reporting, although below what a detailed gc log file or detailed JFR profile can provide. I put it behind this guard mostly as I did see this as beta functionality that to some extent is going hand in hand with the profile triggering. But given it should not increase the binary size we could pull it out so that it is included in the binary, but disabled by default until we feel it can come out of beta

PR #1638, opened 2021-04-19
Remove core dep from smoke tests

trask:


test/smoke/testApps/Sampling/build.gradle
```diff
@@ -1,7 +1,7 @@
 apply plugin: 'war'
 
 dependencies {
-    implementation aiWebJar
+    implementation aiCoreJar
```
kryalama, line 4: Is this a special case ? Still have core dependency?

trask: this test uses the 2.x SDK to send telemetry, to verify that telemetry from 2.x SDK is also sampled correctly

kryalama (approved):

heyams:
common schema (CS) is inside core.. now removing core dep will lead to create duplicate common schema data models in smoke/..
if we want to update CS, it needs to be updated in both places...

heyams:
should we create a common module to avoid creating duplicate codes?

trask:
> common schema (CS) is inside core.. now removing core dep will lead to create duplicate common schema data models in smoke/..
> if we want to update CS, it needs to be updated in both places...

sorry, I meant to explain the motivation for this change above but did not 😞

I'm removing the common schema that is inside of core as part of updating to use the Azure SDK swagger-based exporter

I want this change to go in separately, and then when we switch over to the Azure SDK swagger-based exporter, there will no need to change the smoke tests (which will be good confirmation that we didn't break anything)

heyams (approved):

PR #1644, opened 2021-04-21
Add ability to configure JFR profile settings

johnoliver:
Adds a configuration parameter `removeEnvironmentData` that removes some documented environment data from profiles.

Adds the ability to specify a jfc configuration file that will be used to configure jfr recordings.

- [ ] CHANGELOG.md updated


agent/agent-profiler/docs/README.md
```diff
@@ -70,8 +70,25 @@ Profiler may also be enabled via adding a config section to applicationinsights.
...
+
+`removeEnvironmentData` - If set to true (and no custom profile are provided), memory and CPU profiling will 
+use a profile similar to the default `profile` configuration with the following settings disabled:
+- jdk.JVMInformation
+- jdk.InitialSystemProperty
+- jdk.OSInformation
+- jdk.InitialEnvironmentVariable
+- jdk.SystemProcess
+
+`customMemoryProfile` - Path to a custom jfc configuration file. This configuration will be used in the event of a
```
trask, line 84: what do you think of renaming to `memoryTriggeredSettings` (and `cpuTriggeredSettings`)?

"settings" because it matches the JFR command-line option, and avoids multiple meanings of "profile"

and also, what do you think about making it have default value `profile` (corresponding to the JFR option), and supporting another non-file value `profile-without-env-data` (or similar), and getting rid of `removeEnvironmentData`?

trask (approved):
so sorry for the delayed review, only thoughts about configuration properties below, everything else looks great, thx again!

agent/agent-profiler/docs/README.md
johnoliver: sounds good

trask:
hey @johnoliver, don't worry about the test failure here, we're having problems with that one (unrelated to your PR), i'll get it fixed and get this PR merged in the next couple of days

PR #1646, opened 2021-04-22
Fix 79 component governance security alerts from low to critical

heyams:
Fix [TASK 9744451](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/AEP/Cobalt/2104?workitem=9744451)


trask (approved):
👍

PR #1647, opened 2021-04-22
Fix for jdbc target and duplicate headers

### LINKED ISSUE #1640: Spring webclient instrumentation
We have project where we using java 11, spring boot(webclient/webflux/netty), Application insights agent 3.0.3.

If we specify custom traceparent header in webclient, auto-instrumentation from application insights agent appends additional value to traceparent header:
`traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-b9c7c989f97918e1-01, 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01` .

Is this expected behavior? Problem with this that service which is called by webclient now returns 400, because `traceparent `header is invalid.

@PavelKhatsimtsau I can see that you managed to disable webclient instrumentation with "_otel.instrumentation.spring.webflux.client.enabled=false_", but we could not to disable it, we still see instrumentation in application insights. Is there some additional things that needs to be done to disable this instrumentation?

trask:
Update otel fork submodule, see: #1645

Updates Jdbc smoke tests due to the new (better) target attribute.

Fixes an unrelated issue where the Jdbc smoke tests just started failing due to recent update in mysql image that requires enabling TLSv1.3 in the driver

Resolves #1640

Attaching SNAPSHOT release from this PR: [applicationinsights-agent-3.0.4-BETA-SNAPSHOT.jar.zip](https://github.com/microsoft/ApplicationInsights-Java/files/6362473/applicationinsights-agent-3.0.4-BETA-SNAPSHOT.jar.zip)

### DIFF
--- a/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
+++ b/agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
@@ -509,15 +509,29 @@ public class Exporter implements SpanExporter {
 
     private static int getDefaultPortForDbSystem(String dbSystem) {
         switch (dbSystem) {
-            // TODO replace these with constants from OpenTelemetry API after upgrading to 0.10.0
-            // TODO add these default ports to the OpenTelemetry database semantic conventions spec
-            // TODO need to add more default ports once jdbc instrumentation reports net.peer.*
-            case "mongodb":
+            // jdbc default ports are from io.opentelemetry.javaagent.instrumentation.jdbc.JdbcConnectionUrlParser
+            // TODO make these ports constants (at least in JdbcConnectionUrlParser) so they can be used here
+            case SemanticAttributes.DbSystemValues.MONGODB:
                 return 27017;
-            case "cassandra":
+            case SemanticAttributes.DbSystemValues.CASSANDRA:
                 return 9042;
-            case "redis":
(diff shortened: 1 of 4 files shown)


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -509,15 +509,29 @@ private void applyMessagingClientSpan(Attributes attributes, RemoteDependencyTel
...
                 return 27017;
-            case "cassandra":
+            case SemanticAttributes.DbSystemValues.CASSANDRA:
                 return 9042;
-            case "redis":
+            case SemanticAttributes.DbSystemValues.REDIS:
                 return 6379;
+            case SemanticAttributes.DbSystemValues.MARIADB:
+            case SemanticAttributes.DbSystemValues.MYSQL:
+                return 3306;
```
heyams, line 522: Does otel-java have constants for these ports? It will be nice if they do.  Or also where is the spec for defining these port numbers?

heyams (approved):

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: see the comment above

PR #1653, opened 2021-04-25
Use Azure SDK exporter

trask:
This is a big change that needs some time to bake, so bumping the version to 3.1.0, and will create 3.0.x branch to continue services releases for 3.0.x until this is ready for prime time.

Many things still to do:
* Http proxy configured in json is not used yet
* Data is sent to path `/v2//track` instead of `/v2/track` (see https://github.com/Azure/azure-rest-api-specs/pull/14176)
* Data is not being sent gzipped
* ~Data is not being sent with ndjson~
* Batch sending (using the underlying swagger client) is not implemented yet
* Truncating telemetry fields is not implemented (e.g. see the `@Ignore(d)` JDBC smoke test)
* Store telemetry to disk on failure and retry and throttling (this is significant work, not going to be in this PR)
* All friendly error messages need to be revisited (many got deleted in telemetry channel)
* Note there's a breaking configuration change:
  * in order to use `-Dhttps.proxyHost` and `-Dhttps.proxyPort`, `-Djava.net.useSystemProxies` must also be set now
    (https://docs.microsoft.com/en-us/azure/developer/java/sdk/proxying)
* Check jar file size
* Performance testing (there were some low-level memory optimization done in previous json bindings, make sure new stuff is at least as good)
* Reconcile `com.microsoft.applicationinsights.agent.Exporter` and `com.azure.monitor.opentelemetry.exporter.AzureMonitorTraceExporter` (may need to wait for OpenTelemetry logs)
* Lots of smaller misc `FIXME (trask)` comments left lying around to deal with


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/instrumentation/sdk/TelemetryClientClassFileTransformer.java
```diff
@@ -722,14 +706,14 @@ public static void main(String[] args) {
     @SuppressWarnings("unused")
     public static class TC {
 
-        private TelemetryConfiguration configuration;
+        private com.microsoft.applicationinsights.TelemetryConfiguration configuration;
```
heyams, line 709: just curious to know why removing imports and using explicit types?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -85,7 +85,7 @@
 
     public static class Sampling {
 
-        public double percentage = 100;
+        public float percentage = 100;
```
heyams, line 88: We use double everywhere.. wondering why choose to use float here?

core/src/main/java/com/microsoft/applicationinsights/InstrumentationKey.java
```diff
@@ -0,0 +1,16 @@
+package com.microsoft.applicationinsights;
```
heyams, line 1: All the new classes are missing copyrights.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -152,22 +136,16 @@ private static void start(Instrumentation instrumentation) {
         }
         AppIdSupplier appIdSupplier = AppIdSupplier.INSTANCE;
 
-        TelemetryConfiguration configuration = TelemetryConfiguration.getActiveWithoutInitializingConfig();
-        TelemetryConfigurationFactory.INSTANCE.initialize(configuration, buildXmlConfiguration(config));
-        configuration.getContextInitializers().add(new SdkVersionContextInitializer());
-        configuration.getContextInitializers().add(new ResourceAttributesContextInitializer(config.customDimensions));
+        TelemetryClient telemetryClient = TelemetryClient.initActive(config.customDimensions, buildXmlConfiguration(config));
```
heyams, line 139: should we track config.customDimensions as a feature in statsbeat? 
I wasn't aware of this config before.  if nobody uses it, we can consider removing it?

does telemetry processor enable users to add custom dimensions?

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/instrumentation/sdk/TelemetryClientClassFileTransformer.java
trask: this is commented out code, wanted to it compile when uncommenting it, without having to add the import statements after uncommenting it

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
trask: the new TelemetryItem.setSampleRate takes a float, so that change rippled to several places

core/src/main/java/com/microsoft/applicationinsights/InstrumentationKey.java
trask: 👍 I'll add in separate PR  (#1678 is based on this PR already so don't want to cause it problems)

PR #1660, opened 2021-04-29
Statsbeat + smoke test

heyams:
- send network statsbeat every 15 mins
- send attach statsbeat every 15 mins
- send feature statsbeat daily
- separate statsbeats from regular metricTelemetry to its own buffer
- query azure metadata service every 15 mins
- add unit tests
- add smoke test


core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,172 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.*;
+
+public abstract class BaseStatsbeat {
```
trask, line 39: instead of a base class, can this be passed into the subclasses as a helper (e.g. "common properties" and/or "statsbeat sender")

composition often leads to nicer design over inheritance

and you can add common interface for send/reset if you need that

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(BaseStatsbeat.class);
+    protected final TelemetryClient telemetryClient;
+    protected String resourceProvider;
+    protected String operatingSystem;
+    protected ScheduledExecutorService scheduledExecutor;
```
trask, line 45: probably all instances can share a single thread, can make this static

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,194 @@
...
+public class NetworkStatsbeat extends BaseStatsbeat {
+
+    private static final Logger logger = LoggerFactory.getLogger(NetworkStatsbeat.class);
+    private static volatile Set<String> instrumentationList = new HashSet<>(64);
+    private static final AtomicLong requestSuccessCount = new AtomicLong(0);
+    private static final AtomicLong requestFailureCount = new AtomicLong(0);
+    private static volatile List<Double> requestDurations = new ArrayList<>();
+    private static final AtomicLong retryCount = new AtomicLong(0);
+    private static final AtomicLong throttlingCount = new AtomicLong(0);
+    private static final AtomicLong exceptionCount = new AtomicLong(0);
```
trask, line 52: these probably don't need to be static?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,194 @@
...
+        if (exceptionCount.get() != 0) {
+            MetricTelemetry exceptionCountSt = createStatsbeatTelemetry(EXCEPTION_COUNT, exceptionCount.get());
+            exceptionCountSt.getProperties().put(CUSTOM_DIMENSIONS_INSTRUMENTATION, instrumentation);
+            telemetryClient.track(exceptionCountSt);
+            logger.debug("send a NetworkStatsbeat{}: {}", EXCEPTION_COUNT, exceptionCountSt);
+        }
+    }
+
+    @Override
+    protected synchronized void reset() {
```
trask, line 108: i think there's a decent size race condition between send and reset, another option is to have a static `current` or `active` instance, and throw that away each time, that narrows the race condition down to these two lines (which we can decide to synchronize or not)

```
NetworkStatsbeat local = active
active = new NetworkStatsbeat()
```
then send `local`

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatModule.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final Object lock = new Object();
+
+    public static StatsbeatModule getInstance() {
+        synchronized (lock) {
+            if (instance == null) {
+                instance = new StatsbeatModule();
+            }
+        }
+        return instance;
+    }
```
trask, line 41: is there need for lazy init here?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,172 @@
...
+        runtimeVersion = System.getProperty("java.version");
+    }
+
+    protected abstract void send();
+
+    protected abstract void reset();
+
+    protected MetricTelemetry createStatsbeatTelemetry(String name, double value) {
+        MetricTelemetry telemetry = new MetricTelemetry(name, value);
+        telemetry.setTelemetryName(STATSBEAT_TELEMETRY_NAME);
```
trask, line 136: interesting, I'd like to understand this more

core/src/main/java/com/microsoft/applicationinsights/telemetry/MetricTelemetry.java
```diff
@@ -94,6 +94,11 @@ public void setName(String name) {
         metric.setName(name);
     }
 
+    @Override
+    public void setTelemetryName(String telemetryName) {
+        super.setTelemetryName(telemetryName);
+    }
+
```
trask, line 101: probably don't need the changes in this class anymore

core/src/main/java/com/microsoft/applicationinsights/channel/concrete/inprocess/InProcessTelemetryChannel.java
```diff
@@ -64,13 +67,21 @@ protected boolean doSend(Telemetry telemetry) {
...
             throw new IllegalStateException("Telemetry was previously used: " + telemetry);
         }
-        telemetryBuffer.add(telemetry);
+
+        // TODO Prepare for AAD support for Statsbeat iKey
+        if (telemetry instanceof MetricTelemetry) {
+            MetricTelemetry mt = (MetricTelemetry) telemetry;
+            if (STATSBEAT_TELEMETRY_NAME.equalsIgnoreCase(mt.getTelemetryName())) {
+                statsbeatBuffer.add(telemetry);
+            }
```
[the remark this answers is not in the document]
heyams: a good catch.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        return instance;
+    }
+
+    private AzureMetadataService() {
+        endpoint = String.format("%s?%s&%s", BASE_URL, API_VERSION, JSON_FORMAT);
+        scheduledExecutor.scheduleAtFixedRate(new InvokeMetadataServiceTask(), DEFAULT_STATSBEAT_INTERVAL, DEFAULT_STATSBEAT_INTERVAL, TimeUnit.SECONDS);
+    }
+
+    public static void parseJsonResponse(String response) throws IOException {
```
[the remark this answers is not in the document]
heyams: I moved it out for unit testing.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AzureMetadataService.class);
+    private static AzureMetadataService instance;
+    private static final String API_VERSION = "api-version=2017-08-01"; // this version has the smallest payload.
+    private static final String JSON_FORMAT = "format=json";
+    private static final String BASE_URL = "http://169.254.169.254/metadata/instance/compute";
+
+    private String endpoint;
+    private static final ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor(ThreadPoolUtils.createDaemonThreadFactory(AzureMetadataService.class));
+
+    public static AzureMetadataService getInstance() {
```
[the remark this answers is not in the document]
heyams: it has a scheduler and run every 15 mins..

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,75 @@
...
+        private void sendAsync() {
+            HttpGet request = new HttpGet(endpoint);
+            request.addHeader("Metadata", "true");
+            HttpResponse response;
+            try {
+                response = LazyHttpClient.getInstance().execute(request);
+                if (response != null) {
+                    parseJsonResponse(response.toString());
+                }
+            } catch (Exception ex) {
```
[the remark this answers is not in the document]
heyams: yes, I'll add a todo. I think if it's SocketException, it's sure to be non vm.  I need to do more testing to confirm it.  Backoff and then retry is needed for sporadic failures.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(BaseStatsbeat.class);
+    protected final TelemetryClient telemetryClient;
+    protected String resourceProvider;
+    protected String operatingSystem;
+    protected ScheduledExecutorService scheduledExecutor;
+    protected final long interval;
+
+    private String customerIkey;
+    private String version;
+    private String runtimeVersion;
```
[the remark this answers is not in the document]
heyams: final requires parameter to be passed into the constructor.. this is a private field and there is no setter.. it won't get changed once being initialized.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private final Object lock = new Object();
+
+    public BaseStatsbeat(TelemetryClient telemetryClient, long interval) {
+        this.telemetryClient = telemetryClient;
+        this.interval = interval;
+        initializeCommonProperties();
```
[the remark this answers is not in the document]
heyams: i will use a object to wrap around common properties and make it final.  yes, it will be calculated for each subclass.  I will cache it.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static void parseJsonResponse(String response) throws IOException {
+        if (response != null) {
+            Moshi moshi = new Moshi.Builder().build();
+            JsonAdapter<MetadataInstanceResponse> jsonAdapter = moshi.adapter(MetadataInstanceResponse.class);
+            StatsbeatModule.getInstance().getAttachStatsbeat().updateMetadataInstance(jsonAdapter.fromJson(response));
+        }
+    }
+
+    private class InvokeMetadataServiceTask implements Runnable {
```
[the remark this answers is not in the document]
heyams: no, i refactored it so that it will need to access a method in the outer class.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
heyams: we can talk about it when you're back.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -152,10 +155,15 @@
...
     public static class Heartbeat {
-
         public long intervalSeconds = MINUTES.toSeconds(15);
     }
 
+    public static class Statsbeat {
+        public String connectionString = "InstrumentationKey=c4a29126-a7cb-47e5-b348-11414998b11e;IngestionEndpoint=https://westus-0.in.applicationinsights.azure.com/"; //workspace-aistatsbeat
+        public long intervalSeconds = DEFAULT_STATSBEAT_INTERVAL;
+        public long featureIntervalSeconds = FEATURE_STATSBEAT_INTERVAL;
+    }
```
trask, line 165: are these configurations exposed just for testing? or do we need to document them / expect users to need them?

also, how will this work in gov clouds that cannot access public cloud?

core/src/main/java/com/microsoft/applicationinsights/internal/config/connection/ConnectionString.java
```diff
@@ -35,7 +43,35 @@ public static void parseInto(String connectionString, TelemetryConfiguration tar
...
+            } catch (URISyntaxException e) {
+                throw new InvalidConnectionStringException(Keywords.ENDPOINT_SUFFIX + " is invalid: " + suffix, e);
+            }
+        }
+
+        // set explicit endpoints for Statsbeat
+        String statsbeatEndpoint = kvps.get(Keywords.INGESTION_ENDPOINT);
+        if (!Strings.isNullOrEmpty(statsbeatEndpoint)) {
+            config.getEndpointProvider().setStatsbeatEndpoint(toUriOrThrow(statsbeatEndpoint, Keywords.INGESTION_ENDPOINT));
+        }
```
trask, line 74: if we don't expect users to change the statsbeat endpoint, we could configure instrumentationKey/ingestionEndpoint separately, and avoid parsing here

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,109 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.*;
+
+public class AttachStatsbeat extends BaseStatsbeat {
+
+    private String resourceProviderId;
+    private MetadataInstanceResponse metadataInstanceResponse;
+
+    public AttachStatsbeat(TelemetryClient telemetryClient, long interval) {
+        super(telemetryClient, interval);
+        initResourceProviderId();
```
trask, line 37: it's not ideal to call instance methods of a class during construction, since the class is still under construction, better if this method can be static and have it return the resourceProviderId instead of having the method set it directly
```suggestion
        resourceProviderId = initResourceProviderId();
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,109 @@
...
+        AzureMetadataService.getInstance().initialize(interval);
+    }
+
+    @Override
+    protected void send() {
+        if (resourceProviderId != null) {
```
trask, line 43: can this ever be non-null?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,109 @@
...
+    public void updateMetadataInstance(MetadataInstanceResponse response) {
+        if (response != null) {
+            metadataInstanceResponse = response;
+            updateResourceProvider(RP_VM);
+            updateOperatingSystem();
+        }
+    }
+
+    protected void updateResourceProvider(String rp) {
+        CustomDimensions.getInstance().getProperties().put(CUSTOM_DIMENSIONS_RP, rp);
```
trask, line 70: I think it's a little confusing to have CustomDimensions updated from inside of AttachStatsbeat, since it also needs to be updated for the others, I think would be clearer to treat CustomDimensions as its own top-level concept and set it directly inside of parseJsonResponse

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,109 @@
...
+        initResourceProviderId();
+    }
+
+    // osType from the Azure Metadata Service has a higher precedence over the running app’s operating system.
+    private void updateOperatingSystem() {
+        String osType = metadataInstanceResponse.getOsType();
+        if (osType != null && !"unknown".equalsIgnoreCase(osType)) {
+            CustomDimensions.getInstance().getProperties().put(CUSTOM_DIMENSIONS_OS, osType);
```
trask, line 78: same here

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,68 @@
...
+                    AzureMetadataService.this.parseJsonResponse(response.toString());
+                }
+            } catch (JsonEncodingException jsonEncodingException) {
+                // When it's not VM/VMSS, server does not return json back, and instead it returns text like the following:
+                // "<br />Error: NetworkUnreachable (0x2743). <br />System.Net.Sockets.SocketException A socket operation was attempted to an unreachable network 169.254.169.254:80".
+                logger.debug("This is not running from an Azure VM or VMSS. Shut down AzureMetadataService scheduler.");
+                scheduledExecutor.shutdown();
+                return;
+            } catch (Exception ex) {
+                // TODO add backoff and retry if it's a sporadic failure
```
trask, line 63: is it only going to retry every 15 min currently?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,68 @@
...
+
+        @Override
+        public void run() {
+            HttpGet request = new HttpGet(endpoint);
+            request.addHeader("Metadata", "true");
+            try {
+                HttpResponse response = LazyHttpClient.getInstance().execute(request);
+                if (response != null) {
+                    AzureMetadataService.this.parseJsonResponse(response.toString());
+                }
```
trask, line 55: does this keep running even after success? can you add comment why that's needed?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,82 @@
...
+            public void run() {
+                try {
+                    send();
+                }
+                catch (Exception e) {
+                    logger.error("Error occurred while sending statsbeat", e);
+                }
+            }
+        };
+    }
```
trask, line 77: the method name is confusing, maybe `createStatsbeatSender`? or maybe better to extract as named inner class `StatsbeatSender` and no need for method, just call `new StatsbeatSender()` above

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -0,0 +1,71 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.StatsbeatHelper.FEATURE_MAP;
+
+public class FeatureStatsbeat extends BaseStatsbeat {
+
+    private volatile Set<String> featureList = new HashSet<>(64);
+
+    public FeatureStatsbeat(TelemetryClient telemetryClient, long interval) {
+        super(telemetryClient, interval);
+        initFeatureList();
+        scheduledExecutor.scheduleAtFixedRate(sendStatsbeat(), interval, interval, TimeUnit.SECONDS);
```
trask, line 43: already being scheduled in super class?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -0,0 +1,71 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_OTHER;
+import static com.microsoft.applicationinsights.internal.statsbeat.StatsbeatHelper.FEATURE_MAP;
+
+public class FeatureStatsbeat extends BaseStatsbeat {
+
+    private volatile Set<String> featureList = new HashSet<>(64);
+
+    public FeatureStatsbeat(TelemetryClient telemetryClient, long interval) {
+        super(telemetryClient, interval);
+        initFeatureList();
```
trask, line 42: similar to comment above

```suggestion
        featureList = initFeatureList();
```

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelperTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class StatsbeatHelperTest {
+
+    private static final Set<String> instrumentations = new HashSet<String>() {{
+        add("io.opentelemetry.javaagent.jdbc");
+        add("io.opentelemetry.javaagent.tomcat-7.0");
+        add("io.opentelemetry.javaagent.http-url-connection");
+    }};
+
+    private long instrumentation = 144115188077961216L;
```
trask, line 19: same here

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelper.java
```diff
@@ -0,0 +1,227 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_ADOPT_OPENJDK;
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_MICROSOFT;
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_ORACLE;
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_OTHER;
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_REDHAT;
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_ZULU;
+
+public final class StatsbeatHelper {
+
+    private static final Map<String, Integer> INSTRUMENTATION_MAP = new HashMap<String, Integer>() {{
```
trask, line 38: can you replace this form of map initialization with static block? this is a very uncommon style and will confuse many Java developers 😄

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        if (local.exceptionCount.get() != 0) {
+            MetricTelemetry exceptionCountSt = createStatsbeatTelemetry(EXCEPTION_COUNT, local.exceptionCount.get());
+            exceptionCountSt.getProperties().put(CUSTOM_DIMENSIONS_INSTRUMENTATION, instrumentation);
+            telemetryClient.track(exceptionCountSt);
+        }
+    }
+
+    // TODO create a new instrumentation set automatically when its capacity reaches 64
```
trask, line 96: I'm not clear why this TODO is needed?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,190 @@
...
+
+        if (local.requestDurations.size() != 0) {
+            return sum / local.requestDurations.size();
+        }
+
+        return  sum;
+    }
+
+    private static class IntervalMetrics {
+        private volatile Set<String> instrumentationList = new HashSet<>(64);
```
trask, line 182: unless reason to pre-allocate? also can be final, and should use concurrent data struct (or synchronize all access)
```suggestion
        private final Set<String> instrumentationList = Collections.newSetFromMap(new ConcurrentHashMap<>());
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,190 @@
...
+        private final AtomicLong requestSuccessCount = new AtomicLong(0);
+        private final AtomicLong requestFailureCount = new AtomicLong(0);
+        private volatile List<Double> requestDurations = new ArrayList<>();
```
trask, line 185: let's track total and count separately (then can get average), tracking all durations is very memory expensive on high throughput app

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -152,10 +156,16 @@
     }
 
     public static class Heartbeat {
-
         public long intervalSeconds = MINUTES.toSeconds(15);
     }
 
+    public static class Statsbeat {
+        public String ikey = "c4a29126-a7cb-47e5-b348-11414998b11e;IngestionEndpoint"; //workspace-aistatsbeat
+        public String endpoint = ConnectionString.Defaults.INGESTION_ENDPOINT; // this supports the government cloud
```
trask, line 164: I thought gov clouds required a different ingestion endpoint (at least those gov clouds that do not have access to the public internet)?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
```diff
@@ -152,10 +156,16 @@
     }
 
     public static class Heartbeat {
-
         public long intervalSeconds = MINUTES.toSeconds(15);
     }
 
+    public static class Statsbeat {
+        public String ikey = "c4a29126-a7cb-47e5-b348-11414998b11e;IngestionEndpoint"; //workspace-aistatsbeat
```
trask, line 163: also may as well spell it out completely `instrumentationKey`
```suggestion
        public String ikey = "c4a29126-a7cb-47e5-b348-11414998b11e"; //workspace-aistatsbeat
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public void updateMetadataInstance(MetadataInstanceResponse response) {
+        if (response != null) {
+            metadataInstanceResponse = response;
+            resourceProviderId = initResourceProviderId(RP_VM);
+        }
+    }
+
+    protected String initResourceProviderId(String resourceProvider) {
```
trask, line 70: since this is being called from constructor, I think it's less confusing to make it static and pass in `null` for `metadataInstanceResponse`

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
```diff
@@ -0,0 +1,93 @@
...
+    public void setResourceProviderId(String resourceProviderId) {
+        this.resourceProviderId = resourceProviderId;
+    }
+
+    public MetadataInstanceResponse getMetadataInstanceResponse() {
+        return metadataInstanceResponse;
+    }
+
+    public void updateMetadataInstance(MetadataInstanceResponse response) {
+        if (response != null) {
```
trask, line 64: can `response` be `null` here?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AzureMetadataService.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public void parseJsonResponse(String response) throws IOException {
+        if (response != null) {
+            MetadataInstanceResponse metadataInstanceResponse = jsonAdapter.fromJson(response);
+            StatsbeatModule.getInstance().getAttachStatsbeat().updateMetadataInstance(metadataInstanceResponse);
+            CustomDimensions.getInstance().getProperties().put(CUSTOM_DIMENSIONS_RP, RP_VM);
+
+            // osType from the Azure Metadata Service has a higher precedence over the running app’s operating system.
+            String osType = metadataInstanceResponse.getOsType();
+            if (osType != null && !"unknown".equalsIgnoreCase(osType)) {
```
trask, line 51: this will also check the `null` condition:
```suggestion
            if (!"unknown".equalsIgnoreCase(osType)) {
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/BaseStatsbeat.java
```diff
@@ -0,0 +1,78 @@
...
+            try {
+                send();
+            }
+            catch (Exception e) {
+                logger.error("Error occurred while sending statsbeat", e);
+            }
+        }
+    }
+
+    protected long getInterval() {
```
trask, line 75:
```suggestion
    // only used by tests
    long getInterval() {
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+        properties.put(CUSTOM_DIMENSIONS_OS, operatingSystem);
+        properties.put(CUSTOM_DIMENSIONS_LANGUAGE, LANGUAGE);
+        properties.put(CUSTOM_DIMENSIONS_VERSION, version);
+    }
+
+    public ConcurrentMap<String, String> getProperties() {
+        return properties;
+    }
+
+    public static synchronized void reset() {
```
trask, line 111: also, is synchronized needed?
```suggestion
    // only used by tests
    static synchronized void reset() {
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+        return instance;
+    }
+    
+    private CustomDimensions() {
+        initialize();
+    }
+
+    private void initialize() {
```
trask, line 57: I don't think separate method adds value in this case, since outer method is only single line

also, then can make fields above `final`?

```suggestion
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final Object lock = new Object();
+    
+    public static CustomDimensions getInstance() {
+        synchronized (lock) {
+            if (instance == null) {
+                instance = new CustomDimensions();
+            }
+        }
+        return instance;
+    }
```
trask, line 51: do we need lazy init?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensions.java
```diff
@@ -0,0 +1,114 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.*;
+
+public final class CustomDimensions {
+
+    private static CustomDimensions instance;
+    private String resourceProvider;
+    private String operatingSystem;
+    private String customerIkey;
+    private String version;
+    private String runtimeVersion;
```
trask, line 40: these should be final or volatile

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/Configuration.java
heyams: I confirmed with Krishna, this end point will be used in gov clouds

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelperTest.java
```diff
@@ -0,0 +1,42 @@
...
+        add("io.opentelemetry.javaagent.http-url-connection");
+    }};
+
+    private long instrumentation = 144115188077961216L; // 2^13 + 2^21 + 2^57 (Exponents are keys from StatsbeatHelper.INSTRUMENTATION_MAP.)
+
+    private static final Set<String> features = new HashSet<String>() {{
+        add(JAVA_VENDOR_ZULU);
+    }};
+
+    private long feature = 2L;
```
trask, line 25:
```suggestion
    private static final long EXPECTED_FEATURES_BITMAP = 2L;
```

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelperTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static com.microsoft.applicationinsights.internal.statsbeat.Constants.JAVA_VENDOR_ZULU;
+import static org.junit.Assert.assertEquals;
+
+public class StatsbeatHelperTest {
+
+    private static final Set<String> instrumentations = new HashSet<String>() {{
+        add("io.opentelemetry.javaagent.jdbc");
+        add("io.opentelemetry.javaagent.tomcat-7.0");
+        add("io.opentelemetry.javaagent.http-url-connection");
+    }};
```
trask, line 17: can you change this also to static initializer? (or guava ImmutableSet if we have that on classpath here)

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeatTest.java
```diff
@@ -0,0 +1,121 @@
...
+        StatsbeatModule.getInstance().initialize(new TelemetryClient(), DEFAULT_STATSBEAT_INTERVAL, FEATURE_STATSBEAT_INTERVAL);
+        networkStatsbeat = StatsbeatModule.getInstance().getNetworkStatsbeat();
+    }
+
+    @Test
+    public void testAddInstrumentation() {
+        networkStatsbeat.addInstrumentation("io.opentelemetry.javaagent.jdbc");
+        networkStatsbeat.addInstrumentation("io.opentelemetry.javaagent.tomcat-7.0");
+        networkStatsbeat.addInstrumentation("io.opentelemetry.javaagent.http-url-connection");
+        assertEquals(networkStatsbeat.getInstrumentation(), 144115188077961216L); // 2^13 + 2^21 + 2^57 (Exponents are keys from StatsbeatHelper.INSTRUMENTATION_MAP.)
```
trask, line 32:
```suggestion
        assertEquals(networkStatsbeat.getInstrumentation(), 2^13 + 2^21 + 2^57); // Exponents are keys from StatsbeatHelper.INSTRUMENTATION_MAP
```

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/FeatureStatsbeatTest.java
```diff
@@ -0,0 +1,38 @@
...
+        StatsbeatModule.getInstance().initialize(new TelemetryClient(), DEFAULT_STATSBEAT_INTERVAL, FEATURE_STATSBEAT_INTERVAL);
+        featureStatsbeat = StatsbeatModule.getInstance().getFeatureStatsbeat();
+    }
+
+    @Test
+    public void testFeatureList() {
+        String javaVendor = System.getProperty("java.vendor");
+        final Set<String> features = new HashSet<String>() {{
+            add(javaVendor);
+        }};
```
trask, line 29:
```suggestion
        Set<String> features = Collections.singleton(javaVendor);
```

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensionsTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class CustomDimensionsTest {
+
+    @Before
+    public void setup() {
+        CustomDimensions.reset();
+    }
+
+    @Test
+    public void testResourceProvider() {
+        assertEquals(CustomDimensions.getInstance().getProperties().get(CUSTOM_DIMENSIONS_RP), UNKNOWN);
```
trask, line 27: the assertions are reversed in this test class, confusingly assertEquals takes expected as first arg

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/AttachStatsbeat.java
heyams: static will require helper methods static which failed the mockito tests.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/CustomDimensions.java
heyams: yes, it can be reset to null.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public void incrementThrottlingCount() {
+        current.throttlingCount.incrementAndGet();
+    }
+
+    public void incrementExceptionCount() {
+        current.exceptionCount.incrementAndGet();
+    }
+
+    public long getRequestSuccessCount() {
```
trask, line 145: can you mark the methods that are only used by tests (and make package protected if possible)?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/MetadataInstanceResponse.java
```diff
@@ -0,0 +1,76 @@
...
+    public transient String location;
+    public transient String name;
+    public transient String offer;
+    public transient String platformFaultDomain;
+    public transient String platformUpdateDomain;
+    public transient String publisher;
+    public transient String sku;
+    public transient String tags;
+    public transient String version;
+    public transient String vmSize;
```
trask, line 40: why transient?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -0,0 +1,67 @@
...
+        super(telemetryClient, interval);
+
+        // track java distribution
+        String javaVendor = System.getProperty("java.vendor");
+        if (javaVendor != null && !javaVendor.isEmpty()) {
+            if (FEATURE_MAP.get(javaVendor) == null) {
+                featureList.add(JAVA_VENDOR_OTHER);
+            } else {
+                featureList.add(javaVendor);
+            }
```
trask, line 50: just a more conventional way to write the same:
```suggestion
            if (FEATURE_MAP.contains(javaVendor)) {
                featureList.add(javaVendor);
            } else {
                featureList.add(JAVA_VENDOR_OTHER);
            }
```

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelper.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private static Long getPowerOf2(int exponent) {
+        long result = 1L;
+        while (exponent != 0) {
+            result *= 2;
+            exponent--;
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
heyams: why didn't i do that in the first place.. forgot about such a method exists then.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelper.java
```diff
@@ -0,0 +1,230 @@
...
+
+    private static long encode(Set<String> list, Map<String, Integer> map) {
+        long number = 0L;
+        for (String item : list) {
+            int index = map.get(item);
+            number |= getPowerOf2(index);
+        }
+
+        return number;
```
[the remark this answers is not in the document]
heyams: kusto cluster doesn't support BitSet.. what is the benefit using BitSet?

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/MetadataInstanceResponse.java
heyams: tell json parser to skip passing those values.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelper.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    private static Set<String> decode(long num, Map<Integer, String> decodedMap) {
```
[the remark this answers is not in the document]
heyams: decode is used by both instrumentation and feature.. thought it's better to be in an utils class.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatHelper.java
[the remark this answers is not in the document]
trask: it's going to do more efficient (bitwise) powers of 2, and then you can get the long back from it at the end, it's designed for this kind of thing, esp since we may want 2 longs in the future

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/MetadataInstanceResponse.java
trask: what are these fields used for?

trask (approved):
🎉🎉🎉

PR #1661, opened 2021-04-29
fix instrumental jdbc sybase

ppramos:
Fix jdbc instrumentation
Recursive statements inside Connection#getMetaData

Connection#getMetaData() may execute a Statement or PreparedStatement to retrieve DB info. This happens before the DB CLIENT span is started (and put in the current context), so this instrumentation runs again and the shouldStartSpan() check always returns true - and so on until we get a StackOverflowError using CallDepth prevents this, because this check happens before Connection#getMetadata() is called - the first recursive Statement call is just skipped and we do not create a span for it


trask:
hi @ppramos, so sorry missed this PR, could you submit this upstream to https://github.com/open-telemetry/opentelemetry-java-instrumentation instead?

trask:
closing, IIRC this has been fixed upstream

PR #1662, opened 2021-04-29
Introduce role to application rp configuration.

Incarnation-p-lee:
* Role doesn't support hot load.
* Role can be overriden by environment variables.

Signed-off-by: Pan Li <panli@microsoft.com>

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
@@ -334,7 +334,7 @@ public class ConfigurationBuilder {
     }
 
     // never returns empty string (empty string is normalized to null)
-    private static String getEnvVar(String name) {
+    protected static String getEnvVar(String name) {
         return trimAndEmptyToNull(System.getenv(name));
     }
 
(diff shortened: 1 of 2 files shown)


trask (approved):
thx @Incarnation-p-lee!

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/configuration/ConfigurationBuilder.java
```diff
@@ -285,6 +285,14 @@ private static void overlayRpConfiguration(Configuration config, RpConfiguration
         if (rpConfiguration.ignoreRemoteParentNotSampled != null) {
             config.preview.ignoreRemoteParentNotSampled = rpConfiguration.ignoreRemoteParentNotSampled;
         }
+        if (isTrimEmpty(config.role.name)) {
+            // only use rp configuration role name as a fallback, similar to WEBSITE_SITE_NAME
+            config.role.name = rpConfiguration.role.name;
+        }
+        if (isTrimEmpty(config.role.instance)) {
+            // only use rp configuration role name as a fallback, similar to WEBSITE_INSTANCE_ID
+            config.role.instance = rpConfiguration.role.instance;
```
Incarnation-p-lee, line 294: If so, looks customer cannot override this by env variables ?

Incarnation-p-lee, line 294: Azure Spring Cloud will set the role name from applicationinsights-rp.json by default. But we should allow customer to override the default behavior by env variables like ROLE_NAME and ROLE_INSTANCE.

Currently looks rpConfiguration will always win, right ?

heyams:
We need to update azure docs, right? @trask.  I am going to merge it in.

Incarnation-p-lee:
@heyams  Please wait a moment ? Looks there is one question as above need to address with @trask .

trask:
it will only set role name/instance from rp if it wasn't set by user, I sent PR to add test for this #1663, have a look and let us know if you still see any issue, thx!

Incarnation-p-lee:
Great! @trask just look back your changes, it is what we wanted. Thanks a lot.

PR #1667, opened 2021-05-04
Improve startup time

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/LoggingConfigurator.java
```diff
@@ -0,0 +1,269 @@
...
+import java.nio.file.Path;
+import java.util.Locale;
+
+import static org.slf4j.Logger.ROOT_LOGGER_NAME;
+
+class LoggingConfigurator {
+
+    private final LoggerContext loggerContext;
+
+    public Level level;
```
heyams, line 34: make instance variables private.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/LoggingConfigurator.java
```diff
@@ -0,0 +1,269 @@
...
+        // Diagnostics IPA log file location. Disabled by default.
+        String diagnosticsOutputDirectory = System.getenv(DiagnosticsHelper.APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY);
+        if (diagnosticsOutputDirectory != null && !diagnosticsOutputDirectory.isEmpty()) {
+            rootLogger.addAppender(configureDiagnosticAppender(diagnosticsOutputDirectory));
+        }
+
+        if (DiagnosticsHelper.isOsWindows()) {
+            rootLogger.addAppender(configureEtwAppender());
+        }
+        // TODO what about linux?
```
heyams, line 81: a good question.  can we create a work item to track this? i thought app services have something already.. will need to follow up with Jenny.

heyams (approved):
Added a couple of comments. 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wasbootstrap/LoggingConfigurator.java
trask: I think this is already tracked here: https://msazure.visualstudio.com/One/_workitems/edit/9606582

PR #1692, opened 2021-05-18
Fix bug in initialising JFR integration when used with older Java 8

johnoliver:
Older Java 8s do not have the Mxbean that we used to control JFR. This fixes the initialisation where if the Mxbean is not available it will try using the alternate Diagnostic Command method.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -23,7 +23,7 @@ ext {
 
             azureStorageBlob                 : "12.11.0-beta.3",
             azureClientRuntime               : "2.1.1",
-            microsoftJfrStreaming            : "1.1.0",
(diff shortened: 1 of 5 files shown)


gradle/dependencies.gradle
```diff
@@ -23,7 +23,7 @@ ext {
 
             azureStorageBlob                 : "12.11.0-beta.3",
             azureClientRuntime               : "2.1.1",
-            microsoftJfrStreaming            : "1.1.0",
+            microsoftJfrStreaming            : "1.2.0",
```
trask, line 26: probably need to update the lock files

```
./gradlew generateLockfiles --write-locks
```

trask (approved):

PR #1700, opened 2021-05-20
Use envelope name as the telemetry name and reduce payload

heyams:
Fix [#9860567](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/AEP/Cobalt/2105?workitem=9860567)

sample envelopes with this change:
image
image


trask (commented):
👍👍 can we make this in `aad` branch instead? since the implementation will be different over there, and not in a rush for this change

heyams:
don't we want to use the shorter telemetry name in the master branch? I can cherry-pick into the aad branch.

trask:
trying to reduce merge conflicts by limiting non-critical work in the master branch. planning to get the aad branch finished and merged into master in the next month or two.

trask (approved):

PR #1702, opened 2021-05-21
Truncate exception telemetry's parsedStack when there is a deep stack trace exceeding 65536 bytes

### LINKED ISSUE #482: ExceptionTelemetry with large / deep stack traces are silently omitted

heyams:
2021-05-20 17:44:13.245-07 DEBUG c.m.applicationinsights.agent - ############# current stack depth: 8943
2021-05-20 17:44:13.266-07 DEBUG c.m.a.telemetry.ExceptionTelemetry - parsedStack is exceeding 65536 bytes capacity. It is truncated from full **1024** frames to partial **308** frames.
@trask method call depth (8943) does not equal to the number of stack frames (total 1024). What i showed you earlier in Fiddler is the full stack.

When we add stack frame to ExceptionTelemetry, it's already in the reversed order.

Fix #482

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
@@ -230,6 +230,7 @@ public final class ExceptionTelemetry extends BaseSampleSourceTelemetry<Exceptio
      * @param stackFrame
      * @return the stack frame length for only the strings in the stack frame.
      */
+     // this is the same logic used to limit length on the Breeze side
     private static int getStackFrameLength(StackFrame stackFrame)
     {
         int stackFrameLength = (stackFrame.getMethod() == null ? 0 : stackFrame.getMethod().length())
--- a/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
@@ -25,6 +25,8 @@ import com.google.common.base.Strings;
 import com.microsoft.applicationinsights.internal.schemav2.ExceptionData;
 import com.microsoft.applicationinsights.internal.schemav2.ExceptionDetails;
 import com.microsoft.applicationinsights.internal.schemav2.StackFrame;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


trask:
can you test with an exception that is captured by auto-instrumentation? I think that will go through a different code path and not get fixed by this change

heyams:
Breeze accepts exceptions captured by auto-instrumentation that exceed 65536 bytes.  I've sent an email to Oleg to confirm if that's a bug.. because of it's insistent.  [here](https://github.com/cijothomas/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/ExceptionDetails.bond#L25-L27) clearly indicates that both parsedStack and stack should have the same constraint.. but based on my testing, "stack" doesn't have this constraint.  that's why it doesn't have any impact on auto-instrumentation yet.

trask (approved):

core/src/main/java/com/microsoft/applicationinsights/telemetry/ExceptionTelemetry.java
```diff
@@ -215,4 +225,16 @@ public String getEnvelopName() {
     public String getBaseTypeName() {
         return BASE_TYPE;
     }
+
+    /***
+     * @param stackFrame
+     * @return the stack frame length for only the strings in the stack frame.
+     */
```
trask, line 232:
```suggestion
     */
     // this is the same logic used to limit length on the Breeze side
```

trask (approved):

heyams:
@trask test:smoke:testApps:WebFlux kept failing.. i reran the job once.. i will let you take a look.. this is keep coming back.

PR #1705, opened 2021-05-24
AAD branch: fix shutdown hook

trask:
And improvements to the telemetry pipeline that will help when adding falling back to disk.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
@@ -171,32 +171,7 @@ public class AiComponentInstaller implements ComponentInstaller {
 
         // this is currently used by Micrometer instrumentation in addition to 2.x SDK
         BytecodeUtil.setDelegate(new BytecodeUtilImpl());
-        Runtime.getRuntime().addShutdownHook(new Thread() {
-            @Override
-            public void run() {
-                startupLogger.debug("running shutdown hook");
-                CompletableResultCode otelFlush = OpenTelemetryConfigurer.flush();
-                CompletableResultCode result = new CompletableResultCode();
-                otelFlush.whenComplete(() -> {
-                        CompletableResultCode batchingClientFlush = telemetryClient.flushChannelBatcher();
-                        batchingClientFlush.whenComplete(() -> {
-                            if (otelFlush.isSuccess() && batchingClientFlush.isSuccess()) {
-                                result.succeed();
-                            } else {
-                                result.fail();
-                            }
-                        });
-                });
-                result.join(5, SECONDS);
-                if (result.isSuccess()) {
-                    startupLogger.debug("flushing telemetry on shutdown completed successfully");
-                } else if (Thread.interrupted()) {
-                    startupLogger.debug("interrupted while flushing telemetry on shutdown");
-                } else {
-                    startupLogger.debug("flushing telemetry on shutdown has taken more than 5 seconds, shutting down anyways...");
-                }
-            }
-        });
+        Runtime.getRuntime().addShutdownHook(new ShutdownHook(telemetryClient));
 
         RpConfiguration rpConfiguration = MainEntryPoint.getRpConfiguration();
         if (rpConfiguration != null) {
@@ -315,5 +290,38 @@ public class AiComponentInstaller implements ComponentInstaller {
         paramXml.setName(name);
         paramXml.setValue(value);
         return paramXml;
+    }
+
+    private static class ShutdownHook extends Thread {
+        private final TelemetryClient telemetryClient;
+
+        public ShutdownHook(TelemetryClient telemetryClient) {
+            this.telemetryClient = telemetryClient;
+        }
+
+        @Override
+        public void run() {
+            startupLogger.debug("running shutdown hook");
+            CompletableResultCode otelFlush = OpenTelemetryConfigurer.flush();
+            CompletableResultCode result = new CompletableResultCode();
+            otelFlush.whenComplete(() -> {
+                    CompletableResultCode batchingClientFlush = telemetryClient.flushChannelBatcher();
+                    batchingClientFlush.whenComplete(() -> {
+                        if (otelFlush.isSuccess() && batchingClientFlush.isSuccess()) {
+                            result.succeed();
+                        } else {
+                            result.fail();
+                        }
+                    });
+            });
+            result.join(5, SECONDS);
+            if (result.isSuccess()) {
+                startupLogger.debug("flushing telemetry on shutdown completed successfully");
+            } else if (Thread.interrupted()) {
+                startupLogger.debug("interrupted while flushing telemetry on shutdown");
+            } else {
+                startupLogger.debug("flushing telemetry on shutdown has taken more than 5 seconds, shutting down anyways...");
+            }
+        }
     }
 }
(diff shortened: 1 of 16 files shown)


kryalama:
Awesome

kryalama (approved):

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
```diff
@@ -0,0 +1,29 @@
...
+        if (byteBuffer != null) {
+            return byteBuffer;
+        }
+        return ByteBuffer.allocate(BYTE_BUFFER_SIZE);
+    }
+
+    void offer(List<ByteBuffer> byteBuffers) {
+        // TODO(trask) batch offer?
+        for (ByteBuffer byteBuffer : byteBuffers) {
+            queue.offer(byteBuffer);
```
heyams, line 26: can you check bool return type?  when the queue is full, it doesn't throw an exception instead it returns false and byteBuffer is not added to the queue, which might lead to data lost.

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
```diff
@@ -0,0 +1,119 @@
...
+import reactor.util.context.Context;
+
+import java.io.IOException;
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.util.Iterator;
+import java.util.List;
+
+// TODO performance testing
+class LowLevelClient {
```
heyams, line 23: Can we rename it to ApplicationInisightsClient?

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+
+class ByteBufferPool {
+
+    private static final int BYTE_BUFFER_SIZE = 65536;
+    private static final int MAX_RETAINED = 10;
```
heyams, line 11: is this too small? the old channel buffer supports up to 1000.

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
```diff
@@ -0,0 +1,119 @@
...
+
+    List<ByteBuffer> encode(List<TelemetryItem> telemetryItems) throws IOException {
+        ByteBufferOutputStream out = new ByteBufferOutputStream(byteBufferPool);
+        for (Iterator<TelemetryItem> i = telemetryItems.iterator(); i.hasNext(); ) {
+            mapper.writeValue(out, i.next());
+            if (i.hasNext()) {
+                out.write('\n');
+            }
+        }
+        out.close(); // closing ByteBufferOutputStream is not really needed, but this makes LGTM happy
```
heyams, line 76: it would be nice if calling close() in a finally block in the case of exception, it will get close properly.  is it ok to not make this throwing an IOException and handle exception here instead of line 54 - 59?  if there is an exception thrown from line 69 - 75, bytebufferoutputstream is not properly closed yet.

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
```diff
@@ -0,0 +1,119 @@
...
+
+        // TODO(trask) subscribe with listener
+        //  * retry on first failure (may not need to worry about this if retry policy in pipeline already, see above)
+        //  * write to disk on second failure
+        CompletableResultCode result = new CompletableResultCode();
+        pipeline.send(request)
+                .contextWrite(Context.of(Tracer.DISABLE_TRACING_KEY, true))
+                .subscribe(response -> {
+                    // TODO parse response, looking for throttling, partial successes, etc
+                    System.out.println("on response: " + response);
```
heyams, line 107: can we use logger instead of sout?

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
```diff
@@ -0,0 +1,119 @@
...
+                }, error -> {
+                    System.out.println("on error...");
```
heyams, line 109: logger please.

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
```diff
@@ -0,0 +1,119 @@
...
+                    byteBufferPool.offer(byteBuffers);
+                    result.fail();
+                }, () -> {
+                    System.out.println("on complete...");
```
heyams, line 113: same here. please use logger.

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
trask: 👍 commented it out, this is still work-in-progress

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
trask: clarified comment, and returned buffers to pool on exception

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
trask: this is still 650kb of memory, I think that should be enough for most requests?

core/src/main/java/com/microsoft/applicationinsights/LowLevelClient.java
trask: renamed to TelemetryChannel

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
trask: data has already been sent, so no data lost. it's ok if pool is full, the byte buffer will be GC'd in that case

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
```diff
@@ -0,0 +1,29 @@
+package com.microsoft.applicationinsights;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
+
+class ByteBufferPool {
+
+    private static final int BYTE_BUFFER_SIZE = 65536;
```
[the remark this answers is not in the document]
trask: multiple of these can make up a single request, so is not a hard limit, just limit of each individual buffer

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -173,14 +178,25 @@ private static void start(Instrumentation instrumentation) {
...
                 startupLogger.debug("running shutdown hook");
-                try {
-                    telemetryClient.flush();
-                    telemetryClient.shutdown(5, SECONDS);
-                    startupLogger.debug("completed shutdown hook");
-                } catch (InterruptedException e) {
-                    startupLogger.debug("interrupted while flushing telemetry during shutdown");
-                } catch (Throwable t) {
-                    startupLogger.debug(t.getMessage(), t);
+                CompletableResultCode otelFlush = OpenTelemetryConfigurer.flush();
```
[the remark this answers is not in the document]
trask: extracted, it's not needed anywhere else, but it is a bit long for an anonymous inner class

trask:
@heyams I think I addressed all comments, PTAL 👍

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
heyams: for perf counters metric type, it collects every mins.. i have seen average up to 80+ telemetries per request.

image

heyams:
image
this is how fiddler tracks the request size.. it's average around ~41,000 bytes. I think 640KB is good enough?

core/src/main/java/com/microsoft/applicationinsights/ByteBufferPool.java
trask: multiple telemetries will fit in one bytebuffer

trask:
> this is how fiddler tracks the request size.. it's average around ~41,000 bytes. I think 640KB is good enough?

ya, it can still handle larger, it just won't use cached buffers in that case

heyams (approved):
👍

PR #1706, opened 2021-05-25
Emit Profiler start/stop events

johnoliver:
Emit profiler events that allow the back end to monitor when profiling starts/stops.

### DIFF
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/profiler/ProfilerServiceInitializer.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/profiler/ProfilerServiceInitializer.java
@@ -33,6 +33,7 @@ import com.microsoft.applicationinsights.profiler.ProfilerServiceFactory;
 import com.microsoft.applicationinsights.profiler.config.AlertConfigParser;
 import com.microsoft.applicationinsights.profiler.config.ServiceProfilerServiceConfig;
 import com.microsoft.applicationinsights.telemetry.EventTelemetry;
+import com.microsoft.applicationinsights.telemetry.TraceTelemetry;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/profiler/ProfilerServiceInitializer.java
```diff
@@ -162,19 +163,21 @@ static UploadCompleteHandler sendServiceProfilerIndex(TelemetryClient telemetryC
             event.getProperties().putAll(done.getServiceProfilerIndex().getProperties());
             event.getMetrics().putAll(done.getServiceProfilerIndex().getMetrics());
             telemetryClient.track(event);
+            telemetryClient.track(new TraceTelemetry("StopProfiler succeeded."));
```
trask, line 166: can you add a comment here that this is a very specific message that the backend looks for (as opposed to a diagnostic message for the user)?

core/src/main/java/com/microsoft/applicationinsights/internal/profiler/ProfilerServiceInitializer.java
```diff
@@ -162,19 +163,21 @@ static UploadCompleteHandler sendServiceProfilerIndex(TelemetryClient telemetryC
...
             GcEventMonitor.GcEventMonitorConfiguration gcEventMonitorConfiguration) {
-        return AlertingServiceFactory.create(alertAction(), TelemetryObservers.INSTANCE, telemetryClient, alertServiceExecutorService, gcEventMonitorConfiguration);
+        return AlertingServiceFactory.create(alertAction(telemetryClient), TelemetryObservers.INSTANCE, telemetryClient, alertServiceExecutorService, gcEventMonitorConfiguration);
     }
 
-    private static Consumer<AlertBreach> alertAction() {
+    private static Consumer<AlertBreach> alertAction(TelemetryClient telemetryClient) {
         return alert -> {
             if (profilerService != null) {
+                telemetryClient.track(new TraceTelemetry("StartProfiler triggered."));
```
trask, line 180: and here

trask (approved):

PR #1716, opened 2021-05-27
Fix App Map display of unknown dependencies

trask:


agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
```diff
@@ -322,6 +322,20 @@ private void applySemanticConventions(SpanData span, RemoteDependencyTelemetry r
             applyServiceBusSpan(attributes, remoteDependencyData);
             return;
         }
+
+        String target = getTargetFromPeerAttributes(attributes, -1);
```
heyams, line 326: defaultPort = 0?

heyams: I have seen other places using 0.

heyams (approved):

agent/exporter/src/main/java/com/microsoft/applicationinsights/agent/Exporter.java
trask: and added clarifying comment

PR #1717, opened 2021-05-27
aad branch update otel

trask:
Updates to latest OTel, this is companion PR to #1711


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -74,7 +75,7 @@ public static void setInstrumentation(Instrumentation inst) {
     }
 
     @Override
-    public void beforeByteBuddyAgent() {
+    public void beforeByteBuddyAgent(Config config) {
```
kryalama, line 78: are we using this config? is this for future usage?

trask:
Closing, going to update main branch instead, since need some of the fixes in latest otel sooner

PR #1718, opened 2021-05-28
add support for sse redirect and http proxy settings

kryalama:
add support for sse redirect
add support for http proxy settings
use LazyAzureHttpClient in telemetry channel
Fix bug based on aad enabled flag


trask:
> also add v2.1/track for aad auth path and v2/track for non aad auth path (this is temporary fix until Breeze is stabilised)

Is this what other SDKs are doing?

kryalama:
> > also add v2.1/track for aad auth path and v2/track for non aad auth path (this is temporary fix until Breeze is stabilised)
> 
> Is this what other SDKs are doing?

synced with other sdks. If the SDK already supports Stamp Specific redirects(SSR), recommendation is to use v2.1/track. But for Python, SSR not yet supported , hence the implementation is different from Nodejs and Java.

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
```diff
@@ -36,21 +37,24 @@
 
     private final HttpPipeline pipeline;
     private final URL endpoint;
+    private final HttpPipelinePolicy authenticationPolicy;
```
trask, line 40: is a field needed for this? looks like it's only used in the constructor

core/src/main/java/com/microsoft/applicationinsights/internal/authentication/AzureMonitorRedirectPolicy.java
```diff
@@ -0,0 +1,71 @@
...
+import reactor.core.publisher.Mono;
+
+import java.net.HttpURLConnection;
+
+public final class AzureMonitorRedirectPolicy implements HttpPipelinePolicy {
+
+    private static final int PERMANENT_REDIRECT_STATUS_CODE = 308;
+    // Based on Stamp specific redirects design doc
+    private static final int MAX_REDIRECT_RETRIES = 10;
+    private final ClientLogger logger = new ClientLogger(AzureMonitorRedirectPolicy.class);
```
trask, line 19: let's use slf4j since that's our norm in this repo

core/src/main/java/com/microsoft/applicationinsights/internal/authentication/AzureMonitorRedirectPolicy.java
```diff
@@ -0,0 +1,71 @@
...
+    private String redirectedEndpointUrl;
```
trask, line 20: since it's likely accessed from multiple threads
```suggestion
    private volatile String redirectedEndpointUrl;
```

core/src/main/java/com/microsoft/applicationinsights/internal/authentication/AzureMonitorRedirectPolicy.java
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     *  Function to process through the HTTP Response received in the pipeline
+     *  and retry sending the request with new redirect url.
+     */
+    private Mono<HttpResponse> attemptRetry(final HttpPipelineCallContext context,
+                                            final HttpPipelineNextPolicy next,
+                                            final HttpRequest originalHttpRequest,
+                                            final int retryCount) {
+        // make sure the context is not modified during retry, except for the URL
+        context.setHttpRequest(originalHttpRequest.copy());
```
trask, line 36: does this need to pass around `originalHttpRequest` or can it use `context.getHttpRequest()` each time?

kryalama: I believe we can use context.getHttpRequest, but I followed the same standard of setting the context from https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/RetryPolicy.java#L96

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -123,10 +123,13 @@ private static void start(Instrumentation instrumentation) {
...
-        Configuration.AadAuthentication authentication = config.preview.authentication;
-        AadAuthentication.init(authentication.type, authentication.clientId, authentication.keePassDatabasePath,
-                authentication.tenantId, authentication.clientSecret, authentication.authorityHost);
-
+        if(config.preview.authentication.enabled) {
+            Configuration.AadAuthentication authentication = config.preview.authentication;
+            AadAuthentication.init(authentication.type, authentication.clientId, authentication.keePassDatabasePath,
+                    authentication.tenantId, authentication.clientSecret, authentication.authorityHost);
+        } else {
+            AadAuthentication.init(null, null,null,null,null,null);
```
trask, line 131: do we have to call `init` in this case?

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
```diff
@@ -39,18 +44,22 @@
...
         policies.add(new RetryPolicy());
         // TODO handle authentication exceptions
         HttpPipelinePolicy authenticationPolicy = AadAuthentication.getInstance().getAuthenticationPolicy();
         if (authenticationPolicy != null) {
             policies.add(authenticationPolicy);
         }
-        pipeline.policies(policies.toArray(new HttpPipelinePolicy[0]));
-        this.pipeline = pipeline.build();
+        // Add Logging Policy
+        policies.add(new HttpLoggingPolicy(new HttpLogOptions()));
```
trask, line 60: is this for debugging?

core/src/main/java/com/microsoft/applicationinsights/internal/authentication/AadAuthentication.java
```diff
@@ -121,6 +123,8 @@ public HttpPipeline newHttpPipeLineWithAuthentication() {
         if(authenticationPolicy != null) {
             policies.add(authenticationPolicy);
         }
+        // Add Logging Policy
+        policies.add(new HttpLoggingPolicy(new HttpLogOptions()));
```
trask, line 127: only for debugging?

core/src/main/java/com/microsoft/applicationinsights/internal/authentication/AzureMonitorRedirectPolicy.java
```diff
@@ -0,0 +1,74 @@
+package com.microsoft.applicationinsights.internal.authentication;
+
+// This is a copy from Azure Monitor Open Telemetry Exporter SDK AzureMonitorRedirectPolicy
```
trask, line 3: can you move this comment down to the class declaration?

test/smoke/testApps/WebFlux/src/main/java/com/microsoft/applicationinsights/smoketestapp/TestController.java
```diff
@@ -20,7 +20,7 @@ public String root() {
         CompletableFuture<String> completableFuture = new CompletableFuture<>();
         Executors.newSingleThreadExecutor().execute(() -> {
             try {
-                Thread.sleep(500);
+                Thread.sleep(10000);
```
trask, line 23: is this change needed?

test/smoke/testApps/WebFlux/src/main/java/com/microsoft/applicationinsights/smoketestapp/TestController.java
```diff
@@ -38,7 +38,7 @@ public String root() {
...
```
trask, line 41: is this change needed?

test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
```diff
@@ -291,7 +291,7 @@ public void testTrackPageView() throws Exception {
 
         assertNotNull(pv2);
         assertEquals(new Duration(123456), pv2.getDuration());
-        assertEquals("2010-10-10T00:00:00.000+0000", pvdEnvelope2.getTime());
+        assertEquals("2010-10-10T00:00:00Z", pvdEnvelope2.getTime());
```
trask, line 294: what changed? or did we miss this failure previously?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
kryalama: had to add this because getInstance() will throw an IllegalStateException

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
kryalama: yes this will help customers debug and will show any exceptions thrown in the httppipeline

test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
kryalama: After latest otel update on the aad branch, I encountered this smoketest failing, hence the change.

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
trask: how do customers enable it?

trask:
> * 2 for Unused format argument

@kryalama when you have a chance, also check out the lgtm alert above

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
kryalama: through AZURE_LOG_LEVEL env variable.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
heyams: if authentication is disabled, AadAuthentication.init doesn't make sense.  It sounds like enabled needs to be checked before you can call AadAuthentication.getInstance() or don't throw IllegalStateException in getInstance(), but then instance can be null.  
It seems that AadAuthentication should not be a singleton.

kryalama: even I didnot like calling init with all nulls. Is there an alternative way to implement this?

trask: given time constraints I think it's ok to put a couple of TODO comments explaining that it's not ideal and that we want to revisit

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
trask: have you found this helpful for debugging? if so let's keep it but not put it in public facing docs, and add comment in code explaining how to enable it

test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
trask: I'm not sure this is the correct breeze format..

trask: nevermind, this is good 👍

PR #1721, opened 2021-06-01
add a unique id for GC events so that events from a single JVM can be grouped

johnoliver:


core/src/main/java/com/microsoft/applicationinsights/internal/profiler/GcEventMonitor.java
```diff
@@ -24,6 +25,9 @@
 public class GcEventMonitor {
     private static final Logger LOGGER = LoggerFactory.getLogger(GcEventMonitor.class);
 
+    //TODO if/when Application Insights adds a unique ID that represents a single JVM, pull that ID here
+    private static final String JVM_INSTANCE_UID = UUID.randomUUID().toString();
```
trask, line 29: does [role instance](https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#cloud-role-instance) work for this? or is it required to get a different different value on each restart of a single JVM?

johnoliver: it needs to be different for every restart as the gc starts again from scratch every time the JVM is restarted, and we need to analyse single JVMs

core/src/main/java/com/microsoft/applicationinsights/internal/profiler/GcEventMonitor.java
```diff
@@ -24,6 +25,9 @@
 public class GcEventMonitor {
     private static final Logger LOGGER = LoggerFactory.getLogger(GcEventMonitor.class);
 
+    //TODO if/when Application Insights adds a unique ID that represents a single JVM, pull that ID here
```
trask, line 28:
```suggestion
    // a unique jvm_instance_id is needed for every restart as the gc starts again from scratch every time
    // the JVM is restarted, and we need to analyze single JVM execution
    //TODO if/when Application Insights adds a unique ID that represents a single JVM, pull that ID here
```

trask (approved):

PR #1740, opened 2021-06-12
Jmx config - support multiple attributes and wildcard in names

MS-jgol:
@trask 
Added the wildcard in names test
To truly support wildcard in names we have to either dynamically generate the metric display names when the metrics are being fetched from mbean server, or send them as dimensions.


trask:
closing, we are going to align with OpenTelemetry custom JMX metrics configuration options once that is implemented

PR #1741, opened 2021-06-14
Update to latest otel

trask:


heyams:
should we update these to AAD branch too? switching between main and aad will cause otel-fork to be out of sync.

heyams (approved):

trask:
> should we update these to AAD branch too? switching between main and aad will cause otel-fork to be out of sync.

👍 i'll merge `main` to `aad` after merging this

PR #1744, opened 2021-06-18
Disk Persistence

heyams:
default offline storage folder: 
windows: 
C:\Users\{USERNAME}\AppData\Local\Temp\applicationinsights

linux:
/var/temp/applicationinsights

filename is in this format:
transmission-{System.currentMilliSeconds}-{some-hash-appended-by-createTempFile}.trn

Write raw byte[] to disk when network fails. 
Each write will create a temp file, and after the write is completed, rename it to have the permanent extension ".trn", and then add it to the concurrent linked queue.
Resend persisted telemetries from disk every 30 seconds if there is any in the persisted files queue, which is thread safe.
Persisted file queue will always load the oldest file first and newest file last. 
When sending the persisted telemetries fails, it will get written to disk again and repeat the process again.

In the case of multiple instances running on the same machine, when writing to disk happens roughly at the same time, unique filename will avoid diff processes creating the same file at the same time.  That's why adding an extra component -UUID to the filename to keep its uniqueness. 

Offline storage have the max capacity of 50MB.  Drop telemetries when there is no more space. Warn customers in this case. 
// should we add a config to let customers increase the max file size?

writing and loading tests can't share the same source file even though they're identical.  race condition will fail one test over the other. 

Updated: 
- ~~use the queue as cache.. when the app crashes, data stored in the queue will get lost.. in that case, reload everything from the disk and re-init the cache.~~
-~~create subfolder per process id. subfolder name is the sha256Hash of "username@processId".  in the case of multiple apps, single folder is hard to track the capacity per app.  50MB capacity is default per application.~~

removed subfolder implementation since java doesn't have a way to uniquely identify each app and in the multiple processes environment, all apps will write to the same folder and share the total capacity of 50MB.


core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
```diff
@@ -82,18 +93,20 @@ CompletableResultCode send(List<TelemetryItem> telemetryItems) {
...
+                    gzipOut.write('\n');
                 }
             }
         } catch (IOException e) {
             byteBufferPool.offer(out.getByteBuffers());
             throw e;
         }
-        out.close(); // closing ByteBufferOutputStream is a no-op, but this line makes LGTM happy
+
+        out.close();
```
trask, line 109: any reason to remove the comment?

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static com.microsoft.applicationinsights.internal.persistence.PersistenceHelper.TEMPORARY_FILE_EXTENSION;
+
+/**
+ * This class manages loading a list of {@link ByteBuffer} from the disk.
+ */
+public class LocalFileLoader {
+
+    private static final Logger logger = LoggerFactory.getLogger(LocalFileLoader.class);
+    private static final long INTERVAL = TimeUnit.SECONDS.toSeconds(30); // send persisted telemetries from local disk every 30 seconds.
```
trask, line 28: I think it was mentioned that other SDKs don't do this, and instead check disk after each successful send?

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor(ThreadPoolUtils.createDaemonThreadFactory(LocalFileLoader.class));
+    private static TelemetryChannel telemetryChannel;
+    private static final LocalFileLoader INSTANCE = new LocalFileLoader();
+
+    /**
+     * Track a list of active filenames persisted on disk.
+     * FIFO (First-In-First-Out) read will avoid an additional sorting at every read.
+     */
+    private static final Queue<String> PERSISTED_FILES_QUEUE = new ConcurrentLinkedDeque<>();
```
trask, line 37: typically just constants use UPPER_CASE names

(see https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names for the definition of "but what is a constant?" that otel follows)

```suggestion
    private static final Queue<String> persistedFilesQueue = new ConcurrentLinkedDeque<>();
```

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * This class manages loading a list of {@link ByteBuffer} from the disk.
+ */
+public class LocalFileLoader {
+
+    private static final Logger logger = LoggerFactory.getLogger(LocalFileLoader.class);
+    private static final long INTERVAL = TimeUnit.SECONDS.toSeconds(30); // send persisted telemetries from local disk every 30 seconds.
+    private static final ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor(ThreadPoolUtils.createDaemonThreadFactory(LocalFileLoader.class));
+    private static TelemetryChannel telemetryChannel;
```
trask, line 30: do you mean for this to be instance field?

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/PersistenceHelper.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final long MAX_FILE_SIZE_IN_BYTES = 52428800; // 50MB
+    private static final String PREFIX = "transmission";
+    static final String PERMANENT_FILE_EXTENSION = ".trn";
+    static final String TEMPORARY_FILE_EXTENSION = ".tmp";
+
+    /**
+     * Windows: C:\Users\{USER_NAME}\AppData\Local\Temp\applicationinsights
+     * Linux: /var/temp/applicationinsights
+     */
+    static final File DEFAULT_FOlDER = new File(LocalFileSystemUtils.getTempDir(), "applicationinsights");
```
trask, line 28:
```suggestion
    static final File DEFAULT_FOLDER = new File(LocalFileSystemUtils.getTempDir(), "applicationinsights");
```

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/PersistenceHelper.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Collection;
+import java.util.UUID;
+
+final class PersistenceHelper {
+
+    private static final Logger logger = LoggerFactory.getLogger(PersistenceHelper.class);
+
+    private static final long MAX_FILE_SIZE_IN_BYTES = 52428800; // 50MB
```
trask, line 19: is this what max size used to be? for some reason i recall 10mb

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/PersistenceHelper.java
```diff
@@ -0,0 +1,89 @@
...
+     * Windows: C:\Users\{USER_NAME}\AppData\Local\Temp\applicationinsights
+     * Linux: /var/temp/applicationinsights
+     */
+    static final File DEFAULT_FOlDER = new File(LocalFileSystemUtils.getTempDir(), "applicationinsights");
+
+    static File createTempFileWithUniqueName() {
+        File file = null;
+        try {
+            String prefix = PREFIX + "-" + System.currentTimeMillis() + "-" + UUID.randomUUID().toString().replaceAll("-", "");
+            file = File.createTempFile(prefix, null, DEFAULT_FOlDER);
```
trask, line 34: I don't think you need UUID here, because `createTempFile` will guarantee uniqueness for you

core/src/test/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoaderTests.java
```diff
@@ -0,0 +1,303 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+
+import static com.microsoft.applicationinsights.internal.persistence.PersistenceHelper.DEFAULT_FOlDER;
+import static org.junit.Assert.*;
+
+public class LocalFileLoaderTests {
```
trask, line 18: for the tests, what about just verify reading/writing a random file? i don't think LocalFileLoader and LocalFileWriter even care what the contents of the bytes/file are?

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
heyams: yeah, but raj proposed this is better.  this is how he implemented for otel-dotnet exporter.  we can talk about it. when there is no offline storage, it does nothing.

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/PersistenceHelper.java
heyams: this is what we have in main branch: 
 public final static int DEFAULT_CAPACITY_MEGABYTES = 10;
 private final static int MAX_CAPACITY_MEGABYTES = 1000; 
 private final static int MIN_CAPACITY_MEGABYTES = 1;

dotnet sdk is 50MB and same for otel-dotnet.

trask: ok, so our default was 10mb, but makes sense to me to align with the dotnet default, let's check default with Python and Node also

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
heyams: it's using the factory pattern.  not a singleton.

core/src/test/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoaderTests.java
heyams: added a test for random content.

core/src/main/java/com/microsoft/applicationinsights/internal/persistence/LocalFileLoader.java
```diff
@@ -0,0 +1,166 @@
...
+            persistedFilesCache.addAll(files.stream().map(File::getName).collect(Collectors.toList()));
+        }
+
+        String fileToBeLoaded = persistedFilesCache.poll();
+
+        return fileToBeLoaded != null ? fileToBeLoaded : null;
+
+    }
+
+    List<File> sortPersistedFiles(Collection<File> files) {
```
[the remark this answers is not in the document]
heyams: this is no longer used.. deleted. yup, lastmodified works too.. milliseconds in filename also works.. millisecs is required in filename in order to make it unique.

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
```diff
@@ -116,25 +144,59 @@ private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
...
+                    }
                     result.succeed();
                 });
         return result;
     }
+
+    private void parseResponseCode(int statusCode) {
+        switch(statusCode) {
+            case HttpStatus.SC_UNAUTHORIZED:
+            case HttpStatus.SC_FORBIDDEN: {
```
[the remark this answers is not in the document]
heyams: main branch doesn't include this as retry though.. need to check with other sdks.

core/src/main/java/com/microsoft/applicationinsights/TelemetryChannel.java
```diff
@@ -116,25 +144,59 @@ private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
         // TODO(trask)
         //  not setting User-Agent header at all would be a better option, but haven't figured out how to do that yet
         request.setHeader("User-Agent", "");
+        request.setHeader("Content-Encoding", "gzip");
 
         // TODO(trask) subscribe with listener
         //  * retry on first failure (may not need to worry about this if retry policy in pipeline already, see above)
         //  * write to disk on second failure
         CompletableResultCode result = new CompletableResultCode();
+        List<ByteBuffer> finalByteBuffers = byteBuffers;
```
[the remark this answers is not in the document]
heyams: it should be final.. lambda expression requires a final temp var.

core/src/main/java/com/microsoft/applicationinsights/ByteBufferOutputStream.java
```diff
@@ -8,13 +8,13 @@
 
 class ByteBufferOutputStream extends OutputStream {
 
-    private final ByteBufferPool byteBufferPool;
+    private final AppInsightsByteBufferPool byteBufferPool;
 
     private final List<ByteBuffer> byteBuffers = new ArrayList<>();
```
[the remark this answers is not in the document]
heyams: remove is handled by AppInsightsByteBufferPool's queue.  linkedList works better if there is more remove.. in this case, no.

PR #1750, opened 2021-06-22
Add disk persistence integration test 

heyams:


core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -205,8 +207,10 @@ public BatchSpanProcessor getChannelBatcher() {
         if (channelBatcher == null) {
             synchronized (channelInitLock) {
                 if (channelBatcher == null) {
-                    TelemetryChannel channel = TelemetryChannel.create(endpointProvider.getIngestionEndpoint());
-                    LocalFileLoader.init(channel);
+                    LocalFileCache localFileCache = new LocalFileCache();
```
heyams, line 210: will this cause to create new queue for each request? that's not intended, right?

trask: this only happens once (during initialization under the `channelInitLock`)

PR #1753, opened 2021-06-22
Track AAD enablement via Statsbeat

heyams:
Fix [#10122256](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/AEP/Cobalt/2106?workitem=10122256).

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
@@ -204,7 +204,7 @@ public class AiComponentInstaller implements AgentListener {
         }
 
         // initialize StatsbeatModule
-        StatsbeatModule.get().start(telemetryClient, config.internal.statsbeat.intervalSeconds, config.internal.statsbeat.featureIntervalSeconds);
+        StatsbeatModule.get().start(telemetryClient, config.internal.statsbeat.intervalSeconds, config.internal.statsbeat.featureIntervalSeconds, config.preview.authentication.enabled);
     }
 
     private static GcEventMonitor.GcEventMonitorConfiguration formGcEventMonitorConfiguration(Configuration.GcEventConfiguration gcEvents) {
--- a/core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/Feature.java
+++ b/core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/Feature.java
@@ -34,7 +34,7 @@ enum Feature {
     JAVA_VENDOR_ADOPT_OPENJDK(3),
     JAVA_VENDOR_REDHAT(4),
     JAVA_VENDOR_OTHER(5),
-    AAD_ON(6);
+    AAD(6);
 
     private static final Map<String, Feature> features;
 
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatModule.java
```diff
@@ -41,8 +41,6 @@
 
     private final NetworkStatsbeat networkStatsbeat;
     private final AttachStatsbeat attachStatsbeat;
-    // keeping this as field for now
-    @SuppressWarnings("unused")
```
trask, line 45: 👍

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatTestUtils.java
```diff
@@ -78,6 +78,8 @@
         FEATURE_MAP_DECODING.put(3, Feature.JAVA_VENDOR_ADOPT_OPENJDK);
         FEATURE_MAP_DECODING.put(4, Feature.JAVA_VENDOR_REDHAT);
         FEATURE_MAP_DECODING.put(5, Feature.JAVA_VENDOR_OTHER);
+        FEATURE_MAP_DECODING.put(6, Feature.AAD_ON);
+        FEATURE_MAP_DECODING.put(7, Feature.AAD_OFF);
```
trask, line 82: do we need two features for this, or can we just track "ON"?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
```diff
@@ -204,7 +204,7 @@ private static void start(Instrumentation instrumentation) {
         }
 
         // initialize StatsbeatModule
-        StatsbeatModule.get().start(telemetryClient, config.internal.statsbeat.intervalSeconds, config.internal.statsbeat.featureIntervalSeconds);
+        StatsbeatModule.get().start(telemetryClient, config.internal.statsbeat.intervalSeconds, config.internal.statsbeat.featureIntervalSeconds, config.preview.authentication.enabled);
```
trask, line 207: I think this should check `config.preview.authentication.enabled && config.preview.authentication.type != null`

@kryalama can you remind me why we need `enabled` instead of just checking if `type` is present?

core/src/test/java/com/microsoft/applicationinsights/internal/statsbeat/StatsbeatTestUtils.java
heyams: we need both.. similar to java vendor. we need to track all the possible vendors. Key is static.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/AiComponentInstaller.java
heyams: why check both? @kryalama?

kryalama: Either config.preview.authentication.enabled or config.preview.authentication.type != null  should be enough. If  config.preview.authentication.enabled is true then the agent expects a valid authentication type. We have the enabled flag to be consistent with other configuration types and to skip config validation if the enabled flag is false.

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/Feature.java
```diff
@@ -33,17 +33,19 @@
     JAVA_VENDOR_MICROSOFT(2),
     JAVA_VENDOR_ADOPT_OPENJDK(3),
     JAVA_VENDOR_REDHAT(4),
-    JAVA_VENDOR_OTHER(5);
+    JAVA_VENDOR_OTHER(5),
+    AAD_ON(6);
```
trask, line 37: let's call this the `AAD` feature, and it's either on `1` or off `0`
```suggestion
    AAD(6);
```

trask (approved):

kryalama (approved):

PR #1754, opened 2021-06-23
More FIXMEs

trask:


core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -50,13 +50,13 @@ public DefaultQuickPulseDataFetcher(ArrayBlockingQueue<HttpRequest> sendQueue, T
         sdkVersion = getCurrentSdkVersion();
         StringBuilder sb = new StringBuilder();
 
-        // FIXME (trask) what about azure functions consumption plan where role name not available yet?
+        // FIXME (heya) what about azure functions consumption plan where role name not available yet?
```
heyams, line 53: it will get lazily set later. It's already handled in LazyConfigurationAccessor.setWebsiteSiteName.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -215,27 +213,74 @@ private void addRequest(RequestData requestTelemetry) {
...
-        return Duration.ofDays(days)
-                .plusHours(hours)
-                .plusMinutes(minutes)
-                .plusSeconds(seconds)
-                .plusNanos(microseconds * 1000L)
-                .toMillis();
+        return startingAtDaysOrHours(duration);
+    }
+
+    private static long startingAtDaysOrHours(String duration) {
```
heyams, line 229: where is the doc for this? is breeze so picky about duration? this is new to me :)

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
```diff
@@ -50,13 +50,13 @@ public DefaultQuickPulseDataFetcher(ArrayBlockingQueue<HttpRequest> sendQueue, T
...
         String roleName = telemetryClient.getRoleName();
 
         sb.append("[{");
         sb.append("\"Documents\":[],");
         sb.append("\"Instance\":\"").append(instanceName).append("\",");
-        // FIXME (trask) this seemed to be working when it was always null ikey here??
+        // TODO (trask) instrumentation key is also part of query string, is it needed in both places?
```
heyams, line 59: it always sends null instrumentationKey in the live metric payload. 
ikey is only used in the uri.

core/src/main/java/com/microsoft/applicationinsights/FormattedDuration.java
```diff
@@ -84,36 +84,36 @@ private static void appendDaysHoursMinutesSeconds(StringBuilder sb, long days, l
 
     private static void appendMinTwoDigits(StringBuilder sb, long value) {
         if (value < 10) {
-            sb.append("0");
+            sb.append('0');
```
heyams, line 87: you can also append int directly.  wonder which is faster.

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/QuickPulseDataCollector.java
trask: just reverse engineered from 3.1.1 behavior

core/src/main/java/com/microsoft/applicationinsights/FormattedDuration.java
trask: this is to left pad with zero, e.g. `04` instead of `4`

core/src/main/java/com/microsoft/applicationinsights/internal/quickpulse/DefaultQuickPulseDataFetcher.java
trask: the prefix in this class doesn't get updated at that time tho

kryalama (approved):

PR #1761, opened 2021-06-24
Format round two + checkstyle

trask:
Missed smoke tests for spotless.

Added `.editorconfig` from OTel.

And ran Intellij auto-format on non-java files.

And applied checkstyle rules.


heyams:
@trask can we update README with all this formatting change? is .editorconfig controlled by spotless?

heyams (approved):

trask:
> @trask can we update README with all this formatting change? is .editorconfig controlled by spotless?

👍

PR #1764, opened 2021-06-24
Track more features via Statsbeat

heyams:
default is enabled.. 
track off instead. 

will add other features after this PR.. 
there is no way to pass config (agent-tooling) to core. 
it's not ideal to pass so many arguments to a method.. but i don't want to add dependency. 

~~spotlessApply doesn't apply to test code.~~ look like it works :) @trask


trask:
> there is no way to pass config (agent-tooling) to core.

working on this right now 😁

core/src/main/java/com/microsoft/applicationinsights/internal/statsbeat/Feature.java
```diff
@@ -33,7 +33,15 @@
...
-  AAD(6);
+  AAD(6),
+  Cassandra_OFF(7),
+  JDBC_OFF(8),
+  JMS_OFF(9),
+  KAFKA_OFF(10),
+  MICROMETER_OFF(11),
+  MONGO_OFF(12),
+  REDIS_OFF(13),
+  SPRING_SCHEDULING_OFF(14);
```
trask, line 44: instead of `_OFF` how about `_DISABLED`?

trask (approved):

kryalama (approved):

PR #1767, opened 2021-06-25
Fix legacy propagation interop

trask:


heyams (approved):

heyams:
caw we add a todo or create a work item for legacy sdk smoke test?

trask:
> caw we add a todo or create a work item for legacy sdk smoke test?

done: https://msazure.visualstudio.com/One/_workitems/edit/10237362

PR #1768, opened 2021-06-25
Release 3.2.0-BETA.2

heyams:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+# Version 3.2.0-BETA.2
+* There were lots of refactorings.
```
heyams, line 4: @trask we might need to reword these in the github release notes.

trask: we only need to put user facing changes in the change log
```suggestion
* Lots of internal clean up.
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+* Enable profiler by default.
```
trask, line 5: the profiler is still proof-of-concept (and undocumented), so let's not mention it (also, it's not enabled by default, it's only included by default)
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+* Reduce binary size.
```
trask, line 6: just to clarify that it's not reduced from 3.1.1
```suggestion
* Reduce binary size from 3.2.0-BETA
```

trask (approved):

PR #1772, opened 2021-06-27
Populate operation name on dependencies

trask:


kryalama (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -388,15 +391,23 @@ private void trackTraceAsException(SpanData span, String errorStack) {
   }
 
   private static void setOperationTags(TelemetryItem telemetry, SpanData span) {
-    setOperationTags(telemetry, span.getTraceId(), span.getParentSpanContext().getSpanId());
+    setOperationTags(
```
heyams, line 394: can we simply pass span object? looks like it will require signature change again when something new is required from the span.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/TempGetAttribute.java
```diff
@@ -0,0 +1,76 @@
...
+
+      Field lockField = recordEventsReadableSpanClass.getDeclaredField("lock");
+      lockField.setAccessible(true);
+      Field attributesField = recordEventsReadableSpanClass.getDeclaredField("attributes");
+      attributesField.setAccessible(true);
+
+      lockMethodHandle = lookup.unreflectGetter(lockField);
+      attributesMapMethodHandle = lookup.unreflectGetter(attributesField);
+    } catch (Exception e) {
+      e.printStackTrace();
```
heyams, line 57: can you use logger instead?

trask: 👍 #1857

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: the more general signature is needed from a couple of places in this class

PR #1778, opened 2021-06-30
Shut down azure metadata service when network is not reachable.

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wascore/statsbeat/AzureMetadataService.java
```diff
@@ -94,31 +93,43 @@ void parseJsonResponse(String response) throws IOException {
...
-      // When it's not VM/VMSS, server does not return json back, and instead it returns text like
-      // the following:
-      // "<br />Error: NetworkUnreachable (0x2743). <br />System.Net.Sockets.SocketException A
-      // socket operation was attempted to an unreachable network 169.254.169.254:80".
-      logger.debug(
-          "This is not running from an Azure VM or VMSS. Shut down AzureMetadataService scheduler.");
-      scheduledExecutor.shutdown();
+      response = LazyHttpClient.getInstance().send(request).block();
     } catch (Exception ex) {
       if (ex.getCause()
```
heyams, line 100: with this change, i think we don't need to know the cause.. because it will get shutdown by any exception assuming caused by the endpoint?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/wascore/statsbeat/AzureMetadataService.java
```diff
@@ -94,24 +93,43 @@ void parseJsonResponse(String response) throws IOException {
...
     try {
-      HttpResponse response = LazyHttpClient.getInstance().send(request).block();
-      if (response == null) {
-        // this shouldn't happen, the mono should complete with a response or a failure
-        throw new AssertionError("http response mono returned empty");
+      response = LazyHttpClient.getInstance().send(request).block();
+    } catch (Exception ex) {
+      if (ex.getCause()
+          .toString()
+          .contains("Network is unreachable: no further information: /169.254.169.254:80")) {
```
kryalama, line 102: Is there any other way we can identify that the exception is due to not running on vm or vmss?

heyams: socket exception. this end point doesn't go out to the internet because it's a local linked ip.. non routable.. we are good to assume any exceptions here are likely caused by the azure metadata service.

trask (approved):

PR #1769, opened 2021-06-26
Fix failing pipeline

heyams:

PR #1781, opened 2021-06-30
Fix line endings

trask:
I think this is why diffs in #1769 were weird.

This PR changes files from CRLF to LF, except for `.bat` and `.cmd` files.


heyams (approved):
what did you do to fix it? IntelliJ editor config?

trask:
> what did you do to fix it? IntelliJ editor config?

`dos2unix.exe` 😁

PR #1786, opened 2021-07-02
Fix some todo

heyams:
- send statsbeat in a different batch because its endpoint is not the same as the customer iKey.
- persist statsbeat on disk under "statsbeat" subfolder
- persist regular telemetries on disk under "telemetries" subfolder
- address other todo(s).

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
@@ -36,6 +36,7 @@ import com.azure.monitor.opentelemetry.exporter.implementation.models.RequestDat
 import com.azure.monitor.opentelemetry.exporter.implementation.models.TelemetryEventData;
 import com.azure.monitor.opentelemetry.exporter.implementation.models.TelemetryExceptionData;
 import com.azure.monitor.opentelemetry.exporter.implementation.models.TelemetryItem;
+import com.microsoft.applicationinsights.agent.internal.common.LocalFileSystemUtils;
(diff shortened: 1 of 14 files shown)


agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoaderTests.java
```diff
@@ -165,17 +175,16 @@ public void testWriteGzipRawByte() throws IOException {
...
     writer.writeToDisk(singletonList(ByteBuffer.wrap(result)));
 
     // read gzipped byte[] from disk
-    LocalFileLoader loader = new LocalFileLoader(cache);
+    LocalFileLoader loader = new LocalFileLoader(cache, false);
     byte[] bytes = readTelemetriesFromDiskToBytes(loader);
 
     // ungzip
     ByteArrayInputStream inputStream = new ByteArrayInputStream(result);
-    // TODO (heya) does ungzip need to be larger?
```
heyams, line 177: gzip byte[] is always bigger than the origin byte[].  it works previously was because the text was very small.  gzip happens to be equal to the ungzip.. when the text becomes longer and longer, the gzip byte size can be triple of the original one.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -58,7 +58,6 @@
 
 public class TelemetryClient {
 
-  // TODO (heya) can you confirm these are the same as used in 3.1.1?
```
heyams, line 61: yeah, they're the same.

trask (approved):

PR #1787, opened 2021-07-02
otel fork updated to latest

trask:
~DRAFT, do not merge until 1.4.0 is published~

Need some of these updates on top of 1.3.1, and will update to 1.4.0 before 3.2.0 GA.


heyams:
latest is 1.3.1, correct?

PR #1790, opened 2021-07-03
Add thread name to log capture

### LINKED ISSUE #1699: Capture thread name for instrumented logs.
When logging using logback, log4j2, JUL, the thread name can be included in the output to logfiles (using appender formats etc).

I cannot see any way to capture thread name for the logs sent by the agent to Application Insights.
Can the thread name be included and appear for instance under custom dimensions? It is really important to be able to search and display log messages for a specific thread name.

trask:
Resolves #1699

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
@@ -348,12 +348,13 @@ public class Exporter implements SpanExporter {
     // set message-specific properties
     String level = attributes.get(AI_LOG_LEVEL_KEY);
     String loggerName = attributes.get(AI_LOGGER_NAME_KEY);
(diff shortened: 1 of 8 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -407,6 +410,9 @@ private static void setLoggerProperties(MonitorDomain data, String level, String
     if (loggerName != null) {
       TelemetryUtil.getProperties(data).put("LoggerName", loggerName);
     }
+    if (threadName != null) {
+      TelemetryUtil.getProperties(data).put("ThreadName", loggerName);
```
kryalama, line 414: should be threadName

trask: 😱 👍

kryalama (approved):

PR #1796, opened 2021-07-08
Add data truncation and sanitization to telemetry data

kryalama:
Add data truncation and sanitization to telemetry data


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,136 @@
...
+  public static final int MAX_AVAILABILITY_MESSAGE_LENGTH = 8192;
+  public static final int MAX_METRIC_NAME_SPACE_LENGTH = 256;
+  public static final int MAX_METHOD_NAME_LENGTH = 1024;
+  public static final int MAX_ASSEMBLY_NAME_LENGTH = 1024;
+  public static final int MAX_FILE_NAME_LENGTH = 1024;
+  public static final int MAX_SEQUENCE_LENGTH = 64;
+
+  /** Function to sanitize both key and value in properties. */
+  public static void sanitizeProperties(Map<String, String> properties) {
+    if (properties != null) {
```
trask, line 54: a little more conventional in Java to short-circuit like this (vs having a single return statement)
```suggestion
    if (properties == null) {
      return;
    }
    ...
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,136 @@
...
+        measurements.remove(entry.getKey());
+        String uniqueKey = makeKeyUnique(entry.getValue().getKey(), measurements);
+        measurements.put(uniqueKey, entry.getValue().getValue());
+      }
+    }
+  }
+
+  private SanitizationHelper() {}
+
+  static class Pair {
```
trask, line 128: I think this class would be more readable as a real class named `SanitizedEntry`, with methods getKey() and getValue()

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,136 @@
...
+          String uniqueKey = makeKeyUnique(entry.getValue().getKey(), properties);
+          properties.put(uniqueKey, entry.getValue().getValue());
+        }
+      }
+    }
+  }
+
+  /** Function to create unique key. */
+  private static String makeKeyUnique(String key, Map<String, ?> map) {
+    if (map.containsKey(key)) {
```
trask, line 78: similar here
```suggestion
    if (!map.containsKey(key)) {
      return key;
    }
    ...
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelperTests.java
```diff
@@ -0,0 +1,168 @@
...
+    String longValue = sb.toString();
+    properties.put("key1", longValue);
+    properties.put("key2", "value2");
+    SanitizationHelper.sanitizeProperties(properties);
+    assertThat(properties.get("key1").length()).isEqualTo(SanitizationHelper.MAX_VALUE_LENGTH);
+    assertThat(properties.get("key2")).isEqualTo("value2");
+  }
+
+  @Test
+  // TODO check with team since this test is tightly coupled to implementation details
```
trask, line 146: seems fine to me
```suggestion
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,136 @@
...
+  public static final int MAX_URL_LENGTH = 2048;
+  public static final int MAX_DATA_LENGTH = 8192;
+  public static final int MAX_TARGET_NAME_LENGTH = 1024;
+  public static final int MAX_RUN_LOCATION_LENGTH = 1024;
+  public static final int MAX_AVAILABILITY_MESSAGE_LENGTH = 8192;
+  public static final int MAX_METRIC_NAME_SPACE_LENGTH = 256;
+  public static final int MAX_METHOD_NAME_LENGTH = 1024;
+  public static final int MAX_ASSEMBLY_NAME_LENGTH = 1024;
+  public static final int MAX_FILE_NAME_LENGTH = 1024;
+  public static final int MAX_SEQUENCE_LENGTH = 64;
```
trask, line 50: let's move these constants to the files where they are used, at least for constants that are only used in a single file

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/Strings.java
```diff
@@ -49,4 +49,19 @@ public static boolean isNullOrEmpty(@Nullable String string) {
   }
 
   private Strings() {}
+
+  public static String trimAndTruncate(String str, int maxLength) {
```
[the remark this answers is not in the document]
kryalama: followed the same implementation as of .Net to remove extra spaces. You meant to say the input will never have extra spaces?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelperTests.java
```diff
@@ -0,0 +1,168 @@
...
+  public void testValidPropertiesData2() {
+    Map<String, String> properties = new HashMap<>();
+    properties.put("attribute1", "testValue");
+    properties.put("attribute2", "testValue2");
+    properties.put("sensitiveAttribute1", "sensitiveData1");
+    SanitizationHelper.sanitizeProperties(properties);
+    assertThat(properties.get("attribute1")).isEqualTo("testValue");
+    assertThat(properties.get("attribute2")).isEqualTo("testValue2");
+    assertThat(properties.get("sensitiveAttribute1")).isEqualTo("sensitiveData1");
+  }
```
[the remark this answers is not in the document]
kryalama: my bad, forgot to remove this while debugging smoke tests

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,143 @@
...
+  public static final int MAX_MESSAGE_LENGTH = 32768;
+  public static final int MAX_URL_LENGTH = 2048;
+
+  /** Function to sanitize both key and value in properties. */
+  @SuppressWarnings("ReturnsNullCollection")
+  public static Map<String, String> sanitizeProperties(Map<String, String> properties) {
+    if (properties == null) {
+      return null;
+    }
+    if (!needsSanitizing(properties)) {
```
heyams, line 42: is it better to avoid this iteration? line 47 iteration, we can check if the size has been exceeded, if yes, sanitize.. otherwise go to next element in the list.. i think this is not necessary.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,143 @@
...
+  /** Function to create unique key. */
+  private static String makeKeyUnique(String key, Map<String, ?> map) {
+    if (!map.containsKey(key)) {
+      return key;
+    }
+    String truncatedKey = Strings.truncate(key, MAX_KEY_LENGTH - 3);
+    int candidate = 1;
+    do {
+      key = truncatedKey + candidate;
+      ++candidate;
```
heyams, line 81: what if candidate go to 3+ digit.. that will exceed the limit?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,143 @@
...
+  /** Function to sanitize key. */
+  private static String sanitizeKey(String key) {
+    String sanitizedKey = Strings.trimAndTruncate(key, MAX_KEY_LENGTH);
+    return makeKeyNonEmpty(sanitizedKey);
+  }
+
+  /** Function to return non empty key. */
+  private static String makeKeyNonEmpty(String sanitizedKey) {
+    // TODO check if this is valid. Below code is based on .Net implementation of the same
+    return Strings.isNullOrEmpty(sanitizedKey) ? "empty" : sanitizedKey;
```
heyams, line 100: if this happens many times, empty1, empty2, empty3.... empty34 could end up in the payload.. if the key is empty, it should not get added.. sanitizedKey should not get called in that case.  I wouldn't simply copy logic from .net sdk.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,143 @@
...
+    for (Map.Entry<String, SanitizedEntry<Double>> entry : sanitizedMeasurements.entrySet()) {
+      measurements.remove(entry.getKey());
+      String uniqueKey = makeKeyUnique(entry.getValue().getKey(), measurements);
+      measurements.put(uniqueKey, entry.getValue().getValue());
+    }
+  }
+
+  private SanitizationHelper() {}
+
+  static class SanitizedEntry<V> {
```
heyams, line 126: this is only applied to measurements, not other entry... SanitizedPair?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelperTests.java
```diff
@@ -0,0 +1,189 @@
...
+    Map<String, Double> measurements = new HashMap<>();
+    measurements.put("", 1.0);
+    measurements.put("key2", 2.0);
+    SanitizationHelper.sanitizeMeasurements(measurements);
+    assertThat(measurements).doesNotContainKey("");
+    assertThat(measurements).containsEntry("empty", 1.0);
+    assertThat(measurements).containsEntry("key2", 2.0);
+  }
+
+  @Test
```
heyams, line 88: please add a test for makeUniqueKey.  generate 100 same keys, and then see the result.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
kryalama: as Trask mention this optimization is to avoid extra memory allocation when properties are valid

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
kryalama: there can be only one empty key in the properties hashMap

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
kryalama: removed this completely in the next commit

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
heyams: know hashMap won't allow duplicate keys.. my suggestion is not to add any key-value pair when the key is empty rather than use "empty" as the key in the hashmap.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
heyams: when it's invalid, two iterations.. i am saying for both valid and invalid cases, one iteration is sufficient.

trask: can you show us what you are thinking? I did not see a way to do it in one iteration without memory allocations in the normal case

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
trask: @kryalama is this behavior copied from .NET SDK?

kryalama: yes this is copied from .Net impelementation

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
kryalama: increased the truncation length to 6, which will generate 1million unique keys for a given duplicate key. I think that should be good for this usecase :)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
trask: ok, let's keep it as is then

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/utils/SanitizationHelper.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /** Function to sanitize value. */
+  private static String sanitizeValue(String value) {
+    return Strings.trimAndTruncate(value, MAX_VALUE_LENGTH);
+  }
+
+  /** Function to sanitize key. */
+  private static String sanitizeKey(String key) {
+    String sanitizedKey = Strings.trimAndTruncate(key, MAX_KEY_LENGTH);
```
heyams, line 104: we can check length before trim and truncate.. that will avoid an additional iteration, right?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/Strings.java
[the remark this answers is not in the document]
trask: i'm not convinced that trimming is needed, i'm not sure that's our concern really, we just transport along telemetry, unless breeze has some issue and will reject it if it's not trimmed

trask:
let's wait for https://github.com/microsoft/common-schema/pull/254 to be approved/merged and then sync with that

trask:
> let's wait for [microsoft/common-schema#254](https://github.com/microsoft/common-schema/pull/254) to be approved/merged and then sync with that

this is merged now 👍

trask:
@kryalama the new mapping unlocked some simplifications that we can make to the code, I pushed partial commit to show, can you apply similar to measurements?

trask (approved):
thanks 👍

PR #1799, opened 2021-07-14
Support W3C baggage propagation

trask:


kryalama (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/legacyheaders/DelegatingPropagator.java
```diff
@@ -47,15 +48,16 @@ public static DelegatingPropagator getInstance() {
...
 
-    // important that W3CTraceContextPropagator is last, so it will take precedence if both sets of
-    // headers are present
+    // important that W3CTraceContextPropagator is after AiLegacyPropagator, so it will take
+    // precedence if both sets of headers are present
     delegate =
         TextMapPropagator.composite(
-            AiLegacyPropagator.getInstance(), new ModifiedW3cTraceContextPropagator());
+            AiLegacyPropagator.getInstance(),
+            new ModifiedW3cTraceContextPropagator(),
```
heyams, line 59: just want to have a better understanding about propagator for tracecontext:
what does the difference make between ModifiedW3cTraceContextPropagator vs. W3CBaggagePropagator? 
will they override traceparent headers? looks like W3CBaggagePropagator will have the highest precedence?
Can you also share different set of headers per propagator? links will be helpful. 

It's merely for my own curiosity :)

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/legacyheaders/DelegatingPropagator.java
trask: ModifiedW3cTraceContextPropagator - https://www.w3.org/TR/trace-context/ - this propagates is our normal distributed tracing header (traceparent), plus the tracestate header which we aren't using currently but probably will have some use for in the future
W3CBaggagePropagator - https://www.w3.org/TR/baggage/ - this allows applications to propagate their own stuff "baggage" to their downstream services

PR #1801, opened 2021-07-16
Update App Services internal diagnostics comments

trask:
TODO fix INFO log not showing up in app services extension log

```
LoggerFactory.getLogger(DiagnosticsHelper.DIAGNOSTICS_LOGGER_NAME)
                    .info("Application Insights Codeless Agent {} Attach Successful", version);
```


heyams:
@trask is this still WIP? 
 > TODO fix INFO log not showing up in app services extension log

trask:
> @trask is this still WIP?

just updated, ready now

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/LoggingConfigurator.java
```diff
@@ -96,7 +96,19 @@ private void configureAppSvcs() {
...
         System.getenv(DiagnosticsHelper.APPLICATIONINSIGHTS_DIAGNOSTICS_OUTPUT_DIRECTORY);
     if (diagnosticsOutputDirectory != null && !diagnosticsOutputDirectory.isEmpty()) {
-      rootLogger.addAppender(configureDiagnosticAppender(diagnosticsOutputDirectory));
+      Appender<ILoggingEvent> diagnosticAppender =
+          configureDiagnosticAppender(diagnosticsOutputDirectory);
+
+      // applicationinsights.extension.diagnostics logging should go to extension diagnostic log,
+      // but should not go to normal user-facing log
+      Logger diagnosticLogger = loggerContext.getLogger(DiagnosticsHelper.DIAGNOSTICS_LOGGER_NAME);
+      diagnosticLogger.setLevel(Level.INFO);
```
heyams, line 105: does this imply that it will always be INFO level for app services diagnostic logs? APPLICATIONINSIGHTS_SELF_DIAGNOSTICS_LEVEL is for user-facing logs only? can we make it more clear on the azure doc?

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/LoggingConfigurator.java
trask: > does this imply that it will always be INFO level for app services diagnostic logs? 

yes

> APPLICATIONINSIGHTS_SELF_DIAGNOSTICS_LEVEL is for user-facing logs only?

yes

> can we make it more clear on the azure doc?

extension log is not user-facing, it is just for getting the logs over to our kusto table, so I don't think we need to mention it in azure doc

PR #1803, opened 2021-07-18
Fix azure sdk statsbeat

trask:


heyams (approved):

heyams:
how do i test it to make sure that's the instrumentation name they used for Azure SDK? does Azure SDK need to make a new release for this?

trask:
> how do i test it to make sure that's the instrumentation name they used for Azure SDK?

you can use the [AzureSdk smoke test](https://github.com/microsoft/ApplicationInsights-Java/tree/main/test/smoke/testApps/AzureSdk)

> does Azure SDK need to make a new release for this?

this is the instrumentation name that Azure SDK sends currently, but I think they want to track each Azure SDK separately (e.g. to track CosmosDB, Event Hubs, Service Bus telemetry usage separately) which would require a new Azure SDK release (cc: @lmolkova)

lmolkova:
> > how do i test it to make sure that's the instrumentation name they used for Azure SDK?
> 
> you can use the [AzureSdk smoke test](https://github.com/microsoft/ApplicationInsights-Java/tree/main/test/smoke/testApps/AzureSdk)
> 
> > does Azure SDK need to make a new release for this?
> 
> this is the instrumentation name that Azure SDK sends currently, but I think they want to track each Azure SDK separately (e.g. to track CosmosDB, Event Hubs, Service Bus telemetry usage separately) which would require a new Azure SDK release (cc: @lmolkova)

yes, this is correct and this will take us a bit to ship. I'll keep you folks posted!

PR #1804, opened 2021-07-18
Update submodule to latest 

### LINKED ISSUE #1800: HTTP requests telemetry data is not send by ApplicationInsights-Java after upgrading Tomcat version 9 to 10 (containing javax.* to jakarta.* namespace change)
### Expected behavior
We upgraded our Tomcat version from version 9 to 10, under which our services are now running (Apache Tomcat Version 10.0.8).This upgrade contains a change in the javax.* to jakarta.* namespace.
I'd expect following the upgrade (and possible converting the applicationinsights-agent.jar with the migration tool) that our service would keep on sending http request data to ApplicationInsights through use of the java option: -javaagent:D:\java\appinsights\applicationinsights-agent.jar

### Actual behavior
Following this upgrade, the service is no longer communicating the http request telemetry data to ApplicationInsight, where under Tomcat 9 this worked fine. 
Other data is still processed, TRACE and DEPENDENCY transactions for example, are still present in the transaction search in Azure. 
With the Tomcat10 upgrade there was need for some .jar's to be converted with the jakartaee-migration-1.0.0 tool. This in order to make alterations for the namespace change from javax.* to jakarta.*.
We tried running the applicationinsights-agent3.1.1.jar, both with and without conversion through the migration tool, but in all cases it's not working with Tomcat10.
Running the service under Tomcat9 again instantly restores the sending of http requests data to ApplicationInsights.

### To Reproduce
The above mentioned upgrade of the Tomcat10 version on a Tomcat service using the java agent for ApplicationInsights

### System information
 - Apache Tomcat Version 10.0.8
 - AdoptOpenJDK-16.0.1+9
 - SDK Version: 3.1.1
 - OS type and version: Windows server 2019
 - Using spring-boot? No
 
### Logs
Two applicationinsights.log files are attached at DEBUG level.
1. applicationinsightsNOTCONV.log shows the log with the non converted applicationinsights-agent3.1.1.jar
2. applicationinsights.CONV.log shows the log with the converted applicationinsights-agent3.1.1.jar (converted with jakartaee-migration-1.0.0 tool)
 
In both logging examples, the following steps are logged: 
1. service is started
2. a http request is made and processed by the service
3. service is stopped.

[applicationinsightsNOTCONV.log](https://github.com/microsoft/ApplicationInsights-Java/files/6831583/applicationinsightsNOTCONV.log)
[applicationinsights.CONV.log](https://github.com/microsoft/ApplicationInsights-Java/files/6831586/applicationinsights.CONV.log)

In the Tomcat9HttpRequestLines.txt 2 log lines on the http request telemtry (http.url) are shown from when the service while running under Tomcat9. These lines, for the same http request do not appear in the logs when running Tomcat 10.
[Tomcat9HttpRequestLines.txt](https://github.com/microsoft/ApplicationInsights-Java/files/6831587/Tomcat9HttpRequestLines.txt)

trask:
A few things in this PR:

* Updates to latest OTel, supporting up to OTel API 1.4.1
* Moves 3 instrumentations out of preview, and enables them by default
  * Java 11 HTTP client
  * JAX-WS
  * RabbitMQ
* Pulls in new instrumentations (enabled by default)
  * AsyncHttpClient
  * Google HTTP client
  * JAX-RS client
  * Jetty client
  * Spring RabbitMQ
  * Servlet 5.0 (Resolves #1800)

TODO: submit doc PR


heyams:
can we update Statsbeat.instrumentation with this change?

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/Instrumentations.java
```diff
@@ -109,6 +109,12 @@
     INSTRUMENTATION_MAP.put("io.opentelemetry.java-http-client", 58);
     INSTRUMENTATION_MAP.put("io.opentelemetry.rabbitmq-2.7", 59);
     INSTRUMENTATION_MAP.put("io.opentelemetry.spring-integration-4.1", 60);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.servlet-5.0", 61);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.async-http-client-1.9", 62);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.async-http-client-2.0", 63);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.google-http-client-1.19", 64);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.jetty-httpclient-9.2", 65);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.spring-rabbit-1.0", 66);
```
heyams, line 117: need to send second long.. i will do that later.. that doesn't take us long 😭

PR #1806, opened 2021-07-22
Handle rpc.grpc_status_code

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -657,6 +657,9 @@ private void exportRequest(SpanData span) {
     }
 
     Long httpStatusCode = attributes.get(SemanticAttributes.HTTP_STATUS_CODE);
+    if (httpStatusCode == null) {
+      httpStatusCode = attributes.get(SemanticAttributes.RPC_GRPC_STATUS_CODE);
+    }
```
trask, line 662: this is the only significant change

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -242,7 +242,7 @@ private void exportRemoteDependency(SpanData span, boolean inProc) {
 
     // set dependency-specific properties
     data.setId(span.getSpanId());
-    data.setName(getTelemetryName(span));
+    data.setName(span.getName());
```
trask, line 245: It seems to be just "operation name" where we want special AI-specific naming, I think ok to map directly from span name here (cc: @vishweshbankwar)

heyams (approved):

PR #1809, opened 2021-07-22
Handle partial content success due to Stamp specific redirects

### LINKED ISSUE #1674: Add possibility to use multiple custom destinations(instrumentation keys)
When custom client is configured with own instrumentation key it is ignored and handled by Agent client:
```java
    @Value("${appinsights.analytics.instrumentationkey}")
    private String instrumentationKey;

    @PostConstruct
    public void init() {
        TelemetryConfiguration telemetryConfiguration = TelemetryConfiguration.createDefault();
        telemetryConfiguration.setInstrumentationKey(instrumentationKey);

        client = new TelemetryClient(telemetryConfiguration);
    }
```

Agent configuration:
```json
{
  "connectionString": "InstrumentationKey=another-instrumentation-key;IngestionEndpoint=https://eastus2-0.in.applicationinsights.azure.com/",
  "role": {
    "name": "test-service"
  },
  "selfDiagnostics": {
    "destination": "console"
  }
}

```

The destination(instrumentation key) from custom client should be used, not configured globally in the agent.

**Alternatives**
We were trying to tune the client, but then we have realized that BytecodeUtilImpl.java#L220 configures agent client in the global scope.

**Additional context**
The client is set globally in the source code below  https://github.com/microsoft/ApplicationInsights-Java/blob/3.0.2/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/instrumentation/sdk/BytecodeUtilImpl.java#L220

kryalama:
Add ability to handle partial content success due to Stamp specific redirects

Resolves #1674


agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
```diff
@@ -0,0 +1,201 @@
...
+        createMetricTelemetry("metric" + 1, 1, "00000000-0000-0000-0000-0FEEDDADBEEF"));
+
+    // when
+    List<CompletableResultCode> completableResultCodes = telemetryChannel.send(telemetryItems);
+
+    // then
+    for (CompletableResultCode resultCode : completableResultCodes) {
+      resultCode.join(10, TimeUnit.SECONDS);
+      assertThat(resultCode.isSuccess()).isEqualTo(true);
+    }
```
trask, line 102: can you verify the requests that were sent to the mock http client also?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -60,8 +64,14 @@
       HttpRequest originalHttpRequest,
       int retryCount) {
     // make sure the context is not modified during retry, except for the URL
+    String redirectLocation = null;
```
trask, line 67: see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/style-guideline.md#final-keyword-usage

> Local variables should only be declared `final` if they are not initialized inline
(declaring these vars `final` can help prevent accidental double-initialization).

```suggestion
    final String redirectLocation;
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -74,6 +84,9 @@
                 String responseLocation = httpResponse.getHeaderValue("Location");
                 if (responseLocation != null) {
                   redirectMappings.put(originalHttpRequest.getUrl(), responseLocation);
+                  if (instrumentationKey != null) {
+                    instrumentationKeyMappings.put(instrumentationKey, responseLocation);
+                  }
```
trask, line 89: I think should not populate `redirectMappings` if `instrumentationKey != null`?

(if/else similar to above when reading from the mappings)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -99,4 +112,12 @@ private static boolean shouldRetryWithRedirect(int statusCode, int tryCount) {
         || statusCode == PERMANENT_REDIRECT_STATUS_CODE
         || statusCode == TEMP_REDIRECT_STATUS_CODE;
   }
+
+  private static String getInstrumentationKeyFromContext(HttpPipelineCallContext context) {
+    Object instrumentationKey = context.getData(INSTRUMENTATION_KEY).orElse(null);
+    if (instrumentationKey == null) {
+      return null;
+    }
+    return instrumentationKey.toString();
```
trask, line 121:
```suggestion
    return (String) context.getData(INSTRUMENTATION_KEY).orElse(null);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -162,9 +186,14 @@ private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
     //  * write to disk on second failure
     CompletableResultCode result = new CompletableResultCode();
     final long startTime = System.currentTimeMillis();
+    // Add instrumentation to context to use in redirectPolicy
+    Map<Object, Object> contextKeyValues = new HashMap<>();
+    if (!Strings.isNullOrEmpty(instrumentationKey)) {
+      contextKeyValues.put("instrumentationKey", instrumentationKey);
```
trask, line 192:
```suggestion
      contextKeyValues.put(RedirectPolicy.INSTRUMENTATION_KEY, instrumentationKey);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -162,9 +186,14 @@ private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
     //  * write to disk on second failure
     CompletableResultCode result = new CompletableResultCode();
     final long startTime = System.currentTimeMillis();
+    // Add instrumentation to context to use in redirectPolicy
+    Map<Object, Object> contextKeyValues = new HashMap<>();
+    if (!Strings.isNullOrEmpty(instrumentationKey)) {
```
trask, line 191:
```suggestion
    if (instrumentationKey != null) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -137,7 +160,8 @@ public CompletableResultCode send(List<TelemetryItem> telemetryItems) {
    * Object can be a list of {@link ByteBuffer} or a raw byte array. Regular telemetries will be
    * sent as {@code List<ByteBuffer>}. Persisted telemetries will be sent as byte[]
    */
-  private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
+  private CompletableResultCode internalSend(
+      List<ByteBuffer> byteBuffers, String instrumentationKey) {
```
trask, line 164:
```suggestion
      List<ByteBuffer> byteBuffers, @Nullable String instrumentationKey) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -162,9 +186,14 @@ private CompletableResultCode internalSend(List<ByteBuffer> byteBuffers) {
     //  * write to disk on second failure
     CompletableResultCode result = new CompletableResultCode();
     final long startTime = System.currentTimeMillis();
+    // Add instrumentation to context to use in redirectPolicy
```
trask, line 189:
```suggestion
    // Add instrumentation key to context to use in RedirectPolicy
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -81,7 +85,7 @@ public static TelemetryChannel create(
   }
 
   public CompletableResultCode sendRawBytes(ByteBuffer buffer) {
-    return internalSend(Arrays.asList(buffer));
+    return internalSend(Arrays.asList(buffer), null);
```
trask, line 88: in this case, I think we don't want to store the redirect into `redirectMappings` because we may store the redirect from one ikey, and then pick that up from a different ikey.

one option here would be to use something like "unknown" constant to indicate it's not for a known ikey.

another (maybe better) option would be to configure the RedirectPolicy to *only* use ikey redirects when configuring the telemetry channel

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -35,15 +35,24 @@
 
 // This is mostly a copy from Azure Monitor Open Telemetry Exporter SDK AzureMonitorRedirectPolicy
 public final class RedirectPolicy implements HttpPipelinePolicy {
-
+  private final boolean followInstrumentationKeyForRedirect;
+  // use this only when followInstrumentationKeyForRedirect is true and instrumentation key is null
+  private String cachedUrl;
```
trask, line 40: are you sure this is needed?

kryalama: removed this usage

trask (approved):

PR #1813, opened 2021-07-24
Support system properties for the main "coordinates"

### LINKED ISSUE #1684: Set cloud role name using Java system property
Currently we use Application Insights e2e transaction logging on all our environments, even for local development. Now we have an issue (annoyance) that we have to provide cloud role name of application either using Environment Variable (does not work when you run multiple applications locally) or using JSON configuration file. It would be nice if it would be also possible to set application name (cloud role name) using Java system property, just like we can for configuration file path.

Ability to set cloud role name using Java system property, i.e. `applicationinsights.role.name`.

trask:
Resolves #1684

Supports system properties for the main "coordinates":
* applicationinsights.connection.string
* applicationinsights.role.name
* applicationinsights.role.instance

Also logs warning if using undocumented `APPLICATIONINSIGHTS_JMX_METRICS` (planning to remove support for this undocumented env var in the future).

Also renames `APPLICATIONINSIGHTS_PROFILER_ENABLED` to `APPLICATIONINSIGHTS_PREVIEW_PROFILER_ENABLED` (cc: @johnoliver).

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -41,9 +41,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.CopyOnWriteArrayList;
 import okio.Buffer;
-import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ConfigurationBuilder {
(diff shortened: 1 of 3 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -446,23 +459,48 @@ private static String getWebsiteSiteNameEnvVar() {
     return value;
   }
 
+  public static String overlayWithSysPropEnvVar(
+      String systemPropertyName, String envVarName, String defaultValue) {
+    String value = getSystemProperty(systemPropertyName);
```
heyams, line 464: system properties have a higher precedence over env vars?  should it be the other way around? env vars are user defined properties and i think it has a higher precedence over the default system properties, right? we do want to respect end user's choice?

trask: system properties are also provided by end user

this precedence ordering is same as Spring (and OTel also follows same):

https://docs.spring.io/spring-boot/docs/2.5.2/reference/html/features.html#features.external-config

heyams (approved):

PR #1821, opened 2021-07-30
Replace Moshi with Jackson for json serialization and deserialization in agent-tooling

kryalama:
Replace Moshi with Jackson for json serialization and deserialization in agent-tooling


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -569,6 +574,14 @@ static Configuration getConfigurationFromConfigFile(Path configPath, boolean str
     }
   }
 
+  // Function to remove all non-ASCII characters
+  // Note: Is public, since we have to use this function in tests.
+  public static String cleanJsonContent(String text) {
+    // strips off all non-ASCII characters
+    text = text.replaceAll("[^\\x00-\\x7F]", "");
```
kryalama, line 581: One issue with this implementation is this function will remove any non ascii characters present in the json config values itself. But I am assuming our config will always have ascii characters.

trask (commented):
> One issue with this implementation is this function will remove any non ascii characters present in the json config values itself. But I am assuming our config will always have ascii characters.

I'd rather not make this assumption since it's not very international-friendly, e.g. telemetry processors that match on some international text.

How about if there's a parsing error, checking if there's non-ascii characters and logging to the user that that may be the issue.

Btw, what was the character that was embedded in that file in case that gives us other ideas how best to address?

kryalama:
> > One issue with this implementation is this function will remove any non ascii characters present in the json config values itself. But I am assuming our config will always have ascii characters.
> 
> I'd rather not make this assumption since it's not very international-friendly, e.g. telemetry processors that match on some international text.
> 
> How about if there's a parsing error, checking if there's non-ascii characters and logging to the user that that may be the issue.
> 
> Btw, what was the character that was embedded in that file in case that gives us other ideas how best to address?

I agree. Yeah we can log the non ascii characters string.
Attached the screenshot user shared with us. I was not able to copy or find that special character.
image

heyams:
@kryalama did you find out how customer created that character at the first place? People usually copy and past config from our azure doc.. let's make sure not an issue with our doc.
Let's not modify the json content.. display that char to users will be helpful.. attach will fail in that case. can we list that as a reason of the failure?

kryalama:
Sample Friendly exception thrown:

*************************
Application Insights Java agent failed to start
*************************

Description:
Application Insights Java agent's configuration file C:\git\aijaexp\ApplicationInsights-Java\agent\agent-tooling\build\resources\test\applicationinsights_lenient_malformed.json has a malformed JSON at path $.connectionString


Action:
Learn more about configuration options here: https://go.microsoft.com/fwlink/?linkid=2153358

Note:
Application Insights Java agent identified the following list of non-ASCII strings in the configuration file: 
  "connectionString": à"InstrumentationKey=00000000-0000-0000-0000-000000000000",
    "name": à"Something Good",
    "instance": à"xyz123"
    "percentage": à10
      "enabled": àfalse
      "level": à"error"
    "host": à"myproxy",
      "path": à"/var/log/applicationinsights/abc.log",
Please remove any unexpected characters from the json configuration file

trask (approved):

PR #1822, opened 2021-07-31
Fix netty ByteBuf leak

trask:
Fixes an issue that started in 3.2.0-BETA.1 after swapping out ingestion http client with Azure SDK / reactor netty.

```
ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
```

PR #1823, opened 2021-07-31
Update Azure SDK version

trask:
Just good to stay up-to-date (initially tried this for #1822 but that turned out to be unrelated)


kryalama (approved):

agent/agent-profiler/agent-profiler-api/gradle.lockfile
```diff
@@ -1,7 +1,7 @@
 # This is a Gradle generated file for dependency locking.
 # Manual edits can break the build and are not advised.
 # This file is expected to be part of source control.
-com.azure:azure-core:1.17.0=runtimeClasspath
+com.azure:azure-core:1.18.0=runtimeClasspath
```
heyams, line 4: we don't need to update Statsbeat instrumentation for azure sdk, right?  It's always going to stay as "Azure-OpenTelemetry"?  Just want to double check.  We can remove the comment for this line:

`INSTRUMENTATION_MAP.put("Azure-OpenTelemetry", 6); // bridged by azure-core-1.14 module`

heyams (approved):

agent/agent-profiler/agent-profiler-api/gradle.lockfile
trask: > `INSTRUMENTATION_MAP.put("Azure-OpenTelemetry", 6); // bridged by azure-core-1.14 module`

this comment is related to our instrumentation of azure-core library (when applications we are monitoring use it)

this PR is updating our internal usage of azure-core library (which we now use instead of Apache HttpClient)

PR #1824, opened 2021-07-31
Fix quick pulse disabled option

trask:
disabling quickpulse option is undocumented, but I noticed when trying to disable it as part of investigating #1822


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -176,6 +176,11 @@ synchronized FinalCounters peek() {
   }
 
   public void add(TelemetryItem telemetryItem) {
+    if (telemetryClient == null) {
```
kryalama, line 179: I think this will just stop collecting data. But we also need to stop quick pulse pings?

trask: > we also need to stop quick pulse pings?

that part is already handled

kryalama (approved):

PR #1839, opened 2021-08-06
Fix live metrics sampled counts

trask:
Currently live metrics is under-reporting the request/dependency/exception counts because we are not factoring sampling into consideration.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -185,51 +185,55 @@ public void add(TelemetryItem telemetryItem) {
       return;
     }
 
+    Float sampleRate = telemetryItem.getSampleRate();
+    int itemCount = sampleRate == null ? 1 : Math.round(100 / sampleRate);
```
heyams, line 189: ArithmeticException when sampleRate is 0?

heyams (approved):
👍

PR #1840, opened 2021-08-06
update models based on swagger v2.1 update

kryalama:
update models based on swagger v2.1 update


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/models/MonitorDomain.java
```diff
@@ -52,4 +62,34 @@ public MonitorDomain setVersion(int version) {
...
+  }
+
+  /**
+   * Set the additionalProperties property: The abstract common base of all domains.
+   *
+   * @param additionalProperties the additionalProperties value to set.
+   * @return the MonitorDomain object itself.
+   *     <p>//TODO (kryalama) add data sanitization
+   */
+  public MonitorDomain setAdditionalProperties(Map<String, Object> additionalProperties) {
```
kryalama, line 83: @trask do we need to sanitize properties here for consistency? I dont see any limits specified in swagger file though

trask: we don't use these (and I don't know what they're for) so I think fine to not sanitize

trask (approved):
thx

test/smoke/testApps/CoreAndFilter/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/CoreAndFilterTests.java
```diff
@@ -338,7 +338,7 @@ public void testTrackPageView() throws Exception {
 
     assertNotNull(pv2);
     assertEquals(new Duration(123456), pv2.getDuration());
-    assertEquals("2010-10-10T00:00:00Z", pvdEnvelope2.getTime());
+    assertEquals("1286668800.000000000", pvdEnvelope2.getTime());
```
trask, line 341: does breeze accept this?

trask:
@kryalama CI should be fixed now in main, go ahead and merge main into your PR

trask (approved):

PR #1848, opened 2021-08-11
Fix service bus mapping

trask:
Based on @lmolkova's https://github.com/Azure/azure-sdk-for-java/pull/21668#discussion_r685572619 and https://github.com/Azure/azure-sdk-for-java/pull/21668#pullrequestreview-725874372

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
@@ -164,7 +164,7 @@ public class Exporter implements SpanExporter {
     for (SpanData span : spans) {
       logger.debug("exporting span: {}", span);
       try {
-        export(span);
+        internalExport(span);
         exportingSpanLogger.recordSuccess();
       } catch (Throwable t) {
         exportingSpanLogger.recordFailure(t.getMessage(), t);
@@ -187,7 +187,7 @@ public class Exporter implements SpanExporter {
     return CompletableResultCode.ofSuccess();
   }
 
-  private void export(SpanData span) {
+  private void internalExport(SpanData span) {
     SpanKind kind = span.getKind();
     String instrumentationName = span.getInstrumentationLibraryInfo().getName();
     StatsbeatModule.get().getNetworkStatsbeat().addInstrumentation(instrumentationName);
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -129,20 +99,55 @@
...
+      AttributeKey.longKey("x-opt-enqueued-time");
 
   private static final OperationLogger exportingSpanLogger =
       new OperationLogger(Exporter.class, "Exporting span");
 
   private static final OperationLogger parsingHttpUrlLogger =
       new OperationLogger(Exporter.class, "Parsing http.url");
 
+  static {
+    Set<String> dbSystems = new HashSet<>();
```
heyams, line 118: does otel java sdk have dbSystems defined already? can we refer to that directly if it becomes available? i remember there is a common spec for dbsystems on the otel spec repo.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -129,20 +99,55 @@
...
+    dbSystems.add("sqlite");
+    dbSystems.add("other_sql");
+    dbSystems.add("hsqldb");
+    dbSystems.add("h2");
+
+    SQL_DB_SYSTEMS = Collections.unmodifiableSet(dbSystems);
+
+    // TODO need to keep this list in sync as new semantic conventions are defined
+    // TODO make this opt-in for javaagent
+    Set<String> standardAttributesPrefix = new HashSet<>();
```
heyams, line 135: same here.. i feel that otel should take care of all the standdard attribute prefixes.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: unfortunately there are no otel constants we can use for this, at least not yet

heyams (approved):

PR #1854, opened 2021-08-16
Add inherited attributes preview

### LINKED ISSUE #1743: Using Java Agent with AI SDK
**What we have:**

Right now we are using AI SDK v 2.6.3 for tracking HTTP operations and remote dependencies. We use it in multi-tenant system, and it is important for us to incorporate additional information about tenant (which is passed via request headers) in our telemetry data. 
1. For incoming requests, we use a filter:

```
protected void doFilterInternal(...) {
        ...
        RequestTelemetry requestTelemetry = ThreadContext.getRequestTelemetryContext().getHttpRequestTelemetry();
        requestTelemetry.getProperties().putIfAbsent("TENANT", tenantHolder.getTenant());
        requestTelemetry.setName(tenantHolder.getTenant() + " " + requestTelemetry.getName());
        filterChain.doFilter(httpServletRequest, httpServletResponse);
    }
```
 What we achive is visible distinction between requests for tenants, and ability to filter telemetry for them.
 
2. We also achieved something similiar for remote dependencies by manually sending telemetry data via `RemoteDependencyTelemetry` and `TelemetryClient`. We are able to produce names with tenant information, full URL, time measurments and it basically works perfectly fine

**What we want to achieve:**

We want to keep all of above functionalities, but also have automatic telemetry collected for JMS, Redis, Postgresql etc. so we tried to use SKD with Java Agent v 3.1.0

**Problems:**
1. Despite https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-in-process-agent#override-the-request-telemetry-name-using-the-2x-sdk , this solution doesn't work for us. For above code `getRequestTelemetryContext()` always returns another instance, `HttpRequestTelemetry` is not populated and code is not having any effect on collected telemetry data. When debugging, invoking `requestTelemetry.getName()` after filterChain returns desired name.
2. Our custom telemetry is not correlated with auto-collected, so requests are duplicated, and we couldn't find a way to disable auto collection of only remote dependencies, or to somehow explicitly correlate auto and manual telemetry.

**In short:**

Basically what we want is to preserve HTTP telemetry names in form:

<tenant_name> <http_method> <endpoint/url>
e.g.: Tenant_123 GET /api/v1/get-tenant-info/{tenant_id}

and what we have is:
GET /api/v1/get-tenant-info/{tenant_id} - for incoming requests

HTTP GET - for outgoing request (+ duplicated manual telemetry)

**We use:**

Spring 4.3.30, applicationinsights-web 2.6.3, Application Insights Java Agent 3.1.0

Thanks in advance for any suggestions.

trask:
Resolves #1743


test/smoke/appServers/global-resources/inheritedattributes_applicationinsights.json
```diff
@@ -0,0 +1,11 @@
+{
+  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-0FEEDDADBEEF;IngestionEndpoint=http://host.docker.internal:6060/",
+  "preview": {
+    "inheritedAttributes": [
+      {
+        "key": "tenant",
```
heyams, line 6: can you add tests for other data type supported? i am interested in seeing the structure of Array type, those a bit complicated ones.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/InheritedAttributesSpanProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    if (!(parentSpan instanceof ReadableSpan)) {
+      return;
+    }
+    ReadableSpan parentReadableSpan = (ReadableSpan) parentSpan;
+
+    for (AttributeKey<?> inheritAttributeKey : inheritAttributeKeys) {
+      Object value = TempGetAttribute.getAttribute(parentReadableSpan, inheritAttributeKey);
+      if (value != null) {
+        span.setAttribute((AttributeKey<Object>) inheritAttributeKey, value);
```
heyams, line 61: are inheritAttributes part of the customDimensions in the payload?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -195,11 +196,59 @@ public String toString() {
...
+  public static class InheritedAttribute {
+    public String key;
+    public SpanAttributeType type;
+
+    public AttributeKey<?> getAttributeKey() {
+      switch (type) {
+        case STRING:
+          return AttributeKey.stringKey(key);
+        case BOOLEAN:
+          return AttributeKey.booleanKey(key);
```
heyams, line 215: i checked out #1743, the way they send extra attributes are through part C.  Breeze accepts key/value pair string only.  how other datatypes here will get used?

test/smoke/appServers/global-resources/inheritedattributes_applicationinsights.json
trask: I think non-string attributes will be pretty rare for this feature

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/InheritedAttributesSpanProcessor.java
trask: yes

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: they are all converted to strings, see `Exporter.getStringValue()`

test/smoke/appServers/global-resources/inheritedattributes_applicationinsights.json
heyams: if we expect string all the time, can you refactor SpanAttributeType and InheritedAttribute to support string only?

trask: I want to leave open the possibility for others

heyams: at least remove the array types because Breeze won't accept array? we can add it back when it's supported? i just feel that this might never get used and become confusing later...

trask: arrays are converted to strings, see `Exporter.getStringValue()`

heyams (approved):

PR #1855, opened 2021-08-16
Smoke test cleanup

trask:


test/smoke/framework/testCore/src/main/java/com/microsoft/applicationinsights/smoketest/AiSmokeTest.java
```diff
@@ -839,4 +841,110 @@ protected static void waitForUrlWithRetries(
     return mockedIngestion.getBaseDataForType(index, type);
   }
   // endregion
+
+  protected static Telemetry getTelemetry(int rddCount) throws Exception {
```
heyams, line 845: should we check rddCount can't be greater than 3 based on the [Telemetry](https://github.com/microsoft/ApplicationInsights-Java/pull/1855/files#:~:text=public%20static%20class-,Telemetry,-%7B) class's definition?

heyams (commented):
image
👍

heyams (approved):

kryalama (approved):

PR #1859, opened 2021-08-19
Update Statsbeat

heyams:
- add two disabled configs for Statsbeat
- send network count per host
- send endpoint + host as part of the network statsbeat payload
- send attach on start and change interval to 24 hours
- send azure metadata service on start and change interval to 24 hours
- send instrumentation as a FeatureStatsbeat
- track Statsbeat disabled via the feature list of FeatureStatsbeat


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -163,12 +163,20 @@ public String toString() {
...
+    public boolean disabled = false;
+
     public String instrumentationKey =
         "c4a29126-a7cb-47e5-b348-11414998b11e"; // workspace-aistatsbeat
     public String endpoint =
         DefaultEndpoints.INGESTION_ENDPOINT; // this supports the government cloud
-    public long intervalSeconds = MINUTES.toSeconds(15); // default to 15 minutes
-    public long featureIntervalSeconds = DAYS.toSeconds(1); // default to daily
+    public long shortIntervalSeconds = MINUTES.toSeconds(15); // default to 15 minutes
+    public long longIntervalSeconds = DAYS.toSeconds(1); // default to daily
```
trask, line 179: can user effectively turn off essential statsbeat by making these very big?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -25,18 +25,27 @@
...
 
-class FeatureStatsbeat extends BaseStatsbeat {
+public class FeatureStatsbeat extends BaseStatsbeat {
 
   private static final String FEATURE_METRIC_NAME = "Feature";
+  private static final String INSTRUMENTATION_METRIC_NAME = "Instrumentation";
 
   private final Set<Feature> featureList = new HashSet<>(64);
+  private final Set<String> instrumentationList =
+      Collections.newSetFromMap(new ConcurrentHashMap<>());
```
trask, line 41: 👍 let's use concurrent set for `featureList` too

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -25,18 +25,27 @@
...
+  private final Object lock = new Object();
```
trask, line 42: no need for `lock` since using concurrent set

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/FeatureType.java
```diff
@@ -0,0 +1,27 @@
...
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.statsbeat;
+
+enum FeatureType {
+  Feature,
+  Instrumentation
```
trask, line 26: typical java convention is uppercase enums

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +46,90 @@
...
+      // something goes wrong.
+      // this happens rarely.
+      disabledAll.set(true);
+      return;
     }
 
-    featureStatsbeat.trackConfigurationOptions(config);
+    if (!config.internal.statsbeat.disabled) {
+      if (this.telemetryClient == null) {
+        this.telemetryClient = telemetryClient;
```
trask, line 79: it looks like `telemetryClient` is not initialized when non-essential statsbeat are disabled?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -101,6 +140,37 @@ public NetworkStatsbeat getNetworkStatsbeat() {
...
+  public FeatureStatsbeat getInstrumentationStatsbeat() {
+    return instrumentationStatsbeat;
+  }
+
+  // send network statsbeat whenever redirect happens since url has been changed.
+  // new url is always retrieved from the redirect policy cache map and we don't update the
+  // endpoint.
+  public void sendNetworkStatsbeatOnRedirect(String redirectUrl) {
+    if (!disabledAll.get()) {
+      networkStatsbeat.trackHostOnRedirect(telemetryClient, redirectUrl);
```
trask, line 152: `telemetryClient` is going to be `null` here if `config.internal.statsbeat.disabled` is true and `config.internal.statsbeat.disabledAll` is false (related to comment above)

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
```diff
@@ -132,6 +130,8 @@ public void setup() {
               HttpHeaders httpHeaders = new HttpHeaders(headers);
               return Mono.just(new MockHttpResponse(request, 307, httpHeaders));
             });
+
+    StatsbeatModule.get().setDisabledAll(true); // avoid sending network statsbeat on redirect
```
trask, line 134: can we inject StatsbeatModule, to avoid relying on global state in the test?

test/smoke/testApps/Statsbeat/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/StatsbeatSmokeTest.java
```diff
@@ -64,7 +80,7 @@ public void testStatsbeat() throws Exception {
     assertNotNull(attachData.getProperties().get("language"));
     assertNotNull(attachData.getProperties().get("version"));
     assertNotNull(attachData.getProperties().get("rpId"));
-    assertEquals(8, data.getProperties().size());
+    assertEquals(8, attachData.getProperties().size());
```
trask, line 83: 👍

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/common/TestUtils.java
```diff
@@ -0,0 +1,74 @@
...
+import com.microsoft.applicationinsights.agent.internal.exporter.models.MetricsData;
+import com.microsoft.applicationinsights.agent.internal.exporter.models.MonitorBase;
+import com.microsoft.applicationinsights.agent.internal.exporter.models.TelemetryItem;
+import com.microsoft.applicationinsights.agent.internal.telemetry.FormattedTime;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class TestUtils {
```
trask, line 35: 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: if they look at our code, they can.. we're not going to expose these config in public azure doc.

trask: oh, I forgot `statsbeat` is under `internal` namespace, that's good

```
{
  "internal": {
    "statsbeat": {
      ...
    }
  }
}
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -101,6 +142,32 @@ public NetworkStatsbeat getNetworkStatsbeat() {
...
+      StatsbeatSender sender = new StatsbeatSender(networkStatsbeat, telemetryClient);
+      Thread senderThread = new Thread(sender);
+      senderThread.setDaemon(true);
+      senderThread.start();
```
trask, line 158: same here, does it work to change `initialDelay` instead?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -101,6 +142,32 @@ public NetworkStatsbeat getNetworkStatsbeat() {
...
+    }
+
+    // TODO send optional network statsbeat when 'disabled' is off
+  }
+
+  private void sendAttachStatsbeatOnStart() {
+    StatsbeatSender sender = new StatsbeatSender(attachStatsbeat, telemetryClient);
+    Thread senderThread = new Thread(sender);
+    senderThread.setDaemon(true);
+    senderThread.start();
```
trask, line 168: same here too

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +46,92 @@
...
-      new AzureMetadataService(attachStatsbeat, customDimensions)
-          .scheduleWithFixedDelay(intervalSeconds);
+      AzureMetadataService metadataService =
+          new AzureMetadataService(attachStatsbeat, customDimensions);
+      metadataService.scheduleWithFixedDelay(longIntervalSeconds);
+
+      // start AzureMetadataService on start
+      Thread senderThread = new Thread(metadataService);
+      senderThread.setDaemon(true);
+      senderThread.start();
```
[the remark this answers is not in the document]
heyams: missed that. thanks.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -163,12 +163,20 @@ public String toString() {
   }
 
   public static class Statsbeat {
+    // disabledAll is used internally as an emergency kill-switch to turn off Statsbeat completely
+    // when something goes wrong.
+    public boolean disabledAll = false;
+
+    // disabled is used by customer to turn off non-essential Statsbeat, e.g. disk persistence
+    // operation status, optional network statsbeat, other endpoints except Breeze, etc.
+    public boolean disabled = false;
```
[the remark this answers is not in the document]
heyams: good point.. this will be customer facing config.  
under preview?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -83,6 +87,9 @@ public RedirectPolicy(boolean followInstrumentationKeyForRedirect) {
                   cacheRedirectUrl(
                       responseLocation, instrumentationKey, originalHttpRequest.getUrl());
                   context.setHttpRequest(originalHttpRequest.copy().setUrl(responseLocation));
+                  if (statsbeatModule != null) {
+                    statsbeatModule.sendNetworkStatsbeatOnRedirect(responseLocation);
+                  }
```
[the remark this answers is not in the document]
heyams: we could.. need to mock lots of objects. let me try it.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +46,92 @@
...
+      // this happens rarely.
+      disabledAll.set(true);
+      return;
+    }
+
+    if (this.telemetryClient == null) {
+      this.telemetryClient = telemetryClient;
+    }
+
+    // send essential Statsbeat
```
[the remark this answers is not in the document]
heyams: below is essential statsbeat

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +46,92 @@
...
+    scheduledExecutor.scheduleWithFixedDelay(
+        new StatsbeatSender(instrumentationStatsbeat, telemetryClient),
+        longIntervalSeconds,
+        longIntervalSeconds,
         TimeUnit.SECONDS);
 
     ResourceProvider rp = customDimensions.getResourceProvider();
     // only turn on AzureMetadataService when the resource provider is VM or UNKNOWN.
-    // FIXME (heya) Need to figure out why AzureMetadataService is not reachable from a function app
     // and it's not necessary to make this call.
```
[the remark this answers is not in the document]
heyams: i deleted it intentionally.. it's not reachable from function app by default.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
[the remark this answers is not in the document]
heyams: used Mockito.doNothing()

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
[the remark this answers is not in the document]
heyams: Moved it to preview. we might get rid of this entirely after 6 months.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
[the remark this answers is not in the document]
trask: what about line 112? 

```
    // and it's not necessary to make this call.
```

if that was part of the deleted line then can delete that too

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +46,85 @@
   private final NetworkStatsbeat networkStatsbeat;
   private final AttachStatsbeat attachStatsbeat;
   private final FeatureStatsbeat featureStatsbeat;
+  private final FeatureStatsbeat instrumentationStatsbeat;
 
   private final AtomicBoolean started = new AtomicBoolean();
+  private final AtomicBoolean disabledAll = new AtomicBoolean();
```
trask, line 52: can change this to volatile boolean (not using atomic-ness)

test/smoke/testApps/Statsbeat/src/smokeTest/resources/appServers.txt
```diff
@@ -1,3 +1 @@
-wildfly11
-tomcat85
-jetty9
\ No newline at end of file
+tomcat85
```
trask: looks like accidental commit?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
```diff
@@ -76,7 +75,9 @@
   private TelemetryChannel getTelemetryChannel(boolean followInstrumentationKeyForRedirect)
       throws MalformedURLException {
     List<HttpPipelinePolicy> policies = new ArrayList<>();
-    policies.add(new RedirectPolicy(followInstrumentationKeyForRedirect));
+    StatsbeatModule mockStatsbeatModule = Mockito.mock(StatsbeatModule.class);
+    Mockito.doNothing().when(mockStatsbeatModule).sendNetworkStatsbeatOnRedirect(null);
```
trask, line 79: I think this is the default
```suggestion
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/AzureMetadataService.java
```diff
@@ -68,7 +68,7 @@ void scheduleWithFixedDelay(long interval) {
     // Querying Azure Metadata Service is required for every 15 mins since VM id will get updated
     // frequently.
     // Starting and restarting a VM will generate a new VM id each time.
-    scheduledExecutor.scheduleWithFixedDelay(this, interval, interval, TimeUnit.SECONDS);
+    scheduledExecutor.scheduleWithFixedDelay(this, 0, interval, TimeUnit.SECONDS);
```
trask, line 71: is it important for this to happen right away? would be better to delay it a little bit so it doesn't potentially impact startup time

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
heyams: no, that won't work.. i tried.. that will require more mock object.. it will throw nullpointerexception when executing that method. if condition will blow up.  need to mock the whole chain of configuration.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/AzureMetadataService.java
heyams: 5 sec?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
heyams: why not? what about started? we use atomicboolean in many places. are you concerning that too many atomic operations will add perf overhead?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
trask: can you double check? that doesn't sound right, this is a mock, not a spy

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
trask: started is using it's atomic-ness: `started.getAndSet(true)`

heyams: set and getAndSet are both atomic-ness, correct? both use Varhandle.setVolatile behind the scene.. getAndSet will return the previous value, set doesn't. not sure other differences between them... i probably don't get your point :(

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
heyams: that was my first option. then switched to donothing(). it's a mock, not a spy, but this method works.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatModule.java
```diff
@@ -46,51 +47,81 @@
...
-        intervalSeconds,
+        shortIntervalSeconds,
+        shortIntervalSeconds,
         TimeUnit.SECONDS);
     scheduledExecutor.scheduleWithFixedDelay(
         new StatsbeatSender(attachStatsbeat, telemetryClient),
-        intervalSeconds,
-        intervalSeconds,
+        60,
+        longIntervalSeconds,
```
trask, line 95: it is intentional this is the only one with the short initial delay?

heyams: yes, only attach we need to do this and the azure metadata service.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -213,4 +238,13 @@ private double getRequestDurationAvg() {
...
     }
   }
+
+  /**
+   * e.g. endpointUrl 'https://westus-0.in.applicationinsights.azure.com/v2.1/track' host will
+   * return 'westus-0.in.applicationinsights.azure.com'
+   */
+  static String getHost(String endpointUrl) {
+    assert (endpointUrl != null && !endpointUrl.isEmpty());
+    return endpointUrl.replaceAll("^\\w+://", "").replaceAll("/\\w+.?\\w?/\\w+", "");
```
trask, line 248: instead of regex, can we use simpler indexOf("://") and indexOf("/", ...) and get substring in between those?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -37,165 +40,187 @@
...
-    return current.exceptionCount.get();
-  }
-
-  // only used by tests
-  Set<String> getInstrumentationList() {
-    return current.instrumentationList;
+  private static void addCommonProperties(TelemetryItem telemetryItem, String ikey, String host) {
+    Map<String, String> properties =
+        TelemetryUtil.getProperties(telemetryItem.getData().getBaseData());
+    properties.put("endpoint", BREEZE_ENDPOINT);
```
trask, line 218: just curious what's the purpose of this if it's always the same value?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeatTest.java
```diff
@@ -23,108 +23,106 @@
...
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class NetworkStatsbeatTest {
 
   private NetworkStatsbeat networkStatsbeat;
+  private List<String> ikeys;
```
trask, line 37: it looks like this only ever has 1 ikey in it
```suggestion
  private static final String ikey = "00000000-0000-0000-0000-0FEEDDADBEEF";
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannelTest.java
```diff
@@ -65,18 +62,20 @@
...
 
   @TempDir File tempFolder;
 
-  private TelemetryChannel getTelemetryChannel(boolean followInstrumentationKeyForRedirect)
+  private TelemetryChannel getTelemetryChannel(@Nullable Cache<String, String> ikeyRedirectCache)
       throws MalformedURLException {
     List<HttpPipelinePolicy> policies = new ArrayList<>();
-    policies.add(new RedirectPolicy(followInstrumentationKeyForRedirect));
+
+    policies.add(new RedirectPolicy(ikeyRedirectCache));
```
trask, line 78: then can remove `ikeyRedirectCache` everywhere I think
```suggestion
    policies.add(new RedirectPolicy(Cache.newBuilder().setMaximumSize(5).build()));
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeatTest.java
```diff
@@ -124,5 +124,14 @@ public void testGetHost() {
...
     assertThat(NetworkStatsbeat.getHost(url)).isEqualTo("www.fake-host.com");
+
+    url = "www.fake-host.com/";
+    assertThat(NetworkStatsbeat.getHost(url)).isEqualTo("www.fake-host.com");
+
+    url = "http://fake-host.com";
+    assertThat(NetworkStatsbeat.getHost(url)).isEqualTo("fake-host.com");
+
+    url = "http://fake-host.com/";
+    assertThat(NetworkStatsbeat.getHost(url)).isEqualTo("fake-host.com");
```
trask, line 131: 👍

trask (approved):
👍

PR #1860, opened 2021-08-21
Improve Disk Persistence

heyams:
- purge persisted files older than 48 hours
- delete source file only when it is sent successfully
- add it back to cache when the send fails.
- add unit tests
- fix a couple of bugs where source and temp were deleted permanently, and ListFile returns zero all the time.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileWriter.java
```diff
@@ -123,8 +123,7 @@ private static long getTotalSizeOfPersistedFiles(File telemetryFolder) {
     }
 
     long sum = 0;
-    Collection<File> files =
-        FileUtils.listFiles(telemetryFolder, new String[] {PERMANENT_FILE_EXTENSION}, false);
+    Collection<File> files = FileUtils.listFiles(telemetryFolder, new String[] {"trn"}, false);
```
trask, line 126: oh, were we writing/reading `*..trn` files before? 😁

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -83,28 +84,39 @@
...
-    return create(endpointUrl, localFileWriter, null);
+  public static TelemetryChannel create(
+      URL endpointUrl, LocalFileWriter localFileWriter, LocalFileLoader localFileLoader) {
+    return create(endpointUrl, localFileWriter, localFileLoader, null);
   }
 
   public CompletableResultCode sendRawBytes(ByteBuffer buffer) {
-    return internalSend(singletonList(buffer), null);
+    CompletableResultCode resultCode = internalSend(singletonList(buffer), null);
+    localFileLoader.updateProcessedFileStatus(resultCode.isSuccess());
```
trask, line 105: I think it would be clearer to move calling the `localFileLoader` up to the caller of this method (LocalFileSender) which is also where we already have a reference to the local file loader, and then no need to inject LocalFileLoader into TelemetryChannel

trask (commented):
(I still need to review the LocalFileLoader changes)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileWriter.java
heyams: permanent file name is .trn.. but when we use FileUtils.listFiles, it has to be the file extension without dot.  That's the bug. with ".trn", listfiles always return nothing.

trask: do we have any tests that would/should catch this issue?

heyams: https://github.com/microsoft/ApplicationInsights-Java/pull/1860/files#:~:text=testDeleteFilePermanentlyOnFailure

i found this bug when writing this test. what other test do you have in mind? just test listFiles?  It's from Commons-Utils.

heyams:
```
Finds files within a given directory (and optionally its subdirectories) which match an array of extensions.
Params:
directory – the directory to search in
**extensions – an array of extensions, ex. {"java","xml"}. If this parameter is null, all files are returned.**
recursive – if true all subdirectories are searched as well
Returns:
a collection of java.io.File with the matching files
```

java doc for FileUtils.listFiles is pretty clear.

trask: great, so the new test you added would have caught this?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -87,21 +88,26 @@
   public static TelemetryChannel create(
       URL endpointUrl,
       LocalFileWriter localFileWriter,
+      LocalFileLoader localFileLoader,
```
trask, line 91: this is unused now

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
```diff
@@ -56,12 +57,12 @@ private LocalFileSender(LocalFileLoader localFileLoader, TelemetryChannel teleme
...
   @Override
   public void run() {
-    // TODO (heya) should this loop more aggressively until it reads all telemetry from disk?
-    //  or is it intentional to only send one batch per 30 seconds?
     try {
       ByteBuffer buffer = localFileLoader.loadTelemetriesFromDisk();
       if (buffer != null) {
-        telemetryChannel.sendRawBytes(buffer);
+        CompletableResultCode resultCode = telemetryChannel.sendRawBytes(buffer);
+        resultCode.join(5, TimeUnit.SECONDS); // wait for request to be completed.
```
heyams, line 64: avg statsbeat request is less than 1 second.  is 5 too long and that is also depended on the network speed.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -49,21 +53,24 @@ public LocalFileLoader(LocalFileCache localFileCache, File telemetryFolder) {
   // Load ByteBuffer from persisted files on disk in FIFO order.
   @Nullable
   ByteBuffer loadTelemetriesFromDisk() {
-    // TODO (heya) how does this load files from disk at startup that were left over from prior
-    //  process?
     String filenameToBeLoaded = localFileCache.poll();
-    if (filenameToBeLoaded == null) {
+    if (filenameToBeLoaded == null || filenameToBeLoaded.isEmpty()) {
```
trask, line 57: can `filenameToBeLoaded` be empty string?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -86,20 +93,41 @@ ByteBuffer loadTelemetriesFromDisk() {
...
-      operationLogger.recordFailure("Fail to read telemetry from " + tempFile.getName(), ex);
-      return null;
-    } catch (SecurityException ex) {
-      operationLogger.recordFailure(
-          "Unable to delete " + tempFile.getName() + ". Access is denied.", ex);
-      return null;
-    }
-
+    toBeDeletedFileQueue.add(
+        sourceFile); // mark source file to be deleted when it's sent successfully.
```
trask, line 97:
```suggestion
    // mark source file to be deleted when it's sent successfully.
    toBeDeletedFileQueue.add(sourceFile);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -49,21 +53,24 @@ public LocalFileLoader(LocalFileCache localFileCache, File telemetryFolder) {
...
-    // TODO (heya) how does this load files from disk at startup that were left over from prior
-    //  process?
     String filenameToBeLoaded = localFileCache.poll();
-    if (filenameToBeLoaded == null) {
+    if (filenameToBeLoaded == null || filenameToBeLoaded.isEmpty()) {
       return null;
     }
 
     File tempFile;
+    File sourceFile;
```
trask, line 62: can you add a comment explaining why we are copying the file and then reading from the copy and then deleting the copy (instead of just reading from the source directly)?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -86,20 +93,41 @@ ByteBuffer loadTelemetriesFromDisk() {
...
+      deleteFilePermanentlyOnSuccess();
+    } else {
+      // add the file back to local file cache to be process later.
+      localFileCache.addPersistedFilenameToMap(toBeDeletedFileQueue.poll().getName());
+    }
+  }
+
+  // delete a file on the queue permanently when http response returns success.
+  private void deleteFilePermanentlyOnSuccess() {
+    File file = toBeDeletedFileQueue.poll();
```
trask, line 116: the queue is not clear to me, how do we know the head is the one we want to delete?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -86,20 +93,41 @@ ByteBuffer loadTelemetriesFromDisk() {
...
   }
+
+  // either delete it permanently on success or add it back to cache to be processed again later on
+  // failure
+  public void updateProcessedFileStatus(boolean success) {
+    if (success) {
+      deleteFilePermanentlyOnSuccess();
+    } else {
+      // add the file back to local file cache to be process later.
+      localFileCache.addPersistedFilenameToMap(toBeDeletedFileQueue.poll().getName());
```
trask, line 110: same question about the queue here too

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
```diff
@@ -56,12 +57,12 @@ private LocalFileSender(LocalFileLoader localFileLoader, TelemetryChannel teleme
 
   @Override
   public void run() {
-    // TODO (heya) should this loop more aggressively until it reads all telemetry from disk?
-    //  or is it intentional to only send one batch per 30 seconds?
```
trask, line 60: is this comment addressed now?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -256,17 +258,16 @@ private void writeToDiskOnFailure(
...
-      String instrumentationKey) {
+      int statusCode, String instrumentationKey, List<ByteBuffer> finalByteBuffers) {
     switch (statusCode) {
       case 401: // UNAUTHORIZED
       case 403: // FORBIDDEN
         logger.warn(
             "Failed to send telemetry with status code:{}, please check your credentials",
             statusCode);
-        writeToDiskOnFailure(byteBuffers, finalByteBuffers);
+        if (instrumentationKey != null) {
```
trask, line 268:
```suggestion
        // sending raw bytes won't have any instrumentation key
        if (instrumentationKey != null) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
heyams: yes. will delete that.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
heyams: no, will remove the empty check.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
heyams: after we load a persisted file from the disk, we sent it to breeze right away. base on the http response, if it's successful, we delete it right away; otherwise add it back to the localfilecache.  during this whole process, there will be one file on the tobedeletedQueue.

heyams: LocalFileSender is a single thread scheduler.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
```diff
@@ -0,0 +1,105 @@
...
+      @Nullable Long purgeIntervalSeconds, @Nullable Long expiredIntervalSeconds, File folder) {
+    long interval =
+        purgeIntervalSeconds == null ? TimeUnit.DAYS.toSeconds(1) : purgeIntervalSeconds;
+    long expiredInterval =
+        expiredIntervalSeconds == null ? TimeUnit.DAYS.toSeconds(2) : expiredIntervalSeconds;
+
+    Executors.newSingleThreadScheduledExecutor(
+            ThreadPoolUtils.createDaemonThreadFactory(LocalFilePurger.class))
+        .scheduleWithFixedDelay(
+            new LocalFilePurger(expiredInterval, folder), interval, interval, SECONDS);
```
trask, line 64: for short-lived processes (under 1 day) this will never run?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  LocalFilePurger(long expiredInterval, File folder) {
```
trask, line 67: then we can pass in telemetry + statsbeat folders, and no need for separate test codepath
```suggestion
  LocalFilePurger(long expiredInterval, File... folders) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  // this is used by tests to configure purge interval, expired interval and the test folder which
+  // files are to be purged.
+  static void startPurging(
+      @Nullable Long purgeIntervalSeconds, @Nullable Long expiredIntervalSeconds, File folder) {
+    long interval =
+        purgeIntervalSeconds == null ? TimeUnit.DAYS.toSeconds(1) : purgeIntervalSeconds;
+    long expiredInterval =
+        expiredIntervalSeconds == null ? TimeUnit.DAYS.toSeconds(2) : expiredIntervalSeconds;
```
trask, line 59: I think more conventional to pass defaults directly instead of passing `null`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
trask: what is the resolution? is it not needed to loop more aggressively, or is it now looping more aggressively?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: in that case, there is probably no files need to be purged.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: this overloading is merely for testing and it's package protected.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: i don't see how this can prevent separate test codepath.

trask: sent commit to your branch

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
trask: sent commit to your branch to show how this change makes things simpler, reducing conditionals and code paths

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
heyams: resultCode.join(30, TimeUnit.SECONDS); // wait max 30 seconds for request to be completed.

it's still sending every 30 seconds, but it's less aggressive.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
trask: why not? short-lived processes can still fail when connecting to breeze, and can still write out temp files

heyams: it's not older than 48 hours.  when a process starts up, it will take care purging.

trask: > when a process starts up, it will take care purging.

where is this code? i missed it

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
trask: let's say a process can't connect to breeze for 1 day, that could be `86400 / 5 = 17280` files, which would take `17280 * 30 = 518400` seconds (or 6 days) to send them all at 1 file per 30 seconds?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: they will reattach agent and start a new process and purging will get started again?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: yup, it doesn't require updating the tests.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
heyams: does batch has a capacity limit? we don't handle that case yet. i will add the todo back.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
heyams: what will be a more reasonable interval? short-lived process can be any. 1min?

trask: I would suggest making `initialDelay` small, similar to the statsbeat scheduled tasks, maybe 60 seconds for `initialDelay`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -41,6 +41,8 @@
   private static final OperationLogger operationLogger =
       new OperationLogger(LocalFileLoader.class, "Loading telemetry from disk");
 
+  @Nullable private File toBeDeletedFile;
```
trask, line 44: I think the logic would be easier to follow without this "global" state, by returning it from `loadTelemetriesFromDisk`, and then passing it back in to `updateProcessedFileStatus`

heyams: loadTelemetriesFromDisk is returning ByteBuffer.  are you suggesting to return a struct?

trask: 👍

trask (approved):
👍

PR #1866, opened 2021-08-28
Update to latest otel

### LINKED ISSUE #1472: Handled exceptions being logged as exceptions in agent v3
### Expected behavior
Exceptions that are thrown from Spring Boot controllers and then handled by a Spring boot exception handler shouldn't be logged as `exception` in Application Insights.

### Actual behavior
Application Insights `exceptions` is unnecessarily noisy as there are lots of client errors in the logs.

### To Reproduce
Throw an exception from a Spring Boot controller that is then caught by a method annotated with `org.springframework.web.bind.annotation.ExceptionHandler`.  Even though the exception is then handled by our code it is still logged as an exception.  It appears that the `io.opentelemetry.javaagent.shaded.instrumentation.api.tracer.BaseTracer` intercepts the exception in the `endExceptionally` method after our controller is called and before the Spring Boot exception handler then kicks in.

I had a look to see if we could exclude the exception processor from running, or some way to disable that logging, but didn't see any way in configuration to disable it.

#### Sample Application
Project to replicate the issue is https://github.com/ministryofjustice/prison-estate.git. Need Java 11. Replace with your key for app insights.

To run the application:
```
git checkout pgp-DT-1442-insights-v3.0.2
./gradlew clean assemble && cp applicationinsights.json build/libs
APPLICATIONINSIGHTS_CONNECTION_STRING=InstrumentationKey=<your instrumentation key> java -javaagent:build/libs/applicationinsights-agent-3.0.2.jar -Dspring.profiles.active=dev -jar build/libs/prison-estate-*.jar
```

Then make a request that returns a 404 e.g. `curl http://localhost:8080/prisons/id/ABC`.  Our `PrisonEstateExceptionHandler` will intercept the exception thrown from the service and return a 404 back to the client.  This is normal behaviour and shouldn't cause an exception in `exceptions`.

### System information
SDK Version: 2.6.2
Java Agent Version: 3.0.2
OS type and version: Debian 14
Application Server type and version (if applicable): Spring Boot 2.4.2
Using spring-boot? yes

trask:
A few new features/fixes in here:
* support for suppressing controller spans
* support for 2.x SDK setSuccess() on the request
* reduce verbose otel logging
* fix exception telemetry operation name for exceptions attached directly to a request

Resolves #1472

attaching SNAPSHOT for testing: [applicationinsights-agent-3.2.0-BETA.3-SNAPSHOT.jar.zip](https://github.com/microsoft/ApplicationInsights-Java/files/7080209/applicationinsights-agent-3.2.0-BETA.3-SNAPSHOT.jar.zip)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/TrieImpl.java
```diff
@@ -0,0 +1,126 @@
...
+package com.microsoft.applicationinsights.agent.internal.exporter;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// copied from io.opentelemetry.javaagent.instrumentation.api.util.TrieImpl
+final class TrieImpl<V> implements Trie<V> {
```
kryalama, line 31: Happy to see Trie implementation here :)

kryalama (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/TemporaryNoopMeterConfigurer.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigurableMetricExporterProvider;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import java.util.Collection;
+
+// this is needed to ensure no metrics memory cannot grow unbounded until OTel Java 1.6.0 release
```
heyams, line 32: reading this comment, i still don't get it why use "noop" instead of "none" @trask

trask: if there's no exporter ("none") then the metrics are still collected, but they are never flushed, and so they can grow large if dimensions are not controlled well. this is what's fixed in OTel Java 1.6.0, in that release, if there's no exporter, then metrics won't be collected at all

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -172,6 +173,7 @@ public synchronized void enable(TelemetryClient telemetryClient) {
     counters.set(new Counters());
   }
 
+  @Nullable
```
heyams, line 176: this is mainly used for code analyzers? without it will spotlessApply trigger a warning?
return object type is nullable by default.

trask: I updated errorprone in one of the commits, and that required `@Nullable` on return types for methods that return explicit `null`. Eventually we will want to use `@Nullable` consistently everywhere. Conventionally the default is non-null, and only apply `@Nullable` on parameters, fields, method return values that are nullable.

PR #1875, opened 2021-09-15
Otel fork updated

trask:
Updates to OTel 1.6


.github/workflows/nightly-benchmark-overhead.yml
```diff
@@ -1,4 +1,4 @@
-name: benchmark-overhead tests
+name: Nightly overhead benchmark
```
heyams, line 1: does this generate a report on github somewhere?

heyams:
@trask  can we go through this change in a call tomorrow? i missed kotlin last time.. i think it's helpful if you can walk us through major changes since we're not on otel development.

.github/workflows/nightly-benchmark-overhead.yml
trask: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/gh-pages/benchmark-overhead/results

PR #1876, opened 2021-09-16
Update submodule

trask:
there are a few changes in this PR, check out the individual commits for breakout

uploading SNAPSHOT from CI for testing: [applicationinsights-agent-3.2.0-BETA.4-SNAPSHOT.jar.zip](https://github.com/microsoft/ApplicationInsights-Java/files/7213581/applicationinsights-agent-3.2.0-BETA.4-SNAPSHOT.jar.zip)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -276,6 +281,37 @@ private void exportRemoteDependency(SpanData span, boolean inProc) {
...
+    if (!DEFAULT_HTTP_SPAN_NAMES.contains(name)) {
+      return name;
+    }
+
+    String url = span.getAttributes().get(SemanticAttributes.HTTP_URL);
+    if (url == null) {
+      return name;
+    }
+
+    String path = UrlParser.getPathFromUrl(url);
```
heyams, line 311: does getPathFromUrl take care of empty url? when it's empty, we want to return name as well.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -741,23 +786,26 @@ private void exportRequest(SpanData span) {
       telemetry.getTags().put(ContextTagKeys.AI_LOCATION_IP.toString(), locationIp);
     }
 
-    data.setSource(getSource(attributes));
-
-    if (isAzureQueue(attributes)) {
```
heyams, line 746: is there a reason why this check is removed?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -766,14 +814,16 @@ private void exportRequest(SpanData span) {
...
+  private static String getSource(Attributes attributes, SpanContext spanContext) {
     // this is only used by the 2.x web interop bridge
     // for ThreadContext.getRequestTelemetryContext().getRequestTelemetry().setSource()
     String source = attributes.get(AI_SPAN_SOURCE_KEY);
     if (source != null) {
       return source;
     }
-    source = attributes.get(AI_SPAN_SOURCE_APP_ID_KEY);
+
+    source = spanContext.getTraceState().get("az");
```
heyams, line 825: what is az?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -828,12 +878,17 @@ private static String nullAwareConcat(String str1, String str2, String separator
...
     for (EventData event : span.getEvents()) {
-      boolean lettuce51 =
-          span.getInstrumentationLibraryInfo().getName().equals("io.opentelemetry.lettuce-5.1");
+      String instrumentationLibraryName = span.getInstrumentationLibraryInfo().getName();
+      boolean lettuce51 = instrumentationLibraryName.equals("io.opentelemetry.lettuce-5.1");
       if (lettuce51 && event.getName().startsWith("redis.encode.")) {
         // special case as these are noisy and come from the underlying library itself
         continue;
       }
+      boolean grpc16 = instrumentationLibraryName.equals("io.opentelemetry.grpc-1.6");
```
heyams, line 887: for special case, can you add a comment linked to the PR or issue from otel? it will offer some context.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
heyams: getPathFromUrl returns null currently. it has a todo.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
```diff
@@ -0,0 +1,32 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.exporter;
+
+public class UrlParser {
+
+  public static String getPathFromUrl(String url) {
+    // TODO implement something efficient and add tests
+    return null;
```
heyams, line 28: return url at least for now?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
```diff
@@ -0,0 +1,32 @@
...
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.microsoft.applicationinsights.agent.internal.exporter;
+
+public class UrlParser {
```
heyams, line 24: package protected?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: ya, it's just a placeholder for future work. current no-op implementation preserves existing behavior

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: it's not needed, since only azure queue uses the `AZURE_SDK_ENQUEUED_TIME` attribute that's checked below

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: legacy appId exchange is done using trace state, "az" was the azure monitor "key" that is used

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
trask: intentionally returning `null` which triggers the fall back mechanism which is span name, i'll add some doc to this method

heyams: can you reuse [getHost](https://github.com/microsoft/ApplicationInsights-Java/blob/main/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java#:~:text=static%20String%20getHost(String%20endpointUrl)%20%7B)?

trask: ya it's similar, but we want the path instead of host, and there are a couple of extra edge cases to cover

heyams (approved):

PR #1879, opened 2021-09-21
fix quick pulse memory leak

kryalama:
#TODO: Add unit tests

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
@@ -40,6 +40,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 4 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -139,6 +145,7 @@ static CountAndDuration decodeCountAndDuration(long countAndDuration) {
   private final AtomicReference<Counters> counters = new AtomicReference<>(null);
   private final MemoryMXBean memory;
   private final CpuPerformanceCounterCalculator cpuPerformanceCounterCalculator;
+  private volatile QuickPulseHeaderInfo quickPulseHeaderInfo;
```
trask, line 148: looks like this can just be boolean?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
```diff
@@ -121,7 +126,8 @@ private CountAndDuration(long count, long duration) {
 
     final AtomicLong rddsAndDuations = new AtomicLong(0);
     final AtomicInteger unsuccessfulRdds = new AtomicInteger(0);
-    final List<QuickPulseDocument> documentList = new ArrayList<>();
+    final ArrayBlockingQueue<QuickPulseDocument> documentList =
+        new ArrayBlockingQueue<>(1000, true);
```
trask, line 130: I think `ArrayBlockingQueue` is a bit heavy-weight for our needs (designed for concurrency).

I think `List` with a simple size check before adding should be enough. this is just a fallback safety mechanism now that you have implemented checking for quickpulse `on` before adding

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulseDataCollector.java
kryalama: My initial thought was to use boolean, but I am thinking this is more cleaner.

trask: maybe `QuickPulseStatus` then? (storing header here makes it seem like this class depends on the full header)

trask (approved):
👍

PR #1882, opened 2021-09-24
temporary fix for windows app service performance counters issue

kryalama:
There will be a follow up pr with perf counters included once we get all the information from app services team.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/TelemetryClientInitializer.java
```diff
@@ -95,6 +98,11 @@ public static void initialize(TelemetryClient telemetryClient, Configuration con
     setQuickPulse(configuration, telemetryClient);
   }
 
+  private static boolean isAgentRunningInSandboxEnvWindows() {
+    String qualifiedSdkVersion = PropertyHelper.getQualifiedSdkVersionString();
+    return qualifiedSdkVersion.startsWith("awr") || qualifiedSdkVersion.startsWith("fwr");
```
trask, line 103: can we (easily) cover cases where users are running the standalone agent in App Services?

trask (approved):

PR #1884, opened 2021-09-26
Update http dependency name

trask:
~Built on top of #1890~ EDIT: no longer as #1890 has been merged to `main`

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
@@ -21,6 +21,8 @@
 
 package com.microsoft.applicationinsights.agent.internal.exporter;
 
+import org.checkerframework.checker.nullness.qual.Nullable;
+
 class UrlParser {
 
   /**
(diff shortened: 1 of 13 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
```diff
@@ -21,16 +21,68 @@
...
+
 class UrlParser {
 
   /**
    * Returns the path portion of the url.
    *
    * <p>Returns {@code null} if the path cannot be extracted from url for any reason.
    */
+  @Nullable
   static String getPathFromUrl(String url) {
```
heyams, line 34: It looks like URL.getPath does the same thing? it handles case like ?query and #. Can we convert url string to an URL object and then use getPath directly, and we can get rid of UrlParser?

kryalama (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/UrlParser.java
trask: constructing URL objects is slow: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3699

heyams: it will be nice to compare the perf side by side. was URLStreamHandler.parseURL slow or the constructor of URL? 
inside the constructor it will invoke parseURL from the handler.. I am wondering if the newer version of java has improved it..

heyams (approved):

PR #1889, opened 2021-09-26
Update db mappings

trask:
~Built on top of #1890~ EDIT: no longer as #1890 has been merged to `main`

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
@@ -629,7 +629,7 @@ public class Exporter implements SpanExporter {
         nullAwareConcat(
             getTargetFromPeerAttributes(attributes, getDefaultPortForDbSystem(dbSystem)),
             attributes.get(SemanticAttributes.DB_NAME),
-            "/");
+            " | ");
     if (target == null) {
       target = dbSystem;
     }
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
@@ -620,21 +620,6 @@ public class Exporter implements SpanExporter {
       } else {
         type = "SQL";
       }
-      // keeping existing behavior that was release in 3.0.0 for now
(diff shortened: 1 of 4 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -620,21 +620,6 @@ private static void applyDatabaseClientSpan(
...
-      // these properties, they should only be used if the library being instrumented already
-      // provides them."
-      // also need to discuss with other AI language exporters
-      //
-      // if we go to shorter span name now, and it gets reverted, no way for customers to get the
-      // shorter name back
-      // whereas if we go to shorter span name in the future, and they still prefer more
-      // cardinality, they can get that back using telemetry processor to copy db.statement into
-      // span name
-      telemetry.setName(dbStatement);
```
heyams, line 637: no need to track dbStatement anymore? 

"select * from abc" makes more sense than "select mysql.abc"? (from the tests)

heyams: any link to the spec change?

trask: > any link to the spec change?

ya, https://github.com/microsoft/common-schema/pull/286

heyams: > If [Span.dbStatement](../PartB/traces/Span.md#dbstatement) is not available then use `dbOperation`.
should we keep dbStatement when it's available?

> "select * from abc" makes more sense than "select mysql.abc"? (from the tests)
thoughts?

trask: we still put dbStatement into the data field

heyams: > telemetry.setData(dbStatement);

I see. can we update the test to verify the dbStatement instead of getName()? since dbStatement is no longer part of name.

trask: we verify both

heyams (approved):

PR #1944, opened 2021-11-01
Reduce netty pool size

trask:
attaching snapshot for testing: [applicationinsights-agent-3.2.3-BETA-SNAPSHOT.jar.zip](https://github.com/microsoft/ApplicationInsights-Java/files/7463135/applicationinsights-agent-3.2.3-BETA-SNAPSHOT.jar.zip)

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/LazyHttpClient.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/LazyHttpClient.java
@@ -48,7 +48,7 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
 import org.checkerframework.checker.lock.qual.GuardedBy;
 import reactor.core.publisher.Mono;
-import reactor.netty.resources.ConnectionProvider;
+import reactor.netty.resources.LoopResources;
 
 public class LazyHttpClient implements HttpClient {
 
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/LazyHttpClient.java
```diff
@@ -48,15 +48,14 @@
...
-import reactor.netty.resources.ConnectionProvider;
+import reactor.netty.resources.LoopResources;
 
 public class LazyHttpClient implements HttpClient {
 
   private static final String APPLICATIONINSIGHTS_AUTHENTICATION_SCOPE =
       "https://monitor.azure.com//.default";
 
   private static final HttpClient INSTANCE = new LazyHttpClient();
-  private static final int DEFAULT_MAX_TOTAL_CONNECTIONS = 200;
```
heyams, line 59: 200 to 1? what is the motivation?

trask: 200 is a different number, I don't think we need a non-default value for that.

1 is the number of threads (down from default which is based on cpu-count).

the motivation is in the comment below (to reduce memory consumption, which was a regression in 3.2.0 when we switched http clients)

heyams (approved):

PR #1946, opened 2021-11-03
Handle weird folder names

trask:
Currently the agent fails to start when it's located in a weird folder name like `test%20-test`.


heyams (approved):
👍

PR #1953, opened 2021-11-04
Persist instrumentation key on disk

### LINKED ISSUE #1948: Error sending telemetry items: instrumentationKey must be non-null
I am trying to send logs from a Apache Camel application running in a Docker container to Application Insights.
I have created an Application Insight resource and provided the application with my connection string.
I can connect to the app via the Live Metrics in AI and see that logs are coming in, but when I try to view historic transaction I see no logs.

When looking in the applicationinsights-logs I can see that the agent is starting without complaints. But, after running a while I get this:
```
WARN  c.m.a.a.i.telemetry.TelemetryChannel - In the last 5 minutes, the following operation has failed 20 times (out of 20): Sending telemetry to the ingestion service (telemetry will be stored to disk on failure)::
 * Error sending telemetry items: instrumentationKey must be non-null when using ikey redirect policy (20 times)
```
When looking at the source-code the instrumenation key seems to be missing from the context in com.microsoft.applicationinsights.agent.internal.httpclient.RedirectPolicy

I've tried providing the connection string in the applicationinsights.json-file and via the env-var APPLICATIONINSIGHTS_CONNECTION_STRING without seeing any difference.
I've also tried to provide the instrumentation key explicitly by providing it via APPINSIGHTS_INSTRUMENTATIONKEY without success.

I have tried running a similar setup outside the Docker container, with Camel and the same connection string, and it works fine.

Now, I understand that there may be multiple causes to why I'm not seeing my logs in AI so for this report I'm focusing on the error I get in the logs.

Might be relevant that I get other exceptions before this, so I thought this was related: https://github.com/microsoft/ApplicationInsights-Java/issues/1927 but I have upgraded to v3.2.2 as suggested in that issue.

### Expected behavior
No error regarding missing instrumentation key when connection string has been provided.

### Actual behavior
Getting error "instrumentationKey must be non-null when using ikey redirect policy" when connection string has been provided.

### System information
 - SDK Version: Java 8
 - OS type and version: Linux version 4.19.128-microsoft-standard (oe-user@oe-host) (gcc version 8.2.0 (GCC)) #1 SMP Tue Jun 23 12:58:10 UTC 2020
 - Java 8
 - Using spring-boot? Yes
 - Additional relevant libraries (with version, if applicable): Apache Camel v2.25.1, applicationinsights-agent-3.2.2.

### Logs
[applicationinsights.log](https://github.com/microsoft/ApplicationInsights-Java/files/7467680/applicationinsights.log)
The logs is from 3 different applications. They all run the same Java and Linux versions. You can see that I have tried different configurations. My last run starts at line 513. I get all kinds of fun exceptions during my runs.

heyams:
Fix #1948


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileCache.java
```diff
@@ -35,6 +38,16 @@
...
   private final Queue<String> persistedFilesCache = new ConcurrentLinkedDeque<>();
 
+  public LocalFileCache(File folder) {
+    Collection<File> files = FileUtils.listFiles(folder, new String[] {"trn"}, false);
+    // existing files are not older than 48 hours and need to get added to the queue to be
+    // re-processed.
+    // this will avoid data loss in the case of app crashes and restarts.
+    for (File file : files) {
+      persistedFilesCache.add(file.getName());
+    }
```
trask, line 48: good catch, these were not being read previously?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -92,18 +95,23 @@ PersistedFile loadTelemetriesFromDisk() {
...
-      result = Files.readAllBytes(tempFile.toPath());
+    ByteBuffer ikeyByteBuffer;
+    ByteBuffer result;
+    try (FileInputStream fileInputStream = new FileInputStream(tempFile);
+        FileChannel channel = fileInputStream.getChannel()) {
+      ikeyByteBuffer = ByteBuffer.allocate(36);
+      result = ByteBuffer.allocate((int) (tempFile.length() - 36));
+      channel.read(ikeyByteBuffer);
+      channel.read(result);
+      result.flip();
```
trask, line 106: should not ignore return value from channel.read(), generally have to loop since read can return before fully read

I don't think using ByteBuffer here helps, and would be simpler to read bytes directly from FileInputStream

trask (commented):
what will happen if user upgrades and they still have prior .trn files without the ikey in them?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileCache.java
heyams: correct.

heyams:
> what will happen if user upgrades and they still have prior .trn files without the ikey in them?

if they have .trn that is not older than 48 hours.. it will get sent to App Insights backend.  for anything that is older than 48 hours, purger would take care of it.

trask:
> if they have .trn that is not older than 48 hours.. it will get sent to App Insights backend. for anything that is older than 48 hours, purger would take care of it.

did I miss the code that is handling reading in the old file format with no ikey embedded?

heyams:
> > if they have .trn that is not older than 48 hours.. it will get sent to App Insights backend. for anything that is older than 48 hours, purger would take care of it.
> 
> did I miss the code that is handling reading in the old file format with no ikey embedded?

if old trn that doesn't have instrumentation key, it will not get sent.  in that case it will get purged out after 48 hours.  should i add the purge to include files that are missing instrumentation key?

trask:
> if old trn that doesn't have instrumentation key, it will not get sent. in that case it will get purged out after 48 hours. should i add the purge to include files that are missing instrumentation key?

have you tested this? my guess from the code is that it will cause errors because we will parse the first 36 bytes into an ikey, and then the rest of the json message will be corrupted

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
heyams: Updated.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -95,23 +95,22 @@ PersistedFile loadTelemetriesFromDisk() {
...
+    byte[] telemetryBytes = new byte[rawByteLength];
     try (FileInputStream fileInputStream = new FileInputStream(tempFile);
         FileChannel channel = fileInputStream.getChannel()) {
-      ikeyByteBuffer = ByteBuffer.allocate(36);
-      result = ByteBuffer.allocate((int) (tempFile.length() - 36));
-      channel.read(ikeyByteBuffer);
-      channel.read(result);
-      result.flip();
+      fileInputStream.read(ikeyBytes, 0, 36);
+      fileInputStream.read(telemetryBytes, 0, rawByteLength);
```
trask, line 104: need to check the return value of `read` here also. I'd recommend writing a small `readFully()` method that loops until it's either read the requested number of bytes, or run out of bytes

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -113,6 +112,25 @@ PersistedFile loadTelemetriesFromDisk() {
...
 
+  // reads bytes from a FileInputStream and allocates those into the buffer array byteArray.
+  private static void readFully(
+      FileInputStream fileInputStream, byte[] byteArray, int offset, int length)
+      throws IOException {
+    if (length < 0) {
+      throw new IndexOutOfBoundsException();
+    }
+
+    int i = 0;
```
trask, line 123: I think would help readability
```suggestion
    int totalRead = 0;
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -113,6 +112,25 @@ PersistedFile loadTelemetriesFromDisk() {
...
+    while (i < length) {
+      int count = fileInputStream.read(byteArray, offset + i, length - i);
```
trask, line 125:
```suggestion
      int numRead = fileInputStream.read(byteArray, offset + i, length - i);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -113,6 +112,25 @@ PersistedFile loadTelemetriesFromDisk() {
         tempFile, new String(ikeyBytes, UTF_8), ByteBuffer.wrap(telemetryBytes));
   }
 
+  // reads bytes from a FileInputStream and allocates those into the buffer array byteArray.
+  private static void readFully(
+      FileInputStream fileInputStream, byte[] byteArray, int offset, int length)
```
trask, line 117: doesn't look like need `offset`, it's always `0`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -92,18 +100,57 @@ PersistedFile loadTelemetriesFromDisk() {
...
-    return new PersistedFile(tempFile, ByteBuffer.wrap(result));
+    return new PersistedFile(tempFile, instrumentationKey, ByteBuffer.wrap(telemetryBytes));
+  }
+
+  static boolean isInstrumentationKeyValid(String instrumentationKey) {
+    if (Pattern.matches(INSTRUMENTATION_KEY_REGEX, instrumentationKey.toLowerCase())) {
+      return true;
+    }
+
+    return false;
```
trask, line 135:
```suggestion
    return Pattern.matches(INSTRUMENTATION_KEY_REGEX, instrumentationKey.toLowerCase());
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileSender.java
```diff
@@ -60,7 +60,9 @@ public void run() {
     try {
       LocalFileLoader.PersistedFile persistedFile = localFileLoader.loadTelemetriesFromDisk();
       if (persistedFile != null) {
-        CompletableResultCode resultCode = telemetryChannel.sendRawBytes(persistedFile.rawBytes);
+        assert (persistedFile.instrumentationKey != null);
```
trask, line 63: asserts aren't typically enabled, so rarely used in Java

one option:

```suggestion
        if (persistedFile.instrumentationKey == null) {
          throw new IllegalStateException("...")
        }
```

but better option here is to validate in PersistedFile constructor and throw IllegalArgumentException

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -92,18 +100,53 @@ PersistedFile loadTelemetriesFromDisk() {
       return null;
     }
 
-    byte[] result;
-    try {
-      // TODO (trask) optimization: read this directly into ByteBuffer(s)
-      result = Files.readAllBytes(tempFile.toPath());
+    byte[] ikeyBytes = new byte[36];
+    int rawByteLength = (int) tempFile.length() - 36;
+    byte[] telemetryBytes = new byte[rawByteLength];
```
trask, line 105:
```suggestion
    if (tempFile.length() <= 36) {
       ... just in case ...
       operationLogger.recordFailure(...);
       return null;
    }
    byte[] ikeyBytes = new byte[36];
    int rawByteLength = (int) tempFile.length() - 36;
    byte[] telemetryBytes = new byte[rawByteLength];
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/localstorage/IntegrationTests.java
```diff
@@ -34,36 +35,49 @@
...
+import org.apache.commons.io.FileUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import reactor.core.publisher.Mono;
 
 public class IntegrationTests {
 
+  private static final String INSTRUMENTATION_KEY = "00000000-0000-0000-0000-0FEEDDADBEEF";
+  private static final String PERSISTED_FILENAME = "gziped-raw-bytes.trn";
```
trask, line 65: `gziped` -> `gzipped`

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/localstorage/IntegrationTests.java
```diff
@@ -97,11 +114,70 @@ public void integrationTest() throws InterruptedException {
...
+    expectedGzipByteBuffer.get(ikeyBytes, 0, 36);
+    assertThat(new String(ikeyBytes, UTF_8)).isEqualTo(INSTRUMENTATION_KEY);
+    int length = expectedGzipByteBuffer.remaining();
+    byte[] telemetryBytes = new byte[length];
+
+    expectedGzipByteBuffer.get(telemetryBytes, 0, length);
+    assertThat(loadedPersistedFile.rawBytes).isEqualTo(ByteBuffer.wrap(telemetryBytes));
+  }
+
+  private ByteBuffer getBytebufferFromFile(String filename) throws Exception {
```
trask, line 153: `getBytebufferFromFile` -> `getByteBufferFromFile`

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/localstorage/IntegrationTests.java
```diff
@@ -97,11 +114,70 @@ public void integrationTest() throws InterruptedException {
...
+    in.close();
+
+    return result;
+  }
+
+  private static String ungzip(byte[] rawBytes) throws Exception {
+    if (rawBytes == null) {
+      return null;
+    }
+
```
trask, line 169:
```suggestion
```

trask (commented):
looking good. one last question, which test would have caught this issue?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
```diff
@@ -108,6 +108,8 @@ PersistedFile loadTelemetriesFromDisk() {
       readFully(fileInputStream, ikeyBytes, 36);
       instrumentationKey = new String(ikeyBytes, UTF_8);
       if (!isInstrumentationKeyValid(instrumentationKey)) {
+        fileInputStream
+            .close(); // need to close FileInputStream before delete; otherwise, delete will fail.
```
trask, line 112: can you add a test that would have caught this?

heyams:
> looking good. one last question, which test would have caught this issue?

how to mock a readonly system?

trask:
> > looking good. one last question, which test would have caught this issue?
> 
> how to mock a readonly system? we might need to add a smoke test for readonly?

sorry, by "this issue" I mean the one that this PR addresses

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
heyams: it shouldn't happen.. but if it does somehow like corrupted data, it should be deleted then?

trask: 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileLoader.java
heyams: added.

trask (approved):

PR #1957, opened 2021-11-10
Add logging when purging old telemetry files

trask:
Important to log since this is data loss


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFilePurger.java
```diff
@@ -79,14 +79,21 @@ public void run() {
...
           logger.warn(
               "Fail to delete the expired {} from folder '{}'.", file.getName(), folder.getName());
+        } else {
+          numDeleted++;
         }
       }
     }
+    if (numDeleted > 0) {
+      logger.warn(
+          "{} telemetry file(s) were unable to be ingested and expired after 48 hours", numDeleted);
```
heyams, line 95: let's add folder name to be more clear.

heyams (approved):

PR #1960, opened 2021-11-12
Add span kind to sampling overrides

trask:
it's currently hard to suppress all outgoing http client spans, since those have essentially the same attributes as incoming http server spans.

attaching snapshot for testing: [applicationinsights-agent-3.2.4-BETA-SNAPSHOT.jar.zip](https://github.com/microsoft/ApplicationInsights-Java/files/7540180/applicationinsights-agent-3.2.4-BETA-SNAPSHOT.jar.zip)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -62,6 +63,25 @@ private static boolean isEmpty(String str) {
     return str == null || str.trim().isEmpty();
   }
 
+  public enum SpanKind {
+    @JsonProperty("server")
+    SERVER(io.opentelemetry.api.trace.SpanKind.SERVER),
```
heyams, line 68: is io.opentelemetry.api.trace.SpanKind an enum? can we reuse it rather than create a new one?  if their spankind is updated, we shouldn't need to update ours.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -373,6 +393,8 @@ private static String getDefaultPath() {
   }
 
   public static class SamplingOverride {
+    // TODO (trask) consider making this required when moving out of preview
+    @Nullable public SpanKind spanKind;
```
heyams, line 397: declare it as io.opentelemetry.api.trace.SpanKind?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: our convention is for json enum values to be lowercase, and we can't add `@JsonProperty` to those classes since they are not under our control. I will add a TODO marker here to investigate other options for mapping external annotations to lowercase.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: see comment above

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingOverridesTest.java
```diff
@@ -44,157 +46,243 @@ void shouldSampleByDefault() {
...
+
+    // expect
+    assertThat(sampler.getOverride(SpanKind.SERVER, attributes).getPercentage()).isEqualTo(0);
+  }
+
+  @Test
+  void shouldFilterStrictMatchWithNullSpanKind() {
+    // given
+    List<SamplingOverride> overrides =
+        singletonList(newOverride(null, 0, newStrictAttribute("one", "1")));
```
heyams, line 91: when it's null, it will match any span kind?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingOverridesTest.java
```diff
@@ -44,157 +46,243 @@ void shouldSampleByDefault() {
...
     SamplingOverrides sampler = new SamplingOverrides(overrides);
     Attributes attributes = Attributes.of(AttributeKey.stringKey("one"), "1");
 
     // expect
-    assertThat(sampler.getOverride(attributes).getPercentage()).isEqualTo(0);
+    assertThat(sampler.getOverride(SpanKind.SERVER, attributes).getPercentage()).isEqualTo(0);
```
heyams, line 96: assertThat(sampler.getOverride(SpanKind.CLIENT, attributes).getPercentage))).isEqualTo(0); should work too?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingOverridesTest.java
trask: yes

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingOverridesTest.java
trask: yes

heyams (approved):

PR #1962, opened 2021-11-12
Sort persisted files when resending from disk

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/localstorage/LocalFileCache.java
```diff
@@ -61,4 +67,37 @@ String poll() {
...
+    try {
+      Collections.sort(result, comparator);
+    } catch (RuntimeException ex) {
+      LoggerFactory.getLogger(LocalFileCache.class)
+          .error("Fail to sort a list of persisted files on disk", ex);
+      return null;
+    }
+
+    return result;
+  }
```
trask, line 102: I think ok to keep it simple and just sort by `lastModified`, may not be perfect, but good enough since nothing critically bad will happen if timestamps are incorrect

```suggestion
  private static List<File> sortPersistedFiles(File folder) {
    return FileUtils.listFiles(folder, new String[] {"trn"}, false).stream()
        .sorted(Comparator.comparing(File::lastModified))
        .collect(Collectors.toList());
  }
```

or, for the more complex implementation, something like this:

```suggestion
  private static List<File> sortPersistedFiles(File folder) {
    return FileUtils.listFiles(folder, new String[] {"trn"}, false).stream()
        .sorted(Comparator.comparing(LocalFileCache::timestampFromFileName))
        .collect(Collectors.toList());
  }

  // visible for testing
  static long timestampFromFileName(File file) {
    int endIndex = file.getName().lastIndexOf('-');
    if (endIndex == -1) {
      return 0;
    }
    String timestamp = file.getName().substring(0, endIndex);
    try {
      return Long.parseLong(timestamp);
    } catch (NumberFormatException e) {
      return 0;
    }
  }
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/AggregatingLogger.java
```diff
@@ -177,6 +178,17 @@ public void run() {
                 });
         logger.warn(message.toString());
       }
+
+      if (numSuccesses != 0) {
```
heyams, line 182: ignore this file. i will revert it. push by mistake. logging success might be too much noise.

trask (approved):

PR #1966, opened 2021-11-16
Don't send sampleRate 100

trask:
Sample rate 100 is the default, can reduce ingestion cost by not sending it.


kryalama (approved):

heyams:
can we add a test to cover that payload doesn't include sampling rate when it's by default?

heyams (approved):
👍

PR #1967, opened 2021-11-17
Fix status exception in a readonly file system

heyams:


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/status/StatusFile.java
```diff
@@ -140,7 +140,12 @@ private StatusFile() {}
 
   // visible for testing
   static boolean shouldWrite() {
-    return DiagnosticsHelper.useAppSvcRpIntegrationLogging();
+    return DiagnosticsHelper.useAppSvcRpIntegrationLogging() && !isReadOnlyFileSystem();
```
trask, line 143: Let's init `shouldWrite` value in static initializer, then we can log once something like:

      startupLogger.info(
          "Detected running on a read-only file system, app service status json fiule won't be written. If this is unexpected, please check that the process has write access to the directory: "
              + dir.getAbsolutePath());

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/status/StatusFile.java
```diff
@@ -109,6 +112,14 @@
...
     logDir = initLogDir();
     directory = logDir + STATUS_FILE_DIRECTORY;
+    shouldWrite =
+        DiagnosticsHelper.useAppSvcRpIntegrationLogging() && new File(directory).canWrite();
+    if (!shouldWrite) {
+      LoggerFactory.getLogger(StatusFile.class)
+          .info(
+              "Detected running on a read-only file system. Status json file won't be created. If this is unexpected, please check that process has write access to the directory: {}",
+              directory);
+    }
```
trask, line 122: oh, I think this is too early to log. let's move this to `MainEntryPoint.start()` right after `startupLogger` is configured.

also, I think the condition here needs to be `DiagnosticsHelper.useAppSvcRpIntegrationLogging() && !shouldWrite`, otherwise it will be logged outside of app services also

heyams: shouldWrite includes `DiagnosticsHelper.useAppSvcRpIntegrationLogging()` this check though.

trask: the condition is `!shouldWrite`, so I believe it will include standalone deployments

heyams: MainEntryPoint.start invoked `StatusFile.putValueAndWrite`, which has the shouldWrite check.. i think it's redundant checking twice.  putValueAndWrite is also used elsewhere.  would it be more cleaner to check in all the callers?

before invoking StatusFile methods, check shouldWrite first. Now, i get your point. it should check `DiagnosticsHelper.useAppSvcRpIntegrationLogging()` again when `shouldWrite` is false to log only when it's app service

heyams: another option is to pass startupLogger to StatusFile.

trask: > would it be more cleaner to check in all the callers?

this sounds good, probably just need an AtomicBoolean to only log it once 👍

trask: oh right, I got confused, this is tricky because of the delayed logging init. ya, I kind of like that idea to pass the startupLogger in to StatusFile. if it's null we know it's too early

heyams: i changed it to check shouldWrite in all callers.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/status/StatusFile.java
```diff
@@ -109,6 +117,8 @@
 
     logDir = initLogDir();
     directory = logDir + STATUS_FILE_DIRECTORY;
+    shouldWrite =
+        DiagnosticsHelper.useAppSvcRpIntegrationLogging() && new File(directory).canWrite();
```
trask, line 121: I'm thinking now this may be too early to call `DiagnosticsHelper.useAppSvcRpIntegrationLogging()`, since that relies method relies on setAgentJarFile() which isn't called until MainEntryPoint.

Which is likely why it was in `shouldWrite()` method previously.

Let's cache `isReadOnly` here based on `new File(directory).canWrite()`, and then go back to `shouldWrite()` method.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/status/StatusFile.java
```diff
@@ -109,6 +116,8 @@
 
     logDir = initLogDir();
     directory = logDir + STATUS_FILE_DIRECTORY;
+    File dir = new File(directory);
+    isReadOnly = dir.canRead() && !dir.canWrite();
```
trask, line 120: I think we need to mkdirs here before testing `canRead` and `canWrite` (and then we can remove the mkdirs below)

also I think we can handle `canRead = false` better. maybe we should call the flag `notWritable` and make `notWritable = !dir.canWrite()`? otherwise if you can't read we are still trying to write

heyams: check logDir should work.

trask: > check logDir should work.

👍

I think there's still the separate issue when `dir.canRead()` is `false`. In that case, `isReadOnly` is `false` and we try to write.

maybe we should just remove that check? e.g.

```suggestion
    isReadOnly = !dir.canWrite();
```

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/status/StatusFile.java
```diff
@@ -138,9 +147,20 @@ public static String getLogDir() {
...
-  // visible for testing
-  static boolean shouldWrite() {
-    return DiagnosticsHelper.useAppSvcRpIntegrationLogging();
+  private static boolean shouldWrite() {
+    if (!DiagnosticsHelper.useAppSvcRpIntegrationLogging()) {
+      return false;
+    }
+    if (!writable) {
+      return true;
+    }
```
trask, line 156:
```suggestion
    if (writable) {
      return true;
    }
```

trask (approved):
one question above

Issue #1964, opened 2021-11-16
Add option to disable the logged warning if trace message exceeds Sanitizer.MAX_MESSAGE_LENGTH


Ramblurr:
This PR (https://github.com/microsoft/ApplicationInsights-Java/issues/1021) added warnings when telemetry is truncated. This is fine, but it should be possible to turn the warnings off, or disable the warnings for certain types of telemetry.

In our case we have some large sql queries that trigger this truncation and fill our logs with this warning.

### Expected behavior

We should be able to turn off the TelemetryTruncation logging.

### Actual behavior

Our logs are filled with warnings like:

`WARN  c.m.a.a.i.common.TelemetryTruncation - In the last 5 minutes, the following warning has occurred 188 times: Telemetry was truncated:`

### To Reproduce

1. Use the javaagent
2. Have long sql queries

#### Sample Application

n/a

### System information
Please provide the following information:
 - SDK Version: applicationinsights-agent-3.2.0.jar 
 - OS type and version: Debian Docker
 - Application Server type and version (if applicable): tomcat
 - Using spring-boot? Yes
 - Additional relevant libraries (with version, if applicable): n/a

### Logs

`WARN  c.m.a.a.i.common.TelemetryTruncation - In the last 5 minutes, the following warning has occurred 188 times: Telemetry was truncated:`

### Screenshots

n/a

trask:
hi @Ramblurr, thanks for reporting this!

I agree, the current situation doesn't seem great.

Would it work for you if we limit this to only be logged once (and to truncate the logged value to 80 characters)?

I have submitted PR #1968 to implement this approach (and to log more verbosely at `debug` level).

(we are trying to avoid configuration option proliferation where possible)

Ramblurr:
Thanks for the reply.

This sounds like a reasonable solution. But what does once mean exactly? Once per JVM execution? And once per all truncation notices or just once per trace type?

Happy with any combination of those, but would be good to understand the behavior.

trask:
@Ramblurr thx! it would be once per JVM execution. Well, up to max of 10 times per JVM execution if there are >= 10 different fields that overflow. In your case, if only sql queries are overflowing, it would just be once per JVM execution.

trask:
Resolved by #1968

trask:
[...]

PR #1968, opened 2021-11-18
Reduce noisy truncation logging

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
```diff
@@ -21,28 +21,60 @@
...
+      // this can be expected, so don't want to flood the logs with a lot of these
+      // (and don't want to log the full value, e.g. sql text > 8192 characters)
+      logger.warn(
+          "truncated {} attribute value to {} characters (this message will only be logged once"
+              + " per attribute name, and only for at most 10 different attribute names): {}",
+          attributeName,
+          maxLength,
+          trimTo80(value));
+    }
+    logger.debug(
```
heyams, line 50: should this in the else block?  without it, it will get log twice, warn and debug for the very first time.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
```diff
@@ -21,28 +21,60 @@
...
+    if (alreadyLoggedPropertyKeys.size() < 10 && !alreadyLoggedPropertyKeys.add(propertyKey)) {
+      // this can be expected, so don't want to flood the logs with a lot of these
+      logger.warn(
+          "truncated {} property value to {} characters (this message will only be logged once"
+              + " per property key, and only for at most 10 different property keys): {}",
+          propertyKey,
+          maxLength,
+          trimTo80(value));
+    }
+    logger.debug("truncated {} property value to {} characters: {}", propertyKey, maxLength, value);
```
heyams, line 68: same here.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
```diff
@@ -21,28 +21,60 @@
...
 
   public static String truncateTelemetry(String value, int maxLength, String attributeName) {
     if (value == null || value.length() <= maxLength) {
       return value;
     }
-    warningLogger.recordWarning("truncated " + attributeName + ": " + value);
+    if (!alreadyLoggedAttributeNames.add(attributeName)) {
+      // this can be expected, so don't want to flood the logs with a lot of these
+      // (and don't want to log the full value, e.g. sql text > 8192 characters)
+      logger.warn(
```
heyams, line 43: should we continue to use the AggregatingLogger? it will log for the first 5 mins..  
i think aggregation is still helpful for other customers..  when truncation happens occasionally.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
```diff
@@ -21,28 +21,60 @@
...
-  // which would be memory allocating
-  public static String truncatePropertyValue(String value, int maxLength, String key) {
+  public static String truncatePropertyValue(String value, int maxLength, String propertyKey) {
     if (value == null || value.length() <= maxLength) {
       return value;
     }
-    warningLogger.recordWarning("truncated property[" + key + "]: " + value);
+    if (alreadyLoggedPropertyKeys.size() < 10 && !alreadyLoggedPropertyKeys.add(propertyKey)) {
+      // this can be expected, so don't want to flood the logs with a lot of these
+      logger.warn(
```
heyams, line 61: this is almost identical to line 43. "property" and "attribute" are the only difference.
should we reword it a bit to make it reusable in both cases.

trask: ya, i tried a few variations, but ended up opting for clearer messages/code, with slight duplication

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
trask: i'm not sure i follow "it will log for the first 5 mins"?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
trask: the debug message has more detail (non-trimmed value)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
heyams: it's using a scheduler to run every 5 mins.  it's not possible to log once using aggregated logger unless to stop the scheduler, but we don't have any condition to stop it..  i see why switch from aggregated logger to regular logger.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/common/TelemetryTruncation.java
heyams: in that case, do we need to trim in warn logger level? 
since it's only logging once, i think it's fine to show details.

trask: I think logging an (expected) sql text > 8192 characters (like in #1964) could still be annoying enough for users to ask to suppress it (trying to make this warning not be so bad for the expected cases)

heyams (approved):

kryalama (approved):

PR #1977, opened 2021-11-24
Add telemetry processors masking feature

kryalama:
Added a new telemetry processor action "mask". This feature makes it easier for the customers to mask span or log attributes based on the attribute value regex.
Ex: 
 ```{
        "type": "attribute",
        "actions": [
          {
            "key": "http.url",
            "pattern": "^(?<uriNoCard>.*\\/cardid\\/)(?<cardStart>[0-9]{6})[0-9]{6}(?<cardEnd>[0-9]{4,6}).*",
            "replace": "${uriNoCard}****${cardEnd}",
            "action": "mask"
          }
        ],
        "id": "attributes/mask"
      }
```

Before Telemetry Processors:
http.url: http://example.com/cardid/1234562222227899

After Telemetry Processors:
http.url: http://example.com/cardid/****7899

TODO: Update documentation

### CODE REVIEW

File: agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AttributeProcessor.java [hunk 1]
```diff
@@ -167,6 +169,26 @@ private static SpanData processExtractAction(SpanData span, ProcessorAction acti
...
+    Matcher matcher = actionObj.maskAttribute.pattern.matcher(existingValue);
+    if (!matcher.matches()) {
+      return span;
+    }
+    AttributesBuilder builder = span.getAttributes().toBuilder();
+    String newValue = actionObj.maskAttribute.replace;
+    for (String groupName : actionObj.maskAttribute.groupNames) {
+      newValue = newValue.replaceAll(groupName, matcher.group(groupName));
+    }
+    newValue = newValue.replaceAll("\\$", "");
```
trask, line 187: wondering if this works, relying on Java regular expression's own replacement mechanism (may need to change replacement to `${name}`, e.g. `"${uriNoCard}****${cardEnd}"`)

```suggestion
    Matcher matcher = actionObj.maskAttribute.pattern.matcher(existingValue);
    String newValue = matcher.replaceAll(actionObj.maskAttribute.replace);
    if (newValue.equals(existingValue)) {
      return span;
    }
    AttributesBuilder builder = span.getAttributes().toBuilder();
```

kryalama: thanks for the suggestion. It looks more clean now

trask (approved):

PR #1979, opened 2021-11-25
Attach snapshot to build for testing

trask:


.github/workflows/pr.yaml
```diff
@@ -44,6 +44,10 @@ jobs:
       - name: Test
         # TODO enable build cache, either --build-cache here, or org.gradle.caching=true in gradle.properties
         run: ./gradlew check --stacktrace
+      - name: Upload snapshot
+        uses: actions/upload-artifact@v2
+        with:
+          path: agent/agent/build/libs/applicationinsights-agent-*-SNAPSHOT.jar
```
heyams, line 50: will there be a link available on the pr automatically?  just curious.

heyams (approved):

.github/workflows/pr.yaml
trask: yes! e.g. click into the build below, and go to "Summary" (upper left), then click on "Artifacts" (to the left)

PR #1984, opened 2021-11-28
Fix jackson initialization

trask:


heyams:
fix makes sense. just curious to know any customer issues regarding this? when it's not found, json is corrupted?

heyams (approved):

trask:
> just curious to know any customer issues regarding this?

this was not customer reported, I ran into it myself when testing a command-line app that just emitted telemetry via 2.x SDK (and was very confused about it for a while).

> when it's not found, json is corrupted?

exports fail because it can't find json mapping for `OffsetDateTime` which our model uses for timestamps

william-kirovski-kr:
I am seeing this issue in my logs. It is repeating over and over. I am temporarily suppressing this message, as it seems you have a fix in place for it.

`2021-12-06 21:51:05.447-05:00 WARN c.m.a.a.i.telemetry.TelemetryChannel - Sending telemetry to the ingestion service (telemetry will be stored to disk on failure and retried later):: Error encoding telemetry items: Java 8 date/time type `java.time.OffsetDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.microsoft.applicationinsights.agent.internal.exporter.models.TelemetryItem["time"]) (future warnings will be aggregated and logged once every 5 minutes`

`com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.OffsetDateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.microsoft.applicationinsights.agent.internal.exporter.models.TelemetryItem["time"])`

Do you know the impact, and when this fix will make it into a release?

trask:
hi @william-kirovski-kr! we are planning to a release with this included by end of this week.

I believe the impact is that you will not see any telemetry in the Portal if you are seeing this error.

PR #1986, opened 2021-11-28
Bump submodule to 1.9.0

trask:
Pulling in #1982


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
```diff
@@ -822,6 +822,22 @@ private void exportRequest(SpanData span) {
...
+    }
+    if (span.getStatus().getStatusCode() == StatusCode.OK) {
+      // auto-instrumentation never sets OK, so this is explicit user override
+      return true;
+    }
+    Long statusCode = span.getAttributes().get(SemanticAttributes.HTTP_STATUS_CODE);
+    if (statusCode == null) {
+      return true;
+    }
+    // override default OpenTelemetry mapping of status codes 4xx
```
heyams, line 837: what is the default opentelemetry status code mapping? what about redirect error code that is in the 300 range?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
```diff
@@ -43,7 +43,7 @@
   private static final Logger logger = LoggerFactory.getLogger(RedirectPolicy.class);
   public static final String INSTRUMENTATION_KEY = "instrumentationKey";
 
-  private final Cache<URL, String> redirectMappings = Cache.builder().setMaximumSize(100).build();
+  private final Cache<URL, String> redirectMappings = Cache.bounded(100);
```
heyams, line 46: typo: bound instead of bounded? 
i think it's "cache bound". That's up to opentelemetry sdk to update that.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/Exporter.java
trask: the default otel mapping is
* >= 500 --> `ERROR`
* otherwise --> `UNSET`

prior to 1.9.0, the default otel mapping was
* >= 400 --> `ERROR`
* otherwise --> `UNSET`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/httpclient/RedirectPolicy.java
trask: I think `bounded` is ok also, e.g. "it's a bounded cache"

heyams (approved):

PR #1987, opened 2021-11-29
DRAFT: Fix log sampling behavior

trask:
DRAFT b/c even though it's changing undocumented behavior, we should consider waiting to merge until ready for version bump to 3.3.0.

Logs being captured as spans is a temporary implementation detail that we need to hide, since it will be changing once OpenTelemetry logs are stable and can be used instead.

Currently we only support single default behavior when it comes to log sampling:
* If a log is not part of a trace, capture it.
* If a log is part of a trace, sample it if the trace is being sampled.


trask:
Closing, this was implemented in 3.3.0 when we changed to real OTel logs

PR #1996, opened 2021-12-09
fix duration in live metrics sample telemetry feature

kryalama:


kryalama:
Working on follow up PR to update the same in exporter

trask (approved):
👍

PR #2002, opened 2021-12-13
Update 2.x SDK log4j2 appender to not pull in log4j2 on its own

trask:
The 2.x SDK log4j2 appender is designed for users who are already using log4j2. It should not pull in log4j2 on its own. Users should already be bringing their own version of log4j2 (and should be upgrading that version or applying the mitigation steps from [CVE-2021-44228](https://msrc-blog.microsoft.com/2021/12/11/microsofts-response-to-cve-2021-44228-apache-log4j2/)). The PR ensures that users *have* to bring their own version of log4j2.

### DIFF
--- a/logging/log4j2/build.gradle
+++ b/logging/log4j2/build.gradle
@@ -33,6 +33,6 @@ projectPomDescription = "This module provides a $project.msftAppInsights appende
 // endregion Publishing properties
 
 dependencies {
-    compile group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.15.0'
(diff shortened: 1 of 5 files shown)


heyams (approved):

heyams:
can we drop java 7 support for legacy sdk? that's feature change for some customers... and we don't know how many were on java  7.

heyams:
should we add another statsbeat feature for java version? i think that's more useful than java vendor

trask:
> should we add another statsbeat feature for java version? i think that's more useful than java vendor

👍 it's high cardinality if we capture the full version, but maybe just the major versions 8, 11, 17, and "other"

PR #2011, opened 2021-12-14
Track java version via Statsbeat

heyams:
https://github.com/microsoft/ApplicationInsights-Java/pull/2002#issuecomment-993007916

System.getProperty("java.specification.version") is exactly what we need in this case.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -43,8 +43,8 @@
     // track java distribution
     super(customDimensions);
     this.type = type;
-    String javaVendor = System.getProperty("java.vendor");
-    featureList.add(Feature.fromJavaVendor(javaVendor));
+    featureList.add(Feature.fromJavaVendor(System.getProperty("java.vendor")));
+    featureList.add(Feature.fromJavaVendor(System.getProperty("java.specification.version")));
```
heyams, line 47: will update this.. not fromJavavendor

PR #2012, opened 2021-12-14
Use java.specification.version

heyams:
#2002 (comment)

we did track java.version already.  update it to send  'java.specification.version' to reduce payload.


heyams:
Per offline discussion, we will keep the minor version for now.. it might be useful at some point.

PR #2013, opened 2021-12-15
Fix kafka smoke tests

trask:


trask:
the cp-kafka image was fixed, so this PR is no longer needed 🎉

PR #2033, opened 2022-01-07
SDL compliance

heyams:
Fix https://msazure.visualstudio.com/One/_workitems/edit/12921260

Add CodeQL


trask (approved):

trask:
@heyams we have CodeQL now running daily in github, can we close this?

PR #2035, opened 2022-01-08
Resolve component compliance

heyams:
Fix all the issues found here: https://github-private.visualstudio.com/microsoft/_componentGovernance/161106?_a=alerts&typeId=7302170&alerts-view-option=active

Fix https://msazure.visualstudio.com/One/_workitems/edit/12921256


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/CustomDimensions.java
```diff
@@ -63,7 +63,7 @@
     }
 
     sdkVersion = qualifiedSdkVersion.substring(qualifiedSdkVersion.lastIndexOf(':') + 1);
-    runtimeVersion = System.getProperty("java.version");
+    runtimeVersion = System.getProperty("java.specification.version");
```
trask, line 66: is this intentional?

trask (commented):
it looks like you also updated the submodule in this PR, probably not intentional?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/CustomDimensions.java
heyams: No. not sure why this kept popping up.. i rebased that commit already.. will remove it..

trask (approved):

PR #2037, opened 2022-01-08
http server 4xx configuration

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -228,6 +228,7 @@ public String toString() {
     // this is needed to unblock customer, but is not the ideal long-term solution
     // https://portal.microsofticm.com/imp/v3/incidents/details/266992200/home
     public boolean disablePropagation;
+    public boolean captureHttpServer4xxAsError = true;
```
heyams, line 231: i saw the IcM 282327358 that is asking to treat 4xx response as success not error. rename?

trask: I think this naming is a bit better from otel perspective, as the default in otel is not to capture server 4xx as error (but also not to capture it as success, there's a third option in otel - [UNSET](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L266-L274))

heyams: this is the config customer will be using.. if they wants 4xx as success, they will set this flag to false.  unset will return true. yeah, it works.

heyams (approved):
azure doc needs to be updated.

heyams:
let's track this via statsbeat.

PR #2040, opened 2022-01-10
Bundle OTLP exporters

trask:
Now that otlp exporters are very small, we can bundle them.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/ConfigOverride.java
```diff
@@ -122,20 +122,9 @@ static Config getConfig(Configuration config) {
           "Request-Context");
     } else {
       properties.put("otel.traces.exporter", tracesExporter);
-      // when using another exporter, populate otel.service.name and otel.resource.attributes
-      if (config.role.name != null) {
-        properties.put("otel.service.name", config.role.name);
-      }
-      String resourceAttributes = getProperty("otel.resource.attributes");
```
heyams, line 129: do we need to keep "otel.resource.attributes"?

trask: the OpenTelemetry SDK changed recently, and now picks up env vars and system properties even if we don't set them in our `properties`, so this isn't needed anymore

heyams (approved):

PR #2042, opened 2022-01-11
Resolve component compliance for legacy-sdk

heyams:
Fix issues [here](https://github-private.visualstudio.com/microsoft/_componentGovernance/161106?_a=alerts&typeId=6520417&alerts-view-option=active)


trask (approved):
👍

PR #2047, opened 2022-01-12
Stop collecting oshi perf counters when initialization fails

heyams:
Fix IcM 253155448, 276640835.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/OshiPerformanceCounter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/OshiPerformanceCounter.java
@@ -31,6 +31,7 @@ import oshi.hardware.CentralProcessor;
 import oshi.hardware.CentralProcessor.TickType;
 import oshi.software.os.OSProcess;
 import oshi.software.os.OperatingSystem;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class OshiPerformanceCounter implements PerformanceCounter {
 
@@ -45,6 +46,7 @@ public class OshiPerformanceCounter implements PerformanceCounter {
 
   private volatile OSProcess processInfo;
   private volatile CentralProcessor processor;
+  private static final AtomicBoolean hasError = new AtomicBoolean();
 
   @Override
   public String getId() {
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -55,10 +57,22 @@ public String getId() {
...
+        // e.g. icm 276640835: ExceptionInInitializerError
+        hasError.set(true);
+        return;
+      }
+    }
+
+    // stop collecting oshi perf counters when initialization fails.
+    if (hasError.get()) {
+      return;
     }
```
trask, line 76: should this go at the beginning of the method?

heyams: it doesn't make much diff.. except it avoids checking if it has been init or not..  i can move it.

trask: it will never get here tho?

heyams: 1st init -> has error then return, 
2nd processInfo is not null -> hit line 74

trask: `processorInfo` will still be null on 2nd, because exception was thrown before it could be set

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/OshiPerformanceCounter.java
```diff
@@ -53,12 +55,24 @@ public String getId() {
...
-      SystemInfo systemInfo = new SystemInfo();
-      OperatingSystem osInfo = systemInfo.getOperatingSystem();
-      processInfo = osInfo.getProcess(osInfo.getProcessId());
-      processor = systemInfo.getHardware().getProcessor();
+      try {
+        SystemInfo systemInfo = new SystemInfo();
+        OperatingSystem osInfo = systemInfo.getOperatingSystem();
+        processInfo = osInfo.getProcess(osInfo.getProcessId());
+        processor = systemInfo.getHardware().getProcessor();
+      } catch (Error ex) {
```
trask, line 70: let's log this at `debug` level

trask (approved):

PR #2061, opened 2022-01-20
[RP Integration] Stop sending requests to all endpoints when ikey is empty

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
```diff
@@ -87,7 +87,7 @@ private static void mapToConnectionConfiguration(
     // get ikey
     String instrumentationKey = kvps.get(Keywords.INSTRUMENTATION_KEY);
     if (Strings.isNullOrEmpty(instrumentationKey)) {
-      throw new InvalidConnectionStringException("Missing '" + Keywords.INSTRUMENTATION_KEY + "'");
+      logger.warn("New connection string is null or empty");
```
trask, line 90: I didn't follow why this change?

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
heyams: we don't want to throw an exception when instrumentation key is empty.  null or empty is accepted.

trask: maybe in `parseInto` above we can have a separate path empty connection string, that way we don't lose this validation when connection string is not empty, and hopefully we don't need to `warn` in an expected case

separately, i noticed in this method we probably want to reset all the urls in `EndpointProvider` to the defaults before we start populating them, otherwise we could pick up some values left over from the prior connection string

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
```diff
@@ -39,9 +39,19 @@
 
   private ConnectionString() {}
 
-  public static void parseInto(String connectionString, TelemetryClient targetConfig)
+  public static void parseInto(String connectionString, TelemetryClient telemetryClient)
       throws InvalidConnectionStringException {
-    mapToConnectionConfiguration(getKeyValuePairs(connectionString), targetConfig);
+    if (Strings.isNullOrEmpty(connectionString)) {
+      telemetryClient.setInstrumentationKey(connectionString);
```
trask, line 45: same thing, just since it looks odd to use connection string as an ikey
```suggestion
      telemetryClient.setInstrumentationKey(null);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
```diff
@@ -39,9 +39,19 @@
...
+
+      // reset endpoint to default for ingestion, live metric and profiler
+      EndpointProvider endpointProvider = telemetryClient.getEndpointProvider();
+      endpointProvider.setIngestionEndpoint(endpointProvider.getIngestionEndpoint());
```
trask, line 49: this looks like it's just setting the same value back?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/EndpointProvider.java
```diff
@@ -76,6 +73,14 @@ public URL getAppIdEndpointUrl(String instrumentationKey) {
     return buildAppIdUrl(instrumentationKey);
   }
 
+  // reset endpoint URLs for ingestion, live metric and profiler
+  public void resetEndpointUrls() throws MalformedURLException {
```
trask, line 77: 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
heyams: yup, please check the update. i pushed too quick.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/ConnectionString.java
```diff
@@ -39,9 +39,15 @@
...
 
-  public static void parseInto(String connectionString, TelemetryClient targetConfig)
-      throws InvalidConnectionStringException {
-    mapToConnectionConfiguration(getKeyValuePairs(connectionString), targetConfig);
+  public static void parseInto(String connectionString, TelemetryClient telemetryClient)
+      throws InvalidConnectionStringException, MalformedURLException {
+    if (Strings.isNullOrEmpty(connectionString)) {
+      telemetryClient.setInstrumentationKey(null);
+      // reset endpoint to default for ingestion, live metric and profiler
+      telemetryClient.getEndpointProvider().resetEndpointUrls();
```
trask, line 47: we should do this also in `mapToConnectionConfiguration` below before starting to overwrite them, since we only conditionally overwrite them below

heyams: in maptoconnectionConfiguration, we should reset all? the above doesn't impact statsbeat and profiler.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/quickpulse/QuickPulsePingSender.java
```diff
@@ -91,7 +91,7 @@ public QuickPulsePingSender(
 
   public QuickPulseHeaderInfo ping(String redirectedEndpoint) {
     String instrumentationKey = getInstrumentationKey();
-    if (instrumentationKey == null && "java".equals(System.getenv("FUNCTIONS_WORKER_RUNTIME"))) {
+    if (Strings.isNullOrEmpty(instrumentationKey)) {
```
kryalama, line 94: just curios, is there a reason the function runtime check was removed?

heyams: it is no longer applied to function only. azure spring cloud with empty ikey, no live metrics.  that's why we drop that check.

trask (approved):

PR #2062, opened 2022-01-20
Separate queue for metrics to avoid dropping telemetry

trask:
Creates a separate metrics channel batcher, since metrics are bursty (e.g. reported once a minute) and small so good to have a larger queue size for them.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
@@ -244,7 +244,10 @@ public class Configuration {
 
     public List<InstrumentationKeyOverride> instrumentationKeyOverrides = new ArrayList<>();
 
-    public int exportQueueCapacity = 2048;
+    public int generalExportQueueCapacity = 2048;
+    // metrics get flooded every 60 seconds by default, so need larger queue size to avoid dropping
+    // telemetry (they are much smaller so a larger queue size is ok)
+    public int metricsExportQueueCapacity = 65536;
   }
 
   public static class InheritedAttribute {
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
@@ -61,7 +61,8 @@ public final class BatchSpanProcessor {
       long scheduleDelayNanos,
       int maxQueueSize,
       int maxExportBatchSize,
-      long exporterTimeoutNanos) {
+      long exporterTimeoutNanos,
+      String queueName) {
     MpscArrayQueue<TelemetryItem> queue = new MpscArrayQueue<>(maxQueueSize);
     this.worker =
         new Worker(
@@ -70,7 +71,8 @@ public final class BatchSpanProcessor {
             maxExportBatchSize,
             exporterTimeoutNanos,
             queue,
-            queue.capacity());
+            queue.capacity(),
+            queueName);
     Thread workerThread = new DaemonThreadFactory(WORKER_THREAD_NAME).newThread(worker);
     workerThread.start();
   }
@@ -103,6 +105,7 @@ public final class BatchSpanProcessor {
 
     private final Queue<TelemetryItem> queue;
     private final int queueCapacity;
+    private final String queueName;
     // When waiting on the spans queue, exporter thread sets this atomic to the number of more
     // spans it needs before doing an export. Writer threads would then wait for the queue to reach
     // spansNeeded size before notifying the exporter thread about new entries.
@@ -124,13 +127,15 @@ public final class BatchSpanProcessor {
         int maxExportBatchSize,
         long exporterTimeoutNanos,
         Queue<TelemetryItem> queue,
-        int queueCapacity) {
+        int queueCapacity,
+        String queueName) {
       this.spanExporter = spanExporter;
       this.scheduleDelayNanos = scheduleDelayNanos;
       this.maxExportBatchSize = maxExportBatchSize;
       this.exporterTimeoutNanos = exporterTimeoutNanos;
       this.queue = queue;
       this.queueCapacity = queueCapacity;
+      this.queueName = queueName;
       this.signal = new ArrayBlockingQueue<>(1);
       this.batch = new ArrayList<>(this.maxExportBatchSize);
     }
@@ -138,11 +143,16 @@ public final class BatchSpanProcessor {
     private void addSpan(TelemetryItem span) {
       if (!queue.offer(span)) {
         queuingSpanLogger.recordFailure(
-            "Max export queue size of "
+            "Max "
+                + queueName
+                + " export queue capacity of "
                 + queueCapacity
-                + " has been hit, dropping a span (max export queue size can be increased"
-                + " in the applicationinsights.json configuration file, e.g."
-                + " { \"preview\": { \"exportQueueCapacity\": "
+                + " has been hit, dropping a telemetry record (max "
+                + queueName
+                + " export queue capacity can be increased in the applicationinsights.json"
+                + " configuration file, e.g. { \"preview\": { \""
+                + queueName
+                + "ExportQueueCapacity\": "
                 + (queueCapacity * 2)
                 + " } }");
       } else {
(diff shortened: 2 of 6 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -244,7 +244,10 @@ public String toString() {
 
     public List<InstrumentationKeyOverride> instrumentationKeyOverrides = new ArrayList<>();
 
-    public int exportQueueCapacity = 2048;
+    public int generalExportQueueCapacity = 2048;
+    // metrics get flooded every 60 seconds by default, so need larger queue size to avoid dropping
+    // telemetry (they are much smaller so a larger queue size is ok)
+    public int metricsExportQueueCapacity = 65536;
```
heyams, line 250: 32 times more.. how did you come up with this number?

heyams: how did you find out it got dropped? i wonder if we can utilize statsbeat to catch it.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -244,7 +244,10 @@ public String toString() {
 
     public List<InstrumentationKeyOverride> instrumentationKeyOverrides = new ArrayList<>();
 
-    public int exportQueueCapacity = 2048;
+    public int generalExportQueueCapacity = 2048;
```
heyams, line 247: can we rename it to nonmetricExportQueueCapacity?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -182,7 +185,11 @@ public void trackAsync(TelemetryItem telemetry) {
     // batching, retry, throttling, and writing to disk on failure occur downstream
     // for simplicity not reporting back success/failure from this layer
     // only that it was successfully delivered to the next layer
-    getChannelBatcher().trackAsync(telemetry);
+    if (data instanceof MetricsData) {
```
heyams, line 188: Should it be MetricData instead of MetricsData?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -193,50 +200,69 @@ public void trackStatsbeatAsync(TelemetryItem telemetry) {
...
-                  ikeyEndpointMap,
-                  statsbeatModule,
-                  false,
-                  aadAuthentication);
-
-          if (!readOnlyFileSystem) {
-            LocalFileSender.start(localFileLoader, channel);
-          }
+        if (generalChannelBatcher == null) {
+          generalChannelBatcher = initChannelBatcher(generalExportQueueCapacity, 512, "general");
```
kryalama, line 214: curios why the queue size if 512 here? I thought the default size is 2048?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -193,50 +200,69 @@ public void trackStatsbeatAsync(TelemetryItem telemetry) {
...
 
-          channelBatcher =
-              BatchSpanProcessor.builder(channel).setMaxQueueSize(exportQueueCapacity).build();
+  // metrics get flooded every 60 seconds by default, so need much larger queue size to avoid
+  // dropping telemetry (they are much smaller so a larger queue size and larger batch size are ok)
+  private BatchSpanProcessor getMetricsChannelBatcher() {
+    if (metricsChannelBatcher == null) {
+      synchronized (channelInitLock) {
+        if (metricsChannelBatcher == null) {
+          metricsChannelBatcher = initChannelBatcher(metricsExportQueueCapacity, 2048, "metrics");
```
kryalama, line 227: same question on size

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -266,7 +292,7 @@ public BatchSpanProcessor getStatsbeatChannelBatcher() {
             LocalFileSender.start(localFileLoader, channel);
           }
 
-          statsbeatChannelBatcher = BatchSpanProcessor.builder(channel).build();
+          statsbeatChannelBatcher = BatchSpanProcessor.builder(channel).build("statsbeat");
```
heyams, line 295: Statsbeat is MetricData too and it won't hit the queue capacity since it's using a much lower frequency.. every 15 mins and 24 hours.. but we do need a separate channel for statsbeat.. i wonder if we can keep statsbeat channel the old capacity since it doesn't apply to statsbeat metrics.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: 512 is the max batch size not the queue size

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: nothing changed for the statsbeat channel

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: same answer :)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: are u suggesting renaming it?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: > how did you find out it got dropped?

we log when we drop something

> i wonder if we can utilize statsbeat to catch it.

sure (but not in this PR) 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: I think `general` is ok too

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
heyams: in the metric data.. base type is always 'MetricData'.  maybe rename it later.  I didn't realized we use MetricsData as the data model.

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
```diff
@@ -124,25 +127,32 @@ private Worker(
...
+                + queueName
+                + " export queue capacity of "
                 + queueCapacity
-                + " has been hit, dropping a span (max export queue size can be increased"
-                + " in the applicationinsights.json configuration file, e.g."
-                + " { \"preview\": { \"exportQueueCapacity\": "
+                + " has been hit, dropping a telemetry record (max "
+                + queueName
+                + " export queue capacity can be increased in the applicationinsights.json"
+                + " configuration file, e.g. { \"preview\": { \""
```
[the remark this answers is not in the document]
trask: this is a good idea, we can do in a follow-up

PR #2064, opened 2022-01-21
Fix ingestion error on missing exception message

trask:
Fixes issue where breeze rejects exception telemetry due to empty message:

```
{"itemsReceived":2,"itemsAccepted":1,"errors":[{"index":1,"statusCode":400,"message":"109: Field 'message' on type 'ExceptionDetails' is required but missing or empty. Expected: string, Actual:  "}]}
```


heyams:
can we add a test?

heyams (approved):

PR #2065, opened 2022-01-21
Log ingestion error on 206 response

trask:
Related to #2064, which we never noticed since we didn't log it.

Now the error in #2064 would be logged as:

```
2022-01-21 15:30:10.108-08:00 WARN  c.m.a.a.i.telemetry.TelemetryChannel - Sending telemetry to the ingestion service (telemetry will be stored to disk on failure and retried later): 109: Field 'message' on type 'ExceptionDetails' is required but missing or empty. Expected: string, Actual: undefined (future warnings will be aggregated and logged once every 5 minutes)
```

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
@@ -62,6 +62,7 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 // TODO performance testing
 public class TelemetryChannel {
@@ -72,6 +73,7 @@ public class TelemetryChannel {
 
   private static final AppInsightsByteBufferPool byteBufferPool = new AppInsightsByteBufferPool();
 
+  // TODO (heya) should we suppress logging statsbeat telemetry ingestion issues?
   private static final OperationLogger operationLogger =
       new OperationLogger(TelemetryChannel.class, "Sending telemetry to the ingestion service");
 
@@ -81,7 +83,7 @@ public class TelemetryChannel {
 
   // TODO (kryalama) do we still need this AtomicBoolean, or can we use throttling built in to the
   //  operationLogger?
-  private static final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
+  private final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
   @SuppressWarnings("CatchAndPrintStackTrace")
   private static ObjectMapper createObjectMapper() {
@@ -289,6 +291,7 @@ public class TelemetryChannel {
     return response ->
         response
             .getBodyAsString()
+            .switchIfEmpty(Mono.just(""))
             .subscribe(
                 body -> {
                   int statusCode = response.getStatusCode();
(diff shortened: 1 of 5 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -302,47 +309,52 @@ private void writeToDiskOnFailure(List<ByteBuffer> byteBuffers, String instrumen
...
-          statsbeatModule.getNetworkStatsbeat().incrementRetryCount(instrumentationKey);
-        }
-        break;
       default:
-        // ok
+        operationLogger.recordFailure("received response code: " + statusCode);
+    }
+  }
+
+  private static String getErrorMessageFromPartialSuccessResponse(String body) {
```
heyams, line 343: i wonder how do you generate 206 error code? i haven't run into it..

trask: see #2064

I ran into it doing some testing yesterday:

```
Span span = tracer.spanBuilder("test").startSpan();
span.recordException(new Throwable());
span.end();
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -302,47 +309,52 @@ private void writeToDiskOnFailure(List<ByteBuffer> byteBuffers, String instrumen
...
-        // no need to write to disk again when failing to send raw bytes from the persisted file
-        if (!persisted) {
-          writeToDiskOnFailure(byteBuffers, instrumentationKey);
-        }
-        break;
       case 408: // REQUEST TIMEOUT
       case 500: // INTERNAL SERVER ERROR
       case 503: // SERVICE UNAVAILABLE
       case 429: // TOO MANY REQUESTS
+        // TODO (heya) should we write to disk on any of these response codes?
```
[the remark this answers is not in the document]
heyams: can you update the todo to track data drop via statsbeat?

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -234,115 +257,133 @@ private CompletableResultCode internalSend(
...
+                    case 408: // REQUEST TIMEOUT
+                    case 429: // TOO MANY REQUESTS
+                    case 500: // INTERNAL SERVER ERROR
+                    case 503: // SERVICE UNAVAILABLE
+                      operationLogger.recordFailure(
+                          "received response code "
+                              + statusCode
+                              + " (telemetry will be stored to disk and retried later)");
+                      onFailure.accept(true);
+                      break;
```
trask, line 314: this is the list of response codes that 2.x used to retry on, I think it's a good list, not sure when we diverged from it

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/localstorage/IntegrationTests.java
```diff
@@ -84,23 +87,26 @@ public void setup() throws Exception {
                       () ->
                           new Exception("this is expected to be logged by the operation logger")));
     } else {
-      // 401 and 403 response codes are the only ones the result in storing to disk
+      // 408, 429, 500, and 503 response codes are the only ones the result in storing to disk
```
heyams, line 90: can we add 401 and 403 back?

heyams (approved):

PR #2068, opened 2022-01-22
Fix CI failures on downloading from download.jboss.org

trask:


trask:
Closing, merged this fix into #2066

PR #2073, opened 2022-01-26
Load connection string from a file

### LINKED ISSUE #2069: Allow reading the connection string from a file
I would like to be able to separate the connection string information from the general configuration and not check it into source code as we have different instances of application insights for different environments

A environment variable and system property that point to the file to read just the Connection string information from

- Currently I'm having to store the whole configuration in Azure Key Vault so that I can mount it as a file to be read by the application using the `APPLICATIONINSIGHTS_CONFIGURATION_FILE` override.

**Additional context**
This is in a container on AKS using the [csi-secrets-store-driver](https://docs.microsoft.com/en-us/azure/aks/csi-secrets-store-driver)

We're running in distroless containers with no shell so can't run an init script to hack around this

heyams:
Fix #2069

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -356,7 +356,6 @@ public class ConfigurationBuilder {
     // load connection string from a file if connection string is in the format of
     // "${file:mounted_connection_string_file.txt}"
     StringSubstitutor stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
-    stringSubstitutor.setEnableSubstitutionInVariables(true);
     config.connectionString =
         overlayConnectionStringFromEnv(stringSubstitutor.replace(config.connectionString));
     if (isTrimEmpty(config.role.name)) {
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -40,6 +40,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import org.apache.commons.text.StringSubstitutor;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.slf4j.LoggerFactory;
 
@@ -352,14 +353,10 @@ public class ConfigurationBuilder {
 
   // visible for testing
   static void overlayFromEnv(Configuration config) throws IOException {
-    // load connection string from a file
-    // connectionString: "${file:mounted_connection_string_file.txt}"
-    String connectionString = FileStringLookup.INSTANCE.lookup(config.connectionString);
-    if (connectionString == null) {
-      connectionString = config.connectionString;
-    }
-
-    config.connectionString = overlayConnectionStringFromEnv(connectionString);
+    // load connection string from a file if connection string is in the format of "${file:mounted_connection_string_file.txt}"
+    StringSubstitutor stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
+    stringSubstitutor.setEnableSubstitutionInVariables(true);
+    config.connectionString = overlayConnectionStringFromEnv(stringSubstitutor.replace(config.connectionString));
     if (isTrimEmpty(config.role.name)) {
       // only use WEBSITE_SITE_NAME as a fallback
       config.role.name = getWebsiteSiteNameEnvVar();
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -37,10 +37,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import org.apache.commons.text.StringSubstitutor;
+import org.apache.commons.text.lookup.StringLookup;
+import org.apache.commons.text.lookup.StringLookupFactory;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.slf4j.LoggerFactory;
 
@@ -355,7 +358,11 @@ public class ConfigurationBuilder {
   static void overlayFromEnv(Configuration config) throws IOException {
     // load connection string from a file if connection string is in the format of
     // "${file:mounted_connection_string_file.txt}"
-    StringSubstitutor stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
+    Map<String, StringLookup> stringLookupMap =
+        Collections.singletonMap(StringLookupFactory.KEY_FILE, FileStringLookup.INSTANCE);
+    StringLookup stringLookup =
+        StringLookupFactory.INSTANCE.interpolatorStringLookup(stringLookupMap, null, false);
+    StringSubstitutor stringSubstitutor = new StringSubstitutor(stringLookup);
     config.connectionString =
         overlayConnectionStringFromEnv(stringSubstitutor.replace(config.connectionString));
     if (isTrimEmpty(config.role.name)) {
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -353,10 +353,12 @@ public class ConfigurationBuilder {
 
   // visible for testing
   static void overlayFromEnv(Configuration config) throws IOException {
-    // load connection string from a file if connection string is in the format of "${file:mounted_connection_string_file.txt}"
+    // load connection string from a file if connection string is in the format of
+    // "${file:mounted_connection_string_file.txt}"
     StringSubstitutor stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
     stringSubstitutor.setEnableSubstitutionInVariables(true);
-    config.connectionString = overlayConnectionStringFromEnv(stringSubstitutor.replace(config.connectionString));
+    config.connectionString =
+        overlayConnectionStringFromEnv(stringSubstitutor.replace(config.connectionString));
     if (isTrimEmpty(config.role.name)) {
       // only use WEBSITE_SITE_NAME as a fallback
       config.role.name = getWebsiteSiteNameEnvVar();
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -352,8 +352,14 @@ public class ConfigurationBuilder {
 
   // visible for testing
   static void overlayFromEnv(Configuration config) throws IOException {
-    config.connectionString = overlayConnectionStringFromEnv(config.connectionString);
+    // load connection string from a file
+    // connectionString: "${file:mounted_connection_string_file.txt}"
+    String connectionString = FileStringLookup.INSTANCE.lookup(config.connectionString);
+    if (connectionString == null) {
+      connectionString = config.connectionString;
+    }
 
+    config.connectionString = overlayConnectionStringFromEnv(connectionString);
     if (isTrimEmpty(config.role.name)) {
       // only use WEBSITE_SITE_NAME as a fallback
       config.role.name = getWebsiteSiteNameEnvVar();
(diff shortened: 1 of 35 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.annotation.Nullable;
+import org.apache.commons.text.lookup.StringLookup;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class FileStringLookup implements StringLookup {
+
+  private static final Logger logger = LoggerFactory.getLogger(FileStringLookup.class);
+  private static final String PREFIX = "${file:";
+  private static final String LESS_STRICT_PREFIX = "file:";
```
ramthi, line 38: What is the use case for supporting the LESS_STRICT_PREFIX ?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,71 @@
...
+    } catch (IOException e) {
+      logger.error(
+          "I/O error occurs when loading connection string from the file '{}' with UTF-8 encoding.",
+          filePath,
+          e);
+      return null;
+    } catch (InvalidPathException e) {
+      logger.error("Invalid file path for the connection string '{}'", filePath, e);
+      return null;
+    }
```
ramthi, line 66:
```suggestion
    } catch (IOException e) {
      logger.error(
          "I/O error occurs when loading connection string from the file '{}' with UTF-8 encoding.",
          filePath,
          e);
    } catch (InvalidPathException e) {
      logger.error("Invalid file path for the connection string '{}'", filePath, e);
    }
    return null;
```

ramthi (changes requested):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
trask: use the `StringSubstitutor`, this will handle the templating `${...}`, and then your `FileStringLookup` doesn't need to

heyams: it's easy to make mistake in the json config.  
here i want to support these two formats:
`${file:file.txt} ` or `file:file.txt` 

the goal here is the get the file name.  that's why I allow less strict.

heyams: will try stringSubstitutor.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
heyams: this is outdated.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -352,8 +353,12 @@ public static void logConfigurationWarnMessages() {
 
   // visible for testing
   static void overlayFromEnv(Configuration config) throws IOException {
-    config.connectionString = overlayConnectionStringFromEnv(config.connectionString);
-
+    // load connection string from a file if connection string is in the format of
+    // "${file:mounted_connection_string_file.txt}"
+    StringSubstitutor stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
+    stringSubstitutor.setEnableSubstitutionInVariables(true);
```
trask, line 359: is this needed?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookupTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @BeforeEach
+  public void setup() throws IOException {
+    file = File.createTempFile("test", "", tempFolder);
+    Writer writer = Files.newBufferedWriter(file.toPath(), UTF_8);
+    writer.write(CONNECTION_STRING);
+    writer.close();
+
+    stringSubstitutor = new StringSubstitutor(FileStringLookup.INSTANCE);
+    stringSubstitutor.setEnableSubstitutionInVariables(true);
```
trask, line 53:
```suggestion
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilderTest.java
```diff
@@ -135,4 +135,69 @@ void testRpConfigurationOverlayWithEnvVarAndSysPropPopulated() throws Exception
...
+  }
+
+  @Test
+  void testConnectionStringEnvVarHasHigherPrecedenceOverFileLookup() throws Exception {
+    File connectionStringFile =
+        new File(
+            getClass()
+                .getClassLoader()
+                .getResource("file-look-up-connection-string.txt")
+                .getPath());
```
trask, line 190: consolidate on one way to test from file? e.g. either resource file like here or temp file like in other test?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,60 @@
...
+
+    final String fileName = StringUtils.substringAfter(key, ":");
+    try {
+      return new String(Files.readAllBytes(Paths.get(fileName)), StandardCharsets.UTF_8);
+    } catch (IOException | InvalidPathException e) {
+      logger.error(
+          "Error occurs when reading connection string from the file '{}' with UTF-8 encoding.",
+          fileName,
+          e);
+      return null;
```
trask, line 55: I think rethrow here to fail agent startup with this message

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.LoggerFactory;
+
+final class FileStringLookup implements StringLookup {
+
+  private static final Logger logger = LoggerFactory.getLogger(FileStringLookup.class);
+  static final FileStringLookup INSTANCE = new FileStringLookup();
+
+  @Override
+  @Nullable
+  public String lookup(final String key) {
```
trask, line 42: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/style-guideline.md#final-keyword-usage

```suggestion
  public String lookup(String key) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,60 @@
...
+    if (key == null) {
+      return null;
+    }
+
+    final String fileName = StringUtils.substringAfter(key, ":");
```
trask, line 47:
```suggestion
    String fileName = StringUtils.substringAfter(key, ":");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
ramthi: > it's easy to make mistake in the json config. here i want to support these two formats: `${file:file.txt} ` or `file:file.txt`
> 
> the goal here is the get the file name. that's why I allow less strict.

Got it.  My recommendation is to keep it simple.  If we will not document the less strict things then I don't think we need to anticipate and support the various typos people can make.  If we will document then that is all the more reason to keeping it simple.  I'm not going to block the review on this.  I will let you make the call.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
heyams: yes, otherwise, the result will return "[value]".

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
heyams: rethrow will missing the env var override... i will let setConnectionString handle it.. it will through exception a bit down.

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookupTest.java
heyams: this is required. outcome will be impacted if not.

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilderTest.java
heyams: sure, i can use temp file like FileStringLookup.. too many resource files already.

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookupTest.java
trask: test passes for me when I remove this

heyams: works for me too... but earlier i got "[value]". now i don't know why.  removed.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
trask: existing behavior is to fail on any json configuration error, and not try to recover or handle overrides in that case, I think good to stay consistent with the this current fail-fast behavior

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
trask: @ramthi this was addressed by switching to standard StringSubstitutor (which only handles strict behavior)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
heyams: i see. will do.

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookupTest.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  @AfterEach
+  public void cleanup() throws IOException {
+    Files.delete(file.toPath());
+  }
+
+  @Test
+  public void testGoodFileLookupFormat() {
+    String connectionString = "${file:" + file.getPath() + "}";
```
trask, line 64: let's also test `"${xyz:" + file.getPath() + "}"`, I think this will pass currently

check out StringLookupFactory.interpolatorStringLookup(), you can pass it a singleton map with key "file" and value of your FileStringLookup, then I think it will only capture "file:"

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilderTest.java
```diff
@@ -177,6 +177,10 @@ void testOverlayWithEnvVarWithBadFileStringLookupFormat() throws Exception {
     ConfigurationBuilder.overlayFromEnv(configuration);
     assertThat(configuration.connectionString).isEqualTo(configuration.connectionString);
 
+    configuration.connectionString = "${xyz:" + connectionStringFile.getPath();
```
trask, line 180:
```suggestion
    configuration.connectionString = "${xyz:" + connectionStringFile.getPath() + "}";
```

trask (approved):

ramthi (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/FileStringLookup.java
```diff
@@ -0,0 +1,55 @@
...
+
+  static final FileStringLookup INSTANCE = new FileStringLookup();
+
+  @Override
+  @Nullable
+  public String lookup(String key) {
+    if (key == null) {
+      return null;
+    }
+
```
trask, line 42: I don't think this is needed anymore since switching to StringSubstitutor
```suggestion
```

trask (approved):

PR #2076, opened 2022-01-28
Handle 401 and 403 responses from Breeze

kryalama:
Provided a clear message when breeze responds with 401 or 403 status codes

Sample:
Sending telemetry to the ingestion service: Authentication required. Please provide Azure Active Directory credentials (telemetry will be stored to disk and retried later) (future warnings will be aggregated and logged once every 5 minutes)


trask (approved):
nice, can you paste an example of what the warning message looks like in the PR description?

PR #2078, opened 2022-02-02
Add test for customDimensions for Statsbeat

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/FeatureStatsbeat.java
```diff
@@ -170,5 +170,10 @@ void trackConfigurationOptions(Configuration config) {
         || !config.preview.captureHttpClientHeaders.responseHeaders.isEmpty()) {
       featureList.add(Feature.CAPTURE_HTTP_CLIENT_HEADERS);
     }
+
+    // customDimensions
+    if (config.customDimensions.size() != 0) {
```
trask, line 175:
```suggestion
    if (!config.customDimensions.isEmpty()) {
```

trask (approved):
👍

PR #2084, opened 2022-02-03
Don't modify swagger generated files

trask:
This will be needed in order to share with azure-monitor-opentelemetry-exporter.


heyams:
is this copied directly from azure sdk java exporter? i don't understand why it has diff? 
this will be the base of azure sdk java exporter, right?

trask:
this is basically the same as https://github.com/Azure/azure-sdk-for-java/pull/26880. yes, they will be converged shortly, but not in this PR

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/builders/AbstractTelemetryBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private final TelemetryItem telemetryItem;
+
+  protected AbstractTelemetryBuilder(MonitorDomain data, String telemetryName, String baseType) {
+
+    telemetryItem = new TelemetryItem();
+    telemetryItem.setVersion(1);
+    telemetryItem.setName(telemetryName);
+
+    data.setVersion(2);
```
heyams, line 49: are you going to support the newer version of common schema? this is really old. i remembered it's like 6+?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/builders/AbstractTelemetryBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+
+    MonitorBase monitorBase = new MonitorBase();
+    telemetryItem.setData(monitorBase);
+    monitorBase.setBaseType(baseType);
+    monitorBase.setBaseData(data);
+  }
+
+  public void setTime(OffsetDateTime time) {
```
heyams, line 57: should we match the name in basetelemetry? timestamp?  it's Date instead of OffsetDatetime in basetelemetry. do we need to be consistent?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/builders/ExceptionTelemetryBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+import java.util.Map;
+
+public final class ExceptionTelemetryBuilder extends AbstractTelemetryBuilder {
+
+  private static final int MAX_PROBLEM_ID_LENGTH = 1024;
+
+  private final TelemetryExceptionData data;
+
+  public static ExceptionTelemetryBuilder create() {
```
heyams, line 41: It's a builder.. i wonder why create is needed here?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/builders/MessageTelemetryBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import com.microsoft.applicationinsights.agent.internal.exporter.models.SeverityLevel;
+import com.microsoft.applicationinsights.agent.internal.exporter.utils.SanitizationHelper;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class MessageTelemetryBuilder extends AbstractTelemetryBuilder {
+
+  private final MessageData data;
+
+  public static MessageTelemetryBuilder create() {
```
heyams, line 37: same here why create?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/builders/ExceptionTelemetryBuilder.java
heyams: i see it's not a true builder pattern.. since all the setters don't return the object itself. can we rename all the data model without builder?

kryalama (approved):

PR #2086, opened 2022-02-04
Turn off oshi logging

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/LoggingLevelConfigurator.java
```diff
@@ -62,7 +63,9 @@ public void initLoggerLevels(LoggerContext loggerContext) {
   public void updateLoggerLevel(Logger logger) {
     Level loggerLevel;
     String name = logger.getName();
-    if (name.startsWith("reactor.netty") || name.startsWith("io.netty")) {
+    if (name.startsWith("oshi")) {
+      loggerLevel = Level.OFF;
```
heyams, line 67: completely off @trask?

trask: selfDiagnostic level -> oshi level
* INFO or higher -> OFF
* DEBUG -> INFO
* TRACE -> TRACE

similar to getDefaultLibraryLevel but slightly different

trask (approved):
🤞

PR #2082, opened 2022-02-03
Support Statsbeat in EU regions

heyams:
azure doc draft

https://github.com/MicrosoftDocs/azure-docs-pr/pull/187423

spec: https://microsoft.sharepoint.com/:w:/t/OneDSJavascript/EYRX-qyTul5CvaXYOwNhVAoBVIIUWun4FFoRXklRAtseiA?e=fZZMpa

PR #2087, opened 2022-02-04
Shorten host property in network statsbeat

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -241,23 +241,33 @@ private double getRequestDurationAvg() {
 
   /**
    * e.g. endpointUrl 'https://westus-0.in.applicationinsights.azure.com/v2.1/track' host will
-   * return 'westus-0.in.applicationinsights.azure.com'
+   * return 'westus-0'
    */
   static String getHost(String endpointUrl) {
```
trask, line 246: since this method is not called often, let's use the simpler (and less performant) `new URL(endpointUrl).getHost()` and then parse out the short name from there

heyams: getHost is called every 15 mins for network statsbeat. 

public URL(URL context, String spec, URLStreamHandler handler) is not less performant if you checked their implementation though.

trask: constructing a new URL is expensive relative to not (which is why our `UrlParser` class exists), but every 15 minutes is no problem to use URL

heyams: url.getHost doesn't do what we need. 

"https://fakehost-1.example.com" will return fakehost-1.example.com.

trask: is this method the same as `getGeoWithoutStampSpecific` from #2082?

heyams: it's different.  this method return "westus-0", the other is returning region without the stamp specific "westus".

trask: got it, probably would be nice to leverage one of them to implement the other (strip off the trailing stamp #)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -42,6 +44,9 @@
   private static final String EXCEPTION_COUNT_METRIC_NAME = "Exception Count";
   private static final String BREEZE_ENDPOINT = "breeze";
 
+  private static final Pattern hostPattern =
+      Pattern.compile("^(?:https?://)?(?:www\\.)?([^/\\.]+)");
```
trask, line 48: do we need to support url with missing protocol (no `http://` or `https://`)? does IngestionEndpoint support that?

heyams: this pattern works for all cases.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -241,25 +246,16 @@ private double getRequestDurationAvg() {
...
 
-    int end = endpointUrl.indexOf("/");
-    if (end != -1) {
-      return endpointUrl.substring(0, end);
+    if (matcher.find()) {
+      return matcher.group(1);
     }
 
-    return endpointUrl;
+    return "";
```
trask, line 259: what do u think of returning the full url as the fallback? that way we could understand why it didn't parse
```suggestion
    return endpointUrl;
```

heyams: that will end up sending the full endpointUrl as host to statsbeat.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -42,6 +44,8 @@
   private static final String EXCEPTION_COUNT_METRIC_NAME = "Exception Count";
   private static final String BREEZE_ENDPOINT = "breeze";
 
+  private static final Pattern hostPattern = Pattern.compile("^https?://(?:www\\.)?([^/\\.]+)");
```
trask, line 47:
```suggestion
  private static final Pattern hostPattern = Pattern.compile("^https?://(?:www\\.)?([^/.]+)");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
trask: isn't this a good thing? then we can see what the problem was?

heyams: well, if that's the case, telemetry won't get sent at all.. it's from customer's ingestion endpoint.

trask: I thought this is just for the attribute on the statsbeat telemetry, probably I'm missing something, we can discuss mon

trask (approved):

PR #2098, opened 2022-02-09
add readme and AI operation name span processor from azmon exporter

kryalama:


trask:
@kryalama can we delete `AiOperationNameSpanProcessor` from the agent in this PR?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AiOperationNameSpanProcessor.java
```diff
@@ -19,29 +19,42 @@
...
-import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
 
+/** SpanProcessor implementation to update operation name. */
+// note: operation name for requests is handled during export so that it can use the updated span
+// name from routing instrumentation
+//       if we (only) set operation name on requests here, it would be based on span name at
+// startSpan
 public class AiOperationNameSpanProcessor implements SpanProcessor {
 
+  public static final AttributeKey<String> AI_OPERATION_NAME_KEY =
```
trask, line 38:
```suggestion
  // visible for testing
  static final AttributeKey<String> AI_OPERATION_NAME_KEY =
```

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
```diff
@@ -343,6 +343,12 @@ private static void setOperationTags(AbstractTelemetryBuilder telemetryBuilder,
     setOperationName(telemetryBuilder, span.getAttributes());
   }
 
+  private static void setOperationTags(
+      AbstractTelemetryBuilder telemetryBuilder, String traceId, String parentSpanId) {
+    setOperationId(telemetryBuilder, traceId);
+    setOperationParentId(telemetryBuilder, parentSpanId);
+  }
+
```
trask, line 351: I think we don't need the changes in this file. If we don't, then good to revert them since it will likely cause conflicts with #2116

agent/agent-tooling/src/test/java/com/azure/monitor/opentelemetry/exporter/AiOperationNameSpanProcessorTest.java
```diff
@@ -0,0 +1,203 @@
...
+import reactor.core.publisher.Mono;
+
+public class AiOperationNameSpanProcessorTest {
+
+  private static Tracer configureAzureMonitorExporter(HttpPipelinePolicy validator) {
+    String connectionStringTemplate =
+        "InstrumentationKey=ikey;IngestionEndpoint=https://test.applicationinsights.azure.com/";
+    String connectionString =
+        Configuration.getGlobalConfiguration()
+            .get("APPLICATIONINSIGHTS_CONNECTION_STRING", connectionStringTemplate);
```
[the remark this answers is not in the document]
kryalama: In azure sdk repo, they use env variables for live tests. connectionStringTemplate is default value if the env variables are not present

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AiOperationNameSpanProcessor.java
```diff
@@ -19,29 +19,43 @@
...
+  // visible for testing
+  static final AttributeKey<String> AI_OPERATION_NAME_KEY =
+      AttributeKey.stringKey("applicationinsights.internal.operation_name");
+
   @Override
   public void onStart(Context parentContext, ReadWriteSpan span) {
+    // also check if we should copy parentSpan operation name if the child span already have
+    // operation name
+    if (span.getAttribute(AI_OPERATION_NAME_KEY) != null) {
+      return;
```
trask, line 47:
```suggestion
    if (span.getAttribute(AI_OPERATION_NAME_KEY) != null) {
      // don't copy down the parent span's operation name if the user gave the span an explicit
      // operation name
      return;
```

PR #2109, opened 2022-02-11
Handle 402

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -322,7 +322,8 @@ private CompletableResultCode internalSend(
                               + " (telemetry will be stored to disk and retried later)");
                       onFailure.accept(true);
                       break;
-                    case 439: // Breeze-specific: THROTTLED OVER EXTENDED TIME
+                    case 402: // THROTTLED MONTHLY QUOTA EXCEEDED
+                    case 439: // THROTTLED DAILY QUOTA EXCEEDED
                       // TODO handle throttling
                       operationLogger.recordFailure(
                           "received response code 439 (throttled over extended time)");
```
trask, line 329: let's have different log messages for these

trask (approved):

PR #2104, opened 2022-02-09
Refactor connection string and share

trask:

PR #2116, opened 2022-02-14
Move live metrics to azmon exporter

kryalama:


heyams:
@kryalama you can use draft PR to test out smoke tests and unit tests... when everything is green, you can mark it ready for review.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/QuickPulseDataCollector.java
```diff
@@ -51,7 +50,7 @@
 public enum QuickPulseDataCollector {
   INSTANCE;
 
-  private TelemetryClient telemetryClient;
+  private String instrumentationKey;
```
trask, line 53: we need to support changing the instrumentation key at runtime for Azure Spring Cloud (cc @heyams)
```suggestion
  private Supplier<String> instrumentationKeySupplier;
```

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/EndpointProvider.java
```diff
@@ -0,0 +1,155 @@
...
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+package com.azure.monitor.opentelemetry.exporter.implementation.utils;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class EndpointProvider {
```
trask, line 27: I deleted this class in #2104, can you migrate your code to the new connection string code that's already been moved?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/LazyHttpClient.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import org.checkerframework.checker.lock.qual.GuardedBy;
+import reactor.core.publisher.Mono;
+import reactor.netty.resources.LoopResources;
+
+public class LazyHttpClient implements HttpClient {
```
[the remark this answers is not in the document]
kryalama: My bad, missed this one.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/TelemetryMappingHelper.java
```diff
@@ -0,0 +1,797 @@
...
+import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import reactor.util.annotation.Nullable;
+
+public final class TelemetryMappingHelper {
```
trask, line 56: can you leave these methods in the Exporter for now and make them public where needed, in order to keep diff smaller because it's hard to review these large diffs for accuracy

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/QuickPulsePingSender.java
```diff
@@ -19,68 +19,70 @@
...
+  private final String roleName;
+  private final String instrumentationKey;
+  private final String endPointUrl;
   private long lastValidTransmission = 0;
 
   public QuickPulsePingSender(
       HttpPipeline httpPipeline,
-      TelemetryClient telemetryClient,
+      String roleName,
+      String instrumentationKey,
```
trask, line 74: I think this needs `Supplier<String>` also since it can change at runtime in Azure Spring Cloud

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
```diff
@@ -116,6 +118,14 @@ public void configure(SdkTracerProviderBuilder tracerProvider, ConfigProperties
...
     }
 
+    // Add quick pulse
+    tracerProvider.addSpanProcessor(
+        new LiveMetricsSpanProcessor(
+            telemetryClient.getAadAuthentication(),
+            telemetryClient.getRoleName(),
+            telemetryClient.getInstrumentationKey(),
+            telemetryClient.getRoleInstance(),
+            ConnectionString.parse(configuration.connectionString).getLiveEndpoint().toString()));
```
trask, line 128: let's not add the span processor to the agent in this PR, 2.x bridged data won't flow through this, and we already flush the BatchSpanProcessor very fast in the agent and batch at a lower layer

kryalama: Adding the LiveMetricsSpanProcessor is the only way we can enable quickpulse to the agent (since we removed the old code). Do you recommend any other alternative?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/LiveMetricsSpanProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class LiveMetricsSpanProcessor implements SpanProcessor {
+
+  private static final ClientLogger LOGGER = new ClientLogger(AzureMonitorTraceExporter.class);
+  private String instrumentationKey;
+
+  private LiveMetricsSpanProcessor() {}
+
+  public LiveMetricsSpanProcessor(
+      AadAuthentication aadAuthentication,
```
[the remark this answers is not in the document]
kryalama: thats a good idea. In the future we can modify/add different authentication types

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
trask: I'm not clear what you mean, `QuickPulseDataCollector.INSTANCE.add(telemetryItem)` is still being called from TelemetryClient so this isn't needed (and is probably causing duplicate live metrics?)

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/LazyHttpClient.java
```diff
@@ -19,7 +19,7 @@
  * DEALINGS IN THE SOFTWARE.
  */
 
-package com.microsoft.applicationinsights.agent.internal.httpclient;
+package com.azure.monitor.opentelemetry.exporter.implementation.utils;
```
trask, line 22: do we need LazyHttpClient in azure-monitor-opentelemetry-exporter?

kryalama: The quickpulse ping sender and data sender use httpclient to send quickpulse pings/data. Other alternative is to initialize quickpulse with httpclient. Do you recommend this approach?

trask: ya, if possible follow the approach of the exporter which only uses LazyHttpClient for agent side

trask: this will also allow us to not need the `AadAuthentication` classes I believe

trask (approved):

PR #2119, opened 2022-02-15
Do not send metric NaN values

trask:


kryalama:
@trask can we please add a unit test for this scenario?

kryalama (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -155,27 +158,26 @@ public void trackAsync(TelemetryItem telemetry) {
...
-                      }
-                    }
-                    return true;
-                  })
-              .collect(Collectors.toList());
-
-      if (filteredPoints.isEmpty()) {
+      if (metricsData.getMetrics().isEmpty()) {
+        // this is unexpected
+        logger.debug("MetricsData has no metric point");
```
heyams, line 163: should it throw an exception if it's unexpected?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -155,27 +158,26 @@ public void trackAsync(TelemetryItem telemetry) {
...
+        for (MetricFilter metricFilter : metricFilters) {
+          if (!metricFilter.matches(metricName)) {
+            // user configuration filtered out this metric name
+            return;
+          }
+        }
+      }
+
+      if (!Double.isFinite(point.getValue())) {
+        // breeze doesn't like these values
```
heyams, line 178: do we drop that data point and then send the rest to breeze? here because of bad NaN we drop the whole metric. i wonder how other sdks handle it.

heyams: unless that data point is the key dimension for that metric. it makes sense to drop it. without it metric doesn't make sense.

heyams:
can we add a test for NaN MetricDataPoint?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: ya, metric value is required

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: I was being conservative, but feel pretty confident, so bumped it up to an exception

trask:
> can we add a test for NaN MetricDataPoint?

it's tricky currently b/c `TelemetryClient` is not very testable. I added a `TODO` in the code, and will revisit in the `vnext` branch after further TelemetryClient refactoring/simplifications

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: > I wonder how other sdks handle it

https://github.com/microsoft/Telemetry-Collection-Spec/issues/29

Issue #2112, opened 2022-02-11
agent eating up almost 100% CPU


Sachin1O1:
### Expected behavior
Describe what you expected to happen.

### Actual behavior
Describe what actually happened that was unexpected.

#### Sample Application
If applicable, provide a sample application which reproduces the issue.

### System information
Please provide the following information:
 - SDK Version: 3.2.4
 - OS type and version: Linux
 - Application Server type and version (if applicable): tomcat9-jre11
 - Using spring-boot?
 - Additional relevant libraries (with version, if applicable):

### Logs
2022-02-09 21:33:44.516Z WARN  c.m.a.a.i.q.QuickPulsePingSender - Pinging live metrics endpoint: io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: rt.services.visualstudio.com/23.100.122.113:443 (future warnings will be aggregated and logged once every 5 minutes)
reactor.core.Exceptions$ReactiveException: io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: rt.services.visualstudio.com/23.100.122.113:443
	at reactor.core.Exceptions.propagate(Exceptions.java:392)
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:97)
	at reactor.core.publisher.Mono.block(Mono.java:1706)
	at com.microsoft.applicationinsights.agent.internal.quickpulse.QuickPulsePingSender.ping(QuickPulsePingSender.java:116)
	at com.microsoft.applicationinsights.agent.internal.quickpulse.QuickPulseCoordinator.ping(QuickPulseCoordinator.java:107)
	at com.microsoft.applicationinsights.agent.internal.quickpulse.QuickPulseCoordinator.run(QuickPulseCoordinator.java:63)
	at java.base/java.lang.Thread.run(Unknown Source)
	Suppressed: java.lang.Exception: #block terminated with an error
		at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:99)
		... 5 common frames omitted
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: rt.services.visualstudio.com/23.100.122.113:443
Caused by: java.net.ConnectException: finishConnect(..) failed: Connection refused
[...]

PR #2121, opened 2022-02-16
Attempting to address BatchSpanProcessor overhead

trask:
For #2112


trask:
Closing in favor of #2165

PR #2127, opened 2022-02-17
Fix reverse lookup

trask:


heyams (approved):

heyams:
@trask Srikanth confirmed it fixed the high latency issue for app service. Let's cherry-pick the fix to main branch 👍

PR #2165, opened 2022-03-09
Add BSP configuration option

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
```diff
@@ -161,9 +163,20 @@ private static BatchSpanProcessor createExporter(Configuration configuration) {
     }
 
     // using BatchSpanProcessor in order to get off of the application thread as soon as possible
-    // using batch size 1 because need to convert to SpanData as soon as possible to grab data for
-    // live metrics. the real batching is done at a lower level
-    return BatchSpanProcessor.builder(currExporter).setMaxExportBatchSize(1).build();
+    BatchSpanProcessorBuilder builder = BatchSpanProcessor.builder(currExporter);
+
+    String delayMillisStr = System.getenv("APPLICATIONINSIGHTS_PREVIEW_BSP_DELAY_MILLIS");
```
heyams, line 168: can we not use acronym? it will be long but will be clearer.

heyams: do we want to make this public on azure doc? I know it will be experimental at first.

trask: I renamed it to `APPLICATIONINSIGHTS_PREVIEW_BSP_SCHEDULE_DELAY` to align with otel naming of the same property:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#batch-span-processor

Though I'm hoping this is only for testing and we can find a one-size-fits-all option and remove this config knob.

heyams (approved):

PR #2177, opened 2022-03-12
Add process cpu normalized

trask:
~currently built on top of #2176, will rebase once that is merged~

~One outstanding question about whether we should apply the config option to system cpu percentage also.~

requires doc update

### DIFF
--- a/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/QuickPulse.java
+++ b/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/QuickPulse.java
@@ -46,7 +46,7 @@ public class QuickPulse {
       Supplier<String> instrumentationKey,
       @Nullable String roleName,
       @Nullable String roleInstance,
-      boolean reportNonNormalizedProcessorTime) {
+      boolean useNormalizedValueForNonNormalizedCpuPercentage) {
 
     QuickPulse quickPulse = new QuickPulse();
 
@@ -66,7 +66,7 @@ public class QuickPulse {
                   instrumentationKey,
                   roleName,
                   roleInstance,
-                  reportNonNormalizedProcessorTime);
+                  useNormalizedValueForNonNormalizedCpuPercentage);
             });
 
     return quickPulse;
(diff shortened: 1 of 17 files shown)


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/QuickPulse.java
```diff
@@ -77,7 +83,8 @@ private void initialize(
       Supplier<URL> endpointUrl,
       Supplier<String> instrumentationKey,
       @Nullable String roleName,
-      @Nullable String roleInstance) {
+      @Nullable String roleInstance,
+      boolean backCompatNonNormalizedCpuPercentage) {
```
heyams, line 87: this is confusing.. backCompatNonNormalizedCpuPercentage is used normalized cpu percentage.. 
can you rename to usedNormalizedCpuPercent? and add the explaination for backward compatibility of non-normalized cpu percentage?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -257,6 +257,11 @@ public String toString() {
     // telemetry (they are much smaller so a larger queue size is ok)
     public int metricsExportQueueCapacity = 65536;
 
+    // unfortunately the Java SDK behavior has always been to report the "% Processor Time" number
+    // as "normalized" (divided by # of CPU cores), even though it should be non-normalized
+    // maybe this can be fixed in 4.0 (would be a breaking change)
+    public boolean backCompatNonNormalizedCpuPercentage;
```
heyams, line 263: usedNormalizedCpuPercentage?

trask: ya, this is confusing, I added more comments, I'm not sure what to do about the name, we already report a normalized cpu percentage so I'm not sure `useNormalizedCpuPercentage` makes it any clearer.

maybe `useNormalizedValueForNonNormalizedCpuPercentage` (with default `true` representing the backwards compatible value)? kinda long but... 😂

heyams: useNormalizedValueForNonNormalizedCpuPercentage sounds good although it's a bit long but it's clearer.

heyams (approved):

PR #2178, opened 2022-03-12
Add APPLICATIONINSIGHTS_PROXY env var

trask:
requires doc update


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -450,6 +461,45 @@ private static String overlayConnectionStringFromEnv(String connectionString) {
...
+
+    String userInfo = proxyUrl.getUserInfo();
+    if (userInfo != null) {
+      String[] usernamePassword = userInfo.split(":", 2);
+      if (usernamePassword.length == 2) {
+        proxyFromEnv.username =
+            URLDecoder.decode(usernamePassword[0], StandardCharsets.UTF_8.toString());
+        proxyFromEnv.password =
+            URLDecoder.decode(usernamePassword[1], StandardCharsets.UTF_8.toString());
+      }
```
heyams, line 497: should we log warning when applicationinsights_proxy is malformatted?

trask: done

heyams (approved):

PR #2179, opened 2022-03-12
Remove http response draining

trask:
The underlying bug was fixed in azure sdk.


heyams:
When can we deprecate profile api? Breeze has officially deprecated it.

heyams (approved):

PR #2180, opened 2022-03-12
Reduce warnings

### LINKED ISSUE #2172: Warning response 500
### Expected behavior
Describe what you expected to happen.
expected to show only application insight connection error and loss of data
### Actual behavior
Describe what actually happened that was unexpected.
unexpected the warning logs in the pods which are not useful
 WARN  c.m.a.a.i.telemetry.TelemetryChannel - In the last 5 minutes, the following operation has failed 1 times (out of 64): Sending
 * received response code 500 (telemetry will be stored to disk and retried later) (1 times)
### To Reproduce
I tested two pods with the self Diagnostics with ERROR and WARN. 
When I use ERROR as below
"selfDiagnostics": {
    "destination": "file",
    "level": "ERROR"
    }
 I don't see any logs in the applicationinsight.log file.

When I use WARN as below
"selfDiagnostics": {
    "destination": "file",
    "level": "WARN"
    }
I see below warnings in the applicationinsight.log file

2022-03-10 00:00:47.437Z WARN  c.m.a.a.i.telemetry.TelemetryChannel - Sending telemetry to the ingestion service: received response code 500 (telemetry will
2022-03-10 01:40:47.445Z WARN  c.m.a.a.i.telemetry.TelemetryChannel - In the last 5 minutes, the following operation has failed 1 times (out of 64): Sending
 * received response code 500 (telemetry will be stored to disk and retried later) (1 times)
2022-03-10 02:10:47.447Z WARN  c.m.a.a.i.telemetry.TelemetryChannel - In the last 5 minutes, the following operation has failed 1 times (out of 65): Sending
 * received response code 500 (telemetry will be stored to disk and retried later) (1 times)
2022-03-10 03:05:47.449Z WARN  c.m.a.a.i.telemetry.TelemetryChannel - In the last 5 minutes, the following operation has failed 1 times (out of 64): Sending
 * received response code 500 (telemetry will be stored to disk and retried later) (1 times)

#### Sample Application
If applicable, provide a sample application which reproduces the issue.
In the ISAAC application we are using application insight inside the deployment and seeing this unwanted warning in the pod logs.
### System information
 - SDK Version: 
 - OS type and version:
 - Using spring-boot? yes
 - Additional relevant libraries (with version, if applicable):

### Logs

trask:
Resolves #2172


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/logging/DiagnosticTelemetryPipelineListener.java
```diff
@@ -40,12 +40,14 @@
   private static final Logger logger = LoggerFactory.getLogger(FOR_CLASS);
 
   private final OperationLogger operationLogger;
+  private final boolean logWarnings;
```
heyams, line 43: can we rename it to `shouldLogWarnings`?

heyams (approved):

PR #2184, opened 2022-03-13
Add role name overrides

### LINKED ISSUE #2133: Unable to distinguish between applications running with same javaagent in Application Map
### Expected behavior
I expected the agent to detect the different applications running on the same server and detect the role names from web.xml:

```
<env-entry>
		<env-entry-name>azure.application-insights.role.name</env-entry-name>
		<env-entry-type>java.lang.String</env-entry-type>
		<env-entry-value>XyzAPI</env-entry-value>
</env-entry>
```

### Actual behavior
The applications gets bundled into one instance and the calls between them are visualised as an application calling itself. 
Cloud role name is `subscripitions\{someguid}\ApplicationInsightsResourceName`

### System information
 - SDK Version: 3.2.6
 - OS type and version: Windows 10
 - Application Server type and version (if applicable): JBoss EAP 6.1
 - Using spring-boot? Yes 

We looked at setting the cloud role name based on the request url path but it seems this is not possible with current configuration options.

trask:
Resolves #2133

requires doc update

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
@@ -248,6 +248,7 @@ public class Configuration {
     public PreviewStatsbeat statsbeat = new PreviewStatsbeat();
 
     public List<InstrumentationKeyOverride> instrumentationKeyOverrides = new ArrayList<>();
+    public List<RoleNameOverride> roleNameOverrides = new ArrayList<>();
 
     public int generalExportQueueCapacity = 2048;
     // metrics get flooded every 60 seconds by default, so need larger queue size to avoid dropping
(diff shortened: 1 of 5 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -396,7 +400,27 @@ public void validate() {
...
+  public static class RoleNameOverride {
+    public String httpPathPrefix;
+    public String roleName;
+
+    public void validate() {
+      if (httpPathPrefix == null) {
+        // TODO add doc and go link, similar to telemetry processors
+        throw new FriendlyException(
+            "A role name override configuration is missing an \"httpPathPrefix\".",
+            "Please provide an \"httpPathPrefix\" for the instrumentation key override configuration.");
```
heyams, line 417:
```suggestion
            "Please provide a \"httpPathPrefix\" for the role name override configuration.");
```

heyams (approved):

PR #2189, opened 2022-03-16
Don't load expired telemetry

trask:
I've noticed a couple of times getting breeze errors that telemetry has expired, because files are loaded from disk at startup before purge happens

### DIFF
--- a/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileCache.java
+++ b/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileCache.java
@@ -79,14 +79,14 @@ class LocalFileCache {
     int index = name.indexOf('-');
     if (index == -1) {
       logger.debug("unexpected .trn file name: {}", name);
-      return false;
+      return true;
     }
     long timestamp;
     try {
       timestamp = Long.parseLong(name.substring(0, index));
     } catch (NumberFormatException e) {
       logger.debug("unexpected .trn file name: {}", name);
-      return false;
+      return true;
     }
     Date expirationDate = new Date(System.currentTimeMillis() - 1000 * expiredIntervalSeconds);
     Date fileDate = new Date(timestamp);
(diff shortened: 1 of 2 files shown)


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileCache.java
```diff
@@ -63,9 +64,32 @@ File poll() {
...
         .collect(Collectors.toList());
   }
+
+  // files that are older than expiredIntervalSeconds (default 48 hours) are expired
+  static boolean isExpired(File file, long expiredIntervalSeconds) {
+    String name = file.getName();
+    int index = name.indexOf('-');
+    if (index == -1) {
+      logger.debug("unexpected .trn file name: {}", name);
+      return false;
```
heyams, line 82: return true.. treat bad name file as expired and ignore it?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileCache.java
```diff
@@ -63,9 +64,32 @@ File poll() {
...
+    }
+    long timestamp;
+    try {
+      timestamp = Long.parseLong(name.substring(0, index));
+    } catch (NumberFormatException e) {
+      logger.debug("unexpected .trn file name: {}", name);
+      return false;
```
heyams, line 89: same here?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileCache.java
trask: ya, that's what I wrote first, but was concerned about old filename format, but I guess we can't load those anyways b/c we changed disk format (adding ikey)?

heyams: Ya, old format won't work without ikey. i think we can ignore it. otherwise, infinite retry.

heyams (approved):

PR #2194, opened 2022-03-16
Remove mention of ikey

trask:


heyams:
when should we deprecate the env var support for instrumentation key?

trask:
it is already deprecated

https://github.com/microsoft/ApplicationInsights-Java/blob/b2c60e751b298b0fd10b63edbef25376575eb18c/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java#L444-L446

heyams (approved):

PR #2199, opened 2022-03-17
Fix todo

heyams:


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
```diff
@@ -33,7 +33,7 @@
 
 public class LocalStorageTelemetryPipelineListener implements TelemetryPipelineListener {
 
-  static final Set<Integer> RETRYABLE_CODES =
+  public static final Set<Integer> RETRYABLE_CODES =
```
trask, line 36: this change will conflict with #2198, can we merge that first?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeatHttpPipelinePolicy.java
```diff
@@ -58,15 +58,16 @@ public NetworkStatsbeatHttpPipelinePolicy(NetworkStatsbeat networkStatsbeat) {
                 // these are not tracked as success or failure since they are just redirects
               } else if (statusCode == 402 || statusCode == 439) {
                 networkStatsbeat.incrementThrottlingCount(instrumentationKey, host);
+              } else if (StatusCodes.isRetryable(statusCode)) {
+                networkStatsbeat.incrementRetryCount(instrumentationKey, host);
```
trask, line 62: can we rename them in the code so it is more clear?
```suggestion
              } else if (StatusCodes.isRetryableFailure(statusCode)) {
                networkStatsbeat.incrementRetriableFailureCount(instrumentationKey, host);
```

heyams: i still don't get it why it's not clear. isRetryable is to check if a status code is retryable? 
incrementRetryCount will keep track of number of time remote requests have been retried. 

most of the error code are type of failures... we don't say throttle failure, and simply throttle.. 

statsbeat network stats are tracking counts based on status code returned from breeze.

trask: good points, I think I agree, only confusion is really "retry count" sounds like the number of times we retried something, I think "retryable count" is clearer for what we are counting here

what do you think about just changing `incrementRetryCount` to `incrementRetryableCount`?

heyams: i prefer to keep "incrementRetryCount" since that is the metric name we use for the sake of consistency.

trask: ok let's keep it as `incrementRetryCount` for now at least

trask (approved):

PR #2204, opened 2022-03-24
Increase export throughput

trask:
The first commit is reverting the previous implementation (which was a mess from a diff perspective), so recommend reviewing just the second commit which contains the new implementation, which is much better from a diff perspective, and aligns with the approach we will probably end up with upstream in otel.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
```diff
@@ -206,8 +216,12 @@ private void flush() {
         }
       }
       exportCurrentBatch();
-      flushRequested.get().succeed();
-      flushRequested.set(null);
+      CompletableResultCode.ofAll(pendingExports).join(exporterTimeoutNanos, TimeUnit.NANOSECONDS);
+      CompletableResultCode flushResult = flushRequested.get();
+      if (flushResult != null) {
```
heyams, line 221: is null check needed?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
```diff
@@ -252,7 +266,15 @@ private void exportCurrentBatch() {
...
         CompletableResultCode result = spanExporter.send(Collections.unmodifiableList(batch));
-        result.join(exporterTimeoutNanos, TimeUnit.NANOSECONDS);
+        if (pendingExports.size() < maxPendingExports - 1) {
+          pendingExports.add(result);
+          result.whenComplete(
+              () -> {
+                pendingExports.remove(result);
+              });
+        } else {
+          result.join(exporterTimeoutNanos, TimeUnit.NANOSECONDS);
```
heyams, line 276: can we add log here to warn users that the maxPendingExports is reached here?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
trask: I pulled this change in from upstream

heyams: maybe upstream needs to be updated? based on the old code, null check doesn't seem to be necessary.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
```diff
@@ -267,12 +270,17 @@ private void exportCurrentBatch() {
...
         CompletableResultCode result = spanExporter.send(Collections.unmodifiableList(batch));
         if (pendingExports.size() < maxPendingExports - 1) {
+          addAsyncExport.recordSuccess();
           pendingExports.add(result);
           result.whenComplete(
               () -> {
                 pendingExports.remove(result);
               });
         } else {
+          addAsyncExport.recordFailure(
```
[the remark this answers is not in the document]
trask: I don't expect this to be common, it allows 100 concurrent exports

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchSpanProcessor.java
trask: it looks like upstream added the null check to make for nullaway static analysis happy, I'd rather stay in sync with them

heyams (approved):

PR #2211, opened 2022-03-30
OpenTelemetry Metric to Azure Monitor Metric

heyams:
cc @rajkumar-rangaraj

smoke test will be in a separate PR.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/MetricExporterCustomizer.java
```diff
@@ -0,0 +1,48 @@
...
+    TelemetryClient telemetryClient = TelemetryClient.getActive();
+    if (telemetryClient == null) {
+      // agent failed during startup
+      return;
+    }
+    autoConfiguration.addMeterProviderCustomizer(
+        (meterProvider, config) ->
+            meterProvider.registerMetricReader(
+                PeriodicMetricReader.builder(new AzureMonitorMetricExporter(telemetryClient))
+                    .setInterval(Duration.ofMinutes(1))
```
trask, line 45: we have a preview configuration `metricIntervalSeconds` that can be used here

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/InMemoryMetricExporter.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+// test OpenTelemetry integration
+final class InMemoryMetricExporter implements MetricExporter {
```
trask, line 35: let's see if we can use `io.opentelemetry.sdk.testing.exporter.InMemoryMetricExporter` from `io.opentelemetry:opentelemetry-sdk-metrics-testing:1.12.0-alpha`

heyams: in 1.12.0, they made it package protected. I can check alpha.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Map;
+
+final class AzureMonitorMetricsData {
+
+  private static final int VERSION = 2;
+  private static final String DEFAULT_METRIC_AGGREGATION_INTERVAL = "60000";
+  private static final String AGGREGATION_INTERNAL_MS_KEY = "_MS.AggregationIntervalMs";
+  private final MetricsData metricsData = new MetricsData();
+
+  public AzureMonitorMetricsData(MetricData metricData, PointData pointData) {
```
trask, line 48: this looks more like a static util class, no need for state, just a function that takes two params and returns MetricsData

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Override
+  public CompletableResultCode shutdown() {
+    stopped.set(true);
+    return CompletableResultCode.ofSuccess();
+  }
+
+  private TelemetryItem convertOtelMetricToAzureMonitorMetric(MetricData metricData) {
+    TelemetryItem telemetryItem = new TelemetryItem();
+    telemetryItem.setInstrumentationKey(telemetryClient.getInstrumentationKey());
```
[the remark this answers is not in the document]
heyams: I did attempt to use it at first... then later considered this is simpler. newMetricTelemetryBuilder is used everywhere. and it doesn't give me the flexibility to set stuff.. i need to create more interfaces.

I can add the data validation though.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java
heyams: that works too.

PR #2214, opened 2022-04-02
Add smoke test for AzureMonitorMetricExporter

heyams:
This is a continuation of #2211


OpenTelemetryMetric/gradle.lockfile
```diff
@@ -0,0 +1,4 @@
+# This is a Gradle generated file for dependency locking.
+# Manual edits can break the build and are not advised.
+# This file is expected to be part of source control.
+empty=runtimeClasspath
```
trask, line 4: is this needed?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -366,6 +366,14 @@ public RequestTelemetryBuilder newRequestTelemetryBuilder() {
...
   }
 
+  public Map<String, String> getGlobalTags() {
+    return globalTags;
+  }
+
+  public Map<String, String> getGlobalProperties() {
+    return globalProperties;
+  }
+
```
trask, line 376: can remove this now

```suggestion
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsDataTest.java
```diff
@@ -0,0 +1,315 @@
...
+    inMemoryMetricExporter = InMemoryMetricExporter.create();
+    MetricReaderFactory metricReaderFactory =
+        PeriodicMetricReader.newMetricReaderFactory(inMemoryMetricExporter);
+    SdkMeterProvider meterProvider =
+        SdkMeterProvider.builder().registerMetricReader(metricReaderFactory).build();
+
+    OpenTelemetry openTelemetry =
+        OpenTelemetrySdk.builder().setMeterProvider(meterProvider).build();
+    // buildAndRegisterGlobal can be called only once and then use
+    // GlobalOpenTelemetry.get();
```
trask, line 74:
```suggestion
```

settings.gradle
```diff
@@ -102,6 +102,7 @@ include ':test:smoke:testApps:Micrometer'
 include ':test:smoke:testApps:MongoDB'
 include ':test:smoke:testApps:NonDaemonThreads'
 include ':test:smoke:testApps:OpenTelemetryApiSupport'
+include ':test:smoke:testApps:OpenTelemetryMetric'
```
trask, line 105: is this needed?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java
```diff
@@ -0,0 +1,91 @@
...
+        pointBuilder.setDataPointType(DataPointType.MEASUREMENT);
+        pointBuilder.setValue(((DoublePointData) pointData).getValue());
+        break;
+      case HISTOGRAM:
+        pointBuilder.setDataPointType(DataPointType.AGGREGATION);
+        long histogramCount = ((HistogramPointData) pointData).getCount();
+        if (histogramCount <= Integer.MAX_VALUE && histogramCount >= Integer.MIN_VALUE) {
+          pointBuilder.setCount((int) histogramCount);
+        }
+        pointBuilder.setValue(((HistogramPointData) pointData).getSum());
```
trask, line 70: we should capture min/max also after 1.13.0 is released: https://github.com/open-telemetry/opentelemetry-java/pull/4317

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/builders/MetricTelemetryBuilder.java
```diff
@@ -69,8 +69,14 @@ public void setMetricPoint(MetricPointBuilder point) {
     }
   }
 
+  // visible for testing
+  public List<MetricDataPoint> getMetricPoint() {
+    return data.getMetrics();
+  }
```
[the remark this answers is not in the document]
heyams: build() is returning TelemetryItem.. it's not MetricsData.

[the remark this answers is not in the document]
trask: you can get MetricsData from the TelemetryItem, I pushed change to your branch

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsDataTest.java
```diff
@@ -0,0 +1,313 @@
...
+    meter
+        .gaugeBuilder("testDoubleGauge")
+        .setDescription("the current temperature")
+        .setUnit("C")
+        .buildWithCallback(
+            m -> {
+              m.record(20.0, Attributes.of(AttributeKey.stringKey("thing"), "engine"));
+            });
+
+    Thread.sleep(90 * 1000); // wait 90 seconds
```
trask, line 123: let's set the metric interval smaller so the tests don't take so long, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/31f5734be6946b074e41466019382a3cb0e18422/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/LibraryTestRunner.java#L60-L62

heyams: well, opentelemetry sdk metrics take at least 1 min.

trask: that's configurable (see link above)

I pushed change to your branch

heyams: i forgot that part.. cool thanks.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(AzureMonitorMetricExporter.class);
+  private final AtomicBoolean stopped = new AtomicBoolean();
+
+  public AzureMonitorMetricExporter(TelemetryClient telemetryClient) {
+    this.telemetryClient = telemetryClient;
+  }
+
+  static {
+    EXCLUDED_METRIC_NAMES.add("http.server.active_requests");
+    EXCLUDED_METRIC_NAMES.add("http.server.duration");
```
trask, line 60: let's exclude also:

* `http.client.duration`
* `rpc.client.duration`
* `rpc.server.duration`

I'm a little concerned about keeping up-to-date with these. Can we add validation in all of our smoke tests to ensure that extra "custom" metrics are not received?

test/smoke/testApps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTest.java
```diff
@@ -0,0 +1,255 @@
...
+    assertEquals(tags.get("ai.application.ver"), "123");
+
+    // validate base data
+    MetricData md = (MetricData) ((Data<?>) envelope1.getData()).getBaseData();
+    List<DataPoint> dataPointList = md.getMetrics();
+    assertEquals(1, dataPointList.size());
+    DataPoint dp = dataPointList.get(0);
+    double expectedValue = 2.0;
+    double epsilon = Math.ulp(expectedValue);
+    assertEquals(expectedValue, dp.getValue(), epsilon);
```
trask, line 185: I don't think the usage in these tests can have floating point issues, I'd probably go with

```suggestion
    assertEquals(expectedValue, dp.getValue(), 0);
```

(I did similar here in MicrometerTest: https://github.com/microsoft/ApplicationInsights-Java/blob/0c054711d73d803d2e45c58d24216df09cab20bf/test/smoke/testApps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerTest.java#L74)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
heyams: smoke tests took care of it.. otherwise it won't pass. it expected the right metrics. size won't match if custom metrics show up. where will these custom metrics go then? I need more context.

trask: rpc.client.duration and rpc.server.duration should get generated in the gRPC smoke test

http.client.duration should get generated in HttpClient smoke test

trask: what I was thinking, is if we can add validation to _all_ smoke tests, that way if otel starts generating messaging metrics, that would cause our JMS and Kafka smoke tests to fail and we would know to add those to the exclusion list

heyams: that can be a separate pr. will exclude those three.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,112 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(AzureMonitorMetricExporter.class);
+  private final AtomicBoolean stopped = new AtomicBoolean();
+
+  public AzureMonitorMetricExporter(TelemetryClient telemetryClient) {
+    this.telemetryClient = telemetryClient;
+  }
+
+  static {
+    EXCLUDED_METRIC_NAMES.add("http.server.active_requests"); // Micrometer
+    EXCLUDED_METRIC_NAMES.add("http.server.duration"); // Micrometer
```
trask, line 60:
```suggestion
    EXCLUDED_METRIC_NAMES.add("http.server.active_requests"); // Servlet
    EXCLUDED_METRIC_NAMES.add("http.server.duration"); // Servlet
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,112 @@
...
+  public CompletableResultCode shutdown() {
+    stopped.set(true);
+    return CompletableResultCode.ofSuccess();
+  }
+
+  private List<TelemetryItem> convertOtelMetricToAzureMonitorMetric(MetricData metricData) {
+    List<TelemetryItem> telemetryItems = new ArrayList<>();
+    for (PointData data : metricData.getData().getPoints()) {
+      MetricTelemetryBuilder builder = telemetryClient.newMetricTelemetryBuilder();
+      builder.setTime(FormattedTime.offSetDateTimeFromNow());
```
trask, line 106: I believe otel metric data has its own time, let's use that instead of "now"

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,112 @@
...
+      AzureMonitorMetricsData.populatePointAndProperties(builder, metricData, data);
```
trask, line 107: I think it makes sense to pull this method into this class (it's only used by this class, and this class is not so large that it needs to be broken apart)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java
```diff
@@ -0,0 +1,92 @@
...
+import io.opentelemetry.sdk.metrics.data.DoublePointData;
+import io.opentelemetry.sdk.metrics.data.HistogramPointData;
+import io.opentelemetry.sdk.metrics.data.LongPointData;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.data.MetricDataType;
+import io.opentelemetry.sdk.metrics.data.PointData;
+
+final class AzureMonitorMetricsData {
+
+  private static final String DEFAULT_METRIC_AGGREGATION_INTERVAL = "60000";
```
trask, line 38: can we get the interval from the metric data somehow? in the agent we could use the configured metric interval since we configure that, but for using the metric exporter standalone, we don't have control and users could set the reader interval to anything

trask (commented):
are we getting delta or cumulative metrics? i believe the otel default is cumulative, but we want delta, let's try to add some kind of verification in the smoke test for this

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
heyams: i didn't change it.. it will use whatever otel default. .net is the same. should we get some consensus from other platform first?

heyams:
https://github.com/microsoft/ApplicationInsights-Java/pull/2214#pullrequestreview-936813977

I used the otel default. .net does the same. should we get some consensus from other platforms first before changing it to delta?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
heyams: which one:

 ```
datapoint:
this.startEpochNanos
this.epochNanos
```

i think use now makes more sense since that's when azure monitor metric is created.  in my opinion the time in otel metric doesn't matter.

Comment (trask) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java:106]: I believe the "correct" time is the time that the metric was gathered (e.g. for cpu metric it would be what time the cpu metric was read from the O/S), so in this sense I think the otel metric time is better

File: agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java (hunk 13 above)
Comment (heyams) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: interval is set at the reader level.. our exporter doesn't have that info.. metricData doesn't have that either.

Comment (heyams) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: can you propose a feature request to upstream to make that property available in MetricData?

Comment (trask) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: I'm not sure that it's not already available, what about MetricData `epochNanos - startEpochNanos`, I'm thinking that may be the interval

**Comment by trask** [comment]
> [#2214 (review)](https://github.com/microsoft/ApplicationInsights-Java/pull/2214#pullrequestreview-936813977)
> 
> I used the otel default. .net does the same. should we get some consensus from other platforms first before changing it to delta?

all of the existing Application Insights metrics we report as delta, e.g. cpu, so it is surprising to me that we wouldn't continue the same behavior

**Comment by trask** [comment]
> all of the existing Application Insights metrics we report as delta, e.g. cpu, so it is surprising to me that we wouldn't continue the same behavior

I agree we should get consensus, but for this PR let's capture delta since that's most consistent with the existing javaagent metrics (including micrometer)

File: agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java (hunk 13 above)
Comment (trask) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: can you test with epochNanos - startEpochNanos to see if it works?

Comment (trask) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: getting the interval right is important, so I'd like to at least try to get it right in this PR if possible

Comment (heyams) [reply agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricsData.java]: i was able to confirm epochNanos - startEpochNanos is the interval. i tested it with different intervals.. result is persisting.

**Comment by heyams** [comment]
@trask I think I have addressed all your comments.

**Review by trask — APPROVED** [review]
I merged in otel 1.13.0 to vnext and merged vnext into this branch, so let's go ahead and add min/max in this PR now.

`OpenTelemetryMetrics` smoke test looks like it's failing, not sure if I caused that or not.

Other than that this looks good to go 🎉

**Comment by heyams** [comment]
i will add min/max in a separate pr.

File: test/smoke/testApps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTest.java [hunk 14]
```diff
@@ -109,8 +109,8 @@ private void validateHistogramMetric(String name) throws Exception {
     // validate custom dimension
     Map<String, String> properties = md.getProperties();
     int aggregationIntervalMs = Integer.parseInt(properties.get("_MS.AggregationIntervalMs"));
-    assertThat(aggregationIntervalMs, greaterThanOrEqualTo(30000));
-    assertThat(aggregationIntervalMs, lessThanOrEqualTo(90000));
+    assertThat(aggregationIntervalMs, greaterThanOrEqualTo(60000));
+    assertThat(aggregationIntervalMs, lessThanOrEqualTo(180005));
```
trask, line 113: this is a very big range, and makes me think that the `aggregationIntervalMs` is not the interval but the total time since the first interval?

heyams: what do you suggest reverting it back to 1 min? I thought it's test delay.. i didn't think that way. i was able to test interval at the reader level for sec, or even 15 mins..

heyams: or i can remove this assert statement for now..

trask:
🎉

PR #2217, opened 2022-04-05
Sync the two span exporters more

trask:


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
```diff
@@ -227,9 +219,12 @@ private void exportRemoteDependency(
     RemoteDependencyTelemetryBuilder telemetryBuilder = RemoteDependencyTelemetryBuilder.create();
     initTelemetry(telemetryBuilder);
 
+    float samplingPercentage = 100; // TODO
```
heyams, line 222: the baseline datamodel handle this for all telemetry. not sure why you need to do this specifically?

heyams: it would be better to check config's sampling setting first.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
```diff
@@ -593,10 +588,12 @@ private void exportRequest(SpanData span, List<TelemetryItem> telemetryItems) {
 
     Attributes attributes = span.getAttributes();
     long startEpochNanos = span.getStartEpochNanos();
+    float samplingPercentage = 100; // TODO
```
heyams, line 591: same here

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
heyams: i also noticed TelemetryUtil.getSamplingPercentage.

trask: ya, we don't support samplingPercentage in this exporter yet (that's what the TODO is for), I'm just sync'ing the code between the two span exporters as much as possible, then will try to unify the codebase

heyams:
@trask can you fix the conflict? i will take a second look right after.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileSenderTelemetryPipelineListener.java
```diff
@@ -52,4 +53,9 @@ public void onException(
       TelemetryPipelineRequest request, String errorMessage, Throwable throwable) {
     localFileLoader.updateProcessedFileStatus(false, file);
   }
+
+  @Override
+  public CompletableResultCode shutdown() {
+    return CompletableResultCode.ofSuccess();
```
heyams, line 59: should we add a shutdown boolean and set it here?
onResponse should do nothing when shutdown is true.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/logging/DiagnosticTelemetryPipelineListener.java
```diff
@@ -105,6 +106,11 @@ public void onException(TelemetryPipelineRequest request, String reason, Throwab
     }
   }
 
+  @Override
+  public CompletableResultCode shutdown() {
+    return CompletableResultCode.ofSuccess();
```
heyams, line 111: same here.

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/pipeline/TelemetryPipelineListener.java
```diff
@@ -73,5 +86,10 @@ public void onResponse(TelemetryPipelineRequest request, TelemetryPipelineRespon
     @Override
     public void onException(
         TelemetryPipelineRequest request, String errorMessage, Throwable throwable) {}
+
+    @Override
+    public CompletableResultCode shutdown() {
+      return CompletableResultCode.ofSuccess();
```
heyams, line 92: same here.. it's doing nothing.. can we make the interface shutdown optional? not required for all listener where it return success and does nothing?

trask: let's reconsider what to do with these shutdowns once we think more about non-javaagent usage, they aren't really relevant for javaagent since we it lives for the length of the JVM process

heyams (approved):

PR #2221, opened 2022-04-07
Turn off Statsbeat when proxy is used or any exception from the server

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiComponentInstaller.java
```diff
@@ -228,8 +228,11 @@ private static AppIdSupplier start(Instrumentation instrumentation) {
       RpConfigurationPolling.startPolling(rpConfiguration, config, telemetryClient, appIdSupplier);
     }
 
-    // initialize StatsbeatModule
-    statsbeatModule.start(telemetryClient, config);
+    // initialize StatsbeatModule and don't start Statsbeat when proxy is used
+    String proxyHost = config.proxy.host;
+    if (proxyHost == null || proxyHost.isEmpty()) {
+      statsbeatModule.start(telemetryClient, config);
+    }
```
trask, line 235: will this disable statsbeat when `-Dhttps.proxyHost` is used?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -364,19 +362,21 @@ private void handleStatsbeatOnResponse(
       String instrumentationKey, Consumer<Boolean> onFailure, OperationLogger operationLogger) {
 
     return error -> {
-      if (isStatsbeat
-          && (error instanceof UnknownHostException
-              || error instanceof UnresolvedAddressException)) {
-        // when sending a Statsbeat request and server returns an UnknownHostException, it's
-        // likely that it's using AMPLS. In that case, we use the kill-switch to turn off Statsbeat.
+      if (isStatsbeat && error instanceof Exception) {
```
trask, line 365: I'm not sure there's much reason to keep the `instanceof Exception` condition
```suggestion
      if (isStatsbeat) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -364,19 +362,21 @@ private void handleStatsbeatOnResponse(
...
+        // when sending a Statsbeat request and server returns an Exception, it's
+        // likely that it's using AMPLS or other private endpoints. In that case, we use the
+        // kill-switch to turn off Statsbeat.
+        // TODO (heya) track Statsbeat exception count via a new Statsbeat metric
```
trask, line 369: I don't follow the TODO, can you elaborate a bit?

heyams: do you see a value to know when people are running into exception when sending statsbeat? here we just shut it down and forget about it. that is what I meant to track exception count in this case for statsbeat specific requests.

trask: how would we track it if we're shutting down statsbeat?

trask (approved):
👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -364,19 +364,21 @@ private void handleStatsbeatOnResponse(
       String instrumentationKey, Consumer<Boolean> onFailure, OperationLogger operationLogger) {
 
     return error -> {
-      if (isStatsbeat
-          && (error instanceof UnknownHostException
-              || error instanceof UnresolvedAddressException)) {
-        // when sending a Statsbeat request and server returns an UnknownHostException, it's
-        // likely that it's using AMPLS. In that case, we use the kill-switch to turn off Statsbeat.
+      if (isStatsbeat) {
```
hectorhdzg, line 367: If you shutdown Statsbeat with any error you will see a huge drop of the data, Breeze returning errors is quite normal, we made similar changes in Node.js SDK [here](https://github.com/microsoft/ApplicationInsights-node.js/pull/922) and I actually updated the logic as I saw a significant drop of Statbeat with the specific version with that change [here](https://github.com/microsoft/ApplicationInsights-node.js/pull/924)

heyams: we only shut down statsbeat when sending statsbeat request. it doen't impact customers' telemetry though. we have seen that AMPL will run into different exceptions for different apps. we need to shut it down until we can have a definite way of detecting AMPL or any other private endpoints.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: Updated todo.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: @heyams what if we only shut down statsbeat if the first request fails?

(and still suppress logging always for statsbeat errors)

hectorhdzg: What I'm saying is that you will get a huge drop of Statsbeat data for all customers even in not AMPL because network error is quite common for Breeze, not talking about customer telemetry here, @trask suggestion could work as is more about the probability of first Breeze call to fail, I will argue you can also expect a drop of Statsbeat in that case but definitely better, maybe a 5 in a row fails and then disable.

heyams: shut down after 3 errors in a row.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -364,19 +368,23 @@ private Consumer<Throwable> errorHandler(
...
-              || error instanceof UnresolvedAddressException)) {
-        // when sending a Statsbeat request and server returns an UnknownHostException, it's
-        // likely that it's using AMPLS. In that case, we use the kill-switch to turn off Statsbeat.
+      if (isStatsbeat && statsbeatErrorCount.getAndIncrement() >= MAX_STATSBEAT_ERROR_COUNT) {
+        // when sending a Statsbeat request and server returns an Exception 3 times in a row, it's
+        // likely that it's using AMPLS or other private endpoints. In that case, we use the
+        // kill-switch to turn off Statsbeat.
+        // TODO need to figure out a way to detect AMPL or we can let the new ingestion service to
+        // handle this case for us when it becomes available.
+        statsbeatErrorCount.set(0);
```
hectorhdzg, line 377: statsbeatErrorCount should be reset when there is no error, here you should update isStatsbeat flag I believe so this code is never triggered again

heyams: reset should happen after max is reached. after that this block of code shouldn't happen since statsbeat has been shutting down. isStatsbeat is sent for every statsbeat request. does it make sense?  

getAndIncrement will return the previous value and then increment it. when it reaches 3, reset and then shutdown. reset or not doesn't matter since no more statsbeat going out.

hectorhdzg: If you don't reset the retry counter on success if you have 3 errors even if they are not consecutive, you will stop Statsbeat

trask: I think also we only care if it's the very first 3 requests that error

heyams: Updated.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiComponentInstaller.java
```diff
@@ -228,8 +228,15 @@ private static AppIdSupplier start(Instrumentation instrumentation) {
...
-    statsbeatModule.start(telemetryClient, config);
+    // initialize StatsbeatModule and don't start Statsbeat when proxy is used
+    String configProxyHost = config.proxy.host;
+    String systemPropertyProxyHost = System.getProperty("https.proxyHost");
+    if (configProxyHost == null
+        || configProxyHost.isEmpty()
+        || systemPropertyProxyHost == null
+        || systemPropertyProxyHost.isEmpty()) {
+      statsbeatModule.start(telemetryClient, config);
+    }
```
trask, line 239: with no-logging and kill-switch after failures, I don't think we need this anymore?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -264,6 +266,10 @@ private CompletableResultCode internalSend(
                 instrumentationKey,
                 startTime,
                 () -> {
+                  if (isStatsbeat) {
+                    statsbeatErrorCount.set(
+                        0); // reset error count when non-consecutive errors occur
+                  }
```
trask, line 272: how about setting a flag like `atLeastOneStatsbeatSuccess=true` instead

i don't think we need to shutdown if we see 3 consecutive errors if it's been running successfully already

heyams: this is pretty much doing the same thing. when success reset the error counter. if it's 3 consecutive errors, we want to shut it down. i don't get your point.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiComponentInstaller.java
heyams: well, errorhandler is for exception from breeze, errorcode is handled differently. that IcM is returning 403 for statsbeat because of private points not AMPL. Error handler is for AMPL case. we still need to handle proxy case though.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: thinking it a bit more, i might understand now.. 
saying 1st request is successful, 2nd - 4th were errors.. then in that case we don't want to shut it down since we know it's temporary. that's a valid point.

heyams: but what about 4th statsbeat is a success, in that case we don't want to shut it down either. it might be temporary too. 
should i increase the counter to max 10?

hectorhdzg (approved):
LGTM

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiComponentInstaller.java
trask: ah, I was assuming that we would kill-switch on three consecutive "failures" meaning anything other than 200?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -86,6 +85,10 @@
   //  operationLogger?
   private final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
+  private static final int MAX_STATSBEAT_ERROR_COUNT = 10;
+  private static final AtomicInteger statsbeatErrorCount = new AtomicInteger();
+  private static final AtomicBoolean atLeastOneStatsbeatSuccess = new AtomicBoolean();
```
trask, line 90: not using atomicity
```suggestion
  private static final volatile boolean atLeastOneStatsbeatSuccess;
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -365,18 +373,24 @@ private void handleStatsbeatOnResponse(
...
-        // when sending a Statsbeat request and server returns an UnknownHostException, it's
-        // likely that it's using AMPLS. In that case, we use the kill-switch to turn off Statsbeat.
+          && !atLeastOneStatsbeatSuccess.get()
+          && statsbeatErrorCount.getAndIncrement() >= MAX_STATSBEAT_ERROR_COUNT) {
+        // when sending a Statsbeat request and server returns an Exception 10 times in a row, it's
+        // likely that it's using AMPLS or other private endpoints. In that case, we use the
+        // kill-switch to turn off Statsbeat.
+        // TODO need to figure out a way to detect AMPL or we can let the new ingestion service to
+        // handle this case for us when it becomes available.
+        statsbeatErrorCount.set(0);
```
trask, line 383: don't need to reset anymore with new `atLeastOneStatsbeatSuccess` logic
```suggestion
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -86,6 +85,10 @@
   //  operationLogger?
   private final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
+  private static final int MAX_STATSBEAT_ERROR_COUNT = 10;
```
trask, line 88: 10 seems like a lot, I think if the first three requests all fail, that's a pretty solid indicator

heyams: i gave an example responding to your other comments above. i think 10 is better and it's consistent with breeze retry limit.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: i prefer to keep AtomicBoolean since setting on volatile is not atomic. statsbeat is at least every 15 mins. there is no perf gain in this case.

trask: you aren't using atomicness though

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: 10 is breeze redirect limit, not retry limit?

heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/51/files#:~:text=10%20redirects%20triggered%20in%20succession%20(circular%20redirect)%2C%20we%20will%20drop%20the%20data%20and%20count%20it%20as%20an%20exception.

redirect, not retry.

trask: right, so what does that have to do with why we should use 10 here?

trask:
@heyams @hectorhdzg I pushed some changes:

* suppress all logging of statsbeat telemetry failures (introduced a noop logger to make this easier, there were many additional places that would have needed to be wrapped in conditionals)
* if statsbeat is shut down, then stop sending statsbeat telemetry from disk (which will likely keep failing and getting stored back to disk every 30 seconds until they expire after 2 days)
* shut down statsbeat if the first X statsbeat ingestions requests all fail for any reason other than throttling (439/402), and removed the special case for http proxy
  * I believe this is both simpler than special casing for http proxy, and will cover potentially other unknown environmental limitations that we may be unaware of currently

hectorhdzg (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -73,19 +74,17 @@ public class TelemetryChannel {
...
   private final OperationLogger operationLogger;
   private final OperationLogger retryOperationLogger;
 
   // TODO (kryalama) do we still need this AtomicBoolean, or can we use throttling built in to the
   //  operationLogger?
   private final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
-  private final AtomicInteger statsbeatErrorCount = new AtomicInteger();
-  private final AtomicBoolean statsbeatShutdown = new AtomicBoolean();
+  private final AtomicInteger statsbeatUnableToReachBreezeCounter = new AtomicInteger();
```
heyams, line 86: can we rename it to statsbeatUnableToReachServerCounter? breeze is for internal.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -73,19 +74,17 @@ public class TelemetryChannel {
...
+  private final AtomicBoolean statsbeatHasBeenShutdown = new AtomicBoolean();
 
-  private volatile boolean statsbeatAtLeastOneSuccess;
+  private volatile boolean statsbeatHasReachedBreezeAtLeastOnce;
```
heyams, line 89: statsbeatHasReachedServerAtLeastOnce?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -299,6 +298,11 @@ private CompletableResultCode internalSend(
     return result;
   }
 
+  // not including 401/403/503 in this list because those are commonly returned by proxy servers
+  // when they are not configured to allow traffic for westus-0
+  private static final Set<Integer> RESPONSE_CODES_INDICATING_REACHED_BREEZE =
+      new HashSet<>(asList(200, 206, 402, 408, 429, 439, 500));
```
heyams, line 306: 401?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -282,6 +300,11 @@ private CompletableResultCode internalSend(
...
```
heyams, line 306: 307/308

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: I think "Server" is vague, e.g. which server did it reach, a proxy server? I don't mind renaming it to "IngestionService" though then the var name becomes even longer. This is internal code, and the behavior is specific to Breeze, so I think Breeze is ok, but will rename to "IngestionService" if you really prefer that.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: redirect response codes only get here if they are of the infinite redirect loop variety

do infinite redirects count as "probably from breeze"? I'm good either way, can you update the spec to call that out though?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -121,6 +123,10 @@ public CompletableResultCode sendRawBytes(
       String instrumentationKey,
       Runnable onSuccess,
       Consumer<Boolean> onFailure) {
+    if (isStatsbeat && statsbeatHasBeenShutdown.get()) {
```
heyams, line 126: statsbeathasBeenshutdown.get() will never get called since statsbeat has been shutdown.

heyams: i.e. sendRawByte will never get called again after the shut down.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: image

i did cover redirect in the spec here.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -74,18 +76,18 @@ public class TelemetryChannel {
...
-          TelemetryChannel.class, "Sending telemetry to the ingestion service (retry)");
+  private final OperationLogger operationLogger;
+  private final OperationLogger retryOperationLogger;
 
   // TODO (kryalama) do we still need this AtomicBoolean, or can we use throttling built in to the
   //  operationLogger?
   private final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
+  private final AtomicInteger statsbeatUnableToReachBreezeCounter = new AtomicInteger();
+  private final AtomicBoolean statsbeatHasBeenShutdown = new AtomicBoolean();
```
heyams, line 87: this boolean is not needed.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: what about retries from disk?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: ya, I'm asking specifically about "infinite redirect loops" (the ones that we cap at 10 redirects and then fail)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: breeze is going away.. do you plan to update this to a newer system name? I'm ok with IngestionService.

trask: do we know that the new system will be backwards compatible with breeze? if not, then we will have to revisit this anyways at that time

heyams: i think so. let's use ingestionService and not reveal our internal system name in a public repo?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: that is handled by default. I mentioned that in the statsbeat spec.. not in "how to handle statsbeat failures" section though.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: can we make shutdown handle that?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: > not reveal our internal system name in a public repo

we use the term breeze in several other places already if that is your concern

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: you still haven't answered my question though:

> do infinite redirects count as "probably from breeze"?

heyams: image

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: I don't understand, can you explain more what you mean?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: oops.. we did. then yeah it's ok to leave as it is..

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: when we call statsbeatModule.shutdown(), that method will stop the job scheduler + stop the loader statsbeat telemetry from disk?

heyams: or just stop the job scheduler for LocalFileLoader for statsbeat. I think there is a shutdown method too..

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: ok great, so we shouldn't add 307/308 here

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: ah I understand, that's a good idea, not super easy to implement, and will cause (even more) conflicts, are you ok to revisit this in vnext branch after I merge these changes forward?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
```diff
@@ -389,6 +413,21 @@ private Consumer<Throwable> errorHandler(
     };
   }
 
+  private void statsbeatDidNotReachBreeze() {
+    if (statsbeatUnableToReachBreezeCounter.getAndIncrement() >= 10
+        && !statsbeatHasBeenShutdown.getAndSet(true)) {
```
ramthi, line 418: we never set statsbeatHasBeenShutdown to false.  Like @heyams mentioned above, this boolean seems useless.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
ramthi, line 306: 307/308 are in the spec.  The spec hasn't been approved/merged.  We need to be in sync.

trask: see discussion at https://github.com/microsoft/ApplicationInsights-Java/pull/2221#discussion_r848964564

trask: I will add comment in the code also to explain why they are missing

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: this line that you have highlighted sets the boolean, it is not useless

trask: @heyams comment was that it was useless if we implement her suggestion above of shutting down statsbeat retries from disk (which I'd like to defer until `vnext` branch to limit merge conflicts at this point)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: comment added

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: sure, let's add a todo we can change it in vnext.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
heyams: that makes sense for reaching redirect up to 10 times.. but what about 1 time 307/308 that is successful, should we consider that as one time success from breeze?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: TODO added

heyams:
Approved.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
trask: as we discussed offline, I think we ignore the intermediate 307/308s and only factor in the "final" response code (after any redirects)

heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/58/commits/3eb967e786a310c9c34bbb06cac0ab75f59cef50

I updated the spec to reflect this.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java
ramthi: Is [line](https://github.com/microsoft/ApplicationInsights-Java/blob/0dfa662f055e3dc3a54e74f5c1cc3fb9a5eb5ab3/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryChannel.java#L127) the place where we retry from disk?  Sorry, it wasn't obvious for me that that is what sendRawBytes did.  If that is the case then this is all good.  Ship it!!

Issue #2000, opened 2021-12-10
Startup impact  of application insight on microservices startup


Sachin1O1:
What's the minimum / maximum startup impact of application insight on microservices startups?

Sachin1O1:
any update?

trask:
hi @Sachin1O1, it depends a lot on the application and the compute resources. there's also additional overhead on Java 8 because the agent jar is signed, and the Java JIT compiler is not activated until later on in the startup process on Java 8 (this is not an issue on Java 11).

Sachin1O1:
Cool, but still we can have some minimum benchmark for the resources required. What do you say?

Sachin1O1:
Hi @trask , I have some stats on how my service startup time is impacted by the application insight jar

Service startup sample

Sr no | With   jar | Without   jar
--       | --            | --
1        | 260.21    | 102.6
2        | 170         | 103
3        | 274         | 180
4        | 165         | 99

Container resources

resources:
            limits:
              cpu: 300m
              memory: 700Mi
            requests:
              cpu: 100m
              memory: 400Mi


Java

Java 11 (OpenJ9)

As you can see it inc startup time by almost X2

heyams:
I'm curious to know which version of agent were you testing against?  Did you try the latest GA/BETA version?

Sachin1O1:
Hi @heyams,

I am using Application Insights Java 3.2.4 (GA)

Sachin1O1:
@trask @heyams any update on this.

heyams:
i'm working on a debug startup profiler.. so that you can provide thread dump to us for further debugging. we're not able to repro it.

heyams:
@Sachin1O1 can you use this [3.2.6-BETA-SNAPSHOT](https://github.com/microsoft/ApplicationInsights-Java/suites/5209427074/artifacts/159305758) and help us collect thread dump from your app? 

Please set `-Dapplicationinsights.debug.startupProfiling` to `true `in your jvm arguments:
` java -javaagent:applicationinsights-agent-3.2.6-BETA-SNAPSHOT.jar -Dapplicationinsights.debug.startupProfiling=true`. 

Let it run till your app has started. Then send me (helen.yang@microsoft.com) the file called **stacktrace.txt** located in your temp folder. The full path is available in the applicationinsights.log. Something like `C:\Users\${USER-NAME}\AppData\Local\Temp\applicationinsights`

heyams:
[...]

PR #2224, opened 2022-04-10
Faster startup

trask:
Related to #2000

Needs to be documented as potentially breaking change if using `captureControllerSpans`.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -388,6 +388,10 @@ public static class PreviewInstrumentation {
         new DisabledByDefaultInstrumentation();
 
     public DisabledByDefaultInstrumentation vertx = new DisabledByDefaultInstrumentation();
+
+    // this is opt-in because it can cause startup slowness due to expensive matchers
+    public DisabledByDefaultInstrumentation jaxrsAnnotations =
```
heyams, line 393: disabled is default. let's track enabled via Feature statsbeat?

heyams (approved):

.github/workflows/pr.yaml
```diff
@@ -61,6 +61,11 @@ jobs:
           - ":test:smoke:testApps:CoreAndFilter"
           - ":test:smoke:testApps:CoreAndFilter"
           - ":test:smoke:testApps:CoreAndFilter"
+          - ":test:smoke:testApps:CoreAndFilter"
```
heyams, line 64: how does adding this same line fix the smoke tests?

trask: it's failing sporadically, so I'm running it lots of times to get a repro while I'm troubleshooting it

trask: oh, and I also updated azure sdk libraries, so maybe that fixed it?? I'll run a few 10s of more times to see...

trask: the sporadic failure is still there, I removed the debug code, so we can merge this PR since it's currently blocking #2214, and I'll continue troubleshooting the sporadic test seprately

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: added ptal

heyams (approved):

PR #2228, opened 2022-04-14
Merge main into vnext

trask:
merge don't squash

### DIFF
--- a/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
+++ b/agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
@@ -27,12 +27,15 @@ import com.azure.monitor.opentelemetry.exporter.implementation.pipeline.Telemetr
 import com.azure.monitor.opentelemetry.exporter.implementation.pipeline.TelemetryPipelineResponse;
 import com.azure.monitor.opentelemetry.exporter.implementation.utils.StatusCodes;
 import java.io.File;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 public class LocalStorageTelemetryPipelineListener implements TelemetryPipelineListener {
 
   private final LocalFileWriter localFileWriter;
   private final LocalFileSender localFileSender;
   private final LocalFilePurger localFilePurger;
+
+  private final AtomicBoolean shutdown = new AtomicBoolean();
 
   // telemetryFolder must already exist and be writable
   public LocalStorageTelemetryPipelineListener(
@@ -47,8 +50,12 @@ public class LocalStorageTelemetryPipelineListener implements TelemetryPipelineL
   }
 
   public void shutdown() {
-    localFileSender.shutdown();
-    localFilePurger.shutdown();
+    // guarding against multiple shutdown calls because this can get called if statsbeat shuts down
+    // early because it cannot reach breeze and later on real shut down (when running not as agent)
+    if (!shutdown.getAndSet(true)) {
+      localFileSender.shutdown();
+      localFilePurger.shutdown();
+    }
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatTelemetryPipelineListener.java
```diff
@@ -0,0 +1,93 @@
...
+  // reached the 10 redirect threshold, in which case they are considered non-retryable exceptions
+  private static final Set<Integer> RESPONSE_CODES_INDICATING_REACHED_BREEZE =
+      new HashSet<>(asList(200, 206, 402, 408, 429, 439, 500));
+
+  private final StatsbeatModule statsbeatModule;
+  private final Runnable localStorageTelemetryShutdownFunction;
+
+  private final AtomicInteger statsbeatUnableToReachBreezeCounter = new AtomicInteger();
+  // TODO (trask) remove this boolean and shutdown the disk loader for statsbeat instead
+  private final AtomicBoolean statsbeatHasBeenShutdown = new AtomicBoolean();
```
heyams, line 48: we can remove this now?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatTelemetryPipelineListener.java
```diff
@@ -0,0 +1,93 @@
...
+      // shutting down statsbeat because it's unlikely that it will ever get through at this point
+      // some possible reasons:
+      // * AMPLS
+      // * proxy that has not been configured to allow westus-0
+      // * local firewall that has not been configured to allow westus-0
+      //
+      // TODO need to figure out a way that statsbeat telemetry can be sent to the same endpoint as
+      // the customer data for these cases
+      statsbeatModule.shutdown();
+      localStorageTelemetryShutdownFunction.run();
```
heyams, line 90: since you made shutdown take care of loading from disk.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/StatsbeatTelemetryPipelineListener.java
trask: done, had a small ripple effect, ptal

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
```diff
@@ -47,8 +50,12 @@ public LocalStorageTelemetryPipelineListener(
   }
 
   public void shutdown() {
-    localFileSender.shutdown();
-    localFilePurger.shutdown();
+    // guarding against multiple shutdown calls because this can get called if statsbeat shuts down
+    // early because it cannot reach breeze and later on real shut down (when running not as agent)
+    if (!shutdown.getAndSet(true)) {
+      localFileSender.shutdown();
```
heyams, line 56: Can we add a one time log?

trask: it's not uncommon for shutdown to be silent when called multiple times, I think I'd prefer to leave it unless you have a specific reason why you want to log it

heyams: we log when shutting down statsbeat scheduler.. i thought it will nice to log here when shutting down disk persistence. i think you misunderstood what i meant?

trask: ah, as debug?

trask: I went to add it but realized it could be confusing because in this class we don't know if it's statsbeat or not, and so it's not super clear what exactly to log that wouldn't cause more confusion than less. I think i'd prefer to leave it as is if that's ok.

heyams: yeah, it's only for Statsbeat. i'm ok with it.

heyams: i didn't test this part with 3.2.11. might need to test it out later.

heyams (approved):

PR #2233, opened 2022-04-18
Take care of live endpoint suffix.

Incarnation-p-lee:
* The AI portal has export the endpoint of live similar like this `LiveEndpoint=https://xxx.com/quickpulseservice.svc`.
* The existing build URL will always add the suffix `quickpulseservice.svc`, which will result in end to end failure for live metrics.
* quickpulseservice.svc will be added if not present.
* nothing will be done if quickpulseservice.svc is present.

 Signed-off-by: Pan Li <panli@microsoft.com>


trask:
thanks @Incarnation-p-lee! we discussed with the service team and they are fixing the connection string (it shouldn't have the `quickpulseservice.svc` suffix)

Incarnation-p-lee:
@trask Sure, let me close this PR.

PR #2238, opened 2022-04-21
Fix logging capture

trask:
And added a smoke test for the case where logs are not part of an existing trace


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/InheritedAttributesLogProcessor.java
```diff
@@ -71,6 +71,7 @@ private static List<AttributeKey<?>> buildInheritedAttributesList(
   public void emit(LogData log) {
     Span currentSpan = Span.current();
     if (!(currentSpan instanceof ReadableSpan)) {
+      delegate.emit(log);
```
heyams, line 74: this is considered a log not part of the trace?

heyams, line 74: this is considered a log not part of the trace?

trask: yes, we capture logs that are not part of a request also

heyams (approved):

PR #2239, opened 2022-04-21
Suppress statsbeat warnings

trask:
Noticed in local testing that I missed retry from disk was logging statsbeat failures in vnext (not a problem in main).

Also extended suppressing statsbeat warnings a bit further.


agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/logging/DiagnosticTelemetryPipelineListener.java
```diff
@@ -44,12 +44,12 @@ public class DiagnosticTelemetryPipelineListener implements TelemetryPipelineLis
   private static final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
   private final OperationLogger operationLogger;
-  private final boolean shouldLogWarnings;
+  private final boolean shouldLogRetryableFailure;
```
heyams, line 47: can we use the same variable name "suppressWarnings" for consistency?

agent/agent-tooling/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/logging/DiagnosticTelemetryPipelineListener.java
```diff
@@ -44,12 +44,13 @@ public class DiagnosticTelemetryPipelineListener implements TelemetryPipelineLis
   private static final AtomicBoolean friendlyExceptionThrown = new AtomicBoolean();
 
   private final OperationLogger operationLogger;
-  private final boolean shouldLogWarnings;
+  private final boolean suppressWarningsOnRetryableFailures;
```
heyams, line 47: are you going to add another boolean to suppress logging for other error types? I think error type is irrelevant in this case.

trask: the idea here is to suppress retryable failures, since those will be retried later anyways (and then we will log if the retry fails)

heyams (approved):

PR #2246, opened 2022-04-25
Update to latest otel

trask:
Brings jboss-modules mdc instrumentation


heyams:
does logger and metrics exporters need to flush too?

trask:
> does logger and metrics exporters need to flush too?

log exporter is getting flushed.

metrics exporter doesn't need flush

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporter.java
```diff
@@ -119,7 +119,7 @@ private List<TelemetryItem> convertOtelMetricToAzureMonitorMetric(MetricData met
 
     for (PointData pointData : metricData.getData().getPoints()) {
       MetricTelemetryBuilder builder = telemetryClient.newMetricTelemetryBuilder();
-      builder.setTime(FormattedTime.offSetDateTimeFromEpochNanos(pointData.getStartEpochNanos()));
+      builder.setTime(FormattedTime.offSetDateTimeFromEpochNanos(pointData.getEpochNanos()));
```
trask, line 122: @heyams I think this is more expected to report the end time, and is consistent with what we do for micrometer pre-aggregated metrics

heyams (approved):

PR #2271, opened 2022-05-11
Fix pipeline (alt)

trask:


.scripts/copy-from-maven-local.cmd
```diff
@@ -0,0 +1,7 @@
+pushd %~dp0..
+
+mkdir published
```
heyams, line 3: mkdir -p is needed only created when it doesn't exist. same issue on my pr.

trask: `-p` isn't needed here because it's only called once (see https://github-private.visualstudio.com/microsoft/_build/results?buildId=43433&view=artifacts&pathAsName=false&type=publishedArtifacts)

heyams (approved):

PR #2270, opened 2022-05-10
Update to latest otel

trask:
See #2269

PR #2276, opened 2022-05-12
Use upstream micrometer

trask:
currently built on top of #2270 

@heyams see the last commit https://github.com/microsoft/ApplicationInsights-Java/pull/2276/commits/e933f0a3a51421b485a9cbea76ac33d5f4629195

### DIFF
--- a/test/smoke/testApps/Micrometer/build.gradle.kts
+++ b/test/smoke/testApps/Micrometer/build.gradle.kts
@@ -4,5 +4,6 @@ plugins {
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-web:2.1.7.RELEASE")
-  implementation("io.micrometer:micrometer-core:1.4.1")
+  // TODO (trask) do we still need to support earlier micrometer versions?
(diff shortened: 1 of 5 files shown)


test/smoke/testApps/Micrometer/build.gradle.kts
```diff
@@ -4,5 +4,6 @@ plugins {
 
 dependencies {
   implementation("org.springframework.boot:spring-boot-starter-web:2.1.7.RELEASE")
-  implementation("io.micrometer:micrometer-core:1.4.1")
+  // TODO (trask) do we still need to support earlier micrometer versions?
+  implementation("io.micrometer:micrometer-core:1.5.0")
```
heyams, line 8: why not 1.8.5?  seemed outdated.

trask: Ideally we would test on the lowest version we support and on the latest version (this is what upstream otel instrumentation tests do). I'm not sure what's best in general in our repo. For this particular case I think 1.5.0 is good, especially since it's really not that old in Java years at least (2 years).

trask:
Closing for now

PR #2277, opened 2022-05-12
Fix warning message

trask:
Use updated property name


heyams (approved):
how does this fix relate to warning message? it is just using a different key and default is off.

trask:
> how does this fix relate to warning message? it is just using a different key and default is off.

the underlying otel javaagent emits a warning message if you are using the old property

PR #2278, opened 2022-05-13
Fix lazy logging threshold

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiComponentInstaller.java
```diff
@@ -99,7 +99,7 @@ private static AppIdSupplier start() {
       }
     }
     // TODO (trask) should configuration validation be performed earlier?
```
heyams, line 101: i think so. validating the config should be done before anything. addTransformer is not necessary when the config is bad and agent will fail to start.

heyams (approved):

PR #2282, opened 2022-05-13
Drop  _MS.AggregationIntervalMS

heyams:


trask (approved):

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/exporter/AzureMonitorMetricExporterTest.java
```diff
@@ -307,7 +304,6 @@ public void testDoubleHistogram() throws InterruptedException {
     assertThat(metricsData.getMetrics().size()).isEqualTo(1);
     assertThat(metricsData.getMetrics().get(0).getCount()).isEqualTo(1);
     assertThat(metricsData.getMetrics().get(0).getValue()).isEqualTo(25.45);
-    assertThat(metricsData.getProperties().size()).isEqualTo(1);
```
trask, line 310: I think worth asserting isEmpty or size 0

PR #2286, opened 2022-05-16
Update errorprone and use javax.annotation.Nullable consistently

trask:


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/BytecodeUtil.java
```diff
@@ -33,8 +33,10 @@
 import java.util.Collections;
 import java.util.Date;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 // supporting all properties of event, metric, remote dependency and page view telemetry
+@SuppressWarnings("TooManyParameters")
```
heyams, line 39: how many is too many?

heyams (approved):
this is no longer errorprone.. but google's findbugs.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/BytecodeUtil.java
trask: I think the default is 10

PR #2287, opened 2022-05-16
Check if win 10 sdk install needed for etw build

trask:


heyams (approved):
i tested it.. this is has no impact on etw at all. it's safe to delete.

PR #2292, opened 2022-05-18
Clean up logback dependencies

trask:


smoke-tests/apps/TraceLogBack/build.gradle.kts
```diff
@@ -13,5 +13,5 @@ dependencies {
     // and duplicate class files produces lots of warning messages on jetty
     exclude("com.microsoft.azure", "applicationinsights-core")
   }
-  implementation("ch.qos.logback:logback-classic:1.2.8")
+  implementation("ch.qos.logback:logback-classic")
```
heyams, line 16: without specifying version number, it will pick up whatever the latest version?

heyams (approved):

smoke-tests/apps/TraceLogBack/build.gradle.kts
trask: it will use the version specified in `dependencyManagement/build.gradle.kts`

PR #2296, opened 2022-05-19
Separate agent-tooling from azure-monitor-exporter

heyams:

PR #2297, opened 2022-05-19
Combine the two span exporters

trask:
~@heyams there may be small conflicts with #2296, I will merge this after that PR has been merged~


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
```diff
@@ -208,7 +211,40 @@ private static SpanExporter createSpanExporter(
       Configuration configuration,
       boolean captureHttpServer4xxAsError) {
 
-    SpanExporter spanExporter = new Exporter(telemetryClient, captureHttpServer4xxAsError);
+    AzureMonitorTraceExporter azureMonitorTraceExporter =
+        new AzureMonitorTraceExporter(
+            captureHttpServer4xxAsError,
+            telemetryClient::populateDefaults,
+            (event, instrumentationName) -> {
+              boolean lettuce51 = instrumentationName.equals("io.opentelemetry.lettuce-5.1");
```
heyams, line 219: filtering is also better to be done inside the exporter, whichever class that needs to create such a new exporter, will need to pass in the same filter function?

heyams, line 219: filtering is also better to be done inside the exporter, whichever class that needs to create such a new exporter, will need to pass in the same filter function?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
```diff
@@ -129,72 +152,94 @@ public final class AzureMonitorTraceExporter implements SpanExporter {
...
-              new LocalStorageTelemetryPipelineListener(
-                  TempDirs.getSubDir(tempDir, "telemetry"),
-                  pipeline,
-                  LocalStorageStats.noop(),
-                  false));
-    } else {
-      telemetryItemExporter = new TelemetryItemExporter(pipeline, TelemetryPipelineListener.noop());
-    }
-    this.instrumentationKey = instrumentationKey;
+  public AzureMonitorTraceExporter(
```
heyams, line 168: AzureMonitorSpanExporter? logData will go to trace table and this confused me a bit? SpanData will take care of other entities except metrics and logs.

heyams, line 168: AzureMonitorSpanExporter? logData will go to trace table and this confused me a bit? SpanData will take care of other entities except metrics and logs.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
```diff
@@ -208,7 +211,40 @@ private static SpanExporter createSpanExporter(
       Configuration configuration,
       boolean captureHttpServer4xxAsError) {
 
-    SpanExporter spanExporter = new Exporter(telemetryClient, captureHttpServer4xxAsError);
+    AzureMonitorTraceExporter azureMonitorTraceExporter =
+        new AzureMonitorTraceExporter(
+            captureHttpServer4xxAsError,
+            telemetryClient::populateDefaults,
```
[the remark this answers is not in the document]
trask: `TelemetryClient` is pretty javaagent-specific, so trying not to move it over to the azure-monitor-exporter

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
trask: we only have two places where AzureMonitorTraceExporter is created (javaagent and standalone), and they have different behavior as far as event suppression

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorTraceExporter.java
trask: I think this name was decided across languages because it's "trace" signal in OTel terminology. I'm ok to rename, but we should coordinate with other languages.

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/OpenTelemetryConfigurer.java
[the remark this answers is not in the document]
trask: we can't pass TelemetryClient directly since that's in the javaagent code. will definitely revisit this large constructor parameter list once I change javaagent to use the live metrics span processor and get rid of the extra batching pipeline in the javaagent

PR #2299, opened 2022-05-23
Do not send telemetry data if daily quota exceeded

jeanbisutti:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/pipeline/TelemetryPipeline.java
```diff
@@ -132,6 +141,12 @@ private void onResponseBody(
 
     int responseCode = response.getStatusCode();
 
+    int newDailyQuotaExceeded = 402;
+    if (responseCode == newDailyQuotaExceeded) {
+      logger.warn("Stop sending telemetry data because new daily quota exceeded.");
+      this.stopSending = true;
+    }
```
trask, line 148: we'll want to start sending again when the quota gets next day's allotment, can we check the `retry-after` header from the 402 response? https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/ApplicationInsights/sdk_behavior_breeze.md#402

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/pipeline/TelemetryPipeline.java
```diff
@@ -154,4 +165,24 @@ private void onResponseBody(
...
+      }
+    }
+  }
+  private boolean isNewDailyQuotaExceeded(HttpResponse response) {
+    int responseCode = response.getStatusCode();
+    return responseCode == 402;
+  }
+
+  private boolean isRetriedAfter(HttpResponse response) {
+    return response.getHeaderValue("retry-after") != null;
```
heyams, line 186: retry-after is only return for error code 429. It's confirmed that Breeze doesn't return such a response header for 402. I think it's ok to skip this.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/pipeline/TelemetryPipeline.java
```diff
@@ -154,4 +165,24 @@ private void onResponseBody(
       result.fail();
     }
   }
+
+  private void manageDailyQuota(HttpResponse response) {
+    if (isNewDailyQuotaExceeded(response)) {
```
heyams, line 170:
```suggestion
    if (response.getStatusCode() == 402) {
```

jeanbisutti: The method `isNewDailyQuotaExceeded` may give meaning to `response.getStatusCode() == 402`. When we read the code with the method, we quickly see that the condition is for the business use case "new daily quota exceeded". The code may be more explicit?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/pipeline/TelemetryPipeline.java
trask: @heyams does Breeze team have an guidance for 402s in terms of how long we should wait until we try again?

trask: and we should probably update https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/ApplicationInsights/sdk_behavior_breeze.md#402

heyams: yes, i will create a PR there to update doc.  402 is not retriable. Breeze wants us to drop the data and no retry.

jeanbisutti: > retry-after is only return for error code 429. It's confirmed that Breeze doesn't return such a response header for 402. I think it's ok to skip this.

Done with last commit.

When I read the code of `LocalFileSenderTelemetryPipelineListener`, with a 402 status code, the files (data) will be dropped, right?

trask: > yes, i will create a PR there to update doc. 402 is not retriable. Breeze wants us to drop the data and no retry.

sorry, what I meant wasn't about retrying the same request, but for future requests, if we know that the daily limit has been exceeded, we could potentially proactively drop all future requests for some period of time, not even sending them to Breeze once

trask: it's too bad if Breeze doesn't return `retry-after` header for 402s, since they know when the user quota will get reset

trask: oh, I see the confusion, `retry-after` means retry the same request after X time period, while I'm looking for something that tells us we can proactively drop telemetry for X period of time, sounds like that doesn't exist

heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/63

heyams: they might support it in the future.. the new ingestion service.. we need to make a feature ask though.

PR #2298, opened 2022-05-21
Live metrics span processor

trask:

PR #2300, opened 2022-05-23
Move metric exporter to standalone exporter module

heyams:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorMetricExporter.java
```diff
@@ -44,23 +53,53 @@
...
-
-  public AzureMonitorMetricExporter(TelemetryClient telemetryClient) {
-    this.telemetryClient = telemetryClient;
+  private final Consumer<AbstractTelemetryBuilder> defaultPopulator;
+
+  public AzureMonitorMetricExporter(
+      HttpPipeline httpPipeline,
+      URL endpoint,
+      String instrumentationKey,
+      Consumer<AbstractTelemetryBuilder> defaultPopulator) {
```
trask, line 80: I think it may be better to inject `TelemetryItemExporter` here, so that we can share that, instead of creating separate one here

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -243,6 +248,38 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+        true,
+        builder -> {
+          builder.setInstrumentationKey(instrumentationKey);
+          builder.addTag(
+              ContextTagKeys.AI_INTERNAL_SDK_VERSION.toString(), VersionGenerator.getSdkVersion());
+        },
+        (event, instrumentationName) -> false,
+        telemetryItemExporter::send,
+        telemetryItemExporter::flush,
+        telemetryItemExporter::shutdown,
```
trask, line 262: I think for the metric exporter, we have the option of not going through the lower-level BatchItemProcessor (since metrics are already batched), and binding AzureMonitorMetricExporter directly to TelemetryItemExporter (instead of having to pass in method refs here)

heyams: this is under buildTraceExporter().

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -243,6 +248,38 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+   *
+   * @return An instance of {@link AzureMonitorMetricExporter}.
+   * @throws NullPointerException if the connection string is not set on this builder or if the
+   *     environment variable "APPLICATIONINSIGHTS_CONNECTION_STRING" is not set.
+   */
+  public AzureMonitorMetricExporter buildMetricExporter(
+      Consumer<AbstractTelemetryBuilder> defaultPopulator) {
+    initExpoterBuilder();
+    return new AzureMonitorMetricExporter(
+        new MetricDataMapper(instrumentationKey, defaultPopulator), telemetryItemExporter);
```
heyams, line 279: i passed in telemetryItemExporter directly to AzureMonitorMetricExporter.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorMetricExporter.java
```diff
@@ -0,0 +1,91 @@
...
+      LOGGER.verbose("exporting metric: {}", metricData);
+      try {
+        mapper.map(metricData, telemetryItems::addAll);
+        exportingMetricLogger.recordSuccess();
+      } catch (Throwable t) {
+        exportingMetricLogger.recordFailure(t.getMessage(), t);
+      }
+    }
+
+    return telemetryItemExporter.send(telemetryItems);
```
heyams, line 78: I need to handle the same way you did on #2298 for trace exporter::export?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
trask: 🤦‍♂️

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -243,6 +248,38 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+   *     environment variable "APPLICATIONINSIGHTS_CONNECTION_STRING" is not set.
+   */
+  public AzureMonitorMetricExporter buildMetricExporter(
+      Consumer<AbstractTelemetryBuilder> telemetryInitializer) {
+    initExpoterBuilder();
+    return new AzureMonitorMetricExporter(
+        new MetricDataMapper(instrumentationKey, telemetryInitializer), telemetryItemExporter);
+  }
+
+  private void initExpoterBuilder() {
```
trask, line 282:
```suggestion
  private void initExporterBuilder() {
```

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -243,6 +248,38 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+   * Creates an {@link AzureMonitorMetricExporter} based on the options set in the builder. This
+   * exporter is an implementation of OpenTelemetry {@link MetricExporter}.
+   *
+   * @return An instance of {@link AzureMonitorMetricExporter}.
+   * @throws NullPointerException if the connection string is not set on this builder or if the
+   *     environment variable "APPLICATIONINSIGHTS_CONNECTION_STRING" is not set.
+   */
+  public AzureMonitorMetricExporter buildMetricExporter(
+      Consumer<AbstractTelemetryBuilder> telemetryInitializer) {
+    initExpoterBuilder();
```
trask, line 277: is it a problem if user calls both `buildTraceExporter` and `buildMetricExporter` one after the other? how about if they call `buildTraceExporter` and then update some things (e.g. connectionString) and then they call `buildMetricExporter`?

it's probably worth adding tests for these scenarios as I think we want to support them

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorMetricExporter.java
trask: ya. or alternatively we could not send any telemetry if mapper fails for any of them and returning CompletableResultCode.ofFailure() right away in that case, then we don't need that extra handling that I added in the other PR

trask: can you update the AzureMonitorTraceExporter also to match?

heyams: updated.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -244,6 +250,38 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+  /**
+   * Creates an {@link AzureMonitorMetricExporter} based on the options set in the builder. This
+   * exporter is an implementation of OpenTelemetry {@link MetricExporter}.
+   *
+   * @return An instance of {@link AzureMonitorMetricExporter}.
+   * @throws NullPointerException if the connection string is not set on this builder or if the
+   *     environment variable "APPLICATIONINSIGHTS_CONNECTION_STRING" is not set.
+   */
+  public AzureMonitorMetricExporter buildMetricExporter(
+      Consumer<AbstractTelemetryBuilder> telemetryInitializer) {
```
trask, line 278: let's not expose `telemetryInitializer` in the public API, we probably want to steer standalone users to a more "OpenTelemetry"-way like using a SpanProcessor for this kind of functionality

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -287,20 +324,6 @@ public AzureMonitorTraceExporter buildTraceExporter() {
     } else {
       telemetryItemExporter = new TelemetryItemExporter(pipeline, TelemetryPipelineListener.noop());
     }
```
trask, line 326: I think we want to create new `TelemetryItemExporter` on each call to `build*Exporter` so that we always get the latest config values

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
heyams: updated.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
heyams: it's creating new each time `initExporterBuilder()` is called.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
```diff
@@ -0,0 +1,101 @@
...
+
+public class AgentMetricExporter implements MetricExporter {
+
+  private static final Logger logger = LoggerFactory.getLogger(AgentMetricExporter.class);
+
+  private static final OperationLogger exportingMetricLogger =
+      new OperationLogger(AgentMetricExporter.class, "Exporting metric");
+
+  private final MetricDataMapper mapper;
+  private final Consumer<List<TelemetryItem>> telemetryItemsConsumer;
```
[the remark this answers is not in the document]
heyams: ok, that will be consistent for all entities.

agent/azure-monitor-exporter/src/test/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilderTest.java
```diff
@@ -21,14 +21,67 @@
...
 import java.util.stream.Stream;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 /** Unit tests for {@link AzureMonitorExporterBuilder}. */
-public class AzureMonitorExporterBuilderTest {
+public class AzureMonitorExporterBuilderTest extends MonitorExporterClientTestBase {
```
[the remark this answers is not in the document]
heyams: yeah, only after i added two new tests, it started complaining the mock response. will separate them into diff test classes.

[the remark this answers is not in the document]
heyams: after the breakup, it sill requires a mock response.. not sure why it worked before though. locally failed too without it.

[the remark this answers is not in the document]
trask: remove the extends `MonitorExporterClientTestBase`, that's where it's picking up the mock behavior
```suggestion
public class AzureMonitorExporterBuilderTest {
```

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -71,62 +64,53 @@ public AzureMonitorMetricExporter(TelemetryClient telemetryClient) {
...
-      MetricTelemetryBuilder builder = telemetryClient.newMetricTelemetryBuilder();
+      MetricTelemetryBuilder builder = MetricTelemetryBuilder.create();
+      if (telemetryInitializer != null) {
+        telemetryInitializer.accept(builder);
+      }
+      builder.setInstrumentationKey(instrumentationKey);
+      builder.addTag(
+          ContextTagKeys.AI_INTERNAL_SDK_VERSION.toString(), VersionGenerator.getSdkVersion());
       builder.setTime(FormattedTime.offSetDateTimeFromEpochNanos(pointData.getEpochNanos()));
+      builder.setSampleRate(100); // TODO make this configurable?
```
trask, line 105: I think maybe we should not set this, can you check what we were doing previously?

```suggestion
```

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -71,62 +64,53 @@ public AzureMonitorMetricExporter(TelemetryClient telemetryClient) {
...
-    return CompletableResultCode.ofSuccess();
   }
 
   private List<TelemetryItem> convertOtelMetricToAzureMonitorMetric(MetricData metricData) {
     List<TelemetryItem> telemetryItems = new ArrayList<>();
-
     for (PointData pointData : metricData.getData().getPoints()) {
-      MetricTelemetryBuilder builder = telemetryClient.newMetricTelemetryBuilder();
+      MetricTelemetryBuilder builder = MetricTelemetryBuilder.create();
+      if (telemetryInitializer != null) {
```
trask, line 98: let's make `telemetryInitializer` non-null and remove this check. can always pass in a no-op consumer instead of `null`

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
```diff
@@ -244,6 +248,37 @@ public AzureMonitorExporterBuilder credential(TokenCredential credential) {
...
+   * @throws NullPointerException if the connection string is not set on this builder or if the
+   *     environment variable "APPLICATIONINSIGHTS_CONNECTION_STRING" is not set.
+   */
+  public AzureMonitorMetricExporter buildMetricExporter() {
+    initExporterBuilder();
+    return new AzureMonitorMetricExporter(
+        new MetricDataMapper(instrumentationKey, t -> {}), telemetryItemExporter);
+  }
+
+  private void initExporterBuilder() {
```
trask, line 281: last comment, I think this might be clearer to return `TelemetryItemExporter` and not store it into "temporary" field

I'm also ok waiting and seeing how we feel about it over time, it's not public API so we can always change it

```suggestion
  private TelemetryItemExporter createTelemetryItemExporter() {
```

trask (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorExporterBuilder.java
heyams: Updated.

PR #2302, opened 2022-05-27
Remove github action cache

heyams:
got the idea from https://github.com/gradle/gradle-build-action/issues/115

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -72,7 +72,7 @@ jobs:
       - name: Test
         uses: gradle/gradle-build-action@v2
         env:
-          GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: verison1
+          GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: ${{github.workflow}}#${{github.run_number}}:${{github.run_attempt}}-
         with:
           arguments: ${{ matrix.module }}:smokeTest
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -71,6 +71,8 @@ jobs:
           java-version: 11
       - name: Test
         uses: gradle/gradle-build-action@v2
+        env:
+          GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: ${{github.workflow}}#${{github.run_number}}:${{github.run_attempt}}-
```
trask, line 75: I don't think we will get any benefit from cache if the prefix changes on every run

trask (approved):

.github/workflows/pr.yaml
heyams: let's see how long it takes for each run. If it's no different from cache, i think it's ok not to cache? it's all about timing. 
if it takes long time, will revert back to a static string.

heyams: we can keep updating the prefix whenever running into the cache issue.

heyams: will merge it in for now.. azure pipeline build failed due to sporadic link failure. it's taking 30 mins or so, but regular build also takes around 30 mins.. will check against other runs.

PR #2304, opened 2022-05-27
AzureMonitorLogExporter

heyams:
this will fix the commit history of #2303 
rebase led to lots of conflicts.


.github/workflows/pr.yaml
```diff
@@ -71,9 +71,8 @@ jobs:
           java-version: 11
       - name: Test
         uses: gradle/gradle-build-action@v2
-        env:
-          GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX: ${{github.workflow}}#${{github.run_number}}:${{github.run_attempt}}-
         with:
+          cache-disabled: true
```
trask, line 75: the last CI failure was a sporadic `MicrometerTest` failure: https://github.com/microsoft/ApplicationInsights-Java/runs/6619718932?check_suite_focus=true

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/AzureMonitorMetricExporter.java
```diff
@@ -78,14 +85,16 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
     return telemetryItemExporter.send(telemetryItems);
   }
 
+  /** {@inheritDoc} */
   @Override
   public CompletableResultCode flush() {
-    return CompletableResultCode.ofSuccess();
+    return telemetryItemExporter.flush();
```
trask, line 91: 👍

trask (approved):

.github/workflows/pr.yaml
heyams: GRADLE_BUILD_ACTION_CACHE_KEY_PREFIX doesn't seem to work.. it still restores from cache. `cache-disabled: true` seems to work. i no longer find restoring from cache.

heyams: this wasn't so bad, took 26 mins.

image

we can re-evaluate it later if we need to remove this.

PR #2310, opened 2022-06-01
Fix mkdirs

trask:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/TempDirs.java
```diff
@@ -40,7 +40,7 @@ public static File getApplicationInsightsTempDir(Logger logger, String message)
     tempDir = maybeAddUserSubDir(tempDir);
     tempDir = new File(tempDir, "applicationinsights");
 
-    if ((!tempDir.exists() && !tempDir.mkdir())) {
+    if (!tempDir.exists() && !tempDir.mkdirs()) {
```
heyams, line 43: in v3.2.11, we used mkdirs everywhere. IcM 310151045 was throwing exception and customer can't seem to create log temp folder. Did you notice it because of that icm?

heyams (approved):
👍

PR #2319, opened 2022-06-03
Apply metric filtering to AgentMetricExporter

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
```diff
@@ -66,8 +74,15 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
...
-
     for (MetricData metricData : metrics) {
+      if (!Constant.NON_FILTERABLE_METRIC_NAMES.contains(metricData.getName())) {
+        for (MetricFilter metricFilter : metricFilters) {
+          if (!metricFilter.matches(metricData.getName())) {
+            // user configuration filtered out this metric name
+            return CompletableResultCode.ofSuccess();
+          }
+        }
+      }
```
trask, line 85: probably can extract this into common method in `Constant` class, e.g. `boolean shouldSkip(List<MetricFilter>)`, and share with TelemetryClient

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/utils/Constant.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static java.util.Arrays.asList;
+
+import com.microsoft.applicationinsights.agent.internal.perfcounter.MetricNames;
+import java.util.HashSet;
+import java.util.Set;
+
+public final class Constant {
+
+  public static final Set<String> NON_FILTERABLE_METRIC_NAMES =
```
trask, line 32: this may make sense as a constant inside of `MetricFilter`, then doesn't need separate class (it's nice to try to avoid "Constant" classes and put constants closer to the code where they are used)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
heyams: i moved it to MetricFilter as well.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
```diff
@@ -66,8 +73,10 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
       logger.debug("exporter is not active");
       return CompletableResultCode.ofSuccess();
     }
-
     for (MetricData metricData : metrics) {
+      if (MetricFilter.shouldSkip(metricData.getName(), metricFilters)) {
+        return CompletableResultCode.ofSuccess();
```
trask, line 78:
```suggestion
        continue;
```

heyams: good catch

trask (approved):

PR #2322, opened 2022-06-07
Track statusCode and exceptionType via Statsbeat

heyams:
spec https://github.com/microsoft/Telemetry-Collection-Spec/pull/61

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
@@ -116,7 +116,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   long getRequestSuccessCount(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.requestSuccessCount.get();
     }
   }
@@ -124,7 +124,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   long getRequestFailureCount(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.requestFailureCount.get();
     }
   }
@@ -132,7 +132,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   double getRequestDurationAvg(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.getRequestDurationAvg();
     }
   }
@@ -140,7 +140,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   long getRetryCount(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.retryCount.get();
     }
   }
@@ -148,7 +148,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   long getThrottlingCount(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.throttlingCount.get();
     }
   }
@@ -156,7 +156,7 @@ public class NetworkStatsbeat extends BaseStatsbeat {
   // only used by tests
   long getExceptionCount(String ikey) {
     synchronized (lock) {
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(new IntervalMetricsKey(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.exceptionCount.get();
     }
   }
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
@@ -75,32 +75,33 @@ public class NetworkStatsbeat extends BaseStatsbeat {
     doWithIntervalMetrics(
         ikey,
         host,
+        null,
         intervalMetrics -> {
           intervalMetrics.requestSuccessCount.incrementAndGet();
           intervalMetrics.totalRequestDuration.getAndAdd(duration);
         });
   }
 
-  public void incrementRequestFailureCount(String ikey, String host) {
+  public void incrementRequestFailureCount(String ikey, String host, Integer statusCode) {
     doWithIntervalMetrics(
-        ikey, host, intervalMetrics -> intervalMetrics.requestFailureCount.incrementAndGet());
+        ikey, host, statusCode, intervalMetrics -> intervalMetrics.requestFailureCount.incrementAndGet());
   }
 
   // TODO (heya) this is never called
-  public void incrementRetryCount(String ikey, String host) {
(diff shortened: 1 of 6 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeatHttpPipelinePolicy.java
```diff
@@ -57,17 +57,21 @@ public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineN
...
+                // 400 and 404 will be tracked as failure count
+                networkStatsbeat.incrementRequestFailureCount(
+                    instrumentationKey, host, Integer.valueOf(statusCode));
               }
             })
         .doOnError(
             throwable -> {
-              networkStatsbeat.incrementExceptionCount(instrumentationKey, host);
+              networkStatsbeat.incrementExceptionCount(
+                  instrumentationKey, host, throwable.getClass().toString());
```
trask, line 74: `getName` is more common

```suggestion
                  instrumentationKey, host, throwable.getClass().getName());
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeatHttpPipelinePolicy.java
```diff
@@ -57,17 +57,21 @@ public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineN
...
               } else if (StatusCodes.isRetryable(statusCode)) {
-                networkStatsbeat.incrementRetryCount(instrumentationKey, host);
+                networkStatsbeat.incrementRetryCount(
+                    instrumentationKey, host, Integer.valueOf(statusCode));
               } else {
-                // note: 401 and 403 are currently tracked as failures
-                networkStatsbeat.incrementRequestFailureCount(instrumentationKey, host);
+                // 400 and 404 will be tracked as failure count
+                networkStatsbeat.incrementRequestFailureCount(
+                    instrumentationKey, host, Integer.valueOf(statusCode));
```
trask, line 68: even if the method takes `Integer`, it should auto-box this code:
```suggestion
                    instrumentationKey, host, statusCode);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -151,11 +162,13 @@ long getExceptionCount(String ikey) {
...
   }
 
-  private void doWithIntervalMetrics(String ikey, String host, Consumer<IntervalMetrics> update) {
+  private <T> void doWithIntervalMetrics(
+      String ikey, String host, @Nullable T cause, Consumer<IntervalMetrics> update) {
     synchronized (lock) {
       IntervalMetrics intervalMetrics =
           instrumentationKeyCounterMap.computeIfAbsent(ikey, k -> new IntervalMetrics());
       intervalMetrics.host = host;
+      intervalMetrics.cause = cause;
```
trask, line 171: I think this will overwrite prior cause value?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -178,51 +191,60 @@ private void sendIntervalMetric(
...
+      if (cause instanceof Integer) {
+        telemetryBuilder.addProperty("statusCode", ((Integer) cause).toString());
+      } else if (cause instanceof String) {
+        telemetryBuilder.addProperty("exceptionType", (String) cause);
+      }
+    }
   }
 
-  private static class IntervalMetrics {
+  private static class IntervalMetrics<T> {
```
trask, line 247: I'm not sure about using generic for this, but thing may change after addressing the possible issue with `cause` getting overwritten, so let's revisit after that

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -75,32 +76,42 @@ public void incrementRequestSuccessCount(long duration, String ikey, String host
...
         host,
+        null,
         intervalMetrics -> {
           intervalMetrics.requestSuccessCount.incrementAndGet();
           intervalMetrics.totalRequestDuration.getAndAdd(duration);
         });
   }
 
-  public void incrementRequestFailureCount(String ikey, String host) {
+  public void incrementRequestFailureCount(String ikey, String host, Integer statusCode) {
```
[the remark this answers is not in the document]
heyams: i assume no. any status code is returned from breeze will be nonnull. if there is no status code, it will server-side exception?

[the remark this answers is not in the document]
heyams: let me quickly confirm with Madeleine.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -244,7 +242,18 @@ private static <T> void addCommonProperties(
     }
   }
 
-  private static class IntervalMetrics<T> {
+  private static class IntervalMetricsKey {
```
trask, line 245: you'll need to override `hashCode()` and `equals()` to use this as a map key, the easiest way to do this is using auto-value, which we already have in gradle file, but doesn't look like we're using anywhere

```
  compileOnly("com.google.auto.value:auto-value-annotations")
  annotationProcessor("com.google.auto.value:auto-value")
```

```suggestion
    @AutoValue
    private static abstract class IntervalMetricsKey {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -59,116 +62,133 @@ public NetworkStatsbeat(CustomDimensions customDimensions) {
...
         host,
+        null,
         intervalMetrics -> {
           intervalMetrics.requestSuccessCount.incrementAndGet();
           intervalMetrics.totalRequestDuration.getAndAdd(duration);
         });
   }
 
-  public void incrementRequestFailureCount(String ikey, String host) {
+  public void incrementRequestFailureCount(String ikey, String host, Integer statusCode) {
```
trask, line 88: I think it makes the method signature clearer since non-null, and let auto-boxing (or explicit boxing) happen later on
```suggestion
  public void incrementRequestFailureCount(String ikey, String host, int statusCode) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -59,116 +62,133 @@ public NetworkStatsbeat(CustomDimensions customDimensions) {
...
-      IntervalMetrics intervalMetrics = instrumentationKeyCounterMap.get(ikey);
+      IntervalMetrics intervalMetrics =
+          instrumentationKeyCounterMap.get(IntervalMetricsKey.create(ikey, null));
       return intervalMetrics == null ? 0L : intervalMetrics.getRequestDurationAvg();
     }
   }
 
   // only used by tests
-  long getRetryCount(String ikey) {
+  long getRetryCount(String ikey, int cause) {
```
trask, line 146: `cause` -> `statusCode` I think makes it clearer

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -59,116 +62,133 @@ public NetworkStatsbeat(CustomDimensions customDimensions) {
...
 
-  private void doWithIntervalMetrics(String ikey, String host, Consumer<IntervalMetrics> update) {
+  private void doWithIntervalMetrics(
+      String ikey, String host, @Nullable Object cause, Consumer<IntervalMetrics> update) {
     synchronized (lock) {
       IntervalMetrics intervalMetrics =
-          instrumentationKeyCounterMap.computeIfAbsent(ikey, k -> new IntervalMetrics());
+          instrumentationKeyCounterMap.computeIfAbsent(
+              IntervalMetricsKey.create(ikey, cause), k -> new IntervalMetrics());
       intervalMetrics.host = host;
```
trask, line 178: what do you think about `host` into the key also? it seems potentially more correct, and potentially more symmetric with the other properties

heyams: yes, i think it's better that way.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/NetworkStatsbeat.java
```diff
@@ -178,48 +197,74 @@ private void sendIntervalMetric(
...
+    telemetryBuilder.addProperty("host", key.getHost());
+    Object cause = key.getCause();
+    if (cause != null) {
+      // track 'statusCode' for Failure/Retry/Throttle count and track 'exceptionType' for Exception
+      // Count
+      if (cause instanceof Integer) {
+        telemetryBuilder.addProperty("statusCode", ((Integer) cause).toString());
+      } else if (cause instanceof String) {
+        telemetryBuilder.addProperty("exceptionType", (String) cause);
+      }
```
trask, line 250: I think I prefer the more explicit `causeFieldName` + `causeValue`, instead of the single `cause` and relying here on knowing that `Integer` == "statusCode" and `String` == "exceptionType"

trask (approved):
👍 one additional thought below

PR #2325, opened 2022-06-13
Runtime attachment feature

jeanbisutti:
Runtime attachment feature for _Application Insights._

The feature delegates attachment to [_Otel runtime attachment_ project](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/runtime-attach) and allows to manage things specific to _Application Insights_ (_applicationinsights.json_ and _version tracking_).

[A PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/354) was also  created to improve the _Otel runtime attachment_ (more checks and logs).


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -259,7 +259,16 @@ private static String getCodelessSdkNamePrefix() {
           .warn("could not detect os: {}", System.getProperty("os.name"));
       sdkNamePrefix.append("u");
     }
-    sdkNamePrefix.append("r_"); // "r" is for "recommended"
+    if (isRuntimeAttached()) {
+      sdkNamePrefix.append("ra_");
+    } else {
```
heyams, line 264: can you do runtime attach check at the beginning of the method and then return early?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,91 @@
...
+    Optional<File> optionalConfigFile = searchConfigFile();
+    if (optionalConfigFile.isPresent()) {
+      File configFile = optionalConfigFile.get();
+      System.setProperty("applicationinsights.configuration.file", configFile.getAbsolutePath());
+    }
+
+    RuntimeAttach.attachJavaagentToCurrentJVM();
+
+    if (!agentIsAttached()) {
+      System.setProperty(RUNTIME_ATTACHED_PROPERTY, "false");
```
heyams, line 57: can you explain why line 45 and then setting it to false here?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  private static boolean agentIsAttached() {
```
heyams, line 61: I saw you have another pr on opentelemetry. should you wait till that PR is merged and use the upstream to check if agent is attached?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
jeanbisutti: @heyams Could you please explain why? If we return early at the beginning of the method then the os information won't be available for tracking.

heyams: something like this: code between 
"// start runtime attach prefix" and 
"// end runtime attach prefix"

your change will end up with "ara_" or "fra_" depends on which rp integration it was. 
```
@Nullable
  private static String getCodelessSdkNamePrefix() {
    if (!DiagnosticsHelper.isRpIntegration()) {
      return null;
    }
    StringBuilder sdkNamePrefix = new StringBuilder(4);
    // start runtime attach prefix
    if (isRuntimeAttached()) {
      sdkNamePrefix.append("ra_");
      return;
    } // end runtime attach prefix
    sdkNamePrefix.append(DiagnosticsHelper.rpIntegrationChar());
    if (SystemInformation.isWindows()) {
      sdkNamePrefix.append("w");
    } else if (SystemInformation.isLinux()) {
      sdkNamePrefix.append("l");
    } else {
      LoggerFactory.getLogger("com.microsoft.applicationinsights.agent")
          .warn("could not detect os: {}", System.getProperty("os.name"));
      sdkNamePrefix.append("u");
    }
    sdkNamePrefix.append("r_"); // "r" is for "recommended"
    return sdkNamePrefix.toString();
  }
```

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: At this step, if the runtime attachment is not done, the `agent.runtime.attached` property always has a `true` value. So, for consistency it is set to `false`. However, since the agent is not attached, the property value won't be used.... So, I will remove this piece of code. The code will be easier to read without apparently any impact.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: No, we have not to wait until the OTel PR is merged. OTel ```RuntimeAttach``` class exposes only one public method, ```attachJavaagentToCurrentJVM()```.

agent/runtime-attach/README.md
```diff
@@ -0,0 +1,23 @@
+# Java runtime attachment for Application Insights
+
+If you can't update the JVM arguments to attach the Application Insights agent for Java (_
+-javaagent:path/to/applicationinsights-agent-3.2.11.jar_), this project allows you to do attachment
```
heyams, line 4: might want to use a placeholder for version. it will require an update each time a newer version of java jar is released.
something like {LATEST_JAVA_AGENT_VERSION}.jar will work.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
jeanbisutti: Why? The problem is that the prefix has to contain only 4 characters? Another reason?

agent/runtime-attach/README.md
```diff
@@ -0,0 +1,23 @@
+# Java runtime attachment for Application Insights
```
heyams, line 1: or alternatively, we can expose this runtime attach feature on azure doc and we don't need a separate readme for it?

image

under `Advanced configurations`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
heyams: rp integration uses 4 chars. in your case only 3.

agent/runtime-attach/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+  id("ai.java-conventions")
+}
+
+dependencies {
+  implementation("io.opentelemetry.contrib:opentelemetry-runtime-attach:1.14.0-alpha")
+  implementation("com.microsoft.azure:applicationinsights-agent:" + "3.2.11")
```
trask, line 7: similar to https://github.com/microsoft/ApplicationInsights-Java/blob/7eb023e32f5ec4987c09f618f326ff92baf41815/smoke-tests/apps/SpringBootAttachInMain/build.gradle.kts#L6

```suggestion
  implementation(project(":agent:agent", configuration = "shadow"))
```

agent/runtime-attach/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
...
+  testImplementation("io.opentelemetry:opentelemetry-extension-annotations:1.14.0")
```
trask, line 8: to avoid hard-coding this version, check out https://github.com/microsoft/ApplicationInsights-Java/blob/7eb023e32f5ec4987c09f618f326ff92baf41815/agent/instrumentation/applicationinsights-web-2.3/build.gradle.kts#L20-L25

agent/runtime-attach/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+  id("ai.java-conventions")
+}
+
+dependencies {
+  implementation("io.opentelemetry.contrib:opentelemetry-runtime-attach:1.14.0-alpha")
```
trask, line 6: let's add `otelContribAlphaVersion` similar to `otelInstrumentationAlphaVersion`:

(we probably need to keep separate version constants for now since they are released from separate repos not always at the same time)

https://github.com/microsoft/ApplicationInsights-Java/blob/7eb023e32f5ec4987c09f618f326ff92baf41815/dependencyManagement/build.gradle.kts#L17

agent/runtime-attach/src/test/resources/applicationinsights.json
```diff
@@ -0,0 +1,7 @@
+{
+  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-0FEEDDADBEEF;IngestionEndpoint=http://host.docker.internal:6060/",
```
trask, line 2: just to be clear that nothing is listening there
```suggestion
  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-0FEEDDADBEEF;IngestionEndpoint=http://dummy",
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -259,7 +259,16 @@ private static String getCodelessSdkNamePrefix() {
...
+      sdkNamePrefix.append("r_"); // "r" is for "recommended"
+    }
     return sdkNamePrefix.toString();
   }
+
+  private static boolean isRuntimeAttached() {
+    String runtimeAttachedProp = System.getProperty("agent.runtime.attached");
+    return "true".equals(runtimeAttachedProp);
```
trask, line 272:
```suggestion
    return Boolean.getBoolean("agent.runtime.attached");
```

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,93 @@
...
+      LOGGER.warning("Application Insights is already attached. It is not attached a second time.");
+      return;
+    }
+
+    System.setProperty(RUNTIME_ATTACHED_PROPERTY, "true");
+
+    Optional<File> optionalConfigFile = searchConfigFile();
+    if (optionalConfigFile.isPresent()) {
+      File configFile = optionalConfigFile.get();
+      System.setProperty("applicationinsights.configuration.file", configFile.getAbsolutePath());
```
trask, line 57: do you know if we can rely on the classpath resource always having a real "file path"?

since we are just using the string value, do we need to convert `URL` -> `URI` -> `File` -> `String`? or can we use `URL.getFile()`?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,93 @@
...
+import java.net.URL;
+import java.util.Optional;
+import java.util.logging.Logger;
+
+/** This class allows you to attach the Application Insights agent for Java at runtime. */
+public class ApplicationInsights {
+
+  private static final Logger LOGGER = Logger.getLogger(ApplicationInsights.class.getName());
+
+  public static final String RUNTIME_ATTACHED_PROPERTY = "agent.runtime.attached";
```
trask, line 37: it's helpful to scope any system properties, e.g. `applicationinsights.agent.runtime.attached`

and since this is only for internal (as opposed to customer) usage, maybe `applicationinsights.internal.runtime.attached`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
jeanbisutti: Thank you for the explanations. I understand why now. The code is updated.

agent/runtime-attach/README.md
jeanbisutti: Good point @heyams! Done.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: Good point @trask! Code updated

agent/runtime-attach/build.gradle.kts
jeanbisutti: Code updated Please [recheck](https://github.com/microsoft/ApplicationInsights-Java/blob/runtime-attach/agent/runtime-attach/build.gradle.kts)

agent/runtime-attach/build.gradle.kts
jeanbisutti: Good point. Code updated. Please check [this](https://github.com/microsoft/ApplicationInsights-Java/blob/runtime-attach/dependencyManagement/build.gradle.kts) and [this](https://github.com/microsoft/ApplicationInsights-Java/blob/runtime-attach/agent/runtime-attach/build.gradle.kts).

agent/runtime-attach/README.md
jeanbisutti: @trask What do you think? May I create a PR on Azure repo to report the Readme documentation now or later?

heyams: we can do that later when we're ready to make a new release.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,94 @@
...
+      return false;
+    }
+  }
+
+  private static Optional<File> searchConfigFile() {
+
+    URL jsonUrl = ApplicationInsights.class.getResource("/applicationinsights.json");
+
+    boolean jsonFileNotFound = jsonUrl == null;
+    if (jsonFileNotFound) {
```
heyams, line 78:
```suggestion
    if (jsonUrl == null) {
```

agent/runtime-attach/src/test/resources/applicationinsights.json
```diff
@@ -0,0 +1,7 @@
+{
```
heyams, line 1: without this file, it will return Optional.empty. does the test still work without json config?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
trask: this is a style choice, I would have chosen to inline this too, but I also don't mind being flexible when it comes to style choices that aren't covered by spotless, errorprone, or https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/style-guideline.md

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
trask: > I have noticed [an environment variable related to the configuration content](https://github.com/microsoft/ApplicationInsights-Java/blob/7eb023e32f5ec4987c09f618f326ff92baf41815/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java#L368). @trask I propose to add a system property doing the same thing.

this sounds good, it might be nice if we can hide it, maybe a static String in a class in the bootstrap classloader that we can populate from runtime attach and read from ConfigurationBuilder?

agent/runtime-attach/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+
+val otelContribAlphaVersion: String by project
+val otelVersion: String by project
+
+dependencies {
+  implementation("io.opentelemetry.contrib:opentelemetry-runtime-attach:$otelContribAlphaVersion")
+  implementation(project(":agent:agent", configuration = "shadow"))
+  testImplementation("io.opentelemetry:opentelemetry-extension-annotations:$otelVersion")
+  testImplementation("org.assertj:assertj-core:3.23.1")
+  testImplementation("org.junit.jupiter:junit-jupiter-api:.8.2")
```
trask, line 13: `.8.2` looks wrong? probably this version is covered by dependency management (and probably assertj also?)

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,94 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Optional;
+import java.util.logging.Logger;
+
+/** This class allows you to attach the Application Insights agent for Java at runtime. */
+public class ApplicationInsights {
+
+  private static final Logger LOGGER = Logger.getLogger(ApplicationInsights.class.getName());
```
trask, line 35: let's stay with OTel convention of lowercase `logger`

agent/runtime-attach/src/test/resources/applicationinsights.json
```diff
@@ -0,0 +1,7 @@
...
+  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-0FEEDDADBEEF;IngestionEndpoint=http://dummy",
+  "role": {
+    "name": "testrolename",
+    "instance": "testroleinstance"
+  }
```
trask, line 6: the others aren't needed for this
```suggestion
  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-0FEEDDADBEEF;IngestionEndpoint=http://dummy"
```

trask (approved):
👍

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: My intention with the `jsonFileNotFound` variable was to give meaning to the `jsonUrl == null`  technical check.

agent/runtime-attach/src/test/resources/applicationinsights.json
jeanbisutti: The `RuntimeAttachTest` test will fail if we remove this file because _Application Insights_ needs a connection string provided in the json file. The connection string  may also be supplied with a system property or an environment variable, without a json file,  and the code execution path will then use the `Optional.empty()` line.

jeanbisutti:
`RuntimeAttach` and `AgentFileProvider` classes could be removed once the 2 following OTel PRs will be merged and a new OTel release will be done:
* [OTel PR 1](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/354)
* [OTel PR 2](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/359)

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,94 @@
...
+/** This class allows you to attach the Application Insights agent for Java at runtime. */
+public final class ApplicationInsights {
+
+  private static final Logger logger = Logger.getLogger(ApplicationInsights.class.getName());
+
+  static final String RUNTIME_ATTACHED_ENABLED_PROPERTY =
+      "applicationinsights.internal.runtime.attached.enabled";
+
+  static final String RUNTIME_ATTACHED_JSON_PROPERTY =
+      "applicationinsights.internal.runtime.attached.json";
```
trask, line 43:
```suggestion
      "applicationinsights.internal.runtime.attach.json";
```

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+/** This class allows you to attach the Application Insights agent for Java at runtime. */
+public final class ApplicationInsights {
+
+  private static final Logger logger = Logger.getLogger(ApplicationInsights.class.getName());
+
+  static final String RUNTIME_ATTACHED_ENABLED_PROPERTY =
+      "applicationinsights.internal.runtime.attached.enabled";
```
trask, line 40:
```suggestion
      "applicationinsights.internal.runtime.attached";
```

or maybe

```suggestion
      "applicationinsights.internal.runtime.attach.enabled";
```

trask (approved):
nice, a couple of suggestions on the internal property names if they make sense

PR #2326, opened 2022-06-15
Parse resource for role name and role instance

heyams:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,89 @@
...
+
+  // visible for test
+  @Nullable
+  static Map<ContextTagKeys, String> parseRoleNameAndInstance(@Nullable Resource resource) {
+    if (resource == null) {
+      return null;
+    }
+    String serviceName = resource.getAttribute(ResourceAttributes.SERVICE_NAME);
+    if (serviceName == null || serviceName.isEmpty()) {
+      serviceName = Resource.getDefault().getAttribute(ResourceAttributes.SERVICE_NAME);
```
trask, line 57: what is the default service_name? I'm guessing (but not sure) that we don't want to send any rolename if none is configured (the U/X defaults the app map to the name of the AI Portal Resource in this case)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,89 @@
...
+                || ContextTagKeys.AI_CLOUD_ROLE_INSTANCE.toString().equals(stringKey)) {
+              builder.addTag(stringKey, value);
+            }
+          });
+    }
+  }
+
+  // visible for test
+  @Nullable
+  static Map<ContextTagKeys, String> parseRoleNameAndInstance(@Nullable Resource resource) {
```
trask, line 51: can we call this once at startup and cache the resulting map? Resources aren't allowed to change (currently at least) in OpenTelemetry

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,89 @@
...
+    if (resource == null) {
+      return null;
```
trask, line 53: it's nice to return Collections.emptyMap() to avoid extra null checks

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,89 @@
...
+
+  public static void updateRoleNameAndInstance(
+      AbstractTelemetryBuilder builder, Resource resource) {
+    Map<ContextTagKeys, String> resourceMap = ResourceParser.parseRoleNameAndInstance(resource);
+    if (resourceMap != null && !resourceMap.isEmpty()) {
+      resourceMap.forEach(
+          (key, value) -> {
+            String stringKey = key.toString();
+            if (ContextTagKeys.AI_CLOUD_ROLE.toString().equals(stringKey)
+                || ContextTagKeys.AI_CLOUD_ROLE_INSTANCE.toString().equals(stringKey)) {
```
trask, line 42: do we need this check since these are the only two keys we put into the map?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: no, resource per metric data

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: "unknown_service:java". upstream has a bug there. it should not have ":java". dotnet uses "unknown_service". service_name is mandatory and will never be null nor empty.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: i added a comment and removed these check.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: should I submit a PR to fix [it](https://github.com/open-telemetry/opentelemetry-java/blob/d868eff3e1d63a7d46f5866d676570cc2f2be84a/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#:~:text=private%20static%20final%20Resource%20MANDATORY,SERVICE_NAME%2C%20%22unknown_service%3Ajava%22)?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
trask: ya, but resources are immutable, tied to the `OpenTelemetry` instance, and shared across all telemetry.

it's possible for users to create multiple `OpenTelemetry` instances though, so that would be a reason not to cache.

maybe pass in the telemetry builder, so we don't need to create the extra Map on each call (just as an optimization to reduce memory allocation)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
trask: I'm not sure it's a bug, check out the spec:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,87 @@
...
+      // map contains AI_CLOUD_ROLE and AI_CLOUD_ROLE_INSTANCE
+      resourceMap.forEach(
+          (key, value) -> {
+            builder.addTag(key.toString(), value);
+          });
+    }
+  }
+
+  // visible for test
+  static Map<ContextTagKeys, String> parseRoleNameAndInstance(@Nullable Resource resource) {
```
trask, line 48: I don't think Resource can be null
```suggestion
  static Map<ContextTagKeys, String> parseRoleNameAndInstance(Resource resource) {
```

trask:
How will users of standalone exporters set the service.namespace/service.name/service.id?

One option is to provide an `AzureMonitorResourceProvider` that checks all the things, e.g. `WEBSITE_SITE_NAME`, `WEBSITE_INSTANCE_ID`, `APPLICATIONINSIGHTS_ROLE_NAME`, `APPLICATIONINSIGHTS_ROLE_INSTANCE` and creates the Resource.

This doesn't need to block this PR, but let's discuss and align with other langs on the user experience.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/LogDataMapper.java
```diff
@@ -78,6 +79,7 @@ private TelemetryItem createMessageTelemetryItem(LogData log) {
     setTime(telemetryBuilder, log.getEpochNanos());
     setSampleRate(telemetryBuilder, log);
     setExtraAttributes(telemetryBuilder, attributes);
+    ResourceParser.updateRoleNameAndInstance(telemetryBuilder, log.getResource());
```
trask, line 82: for the Javaagent I think this is always going to be `unknown_service:java` and overwrite the configured role name?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
jeanbisutti, line 48: I would suggest to try not to modify method visibility for testing. The reason is that when we do this then we are testing implementation details. The test is then tied to an implementation detail. If we refactor the code (i.e. changing the implementation without changing the behavior), then the tests might have to be updated. So, this may lead to an extra work to maintain the tests and discourage us to refactor. In this case, the test could use the public `updateRoleNameAndInstance` method.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,87 @@
...
+    Map<ContextTagKeys, String> map = new HashMap<>(2);
+    if (roleName != null) {
+      map.put(ContextTagKeys.AI_CLOUD_ROLE, roleName);
+    }
+    if (roleInstance != null) {
+      map.put(ContextTagKeys.AI_CLOUD_ROLE_INSTANCE, roleInstance);
+    }
+
+    return map;
+  }
```
jeanbisutti, line 84: The `parseRoleNameAndInstance` method has about 35 lines. To have an idea about what the method is doing, we have to browse all lines of code. We would not have this issue if the  `parseRoleNameAndInstance` called private methods (as a table of contents containing chapter names).

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: passing in builder.

heyams:
> 

I created a [work item](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/Copper/CY22Q2/Monthly/Jun?workitem=14734329) to track that. it will be a separate PR.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
heyams: refactored.

heyams:
> How will users of standalone exporters set the service.namespace/service.name/service.id?
> 
> One option is to provide an `AzureMonitorResourceProvider` that checks all the things, e.g. `WEBSITE_SITE_NAME`, `WEBSITE_INSTANCE_ID`, `APPLICATIONINSIGHTS_ROLE_NAME`, `APPLICATIONINSIGHTS_ROLE_INSTANCE` and creates the Resource.
> 
> This doesn't need to block this PR, but let's discuss and align with other langs on the user experience.

I created a [work item](https://msazure.visualstudio.com/One/_sprints/taskboard/TelReachSDKTeam/One/Copper/CY22Q2/Monthly/Jun?workitem=14734329) to track that. it will be a separate PR.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/LogDataMapper.java
trask: did you mean to resolve this comment? I'm guessing this may be why some smoke tests are failing

heyams: osvaldo said don't send ai.role.name when service.name is not set, i.e. don't send `unknown_service`. `:java` is the `process.execution.name`. you're correct, it's not a bug on opentelemetry-java sdk.

heyams: tests are failing because test data has null resource.

heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/78#discussion_r898282377

heyams: got it.. populateDefaults is called before update.. this line will override the customer's configuration.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java
```diff
@@ -202,13 +203,17 @@ public TelemetryItem map(SpanData span, float samplingPercentage) {
...
-    telemetryInitializer.accept(telemetryBuilder);
 
     // set standard properties
     setOperationTags(telemetryBuilder, span);
     setTime(telemetryBuilder, span.getStartEpochNanos());
     setSampleRate(telemetryBuilder, samplingPercentage);
+
+    // update tags
+    ResourceParser.updateRoleNameAndInstance(telemetryBuilder, span.getResource());
+    telemetryInitializer.accept(telemetryBuilder);
```
trask, line 214: I think it's clearer to leave initialization of the telemetry builder at the top, before updating the telemetry builder

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,58 @@
...
+      roleName = "[" + serviceNamespace + "]/" + serviceName;
+    } else if (serviceName != null) {
+      roleName = serviceName;
+    } else if (serviceNamespace != null) {
+      roleName = "[" + serviceNamespace + "]";
+    }
+
+    String roleInstance = resource.getAttribute(ResourceAttributes.SERVICE_INSTANCE_ID);
+    if (roleInstance == null) {
+      roleInstance = System.getenv("HOSTNAME"); // default hostname
```
trask, line 46: on the agent side, won't this overwrite role instance provided by the user?

heyams: if you take a look at my most recent commit, this happens before populateDefaults, it will get overwritten by user's config.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java
heyams: i can move line 214 to the top, but inside updateRoleNameAndInstance, i can override the value if it has been set. not sure builder let me retrieve it though.

heyams: or move both to the top and keep the order?

heyams: i moved it back to the top and add a check in the builder.. only override it if it's not set yet. i think it's cleaner this way so that it is not depended on the order of code execution.

heyams: i am going to add a test for that.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
trask: oh, right 👍

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/utils/ResourceParser.java
```diff
@@ -0,0 +1,65 @@
...
+import com.azure.monitor.opentelemetry.exporter.implementation.models.ContextTagKeys;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import java.util.Map;
+
+public final class ResourceParser {
+
+  public static void updateRoleNameAndInstance(
+      AbstractTelemetryBuilder builder, Resource resource) {
+    Map<String, String> existingTags = builder.build().getTags();
```
trask, line 34: this _looks_ expensive call `build` just to get the tags, though it's not really expensive since we have a slightly unusual builder here

I think my recommendation would be to remove the `System.getenv("HOSTNAME")` default from here, since that's what's causing the issue, and move that default into the standalone exporter's "telemetryInitializer"

trask: and then this class is just purely mapping otel `service.*` into ai `cloud.*`

heyams: it still requires builder.build().getTags() check to make sure it doesn't overwrite customer's role name. it doesn't save much. besides hostname is part of the spec for roleinstance's default value.

trask: what if we change the "telemetry initializer" from `Consumer<AbstractTelemetryBuilder>` to `BiConsumer<AbstractTelemetryBuilder, Resource>`, and then we could call `ResourceParser` just from the standalone exporters and not affect the agent?

in the future we _might_ want to also map otel `service.*` attributes in the agent, but I think we will want to think carefully about order of precedence there

heyams: changes are in standalone exporters though. agent is using the standalone exporter. i'm not sure not affect the agent part.

trask: > what if we change the "telemetry initializer" from `Consumer<AbstractTelemetryBuilder>` to `BiConsumer<AbstractTelemetryBuilder, Resource>`, and then we could call `ResourceParser` just from the standalone exporters and not affect the agent?

I sent a couple of commits to your branch to show what I meant, I think it turned out nice, also uncovered a couple of things we had missed previously

trask (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/LogDataMapper.java
```diff
@@ -169,6 +173,9 @@ static void setExtraAttributes(AbstractTelemetryBuilder telemetryBuilder, Attrib
     attributes.forEach(
         (key, value) -> {
           String stringKey = key.getKey();
+          if (stringKey.startsWith("applicationinsights.internal.")) {
```
heyams, line 176: when do we officially deprecated internal attributes?

trask: we don't need to deprecate `internal` attributes, they are just internal implementation details that we don't expose to customers

PR #2329, opened 2022-06-15
Make disk persistence capacity configurable

heyams:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileWriter.java
```diff
@@ -34,24 +34,25 @@
 /** This class manages writing a list of {@link ByteBuffer} to the file system. */
 final class LocalFileWriter {
 
-  // 50MB per folder for all apps.
-  private static final long MAX_FILE_SIZE_IN_BYTES = 52428800; // 50MB
   private static final String PERMANENT_FILE_EXTENSION = ".trn";
 
+  private final long diskPersistenceMaxSizeMb;
```
trask, line 39:
```suggestion
  private final long diskPersistenceMaxSizeBytes;
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -304,6 +304,9 @@ public static class PreviewConfiguration {
     // telemetry (they are much smaller so a larger queue size is ok)
     public int metricsExportQueueCapacity = 65536;
 
+    // disk persistence has a default capacity of 50MB
+    public int diskPersistenceMaxSizeMb = 50;
```
trask, line 308: 2.x SDK enforces a limit of 1 GB:

https://github.com/microsoft/ApplicationInsights-Java/blob/d3a73be1531a258a3627f36542cc36050464fd3b/core/src/main/java/com/microsoft/applicationinsights/internal/channel/common/TransmissionFileSystemOutput.java#L83

I'm not really sure the reason, can you see if Application Insights .NET SDK has a max capacity that users can configure? if they do, we may want to implement that max too

also, I wonder if we should allow `0`, meaning disable disk persistency, which users on read-only filesystems could use (and then we wouldn't need to log warning for those users)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
```diff
@@ -48,9 +49,14 @@ public LocalStorageTelemetryPipelineListener(
...
         new LocalFileLoader(localFileCache, telemetryFolder, stats, suppressWarnings);
-    localFileWriter = new LocalFileWriter(localFileCache, telemetryFolder, stats, suppressWarnings);
+    localFileWriter =
+        new LocalFileWriter(
+            diskPersistenceMaxSizeMb, localFileCache, telemetryFolder, stats, suppressWarnings);
 
-    localFileSender = new LocalFileSender(loader, pipeline, suppressWarnings);
+    // send persisted telemetries from local disk every 30 seconds by default.
+    // if diskPersistenceMaxSizeMb is greater than 50, it will get changed to 10 seconds.
+    long intervalSeconds = diskPersistenceMaxSizeMb > 50 ? 10 : 30;
```
trask, line 58: estimating how long this will take to send a large number of batches from disk:

  600 characters per telemetry
* 512 telemetry per batch
= ~300kb per batch

1gb of telemetry will be ~3000 batches

so sending 1 batch per 10 seconds, this will take 8 hours

which is not bad

(we can always revisit this detail when spec'ing it out with other langs)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: dotnet is default to 50MB as well. they use 30 seconds as the default sending interval. same as us.. it's not configured. we could ignore value that is less than 50? and document it?

heyams: dotnet [source code](https://github.com/microsoft/ApplicationInsights-dotnet/blob/405fd6a9916956f2233520c8ab66110a1f9dcfbc/BASE/src/Microsoft.ApplicationInsights/Channel/InMemoryTransmitter.cs#:~:text=private%20TimeSpan%20sendingInterval%20%3D%20TimeSpan.FromSeconds(30)%3B)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
heyams: i can a todo to revisit this interval.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalFileWriter.java
```diff
@@ -34,24 +34,25 @@
...
   LocalFileWriter(
+      int diskPersistenceMaxSizeMb,
       LocalFileCache localFileCache,
       File telemetryFolder,
       LocalStorageStats stats,
       boolean suppressWarnings) { // used to suppress warnings from statsbeat
     this.telemetryFolder = telemetryFolder;
     this.localFileCache = localFileCache;
     this.stats = stats;
+    this.diskPersistenceMaxSizeBytes = diskPersistenceMaxSizeMb * 1024L * 1024L;
```
jeanbisutti, line 55: If it would not be an issue in terms of allocation, it may be worth introducing  a `DiskPersistenceMaxSize` object type.  It would have advantages:
 - A method could be added to the `DiskPersistenceMaxSize` object type to encapsulate this kind of code: `diskPersistenceMaxSizeMb * 1024L * 1024L;`
 - This would remove the need to add comments (for example in `AzureMonitorExporterBuilder` for this PR)
- This would strengthen the code checks at compilation

heyams: comment is only for default value. creating an object for a primitive type with super simple arithmetic is not worthy in my opinion.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: > dotnet is default to 50MB as well

do they allow user to configure this? and if so, do they have a maximum value that the user can set?

heyams: it's fixed. no configuration.

trask: that makes me a little concerned that there might be a reason behind why .NET SDK doesn't support this (since they typically have so many more users and configuration knobs in general)

@TimothyMothra do you happen to know if there's any specific reason .NET SDK doesn't support configuration of the disk size / do you know if the topic has come up before? thx

TimothyMothra: .NET SDK's disk storage capacity is configurable: [SeverTelemetryChannel](https://github.com/microsoft/ApplicationInsights-dotnet/blob/405fd6a9916956f2233520c8ab66110a1f9dcfbc/BASE/src/ServerTelemetryChannel/ServerTelemetryChannel.cs#L188-L196)

trask: Thx, does the .Net SDK impose any limit on how large the user can configure it to be?

heyams: but wait, is it exposed to customers though? is it available in the public interface?

cijothomas: > Thx, does the .Net SDK impose any limit on how large the user can configure it to be?

Only limit is the maximum value of `long` itself.

heyams: looks like it's public API. based on the code, it can be any number.

trask: awesome, thx all 👍

trask (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/localstorage/LocalStorageTelemetryPipelineListener.java
cijothomas: @trask We allow (in .NET) users to configure how many parallel sends can occur. (Default is 10). They can increase it to a large number, increasing the throughput heavily, subject to physical limitations of the network itself. 

While scenarios like load testing, by increasing the senders, user could send HUGE amounts of data.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: https://github.com/microsoft/Telemetry-Collection-Spec/pull/83

PR #2333, opened 2022-06-21
Update dependencies

trask:
based on running

```
./gradlew dependencyUpdates -Drevision=release
```


heyams:
do we want to include this in the new release?

heyams:
i wonder if it will impact the binary size. Let's include it.

heyams (approved):

PR #2334, opened 2022-06-21
Add dependabot configuration

trask:
this is the same configuration as used in java-instrumentation repo


jeanbisutti (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+registries:
+  gradle-plugin-portal:
+    type: maven-repository
+    url: https://plugins.gradle.org/m2
+    username: dummy # Required by dependabot
```
heyams, line 6: dependabot needs to drop username and password from the yml. a feature ask for them? it's a weird setup and we need to suppress the dummy cred.

heyams (approved):

.github/dependabot.yml
trask: I didn't look into it, just copied this from OpenTelemetry 😄

I removed the creds, and will merge this and see what happens, and will bring them back if needed

trask: ☹️

> Your .github/dependabot.yml contained invalid details
Dependabot encountered the following error when parsing your .github/dependabot.yml:
>
> The property '#/registries/gradle-plugin-portal/' did not contain a required property of 'username'
The property '#/registries/gradle-plugin-portal/' did not contain a required property of 'password'
Please update the config file to conform with Dependabot's specification.

but looks like an empty `""` password may work, curious if cred scanner will ignore that

heyams: worth of trying :)

PR #2341, opened 2022-06-21
Publish runtime-attach artifact

trask:
This is step 1. Step 2 is to update the release scripts.


heyams:
why do we need have a separate artifact for runtime-attach? is it enabled through system property?

trask:
> why do we need have a separate artifact for runtime-attach? is it enabled through system property?

(discussed offline)

heyams (approved):

PR #2368, opened 2022-06-30
Add ability to trigger profiles from request times and add diagnostic engine api

johnoliver:
Fix # .

- [ ] CHANGELOG.md updated


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/ProfilingInitializer.java
```diff
@@ -83,6 +78,17 @@ private static ServiceProfilerServiceConfig formServiceProfilerConfig(
       Configuration.ProfilerConfiguration configuration, File tempDir) {
     URL serviceProfilerFrontEndPoint =
         TelemetryClient.getActive().getConnectionString().getProfilerEndpoint();
+
+    // If the user has overridden their service profiler endpoint use that url
+    if (configuration.serviceProfilerFrontEndPoint != null) {
+      try {
+        serviceProfilerFrontEndPoint = new URL(configuration.serviceProfilerFrontEndPoint);
+      } catch (MalformedURLException e) {
+        // startupLogger.error("Failed to parse url: " +
```
karianna, line 87: Commented out code

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1223,13 +1223,72 @@ public void validate() {
...
+    REGEX
+  }
+
+  public static class SpanFilter {
+    public SpanFilterType type;
+    public String value;
+  }
+
+  public static class SpanAggregationConfig {
+    public int percentile = 95;
```
karianna, line 1236: Why this magic number?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/ProfilerServiceInitializer.java
```diff
@@ -116,9 +133,12 @@ public static synchronized void initialize(
         return;
       }
 
+      // Initialise diagnostic service
+      startDiagnosticEngine();
```
karianna, line 137: Should this be wrapped behind a feature flag?

karianna (changes requested):
Looks good overall - not sure on the Javadoc policy for this repo (lots is missing)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
johnoliver: need to have some default percentile, 95th percentile is just an arbitrary default

karianna (approved):
LGTM

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/QuantileAggregation.java
```diff
@@ -0,0 +1,84 @@
...
+import com.microsoft.applicationinsights.alerting.analysis.TimeSource;
+import com.microsoft.applicationinsights.alerting.analysis.data.TelemetryDataPoint;
+import java.util.OptionalDouble;
+import java.util.function.Consumer;
+
+/** Applies a time window to data and calculates a quantile of the data during that window. */
+public class QuantileAggregation implements Aggregation {
+
+  private final int quantile;
+  private Consumer<Double> consumer;
```
jeanbisutti, line 34: Would it make sense to use  a `java.util.function.DoubleConsumer` instead of a `java.util.function.Consumer<Double>` to ltry to limit heap allocation from Application Insights?

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/RollingAverage.java
```diff
@@ -0,0 +1,71 @@
...
+package com.microsoft.applicationinsights.alerting.analysis.aggregations;
+
+import com.microsoft.applicationinsights.alerting.analysis.TimeSource;
+import com.microsoft.applicationinsights.alerting.analysis.data.TelemetryDataPoint;
+import java.util.OptionalDouble;
+import java.util.function.Consumer;
+
+/** Applies a time window to data and calculates a mean of the data during that window. */
+public class RollingAverage implements Aggregation {
+  private Consumer<Double> consumer;
```
jeanbisutti, line 31: Would it make sense to use a java.util.function.DoubleConsumer instead of a java.util.function.Consumer<Double> to try to limit heap allocation from Application Insights?

agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrUploadService.java
```diff
@@ -58,16 +59,22 @@ public void receive(AlertBreach alertBreach, long timestamp, File file) {
 
     jfrUploader
         .uploadJfrFile(
+            UUID.fromString(alertBreach.getProfileId()),
```
jeanbisutti, line 62: `UUID` seems generated, converted into a String and then, at this point (JFR file uploading) the `String` is reconverted into `UUID`. I may miss something. `UUID` could not be always used and propagated until the JFR loading?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/triggers/AlertTriggerSpanExporter.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(AlertTriggerSpanExporter.class);
+
+  // Next SpanExporter in App insights pipeline
+  private final SpanExporter delegate;
+
+  // Execution context of the execution loop
+  private final Future<?> spanProcessor;
+
+  // Spans to be processed
+  private final LinkedBlockingQueue<SpanData> queue = new LinkedBlockingQueue<>();
```
jeanbisutti, line 59: It could interesting to replace the unbounded queue with a bounded one and control the size of the queue, to avoid maintaining in memory a queue with a very large number of elements.

jeanbisutti: An `ArrayBlockingQueue` might perhaps be preferable [as in LMAX](https://github.com/LMAX-Exchange/disruptor/wiki/Performance-Results#throughput-performance-testing).

jeanbisutti (changes requested):
My main concern would be to replace the unbounded queue with a bounded one.

agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrUploadService.java
johnoliver: AlertBreach does get serialized and read from a number of places. Unfortunately putting a UUID into it would then necessitate adding Jackson libraries to `agent-alerting-api` for it to understand how to serialize/deserialize a UUID. Given the frequent CVEs that you get from Jackson I would prefer not to add it as a dependency for such a small use case. I think the alternative would be to just treat the id from here on out as a string rather than a UUID, however the reality of the api is that it is required to be a UUID.

jeanbisutti (approved):

agent/agent-profiler/agent-alerting/build.gradle.kts
```diff
@@ -6,6 +6,9 @@ dependencies {
   implementation(project(":agent:agent-profiler:agent-alerting-api"))
   implementation("org.slf4j:slf4j-api")
 
+  implementation("com.google.guava:guava")
```
trask, line 9: we don't already have guava in the javaagent, would love not to pull in the extra 2.5 mb if possible

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/triggers/AlertTriggerSpanExporter.java
```diff
@@ -0,0 +1,151 @@
...
+  // Execution context of the execution loop
+  private final Future<?> spanProcessor;
+
+  // Spans to be processed
+  private final BlockingQueue<SpanData> queue;
+  private final Supplier<AlertingSubsystem> alertingSubsystemSupplier;
+
+  @Override
+  public CompletableResultCode export(Collection<SpanData> spans) {
+    // Process spans off of the main thread to avoid any complex logic blocking span processing
```
trask, line 65: this is already off of the main thread (behind a `BatchSpanProcessor`), I think it would be ok to process directly here instead of passing off to queue

core/src/test/java/com/microsoft/applicationinsights/internal/profiler/ProfilerServiceTest.java
```diff
@@ -0,0 +1,275 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import static com.microsoft.applicationinsights.internal.perfcounter.Constants.TOTAL_CPU_PC_METRIC_NAME;
+import static com.microsoft.applicationinsights.internal.perfcounter.jvm.JvmHeapMemoryUsedPerformanceCounter.HEAP_MEM_USED_PERCENTAGE;
+
+public class ProfilerServiceTest {
```
trask, line 70: this file somehow went rogue and isn't under any module

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/triggers/AlertTriggerSpanExporter.java
```diff
@@ -0,0 +1,151 @@
...
+          LOGGER.error("Error while processing span: {}", e.getMessage());
+        } catch (InterruptedException e) {
+          return;
+        }
+      }
+    }
+  }
+
+  public void processSpan(SpanData span) {
+    if (span.getKind() == SpanKind.SERVER) {
```
trask, line 111: I sent a commit to your branch to refactor `SpanDataMapper` and introduced a new method that you may want to use here

```suggestion
    if (SpanDataMapper.isRequest(span)) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/triggers/AlertTriggerSpanExporter.java
```diff
@@ -0,0 +1,151 @@
...
+      double duration = (span.getEndEpochNanos() - span.getStartEpochNanos()) / 1_000_000.0d;
```
trask, line 112: `duration` -> `durationMillis` to help keep track of units

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/AlertingSubsystem.java
```diff
@@ -131,8 +133,17 @@ public void initialize(AlertingConfiguration alertConfig) {
...
+    }
+  }
+
+  public void trackTelemetryDataPoint(@Nullable TelemetryDataPoint value) {
+    if (value != null) {
+      workQueue.add(value);
+    }
   }
 
   /** Block until work queue is empty. */
```
trask, line 149:
```suggestion
  /** Block until work queue is empty. */
  // only used by tests
```

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/Aggregation.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.microsoft.applicationinsights.alerting.analysis.data.TelemetryDataPoint;
+import java.util.OptionalDouble;
+import java.util.function.DoubleConsumer;
+
+/** A process that consumes data points and computes metrics. */
+public interface Aggregation {
+
+  /** Add new data to the aggregation. */
+  double update(TelemetryDataPoint telemetryDataPoint);
```
trask, line 32: it doesn't look like the return value is ever used

```suggestion
  void update(TelemetryDataPoint telemetryDataPoint);
```

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/WindowedAggregation.java
```diff
@@ -19,62 +19,56 @@
...
   private final long windowLengthInSec;
   private final TimeSource timeSource;
-  private final List<TelemetryDataPoint> telemetryDataPoints = new ArrayList<>();
-  private Consumer<Double> consumer;
+  private final List<TelemetryDataPoint> telemetryDataPoints =
+      Collections.synchronizedList(new ArrayList<>());
   private static final int DEFAULT_ROLLING_AVERAGE_WINDOW_IN_SEC =
       Integer.parseInt(
           System.getProperty(
               "applicationinsights.preview.profiler.rolling-average-window-in-sec", "120"));
```
trask, line 41: Since we know the percentile (e.g. 95th) and the threshold (e.g. 5000 millis), maybe we can do something smart like keep 1-second buckets that store just the request count that exceeded the threshold and the total request count. In addition to taking less memory, should be a lot faster to calculate if the 95th percentile exceeded 5000 milliseconds compared to (whatever magic guava Quantiles does). Also maybe this is our way out of the guava dependency 🤔.

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/QuantileAggregation.java
```diff
@@ -0,0 +1,84 @@
...
+    double[] dataPoints =
+        windowedAggregation.getTelemetryDataPoints().stream()
+            .mapToDouble(TelemetryDataPoint::getValue)
+            .toArray();
+
+    if (dataPoints.length == 0) {
+      return OptionalDouble.empty();
+    }
+
+    return OptionalDouble.of(Quantiles.percentiles().index(quantile).compute(dataPoints));
```
trask, line 82: (only if we have to keep guava, may as well not clone the dataPoints array)
```suggestion
    return OptionalDouble.of(Quantiles.percentiles().index(quantile).computeInPlace(dataPoints));
```

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/WindowedAggregation.java
```diff
@@ -19,62 +19,56 @@
...
 
-  public RollingAverage setConsumer(Consumer<Double> consumer) {
-    this.consumer = consumer;
-    return this;
+  public long getWindowLengthInSec() {
+    return windowLengthInSec;
   }
 
-  public double track(TelemetryDataPoint telemetryDataPoint) {
+  public List<TelemetryDataPoint> update(TelemetryDataPoint telemetryDataPoint) {
```
trask, line 65: looks like return value is not used
```suggestion
  public void update(TelemetryDataPoint telemetryDataPoint) {
```

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/data/TelemetryDataPoint.java
```diff
@@ -31,17 +31,23 @@ public class TelemetryDataPoint implements Comparable<TelemetryDataPoint> {
   private final AlertMetricType type;
   private final Instant time;
   private final double value;
+  private final String name;
 
-  public TelemetryDataPoint(AlertMetricType type, Instant time, double value) {
+  public TelemetryDataPoint(AlertMetricType type, Instant time, String name, double value) {
```
trask: I pushed a commit to turn this class into `@AutoValue` (let me know if it needed to implement Comparable)

johnoliver: No, should be fine

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1230,13 +1230,90 @@ public void validate() {
...
+    public SpanTriggerThresholdType type = SpanTriggerThresholdType.GREATER_THAN;
+
+    // Threshold value applied to the output of the aggregation
+    // i.e :
+    //  - For the BreachRatio aggregation, 0.75 means this will trigger if 75% of sample breach the
+    // threshold.
+    //  - For a rolling average aggregation 0.75 will mean this will trigger if the average span
+    // processing time
+    //      breaches 0.75ms
+    public float value = 0.75f;
```
trask, line 1278: just confirming my understanding: this means the default trigger is based on 75th percentile exceeding 5000 milliseconds?

johnoliver: correct

agent/agent-profiler/agent-alerting-api/src/main/java/com/microsoft/applicationinsights/alerting/config/AlertMetricType.java
```diff
@@ -25,5 +25,6 @@ public enum AlertMetricType {
   CPU,
   MEMORY,
   PERIODIC,
-  MANUAL
+  MANUAL,
+  SPAN
```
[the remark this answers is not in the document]
trask: ah, how about `REQUEST` (since only applied to "request" spans)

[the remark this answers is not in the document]
johnoliver: ok

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/filter/AlertSpanFilter.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package com.microsoft.applicationinsights.alerting.analysis.filter;
+
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+
+/** Filters span data based on its name. */
+public abstract class AlertSpanFilter implements Predicate<String> {
```
trask, line 29: ~`AlertSpanFilter` -> `AlertSpanNameFilter`~

hm, maybe `AlertRequestFilter`, or even `AlertRequestNameFilter`

agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/upload/ServiceProfilerUploader.java
```diff
@@ -265,9 +302,13 @@ private static File createZippedTraceFile(UploadContext uploadContext) throws IO
   }
 
   // Deleting file recursively.
-  private static void deletePathRecursive(File fileToDelete) throws IOException {
-    if (fileToDelete.exists()) {
-      deletePathRecursive(fileToDelete.toPath());
+  private static void deletePathRecursive(@Nullable File fileToDelete) throws IOException {
+    if (fileToDelete != null && fileToDelete.exists()) {
+      if (retainJfrFile) {
+        LOGGER.info("JFR file retained at: " + fileToDelete.getAbsolutePath());
```
trask, line 308:
```suggestion
        LOGGER.info("JFR file retained at: {}", fileToDelete.getAbsolutePath());
```

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/windowed/WindowedAggregation.java
```diff
@@ -0,0 +1,112 @@
...
+        if (trackCurrentBucket) {
+          buckets.add(currentBucket);
+        }
+      }
+
+      return currentBucket;
+    }
+  }
+
+  private void gcBuckets(Instant cutoff) {
```
trask, line 105: what does `gc` mean here?

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/filter/AlertSpanFilter.java
johnoliver: I have named this AlertRequestFilter, however applied the NameFilter suffix only to the specific fiter that applies a name filter, this abstract class could be subclassed to create any filter over a span, not just the name

agent/agent-profiler/agent-alerting/src/main/java/com/microsoft/applicationinsights/alerting/analysis/aggregations/windowed/WindowedAggregation.java
johnoliver: was using to to mean garbage collect, renamed it to `removeBucketsBeforeCutoff`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1230,13 +1230,90 @@ public void validate() {
     }
   }
 
+  public enum RequestFilterType {
+    REGEX
```
trask, line 1234: maybe?

```suggestion
    NAME_REGEX
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1230,13 +1230,90 @@ public void validate() {
...
+    // Used by the breach ratio aggregation
+    public int thresholdMs = 5000;
+
+    // Minimum number of samples that must have been collected in order for the aggregation to
+    // produce data. Avoids volatile aggregation output on small sample sizes.
+    public int minimumSamples = 0;
+  }
+
+  public enum RequestAggregationType {
+    BREACH_RATIO
```
trask, line 1254: this has been our convention for configuration enum mappings:

```suggestion
    @JsonProperty("breach-ratio")
    BREACH_RATIO
```

johnoliver: I also applied this to `name-regex` and `greater-than` correct?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1230,13 +1230,93 @@ public void validate() {
...
+  }
+
+  public enum RequestTriggerThrottlingType {
+    COOLDOWN
+  }
+
+  public static class RequestTriggerThrottling {
+    public RequestTriggerThrottlingType type = RequestTriggerThrottlingType.COOLDOWN;
+    public long value = 60000; // in ms
+  }
```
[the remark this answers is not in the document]
johnoliver: An alternate could be RATE, i.e "No more than 2 profiles per hour". Could also do different types of COOLDOWN, i.e exponential backoff. If we were to encode different types of throttle, then it may be more accurate to call this FIXED_DURATION_COOLDOWN, that then leaves scope for other types of cooldown.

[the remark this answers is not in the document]
johnoliver: I have updated it to FIXED_DURATION_COOLDOWN and aliased it to "fixed-duration-cooldown"

trask:
thx 🎉

PR #2371, opened 2022-06-30
Snippet Injection Record

siyuniums:
Try to open a PR


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -102,6 +106,16 @@ public class SecondEntryPoint implements AutoConfigurationCustomizerProvider {
   @Nullable private static BatchSpanProcessor batchSpanProcessor;
   @Nullable private static MetricReader metricReader;
 
+  public String readFile(String resourceName) throws IOException {
```
trask, line 109:
```suggestion
  private static String readFile(String resourceName) throws IOException {
```

trask (approved):
in a follow-up, let's add a json configuration option under "preview" to enable snippet injection (since we don't want it enabled by default)

jeanbisutti:
@trask @siyuniums The tests can't run: `Could not find io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.16.0-alpha-SNAPSHOT.`

siyuniums:
> @trask @siyuniums The tests can't run: `Could not find io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.16.0-alpha-SNAPSHOT.`

Hi Jean, thanks for the reviewing. This PR is only for a code record. The snapshot is a new local version open telemetry which have the snippet injection enabled.

trask:
> Hi Jean, thanks for the reviewing. This PR is only for a code record. The snapshot is a new local version open telemetry which have the snippet injection enabled.

this PR is targeted to the `snippet-injection` branch where we'll track these changes until we're able to get the upstream changes landed and then we'll send PR to propose merging the `snippet-injection` branch to `main`

PR #2391, opened 2022-07-13
Add Java 18 to test matrix

trask:


heyams (approved):

jeanbisutti:
c7961639234bddf600fada4b2dd90d0b9a73d415 commit removes asm transitive dependencies from the agent project:  `./gradlew :agent:agent:dependencies`  => no asm dependency. When the agent is packaged and used for this commit, an asm version that does not support java 18 is still in use (this version is ok until Java 17). Open telemetry java agent 1.15.0 uses an asm version that works until Java 19.

trask:
I saw the issue below earlier with Java 17, some weird incompatibility between log4j 1.2 MDC and Java 17 (I don't think I ever tracked it down), but then the issue went away in later Java 17 releases, I'm guessing they made some fix/accommodation in the JDK, see https://github.com/microsoft/ApplicationInsights-Java/pull/1920/files#diff-c6a2e6bfb7c1b49dd897a039b659bdb53acd65dfddef05473a1dc98d90b23f8d

```
TraceLog4j12Test$Tomcat8Java18Test > testTraceLog4j1_2WithExeption() FAILED
    java.lang.AssertionError: 
    Expecting map:
      {"Logger Message"="This is an exception!", "LoggerName"="smoketestapp", "SourceType"="Logger", "ThreadName"="http-nio-8080-exec-2"}
    to contain entries:
      ["MDC key"="MDC value"]
    but could not find the following map entries:
      ["MDC key"="MDC value"]
```

PR #2396, opened 2022-07-14
Clean up json playback

heyams:


trask (approved):
nice

heyams:
playback is per URI.

PR #2397, opened 2022-07-14
Better Azure Functions consumption plan behavior

trask:


agent/instrumentation/azure-functions/src/main/java/io/opentelemetry/javaagent/instrumentation/azurefunctions/FunctionEnvironmentReloadInstrumentation.java
```diff
@@ -0,0 +1,56 @@
...
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@SuppressWarnings("unused")
+class FunctionEnvironmentReloadInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return named("com.microsoft.azure.functions.worker.handler.InvocationRequestHandler");
```
heyams, line 39: it seems we need to update the typematcher here? it will be different from InvocationInstrumentation.java.

agent/instrumentation/azure-functions/src/main/java/io/opentelemetry/javaagent/instrumentation/azurefunctions/FunctionEnvironmentReloadInstrumentation.java
```diff
@@ -0,0 +1,56 @@
...
+    transformer.applyAdviceToMethod(
+        isMethod().and(named("execute")),
+        FunctionEnvironmentReloadInstrumentation.class.getName() + "$ExecuteAdvice");
+  }
+
+  @SuppressWarnings("PrivateConstructorForUtilityClass")
+  public static class ExecuteAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void methodExit() {
+      AzureFunctions.configureOnce();
```
heyams, line 53: do we need to check any env vars before configure if i recalled correctly from the meeting with function team?

trask: I believe those are the env vars that are checked inside of `configureOnce()`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AzureFunctionsInitializer.java
```diff
@@ -77,6 +72,18 @@ public void lazyLoad() {
...
   }
 
+  private static void disableBytecodeInstrumentation() {
+    Instrumentation instrumentation = InstrumentationHolder.getInstrumentation();
+    ClassFileTransformer transformer = ClassFileTransformerHolder.getClassFileTransformer();
+    if (instrumentation == null || transformer == null) {
+      return;
+    }
+    if (instrumentation.removeTransformer(transformer)) {
+      ClassFileTransformerHolder.setClassFileTransformer(null);
```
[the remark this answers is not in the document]
trask: ya, I initially thought `removeTransformer` was enough, but otel javaagent also uses ClassFileTransformerHolder.getClassFileTransformer() to transform lambda classes (which don't go through normal bytecode instrumentation process), so that's why also need to set that to `null`

jeanbisutti (approved):

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/AzureFunctions.java
```diff
@@ -21,23 +21,31 @@
...
-    }
+  public static boolean hasConnectionString() {
+    return hasConnectionString.get();
   }
 
-  public interface Accessor {
-    void lazyLoad();
+  public static void configureOnce() {
+    if (configure != null) {
+      if (!hasConnectionString.get()) {
```
heyams, line 43: might just reuse the method: !hasConnectionString()

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -539,9 +539,13 @@ private static String getDefaultPath() {
       if (DiagnosticsHelper.useAppSvcRpIntegrationLogging()) {
         return StatusFile.getLogDir() + "/" + DEFAULT_NAME;
       }
-      if (DiagnosticsHelper.useFunctionsRpIntegrationLogging()
-          && !DiagnosticsHelper.isOsWindows()) {
-        return "/var/log/applicationinsights/" + DEFAULT_NAME;
+      if (DiagnosticsHelper.useFunctionsRpIntegrationLogging()) {
+        if (DiagnosticsHelper.isOsWindows()) {
+          // TODO (trask) is there a preference where this should go?
+          return DEFAULT_NAME;
```
heyams, line 545: String tempDirectory = System.getProperty("java.io.tmpdir"), 
tempDirectory + "/applicationinsights/applicationinsights.log"?

PR #2403, opened 2022-07-19
Improve agent already loaded check

jeanbisutti:
Improve agent already loaded check

### DIFF
--- a/agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Agent.java
+++ b/agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Agent.java
@@ -38,7 +38,7 @@ public class Agent {
   private static final AtomicBoolean alreadyLoaded = new AtomicBoolean(false);
 
   public static void premain(String agentArgs, Instrumentation inst) {
-    if (alreadyLoaded.compareAndSet(true, true)) {
+    if (alreadyLoaded.getAndSet(true)) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Agent.java
```diff
@@ -34,19 +35,33 @@ public class Agent {
   // this is to prevent the agent from loading and instrumenting everything twice
   // (leading to unpredictable results) when -javaagent:applicationinsights-agent.jar
   // appears multiple times on the command line
-  private static volatile boolean alreadyLoaded;
+  private static final AtomicBoolean alreadyLoaded = new AtomicBoolean(false);
 
   public static void premain(String agentArgs, Instrumentation inst) {
-    if (alreadyLoaded) {
+    if (alreadyLoaded.compareAndSet(true, true)) {
```
trask, line 41:
```suggestion
    if (alreadyLoaded.getAndSet(true)) {
```

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Agent.java
```diff
@@ -34,19 +35,33 @@ public class Agent {
...
+    try {
+      if (Boolean.getBoolean("applicationinsights.debug.startupProfiling")) {
+        StartupProfiler.start();
+      }
+
+      OpenTelemetryAgent.premain(agentArgs, inst);
+    } finally {
+      if (!agentIsInstalled()) {
+        alreadyLoaded.set(false);
+      }
```
trask, line 54: do you think this is needed? my thinking is that if it failed to install, may as well not try again

jeanbisutti: Removed, as discussed

trask (approved):

heyams (approved):

PR #2407, opened 2022-07-19
Fix live metrics

trask:


heyams (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/quickpulse/model/QuickPulseMetrics.java
```diff
@@ -29,12 +29,12 @@ public class QuickPulseMetrics {
...
   @JsonProperty(value = "Value")
-  private final long value;
+  private final double value;
 
   @JsonProperty(value = "Weight")
-  private final int weight;
+  private final long weight;
 
-  public QuickPulseMetrics(String name, long value, int weight) {
+  public QuickPulseMetrics(String name, double value, long weight) {
```
tokaplan, line 37: weight should be int, same as C# SDK

heyams: will update. thanks @tokaplan

PR #2409, opened 2022-07-20
Track messageId in App Service ETW event and Linux json

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -162,7 +163,7 @@ private static void updateStatusFile(boolean success) {
       StatusFile.putValueAndWrite("AgentInitializedSuccessfully", success, startupLogger != null);
     } catch (Throwable t) {
       if (startupLogger != null) {
-        startupLogger.error("Error writing status.json", t);
+        startupLogger.error("Error writing status.json", t, MessageId.STATUS_FILE_RELATED_ERROR);
```
trask, line 166: I think this changes the meaning, when throwable is last it prints the stack trace

Also, I'm wondering now if it would be better to use MDC for storing/extracting the messageid

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/diagnostics/etw/EtwAppender.java
```diff
@@ -73,6 +76,7 @@ public void start() {
     super.start();
   }
 
+  @SuppressWarnings("SystemOut")
```
trask, line 79: just to not forget
```suggestion
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
heyams: MDC is also a good option

heyams:
@trask i am working on testing it on app service windows and linux now. i had looked at "other" diagnostics error logs, and they have all been covered so far.

PR #2414, opened 2022-07-21
Use more OTel code for runtime attachment

jeanbisutti:
Use more OTel code for runtime attachment


agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -102,4 +95,20 @@ private static boolean agentIsAttached() {
...
+  }
+
+  private static String findAppInsightVersion() {
+    try (InputStream jarAsInputStream =
+        ApplicationInsights.class.getResourceAsStream("/ai.sdk-version.properties")) {
+      Properties props = new Properties();
+      props.load(jarAsInputStream);
+      return props.getProperty("version");
+    } catch (IOException e) {
+      throw new IllegalStateException("Unable to fight Application Insights version", e);
```
heyams, line 111: type: find

heyams (approved):

PR #2417, opened 2022-07-26
Add find-sec-bugs and change preview hash attribute processor from sha1 to sha256

trask:


smoke-tests/framework/build.gradle.kts
```diff
@@ -2,6 +2,12 @@ plugins {
   id("ai.java-conventions")
 }
 
+tasks {
+  named("spotbugsMain") {
+    enabled = false
```
heyams, line 7: should we enable it by default?

heyams: additionally, should the task log for spotbugsMain shows up in `Build / assemble / assemble (pull_request)`?

trask: > should we enable it by default?

it's only disbled here for `:smoke-tests:framework` module since that's just test code

> additionally, should the task log for spotbugsMain shows up in `Build / assemble / assemble (pull_request)`?

it shows up (and fails) the tests builds (which run the `gradle check`)

heyams (approved):

PR #2429, opened 2022-08-02
Add new redirect header for QuickPulse Module

ramanujbhattacharjee93:
Fix # .

- [ ] CHANGELOG.md updated


trask (approved):

trask:
@ramanujbhattacharjee93 is there any context around this change that we should be aware of?

ramanujbhattacharjee93:
> @ramanujbhattacharjee93 is there any context around this change that we should be aware of?

Hey @trask This was a change to fix the redirect header. Earlier we had a problem because while dot Net and Java SDKs worked with the QuickPulseService.svc suffix in the redirect header value, the Node.js SDK did not. 

This is a new header name that will be used by all 3 SDKs now that this issue has been fixed, this is so that older SDKs can continue using the old header, which is only sent to dot NET SDK right because it works with dot NET and we can tell what SDK is dot NET using the invariant version, we dont send the old header to node because it does not work, we dont send it to Java because we cant tell what SDK is Java and what is Node in ping requests. 

Hope this helps clarify

trask:
awesome, thx for the explanation

PR #2437, opened 2022-08-09
Update instrumentation map

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/Instrumentations.java
```diff
@@ -79,20 +79,29 @@ class Instrumentations {
...
     INSTRUMENTATION_MAP.put("io.opentelemetry.vertx-http-client-4.0", 58);
     INSTRUMENTATION_MAP.put("io.opentelemetry.spring-jms-2.0", 59);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.tomcat-10.0", 60);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.jetty-11.0", 61);
+    // TODO (trask) the liberty names should be getting a version soon
+    // see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6449
+    INSTRUMENTATION_MAP.put("io.opentelemetry.liberty", 62);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.liberty-dispatcher", 63);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.methods", 64); // used by "custom instrumentation"
+    INSTRUMENTATION_MAP.put("io.opentelemetry.okhttp-2.2", 65);
```
heyams, line 102: this is exceeding 64 bits. do we want to send second long for anything that is > 64? or clean some up?

trask: ya, that's a good idea to prolong adding the second long

trask: I went through all the indexes (and found more missing instrumentations), and if we reuse the indexes that we retired in 3.3.0, we will be completely maxing out the 64 bits, and one more instrumentation will bump us to a second long.

Since the second long seems inevitable anyways (and probably soon), I think we may as well do that now and avoid reusing which means more version-specific mappings.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/Instrumentations.java
```diff
@@ -69,30 +71,72 @@ class Instrumentations {
...
     INSTRUMENTATION_MAP.put("io.opentelemetry.mongo-3.1", 39); // mongo 4.0 is covered in 3.1
     INSTRUMENTATION_MAP.put("io.opentelemetry.grizzly-2.0", 40);
     INSTRUMENTATION_MAP.put("io.opentelemetry.quartz-2.0", 41);
-    INSTRUMENTATION_MAP.put("io.opentelemetry.apache-camel-2.20", 42);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.apache-camel-2.20", 42); // no usage yet
     INSTRUMENTATION_MAP.put("io.opentelemetry.netty-4.0", 43);
     INSTRUMENTATION_MAP.put("io.opentelemetry.netty-4.1", 44);
     INSTRUMENTATION_MAP.put("io.opentelemetry.okhttp-3.0", 45);
-    INSTRUMENTATION_MAP.put("io.opentelemetry.opentelemetry-annotations-1.0", 46);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.opentelemetry-extension-annotations-1.0", 46);
```
trask, line 88: @heyams can you make this rename in `javaInstrumentationBitMap3_2_3()`, and add the new mappings below?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/Instrumentations.java
```diff
@@ -83,22 +92,51 @@ class Instrumentations {
...
     INSTRUMENTATION_MAP.put("io.opentelemetry.methods", 64); // used by "custom instrumentation"
     INSTRUMENTATION_MAP.put("io.opentelemetry.okhttp-2.2", 65);
     INSTRUMENTATION_MAP.put("io.opentelemetry.opentelemetry-instrumentation-annotations-1.16", 66);
     INSTRUMENTATION_MAP.put("io.opentelemetry.undertow-1.4", 67);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.play-ws-1.0", 68);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.play-ws-2.0", 69);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.play-ws-2.1", 70);
+    INSTRUMENTATION_MAP.put("io.opentelemetry.vertx-kafka-client-3.5", 71);
+
+    // kusto query to find/verify unused
```
heyams, line 115: let's remove the query.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/statsbeat/Instrumentations.java
heyams: i will have a follow-up PR to send second long.

heyams (approved):

PR #2446, opened 2022-08-15
Rename to classic-sdk

trask:


heyams:
can you share the reason behind? is it trying to be in line with azure terminologies?

heyams (approved):

PR #2447, opened 2022-08-15
Support metric namespace

### LINKED ISSUE #1099: Add TrackAvailability method
It would be nice to have the [TrackAvailability](https://docs.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.telemetryclient.trackavailability?view=azure-dotnet) method in Java. Currently it's only available in .NET.

### LINKED ISSUE #2088: Allow passing a namespace with a custom metric
Currently all custom metrics land in the namespace: 'azure.applicationinsights'. It seems to be possible to provide a Metric namespace through the REST API (https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-custom-overview#sample-custom-metric-publication) and in the C# SDK (https://github.com/microsoft/ApplicationInsights-dotnet/blob/405fd6a9916956f2233520c8ab66110a1f9dcfbc/BASE/src/Microsoft.ApplicationInsights/DataContracts/MetricTelemetry.cs) however this is unavailable from the Java SDK.

I would like to set the namespace in a similar way to the C# SDK, an example:

```
var metric = new MetricTelemetry();
metric.setMetricNamespace("<myNamespace>");
metric.setName("<myMetric>");
metric.setValue(50.0);
client.trackMetric(metric);
```

Preferably we would be able to set a default metric namespace on the `TelemetryClient`.

Alternatively we prefix the metric name with the namespace. We are using a shared App Insights instance for over 40 applications. This will result in a huge list of metrics. We would rather namespace them per application (or per group of applications).

trask:
~Currently built on top of #2446, will rebase once that is merged.~

Resolves #2088
Resolves #1099


agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/BytecodeUtil.java
```diff
@@ -300,7 +341,7 @@ void trackDependency(
         @Nullable Date timestamp,
         String name,
         String id,
-        String resultCode,
+        String duration,
```
heyams, line 344: this doesn't seem right. i think you intended to rename `Long totalMillis` to `Long duraiton` (line 345)?

classic-sdk/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
```diff
@@ -58,20 +59,16 @@ public void trackEvent(
       @Nullable Map<String, String> properties,
       @Nullable Map<String, Double> metrics) {
 
-    if (isDisabled()) {
```
heyams, line 61: it seems isDisabled() check is removed everywhere. Should we remove that boolean method too? It's only used in this class.

agent/agent-bootstrap/src/main/java/com/microsoft/applicationinsights/agent/bootstrap/BytecodeUtil.java
trask: ooops!! indeed that is what I meant to do

classic-sdk/core/src/main/java/com/microsoft/applicationinsights/TelemetryClient.java
trask: it was _kind_ of needed, because we used it in the bytecode instrumentation, but I agree that it's confusing, so I removed it, and if we need that feature in the future (to disable 2.x telemetry), it should be a json configuration option anyways

heyams (approved):

PR #2453, opened 2022-08-19
Update to OpenTelemetry 1.17.0

trask:
Includes a fix for role name and ikey overrides.

Also adds new config for upstream option:

```
{
  "instrumentation": {
    "jdbc": {
      "masking": {
        "enabled": false
      }
    }
  }
}
```

Resolves #1989 (disabling the jdbc masking)

PR #2460, opened 2022-08-23
Update smoke tests and fix Kafka running locally

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -192,7 +192,7 @@ public static class Instrumentation {
 
     public EnabledByDefaultInstrumentation azureSdk = new EnabledByDefaultInstrumentation();
     public EnabledByDefaultInstrumentation cassandra = new EnabledByDefaultInstrumentation();
-    public EnabledByDefaultInstrumentation jdbc = new EnabledByDefaultInstrumentation();
+    public JdbcInstrumentation jdbc = new JdbcInstrumentation();
```
jeanbisutti, line 195: It seems possible to make the fields final. It may be converted into constants and the Instrumentation class could be an interface or the class could own a private constructor.

trask: sorry, this code change is part of #2453, this PR is built on top of that one, and I forgot to set the target branch for it to reflect, will rebase this PR after that one is approved/merged

heyams (approved):

PR #2457, opened 2022-08-22
Improve the error message when the profiler can't be initialized

jeanbisutti:
Improve the error message when the profiler can't be initialized

Fix #2379

PR #2462, opened 2022-08-23
Simplify profiler init exception handling

trask:
Noticed this could be simplified while reviewing #2457

### DIFF
--- a/agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrProfiler.java
+++ b/agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrProfiler.java
@@ -114,30 +114,23 @@ public class JfrProfiler implements ProfilerConfigurationHandler, Profiler {
    *     enabled.
    */
   @Override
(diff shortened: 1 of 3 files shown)


agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrProfiler.java
```diff
@@ -114,30 +114,23 @@ public JfrProfiler(ServiceProfilerServiceConfig configuration) {
    *     enabled.
    */
   @Override
-  public boolean initialize(
+  public void initialize(
```
heyams, line 117: can we use initialise ? it seems it's pretty consistent everywhere else.

johnoliver (approved):

agent/agent-profiler/agent-service-profiler/src/main/java/com/microsoft/applicationinsights/serviceprofilerapi/profiler/JfrProfiler.java
trask: initialise is @johnoliver's british spelling 😂

PR #2463, opened 2022-08-24
Better configuration debug logging

trask:
and consolidates some logic between file config parsing and env var / runtime attach config parsing

PR #2465, opened 2022-08-24
Disable profiler-related features for OpenJ9 JVM

jeanbisutti:
Disable profiler-related features for OpenJ9 JVM

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -242,11 +242,11 @@ public class ConfigurationBuilder {
     }
   }
 
-  private static void overlayProfilerEnvVars(Configuration config) {
+  private static void overlayProfilerConfig(Configuration config) {
     if (isOpenJ9Jvm()) {
       configurationLogger.warn(
           "Profiler is not supported for an OpenJ9 JVM. Instead, please use an OpenJDK JVM.");
-      return;
+      config.preview.profiler.enabled = false;
     }
     config.preview.profiler.enabled =
         Boolean.parseBoolean(
@@ -498,7 +498,7 @@ public class ConfigurationBuilder {
     loadJmxMetricsEnvVar(config);
 
     addDefaultJmxMetricsIfNotPresent(config);
-    overlayProfilerEnvVars(config);
+    overlayProfilerConfig(config);
     overlayAadEnvVars(config);
     overlayInstrumentationEnabledEnvVars(config);
   }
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -243,13 +243,23 @@ private static void logWarningIfUsingInternalAttributes(String attributeKey) {
   }
 
   private static void overlayProfilerEnvVars(Configuration config) {
+    if (isOpenJ9Jvm()) {
+      configurationLogger.warn(
+          "Profiler is not supported for an OpenJ9 JVM. Instead, please use an OpenJDK JVM.");
+      return;
+    }
```
trask, line 250: this won't update `profiler.enabled` if it was set via boolean, how about moving this check (and updating the boolean) near the end of `create` method?

jeanbisutti: The `profiler.enabled` will keep its default value, that is to say false. I would prefer to have all the code related to  `profiler.enabled`  intialization in only one place.

trask (approved):

jeanbisutti:
I have added a commit to make clear that the creation of the configuration has 3 steps:

```
public static Configuration create(Path agentJarPath, @Nullable RpConfiguration rpConfiguration)
      throws IOException {
    Configuration config = loadConfigurationFile(agentJarPath);
    logConfigurationWarnings(config);
    overlayConfiguration(agentJarPath, rpConfiguration, config);
    return config;
  }
```

trask:
> I have added a commit to make clear that the creation of the configuration has 3 steps:

nice!

can you review #2463, so that I can merge that as it may have conflicts with this PR?

jeanbisutti:
> > I have added a commit to make clear that the creation of the configuration has 3 steps:
> 
> nice!
> 
> can you review #2463, so that I can merge that as it may have conflicts with this PR?

@trask Done

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -488,7 +506,7 @@ static void overlayFromEnv(Configuration config, Path baseDir) throws IOExceptio
     loadJmxMetricsEnvVar(config);
 
     addDefaultJmxMetricsIfNotPresent(config);
-    overlayProfilerEnvVars(config);
+    overlayProfilerConfig(config);
     overlayAadEnvVars(config);
     overlayInstrumentationEnabledEnvVars(config);
```
trask, line 511: why change ...EnvVars to ...Config?

maybe `overlayProfilerConfigFromEnv`?

trask (approved):

PR #2471, opened 2022-08-27
Connection string overrides

trask:
Built on top of #2468

PR #2472, opened 2022-08-28
Make preagg standard metrics work with ikey and role name overrides

trask:
Built on top of #2471


agent/agent-bootstrap/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/TemporaryMetricsView.java
```diff
@@ -0,0 +1,147 @@
...
+    view.add(SemanticAttributes.HTTP_HOST);
+    view.add(SemanticAttributes.HTTP_SCHEME);
+    view.add(SemanticAttributes.HTTP_FLAVOR);
+    view.add(SemanticAttributes.HTTP_SERVER_NAME);
+    return view;
+  }
+
+  // START APPLICATION INSIGHTS MODIFICATIONS
+
+  static Attributes applyClientDurationAndSizeView(
```
heyams, line 101: i don't get it why these two static methods have to be inside TemporaryMetricsView? it can be elsewhere right? 
Thought we agreed not to include this class.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -161,6 +161,8 @@ public static void updateMetricPointBuilder(
     }
 
     pointBuilder.setValue(pointDataValue);
+    // TODO (heya) why give it the same name as otel metric?
+    //  it seems this field doesn't matter and only _MS.MetricId property matters?
     pointBuilder.setName(metricData.getName());
```
heyams, line 166: dotnet sends name though. that is the display name.

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -170,6 +172,10 @@ public static void updateMetricPointBuilder(
       boolean success = isSuccess(statusCode, captureHttpServer4xxAsError);
       Boolean isSynthetic = attributes.get(IS_SYNTHETIC);
       if (metricData.getName().contains(".server.")) {
+        attributes.forEach(
+            (key, value) ->
+                SpanDataMapper.applyConnectionStringAndRoleNameOverrides(
+                    metricTelemetryBuilder, value, key.getKey()));
```
heyams, line 178: this is applied to all cases.. can you move it to the beginning so that it's applied only once for all cases?
setExtraAttributes can exclude it.

trask: done

agent/agent-bootstrap/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/TemporaryMetricsView.java
```diff
@@ -0,0 +1,147 @@
...
+    applyView(filtered, endAttributes, durationClientView);
+
+    PreAggregatedStandardMetrics.applyHttpClientView(
+        filtered, context, startAttributes, endAttributes);
+
+    return filtered.build();
+  }
+
+  // had to add "2" to this method name because sometimes compilation picks up the upstream version
+  static Attributes applyServerDurationAndSizeView2(
```
heyams, line 114: why it would pick up the upstream version, which is excluded?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -562,12 +571,15 @@ private static SdkMeterProviderBuilder configureMetrics(
...
         PeriodicMetricReader.builder(
-                new AgentMetricExporter(
-                    metricFilters, mapper, telemetryClient.getMetricsBatchItemProcessor()))
-            .setInterval(Duration.ofSeconds(configuration.preview.metricIntervalSeconds))
-            .build();
+            new AgentMetricExporter(
+                metricFilters, mapper, telemetryClient.getMetricsBatchItemProcessor()));
+    int intervalMillis =
+        Integer.getInteger(
+            "applicationinsights.testing.metric-reader-interval-millis",
```
heyams, line 580: can we add comment to indicate that this is for testing purpose only?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -190,11 +198,26 @@ public static void updateMetricPointBuilder(
...
 
+  private static void setExtraAttributes(
+      AbstractTelemetryBuilder telemetryBuilder, Attributes attributes) {
+    attributes.forEach(
+        (key, value) -> {
+          String stringKey = key.getKey();
+          if (SpanDataMapper.applyConnectionStringAndRoleNameOverrides(
+              telemetryBuilder, value, stringKey)) {
+            return;
+          }
```
heyams, line 213: this check is no longer needed.

agent/agent-bootstrap/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/TemporaryMetricsView.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private static Set<AttributeKey> buildActiveRequestsView() {
+    // the list of included metrics is from
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#attributes
+    Set<AttributeKey> view = new HashSet<>();
+    view.add(SemanticAttributes.HTTP_METHOD);
+    view.add(SemanticAttributes.HTTP_HOST);
+    view.add(SemanticAttributes.HTTP_SCHEME);
+    view.add(SemanticAttributes.HTTP_FLAVOR);
```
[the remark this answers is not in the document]
trask: these are for `http.server.active_requests` which we don't use

agent/agent-bootstrap/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/TemporaryMetricsView.java
trask: the problem was happening to me sometimes during compilation, because we don't exclude it until shadowJar, which is after compilation

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
trask: all properties under `applicationinsights.testing.*` are for testing purposes only

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
trask: I think this is needed for custom metrics (so `ai.preview.role_name` can be added as attribute/dimension on custom metrics)

heyams: i see. it was inside ispreaggregated block and it didn't cover custom metric else block.

heyams (approved):

PR #2476, opened 2022-08-29
Update change log for 3.4.0-BETA

trask:


CHANGELOG.md
```diff
@@ -1,5 +1,135 @@
...
+  }
+  ```
+  Note: the `limitPerSecond` can be a decimal value, including values less than 1.
+* Exceptions are no longer captured directly on dependency records, in order to reduce ingestion
+  cost and because exceptions which are logged or bubble up to the request are already captured
+  ([#2423](https://github.com/microsoft/ApplicationInsights-Java/pull/2423)).
+* New versions of the `applicationinsights-core` and `applicationinsights-web` 2.x SDK artifacts
+  have been released, with a reduced API surface area to makes it clear which APIs interop with the
+  3.x Java agent ([#2418](https://github.com/microsoft/ApplicationInsights-Java/pull/2418)).
+  These new versions are no-ops when run without the Java agent. To update your dependencies:
```
heyams, line 59:
```suggestion
  These new versions are no-ops when running without the Java agent. To update your dependencies:
```

CHANGELOG.md
```diff
@@ -1,5 +1,135 @@
...
+  | 2.x dependency | Action | Remarks |
+  |----------------|--------|---------|
+  | `applicationinsights-core` | Update the version to `3.4.0` or later | |
```
heyams, line 62: this is beta 3.4.0-BETA. should we mention the new shim till 3.4.0 GA?

CHANGELOG.md
```diff
@@ -1,5 +1,135 @@
...
+  ```
+  {
+    "sampling": {
+      "limitPerSecond": 5
+    }
+  }
+  ```
+  Note: the `limitPerSecond` can be a decimal value, including values less than 1.
+* Exceptions are no longer captured directly on dependency records, in order to reduce ingestion
+  cost and because exceptions which are logged or bubble up to the request are already captured
```
heyams, line 54: we might need to reword this. it's not very clear to me.

CHANGELOG.md
trask: I think it's good to include here, hopefully someone will give it a spin during BETA, I will update the version mentioned here though

heyams (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,139 @@
...
+  If you do want to sample "standalone" telemetry (e.g. startup logs), you can use sampling
+  overrides, e.g.
+  ```
+  {
+    "preview": {
+      "sampling": {
+        "overrides": [
+          {
+            "telemetryType": "dependency",
+            "includeStandaloneTelemetry": true,
```
heyams, line 29: let's update this to includingStandaloneTelemetry. #2481

PR #2484, opened 2022-08-31
Standard Metric: Fix empty dimension key name in the metric blade on the portal

heyams:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/preaggregatedmetrics/DependencyExtractor.java
```diff
@@ -33,8 +33,8 @@ public final class DependencyExtractor {
 
   // visible for testing
   public static final String DEPENDENCIES_DURATION = "dependencies/duration";
-  public static final String DEPENDENCY_TYPE = "dependency/type";
-  public static final String DEPENDENCY_SUCCESS = "dependency/success";
+  public static final String DEPENDENCY_TYPE = "'Dependency.Type";
```
trask, line 36:
```suggestion
  public static final String DEPENDENCY_TYPE = "Dependency.Type";
```

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/preaggregatedmetrics/DependencyExtractor.java
```diff
@@ -33,8 +33,8 @@ public final class DependencyExtractor {
...
+  public static final String DEPENDENCY_SUCCESS = "Dependency.Success";
   public static final String DEPENDENCY_TARGET = "dependency/target";
```
trask, line 38: this is confusing why some attributes are different format?

trask (approved):

PR #2485, opened 2022-09-01
Add tool to unsign

trask:
TODO still need to write e2e (smoke) test


agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private static final boolean DEBUG = Boolean.getBoolean("applicationinsights.debug.unsign");
+
+  private static final Pattern SIGNATURE_FILE = Pattern.compile("META-INF/.*\\.(RSA|SF)");
+
+  public static void main(String[] args) throws Exception {
+    if (args.length == 1 && args[0].equals("unsign")) {
+      unsign();
+    } else {
+      System.err.println("Incorrect usage");
```
karianna, line 54: Do we want to be more helpful here or deliberately make obtuse for security reasons?

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+          continue;
+        }
+        out.putNextEntry(entry);
+        int read;
+        while ((read = in.read(buffer)) != -1) {
+          out.write(buffer, 0, read);
+        }
+      }
+
+      out.close();
```
karianna, line 92: is out Autoclosed regardless?

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+
+  public static void main(String[] args) throws Exception {
+    if (args.length == 1 && args[0].equals("unsign")) {
+      unsign();
+    } else {
+      System.err.println("Incorrect usage");
+    }
+  }
+
+  private static void unsign() throws Exception {
```
karianna, line 58: Is there some more SRP that could be extracted from this method?  It feels like there is.

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+    return false;
+  }
+
+  private static Path getAgentJar() throws URISyntaxException {
+    CodeSource codeSource = Main.class.getProtectionDomain().getCodeSource();
+
+    if (codeSource == null) {
+      throw new IllegalStateException("could not get agent jar location");
```
karianna, line 147: do we want to sys.err gracefully here or truly bomb out?

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    return Paths.get(codeSource.getLocation().toURI());
+  }
+
+  private static String getJavaExecutable() {
+    return System.getProperty("java.home") + File.separator + "bin" + File.separator + "java";
```
karianna, line 154: Is this always set?

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/Copy.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+
+// this class cannot have any dependencies on any third party libraries
+public class Copy {
+
+  public static void main(String[] args) throws Exception {
```
karianna, line 32: Does the caller deal with Exception?

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
trask: added more helpful messaging 👍

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
trask: `out` wasn't used in try-with-resources, so wasn't, but changed it to use try-with-resources, which is better 👍

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
trask: yup, ptal

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
trask: I think bombing is ok, I'm not sure why code source would be null, so not sure how to even give a more constructive error message (other than asking the user to open a github issue?)

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
trask: this code is gone, it was a fallacy to try to spin up a separate JVM process and overwrite the original agent jar, I mean it worked, but I realized that it could always be a problem if people used that in a script to unsign the jar and copied/moved right after the tool returned, since there was no way (that I could see?) to make the first JVM process not "exit back to the user" until after the second JVM process did it's work

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/internal/Copy.java
trask: this code is gone now

agent/agent/src/main/java/com/microsoft/applicationinsights/agent/Main.java
```diff
@@ -0,0 +1,158 @@
...
+
+    shutdown(executor);
+
+    System.err.println(); // for the "Unsigned......." line
+
+    System.out.println();
+    System.out.println("Unsigned agent jar: " + getRelativePath(unsignedPath));
+  }
+
+  private static void unsign(JarInputStream in, Manifest signedManifest, Path unsignedPath)
```
jeanbisutti, line 89: The responsability of the Main class seems to execute one action type according to a command value (strategy pattern). That would be its only responsibility for me and it would not be necessary that the Main class knows the implementation details of the jar unsigning.

I would have put the code related to the jar unsigning in a specific class.  

If one day we add a feature in the main class, we could immediatly get the idea to create another class and it would help to avoid having a big Main class in the end doing many different things.

trask: done 👍

jeanbisutti (approved):

PR #2486, opened 2022-09-01
Fix config parsing

trask:
This was a regression in 3.4.0-BETA.


agent/agent-tooling/src/test/resources/applicationinsights_with_utf8_bom.json
```diff
@@ -0,0 +1,3 @@
+﻿{
+  "connectionString": "InstrumentationKey=00000000-0000-0000-0000-000000000000"
+}
```
trask, line 3: even though it doesn't show up here, I verified locally after deleting, and pulling back from git, that this file has the byte order marker (BOM) at the beginning, you should be able to see it in Intellij:

image

jeanbisutti (approved):

PR #2490, opened 2022-09-02
Allow suppressing all metrics

### LINKED ISSUE #1905: Turn off  customMetrics, performanceCounters and metrics collected by default
By default agent collects lots of metrics/data:

customMetric	Loaded Class Count
customMetric	% Of Max Heap Memory Used
customMetric	Heap Memory Used (MB)
customMetric	Suspected Deadlocked Threads
customMetric	Current Thread Count
customMetric	GC Total Time
customMetric	GC Total Count
customMetric	HeartbeatState
performanceCounter	Available Bytes
performanceCounter	IO Data Bytes/sec
performanceCounter	% Processor Time
performanceCounter	Private Bytes
metric	Loaded Class Count
metric	% Of Max Heap Memory Used
metric	Heap Memory Used (MB)
metric	Suspected Deadlocked Threads
metric	Current Thread Count
metric	GC Total Time
metric	GC Total Count
metric	HeartbeatState
metric	Available Bytes
metric	% Processor Time
metric	Private Bytes
metric	IO Data Bytes/sec	

I prefer to use standard provided by Spring Boot/Micrometer. Is there a possibility to turn off collecting of the above telemetries?

By default no telemetries should be collected by default. Adding  a config entry like collectStandardJVMMetric should cause collecting the data.

-

**Additional context**
-

trask:
Resolves #1905


heyams:
would it be more convenient to turn off metrics in one kill-switch config for a telemetryKind? customers need to indicate which ones to exclude.. if they want all of them, that's lots of exclude or regular expressions.

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/MetricFilter.java
```diff
@@ -38,8 +25,8 @@ public MetricFilter(Configuration.ProcessorConfig metricFilterConfiguration) {
     this.exclude = new IncludeExclude(metricFilterConfiguration.exclude);
   }
 
-  boolean matches(String metricName) {
-    return !exclude.matches(metricName);
+  boolean exclude(String metricName) {
```
jeanbisutti, line 28: `metricName` parameter can't be `null`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/MetricFilter.java
```diff
@@ -85,14 +72,6 @@ boolean matches(String metricName) {
...
-    if (!MetricFilter.NON_FILTERABLE_METRIC_NAMES.contains(metricName)) {
-      for (MetricFilter metricFilter : metricFilters) {
-        if (!metricFilter.matches(metricName)) {
-          // user configuration filtered out this metric name
-          return true;
-        }
-      }
-    }
-    return false;
+    return metricFilters.stream().anyMatch(metricFilter -> metricFilter.exclude(metricName));
```
jeanbisutti, line 75: I have noticed that this method is called from a [loop](https://github.com/microsoft/ApplicationInsights-Java/blob/d195ed21b89124a1a98f0484ecc8928bf4debb9a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java#L62) in the `AgentMetricExporter` class. If the number of iterations could be high, I would prefer to avoid creating a stream at each iteration, to limit heap allocation.

PR #2496, opened 2022-09-05
Extract Alerting config to common library so that it can be included in AleartBreach

johnoliver:
We require the ability to see the Alert configuration within the AlertBreach. So that we have access to the config schema from the API modules, I have moved the relevant sections from Configuration into the AlertingConfig class in the `agent-alerting-api` module.

The main open question here is, is it acceptable that the Configuration definition gets split up into a separate class/module.

### DIFF
--- a/agent/agent-profiler/agent-alerting-api/build.gradle.kts
+++ b/agent/agent-profiler/agent-alerting-api/build.gradle.kts
@@ -1,7 +1,3 @@
 plugins {
   id("ai.java-conventions")
 }
-
-dependencies {
-  implementation("com.fasterxml.jackson.core:jackson-annotations:2.13.3")
-}
--- a/agent/agent-profiler/agent-alerting-api/build.gradle.kts
+++ b/agent/agent-profiler/agent-alerting-api/build.gradle.kts
@@ -1,3 +1,7 @@
 plugins {
   id("ai.java-conventions")
 }
+
+dependencies {
+  implementation("com.fasterxml.jackson.core:jackson-annotations:2.13.3")
+}
--- a/agent/agent-profiler/agent-alerting-api/gradle.lockfile
+++ b/agent/agent-profiler/agent-alerting-api/gradle.lockfile
@@ -3,6 +3,7 @@
 # This file is expected to be part of source control.
 com.azure:azure-sdk-bom:1.2.6=runtimeClasspath
 com.google.guava:guava-bom:31.1-jre=runtimeClasspath
+com.fasterxml.jackson.core:jackson-annotations:2.13.3=runtimeClasspath
 io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.18.0-alpha=runtimeClasspath
 io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom:1.18.0=runtimeClasspath
 io.opentelemetry:opentelemetry-bom-alpha:1.18.0-alpha=runtimeClasspath
--- a/agent/agent-profiler/agent-alerting-api/gradle.lockfile
+++ b/agent/agent-profiler/agent-alerting-api/gradle.lockfile
@@ -3,7 +3,6 @@
 # This file is expected to be part of source control.
 com.azure:azure-sdk-bom:1.2.6=runtimeClasspath
 com.google.guava:guava-bom:31.1-jre=runtimeClasspath
-com.fasterxml.jackson.core:jackson-annotations:2.13.3=runtimeClasspath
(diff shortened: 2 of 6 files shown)


jeanbisutti:
I may miss something: it would not be possible to propagate the configuration from method to method ([for example the configuration is already injected in `AlertingSubsystem`](https://github.com/johnoliver/ApplicationInsights-Java/blob/283c550afaef30ad01f078145afad7f54a8d1d41/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/ProfilerServiceInitializer.java#L242)) and inject the alerting configuration into `agent-alerting-api` objects?

jeanbisutti:
I may miss something: it would not be possible to propagate the configuration through objects ([for example the configuration is already injected in `AlertingSubsystem`](https://github.com/johnoliver/ApplicationInsights-Java/blob/283c550afaef30ad01f078145afad7f54a8d1d41/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/ProfilerServiceInitializer.java#L242)) and inject the alerting configuration into `agent-alerting-api` objects?

trask:
> I may miss something: it would not be possible to propagate the configuration through objects ([for example the configuration is already injected in `AlertingSubsystem`](https://github.com/johnoliver/ApplicationInsights-Java/blob/283c550afaef30ad01f078145afad7f54a8d1d41/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/ProfilerServiceInitializer.java#L242)) and inject the alerting configuration into `agent-alerting-api` objects?

I believe it's because it would introduce a circular dependency between modules.

@johnoliver would it be horrible to have a "duplicate" configuration inside the profiling module, and create that "duplicate" configuration instance (based on the configuration instance from the json binding)? in other cases in this codebase I have found that decoupling the internal config from the json config actually helpful, and I can optimize the internal config for implementation needs, while leaving the json config optimized for humans

johnoliver:
@trask It would not be horrible no, I was playing the odds here between what you would prefer between duplication and having config defined in multiple places, if you prefer the semi-duplication can certainly go that way

jeanbisutti:
> I believe it's because it would introduce a circular dependency between modules.

 It may be something to fix to avoid this kind of issue.

trask:
> @trask It would not be horrible no, I was playing the odds here between what you would prefer between duplication and having config defined in multiple places, if you prefer the semi-duplication can certainly go that way

let's go with the semi-duplication since that's the approach we've used elsewhere, thx!

jeanbisutti:
> let's go with the semi-duplication since that's the approach we've used elsewhere, thx!

It may have a design issue on Application Insights impending the configuration to be retrieved once and shared with the Java objects of Application Insights. So it would be interesting to understand it to be able to simplify the code.

jeanbisutti:
Hi [@johnoliver](https://github.com/johnoliver)! I have discussed this with [@trask](https://github.com/trask). We agree to continue with the semi-duplication for this PR. We are going to track this configuration issue to try to simplify the developments in the future.

jeanbisutti (approved):

agent/agent-profiler/agent-alerting-api/build.gradle.kts
```diff
@@ -1,3 +1,7 @@
 plugins {
   id("ai.java-conventions")
 }
+
+dependencies {
+  implementation("com.fasterxml.jackson.core:jackson-annotations:2.13.3")
+}
```
trask, line 7: is this still needed?

johnoliver: unfortunately yes, in order to serialize the enums correctly we need access to the JsonProperty annotation

trask: sorry, I think I'm missing something, where does the serialization of AlertingConfig happen?

johnoliver: That serialization happens in the diagnostics so is not in this repository. However you have just reminded me I can use Jacksons mixins to configure the serialization without modifying this class, will remove it from here and use mixins to get the serialization right

trask: > will remove it from here and use mixins

that sounds nice, but if it becomes any trouble, it's ok to keep the annotations here, just add a comment to explain their usage so we don't go removing them thinking they're unused 😅

trask (approved):

PR #2500, opened 2022-09-06
Add start-up diagnostics

jeanbisutti:
Add start-up diagnostics:
* Disable C2 compiler during start-up
* _JVM flags (removed from this PR)_
* Resident-set size
* Native memory tracking

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
@@ -1,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.applicationinsights.agent.internal.init;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+
+class CommandExecutor {
+
+  private CommandExecutor() {}
+
+  static String execute(String[] command) {
+    IllegalStateException exitException = null;
+    String result = null;
+    try {
+      Process process = new ProcessBuilder(command).start();
+      int exitValue = process.waitFor();
+      exitException = buildExitValueExceptionIfNecessary(command, exitValue, process);
+      if (exitException == null) {
+        InputStream inputStream = process.getInputStream();
+        result = toString(inputStream);
+      }
+      process.destroy();
+    } catch (IllegalStateException | IOException | InterruptedException e) {
+      throw combineExceptionsIfNecessary(exitException, e, command);
+    }
+    if (exitException != null) {
+      throw exitException;
+    }
+    return result;
+  }
+
+  static String executeWithoutException(String[] command, Logger startupLogger) {
+    try {
+      return execute(command);
+    } catch (Exception e) {
+      startupLogger.error("Error when executing command " + Arrays.asList(command) + ".", e);
+      if (e.getSuppressed().length == 1) {
+        return e.getMessage() + " (Suppressed: " + e.getSuppressed()[0] + ")";
+      }
+      return e.getMessage();
+    }
+  }
+
+  private static IllegalStateException buildExitValueExceptionIfNecessary(
+      String[] command, int exitValue, Process directivesClearProcess) throws IOException {
+    if (exitValue != 0) {
+      InputStream errorStream = directivesClearProcess.getErrorStream();
+      String error = toString(errorStream);
+      return new IllegalStateException(
+          "Error executing command " + Arrays.asList(command) + ": " + error + ".");
+    }
+    return null;
+  }
+
+  private static String toString(InputStream inputStream) throws IOException {
+    try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream))) {
+      return bufferedReader.lines().collect(Collectors.joining(System.lineSeparator()));
+    }
+  }
+
+  private static IllegalStateException combineExceptionsIfNecessary(
+      IllegalStateException exitValueException, Exception e, String[] command) {
+    IllegalStateException exceptionWithMessage =
+        new IllegalStateException("Error related to the execution of " + Arrays.asList(command) + ".", e);
+    if (exitValueException == null) {
+      return exceptionWithMessage;
+    }
+    exitValueException.addSuppressed(exceptionWithMessage);
+    return exitValueException;
+  }
+}
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
@@ -7,8 +7,10 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 
 class CommandExecutor {
@@ -39,7 +41,7 @@ class CommandExecutor {
   static String executeWithoutException(String[] command, Logger startupLogger) {
     try {
       return execute(command);
-    } catch (Exception e) {
+    } catch (RuntimeException e) {
       startupLogger.error("Error when executing command " + Arrays.asList(command) + ".", e);
       if (e.getSuppressed().length == 1) {
         return e.getMessage() + " (Suppressed: " + e.getSuppressed()[0] + ")";
@@ -48,6 +50,7 @@ class CommandExecutor {
     }
   }
 
+  @Nullable
   private static IllegalStateException buildExitValueExceptionIfNecessary(
       String[] command, int exitValue, Process directivesClearProcess) throws IOException {
     if (exitValue != 0) {
@@ -60,7 +63,8 @@ class CommandExecutor {
   }
 
   private static String toString(InputStream inputStream) throws IOException {
-    try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream))) {
+    try (BufferedReader bufferedReader =
+        new BufferedReader(new InputStreamReader(inputStream, Charset.defaultCharset()))) {
       return bufferedReader.lines().collect(Collectors.joining(System.lineSeparator()));
     }
   }
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
@@ -72,7 +72,7 @@ class CommandExecutor {
   }
 
   private static IllegalStateException combineExceptionsIfNecessary(
-      IllegalStateException exitValueException, Exception e, List<String> command) {
+      @Nullable IllegalStateException exitValueException, Exception e, List<String> command) {
     IllegalStateException exceptionWithMessage =
         new IllegalStateException(
             "Error related to the execution of " + Arrays.asList(command) + ".", e);
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
@@ -9,6 +9,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.Charset;
 import java.util.Arrays;
+import java.util.List;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
@@ -17,20 +18,21 @@ class CommandExecutor {
 
   private CommandExecutor() {}
 
-  static String execute(String[] command) {
+  static String execute(ProcessBuilder processBuilder) {
     IllegalStateException exitException = null;
     String result = null;
     try {
-      Process process = new ProcessBuilder(command).start();
+      Process process = processBuilder.start();
       int exitValue = process.waitFor();
-      exitException = buildExitValueExceptionIfNecessary(command, exitValue, process);
+      exitException =
+          buildExitValueExceptionIfNecessary(processBuilder.command(), exitValue, process);
       if (exitException == null) {
         InputStream inputStream = process.getInputStream();
         result = toString(inputStream);
       }
       process.destroy();
     } catch (IllegalStateException | IOException | InterruptedException e) {
-      throw combineExceptionsIfNecessary(exitException, e, command);
+      throw combineExceptionsIfNecessary(exitException, e, processBuilder.command());
     }
     if (exitException != null) {
       throw exitException;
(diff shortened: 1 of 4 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+  static String execute(ProcessBuilder processBuilder) {
+    IllegalStateException exitException = null;
+    String result = null;
+    try {
+      Process process = processBuilder.start();
+      int exitValue = process.waitFor();
+      exitException =
+          buildExitValueExceptionIfNecessary(processBuilder.command(), exitValue, process);
+      if (exitException == null) {
+        InputStream inputStream = process.getInputStream();
```
trask, line 30: does this work, after the process has exited?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,182 @@
...
+  //    match: [
+  //      "*::*"
+  //    ],
+  //    c2: {
+  //      Exclude: true
+  //    }
+  //  }
+  // ]
+  private static final String APPPLICATIONINSIGHTS_DEBUG_JIT_C_2_DISABLING_ENABLED =
+      "appplicationinsights.debug.jit.c2-disabling.enabled";
```
trask, line 33: this option name is a bit confusing, what do you think of `applicationinsights.experiment.clear-compiler-directives-after-initialization`

(and differentiating `debug` meaning just logging, from `experiment` meaning an experimental option)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,182 @@
...
+  private void generateReport(DiagnosticsReport diagnosticsReport) {
+    if (!diagnosticsReport.isEmpty()) {
+      String diagExport = System.getProperty(APPPLICATIONINSIGHTS_DEBUG_DIAG_EXPORT);
+      if ("console".equals(diagExport) || "file-console".equals(diagExport)) {
+        startupLogger.info("Start-up diagnostics" + File.separator + diagnosticsReport);
+      }
+      // "file" (default value) / "console" / "file-console"
+      if (diagExport == null || "file".equals(diagExport) || "file-console".equals(diagExport)) {
+        saveIntoFile(diagnosticsReport);
+      }
```
trask, line 113: I'd suggest that we just write it to the normal application insights self-diagnostics stream (which is by default console and applicationinsights.log), since that output is already configurable (https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#self-diagnostics)

we could even enable all of these reports automatically when self-diagnostic level is `debug`, similar to how we dump all jmx metrics at that level.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,182 @@
...
+  private static final String APPPLICATIONINSIGHTS_DEBUG_JVM_FLAGS_ENABLED =
+      "appplicationinsights.debug.jvm-flags.enabled";
+
+  // Execute with -XX:NativeMemoryTracking=summary
+  private static final String APPPLICATIONINSIGHTS_DEBUG_NATIVE_MEM_TRACKING_ENABLED =
+      "appplicationinsights.debug.native-mem-tracking.enabled";
+
+  // "file" (default value) / "console" / "file-console"
+  public static final String APPPLICATIONINSIGHTS_DEBUG_DIAG_EXPORT =
+      "appplicationinsights.debug.diag-export";
```
trask, line 44: see comment below, maybe we don't need any of these, and can enable them all automatically at self-diagnostic `debug` level?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
jeanbisutti: Yes, in my tests the diagnostic info (_RSS_, _native memory tracking_, ...) were successfully retrieved from this `InputStream`.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
jeanbisutti: Agree, renaming done

jeanbisutti: I have also put the code related to the JVM compiler in a specific class (single responsibility principle).

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
jeanbisutti: @trask I reply to your two comments about self-diagnostic and the properties added in this PR.

The native memory tracking requires the addition of the `-XX:NativeMemoryTracking=summary` JVM option. I would prefer that the current self diagnostics features remains as simple as possible to configure for our users.

I would like to be able to execute only one type of memory diag (rss, native mem tracking, and perhaps one day heap allocation) to avoid that a measure comes to impact the result of another one. I would also like to avoid that an excessive logging from the self-diagnostic feature comes to pollute the memory measurements. 

We can have different levels of diagnostics. First, the self diagnostics, simple to use. At a second level, the start up profiler and these memory diagnostics at start-up.

For all these reasons, I would prefer to separate the features of this PR from the current self diagnostic.

I can propose to create another PR to add the rss measurement to the current self-diagnostic, to have an idea of the value. _Could you please give me a pointer where to put the code?_ 

> similar to how we dump all jmx metrics at that level

 I did not know that. _Could you please give me a pointer in the code?_ I have removed the retrieval of the JVM flags from this PR. If the current jmx metrics do not contain them, I will add them in another PR. I will also add the classpath content if it is not already retrieved.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private void generateReport(DiagnosticsReport diagnosticsReport) {
+    if (!diagnosticsReport.isEmpty()) {
+      String diagExport = System.getProperty(APPLICATIONINSIGHTS_DEBUG_DIAG_EXPORT);
+      if (diagExport == null || "console".equals(diagExport) || "file-console".equals(diagExport)) {
+        startupLogger.info("Start-up diagnostics" + File.separator + diagnosticsReport);
+      }
+      if ("file".equals(diagExport) || "file-console".equals(diagExport)) {
+        saveIntoFile(diagnosticsReport);
+      }
```
trask, line 83: what do you think about something like this, to reduce number of configuration values?
```suggestion
      boolean exportToFile = Boolean.getBoolean(APPLICATIONINSIGHTS_DEBUG_DIAG_EXPORT_TO_FILE);
      startupLogger.info("Start-up diagnostics" + File.separator + diagnosticsReport);
      if (exportToFile) {
        saveIntoFile(diagnosticsReport);
      }
```

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,134 @@
...
+      if (exportToFile) {
+        saveIntoFile(diagnosticsReport);
+      }
+    }
+  }
+
+  private String findResidentSetSize() {
+    try (Stream<String> lines = Files.lines(Paths.get("/proc/self/status"))) {
+      Optional<String> optionalRss = lines.filter(line -> line.startsWith("VmRSS")).findAny();
+      return optionalRss.orElse("");
```
trask, line 87: I don't find the named var helps with readability here, but ok to keep if you do
```suggestion
      return lines.filter(line -> line.startsWith("VmRSS")).findAny().orElse("");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static String toString(InputStream inputStream) throws IOException {
+    try (BufferedReader bufferedReader =
+        new BufferedReader(new InputStreamReader(inputStream, Charset.defaultCharset()))) {
+      return bufferedReader.lines().collect(Collectors.joining(System.lineSeparator()));
+    }
+  }
+
+  private static IllegalStateException combineExceptionsIfNecessary(
+      IllegalStateException exitValueException, Exception e, List<String> command) {
```
trask, line 75:
```suggestion
      @Nullable IllegalStateException exitValueException, Exception e, List<String> command) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/CommandExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  static String executeWithoutException(ProcessBuilder processBuilder, Logger startupLogger) {
+    try {
+      return execute(processBuilder);
+    } catch (RuntimeException e) {
+      startupLogger.error("Error when executing command " + processBuilder.command() + ".", e);
+      if (e.getSuppressed().length == 1) {
+        return e.getMessage() + " (Suppressed: " + e.getSuppressed()[0] + ")";
+      }
```
trask, line 50: The suppressed message doesn't seem like it would be very interesting typically, and this would allow still getting the full stack trace along with suppressed exception if really needed at debug level.
```suggestion
      startupLogger.debug(e.getMessage(), e);
```

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
```diff
@@ -0,0 +1,46 @@
...
+  //      "*::*"
+  //    ],
+  //    c2: {
+  //      Exclude: true
+  //    }
+  //  }
+  // ]
+  private static final String
+      APPLICATIONINSIGHTS_EXPERIMENT_CLEAR_COMPILER_DIRECTIVES_AFTER_INITIALIZATION =
+          "applicationinsights.experiment.clear-compiler-directives-after-initialization";
```
heyams, line 27:
```suggestion
          "applicationinsights.experiment.clear_compiler_directives_after_initialization";
```

heyams: should we follow the same naming convention? i have seen lost of underline instead of dash. See AiSemanticAttributes.java

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,134 @@
...
+import org.slf4j.Logger;
+
+class StartupDiagnostics {
+
+  public static final String APPLICATIONINSIGHTS_DEBUG_RSS_ENABLED =
+      "applicationinsights.debug.rss.enabled";
+
+  // Execute with -XX:NativeMemoryTracking=summary
+  private static final String APPLICATIONINSIGHTS_DEBUG_NATIVE_MEM_TRACKING_ENABLED =
+      "applicationinsights.debug.native-mem-tracking.enabled";
```
heyams, line 24: same here

heyams:
```suggestion
      "applicationinsights.debug.native_mem_tracking_enabled";
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private JvmCompiler() {}
+
+  static boolean hasToDisableJvmCompilerDirectives() {
+    return Boolean.getBoolean(
+        APPLICATIONINSIGHTS_EXPERIMENT_CLEAR_COMPILER_DIRECTIVES_AFTER_INITIALIZATION);
```
heyams, line 33: such a long system property name, should we consider a shorter name?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the MIT License.
+
+package com.microsoft.applicationinsights.agent.internal.init;
+
+import com.microsoft.applicationinsights.agent.bootstrap.diagnostics.PidFinder;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+class JvmCompiler {
+
+  // To disable C2 compiler during Application Insights set-up, for Java >= 9, do the following
```
heyams, line 11: can you add a note how user can use `APPLICATIONINSIGHTS_EXPERIMENT_CLEAR_COMPILER_DIRECTIVES_AFTER_INITIALIZATION`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
trask: this is a system property name, and not a semantic attribute name

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
trask: this is a hard one to explain succinctly 😅

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/JvmCompiler.java
jeanbisutti: @heyamsthis It's an experimental feature that we would like to test with some users. Following the feedbacks, we could properly document this feature.

jeanbisutti: @heyams It is necessary to add the JVM option explained from line 11 and set the `applicationinsights.experiment.clear-compiler-directives-after-initialization` property to `true`.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public static final String APPLICATIONINSIGHTS_DEBUG_DIAG_EXPORT_TO_FILE =
+      "applicationinsights.debug.diag-export-to-file";
```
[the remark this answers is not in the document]
jeanbisutti: Done

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StartupDiagnostics.java
```diff
@@ -0,0 +1,134 @@
...
+      File tempDir = optionalTempDir.get();
+      File diagFile = new File(tempDir, "diagnostics.txt");
+      write(diagnosticsReport, diagFile);
+    }
+  }
+
+  private void write(DiagnosticsReport diagnosticsReport, File diagFile) {
+    byte[] diagReportAsBytes = diagnosticsReport.toString().getBytes(StandardCharsets.UTF_8);
+    try {
+      Files.write(diagFile.toPath(), diagReportAsBytes);
```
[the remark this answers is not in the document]
jeanbisutti: @trask `Files.writeString` is not availlable in Java 8: https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html

PR #2503, opened 2022-09-07
Build changes: Add an Automatic Module Name to Runtime Attach library

KrogerWalt:
This is the minimum amount of work needed so that the Runtime Attach library can be imported into modular java projects.
This should not cause any problems for anyone using Java 11 or higher, as it just inserts a line into the Jar Manifest. The build chain required Java 11, but targetCompatibility and sourceCompatibility were still set to 1.8. To get rid of multiple warnings about reflection, the targetCompatibility and sourceCompatibility had to be increased to 11. If the warnings are inconsequential, then changes to the Java Conventions file can be omitted.


jeanbisutti:
Hi @KrogerWalt,

Thank you for having reported this issue and many thanks for the fix proposal!

We have users with Java 8 code. So, the source code has to remain on 8 version.

What about a `module-info.java` file opening Application Insights package, see for example [this paper](https://www.baeldung.com/java-illegal-reflective-access).

KrogerWalt:
Ok, thanks for letting me know.

trask:
hi @KrogerWalt! I'm re-opening this, I think adding `Automatic-Module-Name` to the manifest is a probably good idea.

was there a reason you needed to bump the groovy test compilation target from 1.8 to 11?

KrogerWalt:
@trask I was trying several different things to get rid of some errors and warnings. At one point it was warning that something was 1.8 and everything else was 11 so I changed all the 1.8s.  If it will compile with the those set to 1.8 and still insert the Automatic-Module-Name I think it should work.  Let me do some testing and I'll let you know if those changes are even necessary.

trask:
(oops, forgot to actually hit the re-open button 😅)

> Let me do some testing and I'll let you know if those changes are even necessary.

great, let me know, it's been on my todo list to convert the few groovy tests we have over to Java anyways if that becomes a blocker

KrogerWalt:
@trask  Hey, it still works when I revert that. It complains a lot, but it still builds in my app and sends data to AppInsights.
Sorry I missed that. I was playing  error whack-a-mole for so long when I got something that worked I quit while I was ahead.  ;-)

I've updated the main branch so you can merge in just the manifest change and the documentation change.  Thanks for helping me get this sorted out!

agent/runtime-attach/build.gradle.kts
```diff
@@ -20,5 +20,8 @@ tasks {
     from({
       agent.singleFile
     })
+    manifest {
+      attributes("Automatic-Module-Name" to "applicationinsights.runtime.attach")
```
trask, line 24: I believe matching the package name is a good convention
```suggestion
      attributes("Automatic-Module-Name" to "com.microsoft.applicationinsights.attach")
```

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ Migration notes:
 
 Enhancements:
 
+* Automatic module name added to Runtime Attach library to support Modular Java.
```
trask, line 40: can you create a new section at the top for 3.4.0-BETA.2, and move this there?

agent/runtime-attach/build.gradle.kts
KrogerWalt: I used the name that the automatic module naming used, for maximum compatibility.  If we change this, everyone (including us) who has previously set it up with Auto-naming will have to change their code.  I would prefer not to change this to something else as it would cause additional work for everyone using Modular Java who was able to get Automatic Modules to work with their project.

CHANGELOG.md
KrogerWalt: done

agent/runtime-attach/build.gradle.kts
KrogerWalt: I tested this change, and while Gradle build still works and it still runs, IntelliJ reports that it can't find the module when I make the change above.  So while we could still use it if we match the package name, it's more convenient to use the one that Automatic Module Naming picks if it is working in that project so that we can compile in both IntelliJ and Gradle.  

If you still want to match the package name, we should add a note to the Changelog mentioning the difference. Let me know what you decide.  Thanks for the help on this.

trask: thx @KrogerWalt! I'm afraid I don't have a good grasp on Java Modules, would you mind posting a Java Modules test project that uses runtime attach, that I can use to better understand the naming issue related to Intellij?

KrogerWalt: Renaming issue turned out to be a caching bug in the IntelliJ version I was using.  Updating to latest version fixed that.  This was the last hurdle,  so I will accept the change suggestion and add a line to the Changelog.  Thanks!

KrogerWalt:
Ok @trask  the Readme is updated and everything should be ready to go. Thanks again!

trask (approved):
nice, thx @KrogerWalt!

KrogerWalt:
@trask pipeline failed because of a space at the end of a line in the Changelog.  I ran the command it indicated to fix it. Please reapprove.  Thanks!

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+Migration notes:
+
+* If you were using Automatic naming before, you will need to update your module-info.java file.
+  Change ```requires applicationinsights.runtime.attach;``` to
+  ```requires com.microsoft.applicationinsights.runtime.attach;``` and everything should work.
+
+Enhancements:
+
+* Automatic module name enttry added to Runtime Attach library Jar to support Modular Java.
```
heyams, line 13: sorry for the late comment, just fixed a typo https://github.com/microsoft/ApplicationInsights-Java/pull/2516

PR #2505, opened 2022-09-08
Convert all groovy tests to java

trask:


heyams:
👍 was it a manual process or inteillj will automatically convert it for you? just curious.

trask:
Intellij has a feature that will convert Groovy to Java, so I used that, but it probably only did about half the work, but definitely helped

PR #2509, opened 2022-09-09
Update to otel 1.18.0

trask:
Built on #2513


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java
```diff
@@ -410,16 +410,28 @@ private static String getTargetOrNull(Attributes attributes, int defaultPort) {
     if (peerService != null) {
       return peerService;
     }
-    String netPeerName = attributes.get(SemanticAttributes.NET_PEER_NAME);
-    if (netPeerName == null) {
-      return null;
+    String host = attributes.get(SemanticAttributes.NET_PEER_NAME);
+    if (host != null) {
+      Long port = attributes.get(SemanticAttributes.NET_PEER_PORT);
+      if (port != null && port != defaultPort) {
```
jeanbisutti, line 416: Code blocks between lines 416/420 and 428/432 seem similar. What about an extract method? So the getTargetOrNull method could be shorter and faster to read?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -93,6 +94,8 @@ public void init() {
 
       StartupDiagnostics.execute();
 
+      InternalLogger.initialize(Slf4jInternalLogger::create);
```
heyams, line 97: what is this for?

trask: Upstream logging changed significantly in this last release: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6543

heyams (approved):

PR #2517, opened 2022-09-13
Add classpath and input arguments to the self-diagnostics

jeanbisutti:
Add the following info to the self-diagnostics:
* Class path
* Input arguments (what is added by the users: JVM options, properties, ...)


trask (approved):
👍

PR #2534, opened 2022-09-23
Fix missing etw dll in the final binary

heyams:


settings.gradle
```diff
@@ -34,7 +34,7 @@ gradleEnterprise {
 
 rootProject.name = 'applicationinsights-java'
 
-def buildNative = System.getenv("CI") != null && Os.isFamily(Os.FAMILY_WINDOWS)
+def buildNative = project.hasProperty("ai.etw.native.build") != null && Os.isFamily(Os.FAMILY_WINDOWS)
```
trask, line 37: keep

settings.gradle
```diff
@@ -1,5 +1,3 @@
-import org.apache.tools.ant.taskdefs.condition.Os
-
```
trask, line 2: not sure, but you may still need this

trask (approved):

settings.gradle
heyams: it was not needed previously. local build is fine without it.

PR #2535, opened 2022-09-23
Rename telemetryKind to telemetryType

trask:


heyams:
is there reason for the renaming? SpanKind, TelemetryKind kind of rhyme with each other.

trask:
it's more standard application insights terminology: https://learn.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.datacontracts.samplingtelemetryitemtypes?view=azure-dotnet, and also more similar to kusto "itemType"

heyams:
yup, i checked node.js, they're using TelemetryType too.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -622,7 +622,10 @@ public static class SamplingOverride {
 
     // TODO (trask) make this required when moving out of preview
     //   for now the default is both "request" and "dependency" for backwards compatibility
-    @Nullable public SamplingTelemetryKind telemetryKind;
+    @Nullable public SamplingTelemetryType telemetryType;
+
+    // this config option existed in one GA release (3.4.0), and was then replaced by telemetryType
+    @Deprecated @Nullable public SamplingTelemetryType telemetryKind;
```
heyams, line 628:
```suggestion
    @Deprecated @Nullable public SamplingTelemetryType telemetryType;
```

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: this is the old (not deprecated) name, so that we don't break anyone who is already using it

PR #2565, opened 2022-09-28
Remove direct bytebuddy dependency

trask:


heyams (approved):

agent/runtime-attach/gradle.lockfile
```diff
@@ -8,7 +8,7 @@ io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:1.18.0-
 io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom:1.18.0=runtimeClasspath
 io.opentelemetry:opentelemetry-bom-alpha:1.18.0-alpha=runtimeClasspath
 io.opentelemetry:opentelemetry-bom:1.18.0=runtimeClasspath
-net.bytebuddy:byte-buddy-agent:1.12.10=runtimeClasspath
+net.bytebuddy:byte-buddy-agent:1.11.18=runtimeClasspath
```
heyams, line 11: is this correct? should this line be gone too?

trask: this is correct, this dependency (and version) is coming from the upstream artifact: `io.opentelemetry.contrib:opentelemetry-runtime-attach-core:1.18.0-alpha`

PR #2573, opened 2022-09-28
Update errorprone

trask:


buildSrc/src/main/kotlin/ai.errorprone-conventions.gradle.kts
```diff
@@ -59,6 +59,9 @@ tasks {
         disable("UnnecessaryDefaultInEnumSwitch")
         disable("InconsistentOverloads")
 
+        // consider enabling this after removing azure-monitor-exporter from this repo
+        disable("CanIgnoreReturnValueSuggester")
```
heyams, line 63: what is the point of using this anyway? it will force us to use @CheckReturnValue annotation?

heyams: or @SuppressWarnings("CanIgnoreReturnValueSuggester") if enable everywhere.

trask: see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6646#discussion_r978149555

heyams (approved):

PR #2574, opened 2022-09-28
Remove unused dependency

trask:


trask:
> does it need to update lockfiles?

it was only in dependencyManagement (version management), was never used as an actual dependency

PR #2581, opened 2022-10-03
Ability to set the configuration file from the classpath of runtime attachment

### LINKED ISSUE #2576: Appinsights logging for different environments
Azure app insights is working in the way shown below.

Added these two dependencies:
1. implementation("com.microsoft.azure:applicationinsights-agent:3.4.1")
2. implementation("com.microsoft.azure:applicationinsights-runtime-attach:3.4.1")

Invoked the attach() method as mentioned in docs
@SpringBootApplication
public class SpringBootApp {

  public static void main(String[] args) {
    ApplicationInsights.attach();
    SpringApplication.run(SpringBootApp.class, args);
  }
}

I have created applicationinsight.json, added my connection string and placed in src/main/resources.

However, my need is to log local and dev logs to one connection string (resource group) and prod logs to another connection string (resource group).
I have tried creating a new file applicationinsight-local.json, applicationinsight-dev.json in same path. But, that doesnt seem to work.
Dont you offer this functionality.

Thanks.

jeanbisutti:
Fix #2576

### DIFF
--- a/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
+++ b/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
@@ -18,7 +18,7 @@ import java.util.stream.Collectors;
 public final class ApplicationInsights {
 
   private static final String APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE =
-      "applicationinsights.runtimeattach.configuration.file";
+      "applicationinsights.runtime-attach.configuration.file";
 
   private static final Logger logger = Logger.getLogger(ApplicationInsights.class.getName());
 
--- a/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
+++ b/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
@@ -18,7 +18,7 @@ import java.util.stream.Collectors;
 public final class ApplicationInsights {
 
   public static final String APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE =
-      "applicationinsights.runtime-attach.configuration.file";
+      "applicationinsights.runtime-attach.classpath.file";
 
   private static final Logger logger = Logger.getLogger(ApplicationInsights.class.getName());
 
--- a/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
+++ b/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
@@ -39,7 +39,7 @@ public final class ApplicationInsights {
    * requested at the beginning of the main method.
    *
    * @throws ConfigurationException If the file given by the
-   *     applicationinsights.internal.runtime.attached.json property was not found
+   *     applicationinsights.runtime-attach.configuration.classpath.file property was not found
    */
   public static void attach() {
 
--- a/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
+++ b/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
@@ -60,20 +60,35 @@ public final class ApplicationInsights {
   private static Optional<String> findJsonConfig() {
 
     String fileName = findJsonConfigFile();
-    InputStream configContentAsInputStream =
-        ApplicationInsights.class.getResourceAsStream("/" + fileName);
+
+    InputStream configContentAsInputStream = findResourceAsStream(fileName);
+
     if (configContentAsInputStream == null) {
-      throw new ConfigurationException(fileName + " not found");
+      return Optional.empty();
     }
+
+    String json = findJson(configContentAsInputStream);
+    return Optional.of(json);
+  }
+
+  private static String findJson(InputStream configContentAsInputStream) {
     try (InputStreamReader inputStreamReader =
             new InputStreamReader(configContentAsInputStream, StandardCharsets.UTF_8);
         BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {
-      String json = bufferedReader.lines().collect(Collectors.joining(""));
-      return Optional.of(json);
+      return bufferedReader.lines().collect(Collectors.joining(""));
     } catch (IOException e) {
       throw new IllegalStateException(
           "Unexpected issue during loading of JSON configuration file: " + e.getMessage());
     }
+  }
+
+  private static InputStream findResourceAsStream(String fileName) {
+    InputStream configContentAsInputStream =
+        ApplicationInsights.class.getResourceAsStream("/" + fileName);
+    if (configContentAsInputStream == null && isJsonFileConfiguredWithProperty()) {
+      throw new ConfigurationException(fileName + " not found");
+    }
+    return configContentAsInputStream;
   }
 
   public static class ConfigurationException extends IllegalArgumentException {
@@ -83,12 +98,16 @@ public final class ApplicationInsights {
   }
 
   private static String findJsonConfigFile() {
-    String fileFromProperty =
-        System.getProperty(APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE);
-    if (fileFromProperty != null) {
-      return fileFromProperty;
+    if (isJsonFileConfiguredWithProperty()) {
+      return System.getProperty(APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE);
     }
     return "applicationinsights.json";
+  }
+
+  private static boolean isJsonFileConfiguredWithProperty() {
+    String fileFromProperty =
+        System.getProperty(APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE);
+    return fileFromProperty != null;
   }
 
   private static boolean agentIsAttached() {
--- a/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
+++ b/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
@@ -90,7 +90,7 @@ public final class ApplicationInsights {
     InputStream configContentAsInputStream =
         ApplicationInsights.class.getResourceAsStream("/" + fileName);
     if (configContentAsInputStream == null && isJsonFileConfiguredWithProperty()) {
-      throw new ConfigurationException(fileName + " not found");
+      throw new ConfigurationException(fileName + " not found on the class path");
     }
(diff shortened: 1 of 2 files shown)


agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -17,6 +17,9 @@
 /** This class allows you to attach the Application Insights agent for Java at runtime. */
 public final class ApplicationInsights {
 
+  public static final String APPLICATIONINSIGHTS_RUNTIMEATTACH_CONFIGURATION_FILE =
```
heyams, line 20:
```suggestion
  public static final String APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE =
```

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -17,6 +17,9 @@
...
+      "applicationinsights.runtimeattach.configuration.file";
```
heyams, line 21:
```suggestion
      "applicationinsights.runtime.attach.configuration.file";
```

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -69,6 +73,15 @@ private static Optional<String> findJsonConfig() {
     }
   }
 
+  private static String findJsonConfigFile() {
+    String fileFromProperty =
+        System.getProperty(APPLICATIONINSIGHTS_RUNTIMEATTACH_CONFIGURATION_FILE);
+    if (fileFromProperty != null) {
```
heyams, line 79: i think we should check if runtime attach is used here?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -69,6 +73,15 @@ private static Optional<String> findJsonConfig() {
     }
   }
 
+  private static String findJsonConfigFile() {
+    String fileFromProperty =
+        System.getProperty(APPLICATIONINSIGHTS_RUNTIMEATTACH_CONFIGURATION_FILE);
```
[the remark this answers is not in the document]
heyams: or support both?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
trask: I was thinking "runtimeattach" one word since it represents one concept, and we would have other `applicationinsights.runtime.*` properties

However, separating the words does match what we've done with the artifact name `applicationinsights-runtime-attach`.

In OpenTelemetry Instrumentation, we'd probably use `applicationinsights.runtime-attach.configuration.file`, wdyt of that?

heyams: runtime-attach works for me too.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: Done (IntelliJ renaming)

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: Change done with `applicationinsights.runtime-attach.configuration.file`

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: `RUNTIME_ATTACHED_ENABLED_PROPERTY` is always true at this point. This java agent uses this property to detect the runtime attachment and set the appropriate sdk prefix.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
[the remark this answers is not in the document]
jeanbisutti: The configuration of a spring profile with a system property (`-Dspring.profiles.active=dev`) may be more common than the configuration with an environment variable (`export spring_profiles_active=dev`).  We may start with a system property, and following user feedbacks, we could add an env var. The system property would allow the ability to configure applications (JVM) in different ways in the same environment, which would not be possible with an env var. I prefer to wait for user feedback and don't multiply the ways of configuring Application Insights.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -53,8 +56,9 @@ public static void attach() {
 
   private static Optional<String> findJsonConfig() {
 
+    String fileName = findJsonConfigFile();
     InputStream configContentAsInputStream =
-        ApplicationInsights.class.getResourceAsStream("/applicationinsights.json");
+        ApplicationInsights.class.getResourceAsStream("/" + fileName);
```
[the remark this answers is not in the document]
jeanbisutti: We could, but we would have to add the following kind of code to the current test:

```java
  public static void main(String[] args) {
    System.setProperty("applicationinsights.runtime-attach.configuration.file", "applicationinsights-test.json");
    ApplicationInsights.attach();
    SpringApplication.run(SpringBootApp.class, args);
  }
```

The users would instead add a system property with `-Dapplicationinsights.runtime-attach.configuration.file=applicationinsights-test.json`. So, the test would not represent the typical way of configuring the file name. And so, the test would not document the standard way of configuring the file name.

Another option could be to rewrite a test to be able to add `-Dapplicationinsights.runtime-attach.configuration.file=applicationinsights-test.json`. However, this would require a significant time for an added code pretty simple.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
heyams: yup, i didn't see line 44 earlier

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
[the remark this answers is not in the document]
heyams: maybe add a todo for now? it's not required for this PR.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -17,6 +17,9 @@
 /** This class allows you to attach the Application Insights agent for Java at runtime. */
 public final class ApplicationInsights {
 
+  public static final String APPLICATIONINSIGHTS_RUNTIME_ATTACH_CONFIGURATION_FILE =
+      "applicationinsights.runtime-attach.configuration.file";
```
heyams, line 21: can we make this work for a file path as well?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
heyams: we might need to check if such a file exists?

jeanbisutti: This PR was created because a user needs a connection string different for each spring profile/environment (see [#2576](https://github.com/microsoft/ApplicationInsights-Java/issues/2576)), with runtime attachment. If the file is not found then, Application Insights will fail to startup with a message saying that a connection string is missing. The user should easily understand that the file containing the connection string was not found?

trask: I think a message that the resource was not found could be helpful, we do similar for json configuration file:

https://github.com/microsoft/ApplicationInsights-Java/blob/568a2373dc479181b214a7278aeb1dabe35e5a32/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java#L455

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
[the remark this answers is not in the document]
trask: is there some other form of test coverage that may be simpler (but still useful)?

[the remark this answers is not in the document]
jeanbisutti: An approach could be to change the visibility of some methods to be able to test them. However, I don't recommend doing this. The major drawback of this approach is that we are testing implementation details, not a user use case. So, since the test is tied to implementation details, refactoring (changing the code without changing the behavior) could break it (because we remove a method, for example), adding a cost to test maintenance and slowing down and discouraging future refactoring.

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
jeanbisutti: Added since it seems the team preference

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
heyams: are we going with "applicationinsights.runtime-attach.configuration.classpath.file"? can we add a comment to say this file must be in the classpath, otherwise it wouldn't work?

agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -53,22 +63,55 @@ public static void attach() {
...
       throw new IllegalStateException(
           "Unexpected issue during loading of JSON configuration file: " + e.getMessage());
     }
   }
 
+  private static InputStream findResourceAsStream(String fileName) {
+    InputStream configContentAsInputStream =
+        ApplicationInsights.class.getResourceAsStream("/" + fileName);
+    if (configContentAsInputStream == null && isJsonFileConfiguredWithProperty()) {
+      throw new ConfigurationException(fileName + " not found");
```
trask, line 93:
```suggestion
      throw new ConfigurationException(fileName + " not found on the class path");
```

heyams (approved):

PR #2583, opened 2022-10-06
Extract RequestChecker class

jeanbisutti:
[`SpanDataMapper` class](https://github.com/microsoft/ApplicationInsights-Java/blob/main/agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java) contains almost 1000 lines.

This class may have too many responsibilities.

 The check of the request case does not seem to be the responsibility of the `SpanDataMapper` class. So, this PR has extracted several public methods of SpandDataMapper to a `RequestChecker` class.

### DIFF
--- a/agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/RequestChecker.java
+++ b/agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/RequestChecker.java
@@ -9,14 +9,14 @@
 
 package com.azure.monitor.opentelemetry.exporter.implementation;
 
+import static com.azure.monitor.opentelemetry.exporter.implementation.AiSemanticAttributes.JOB_SYSTEM;
+
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.sdk.trace.ReadableSpan;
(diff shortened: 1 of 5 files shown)


trask (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/RequestChecker.java
```diff
@@ -0,0 +1,49 @@
...
+
+public final class RequestChecker {
+
+  private RequestChecker() {}
+
+  public static boolean isRequest(SpanData span) {
+    return isRequest(span.getKind(), span.getParentSpanContext(), span.getAttributes()::get);
+  }
+
+  public static boolean isRequest(ReadableSpan span) {
```
ramthi, line 29: Not related specifically to this PR.  I just noticed there aren't any documentation comments for methods.  Wondering why!

jeanbisutti: My understanding is that we document the methods that our users may use. This one is not susceptible to being used by our users. An example [here](https://github.com/microsoft/ApplicationInsights-Java/blob/568a2373dc479181b214a7278aeb1dabe35e5a32/agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java#L34) of a method that our users can use.

PR #2592, opened 2022-10-13
Fix OFF logging threshold

### LINKED ISSUE #2590: Disabling instrumentation of logging no longer works
We disabled instrumentation of logging (log4j2) by
export APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL=OFF
Disabling worked in version 3.2.4. After upgrading to version 3.3.0, disabling no longer works and the agent transfers all (severity level from 0 to 3) log messages to azure.
The agent in version 3.4.1 behaves the same way.

### Expected behavior
Log messages are not transmitted to azure, when environment variable
APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL is set to "OFF".
If the value is set to "FATAL", the agent only transmits log messages with serverity level 4 (FATAL). 

### Actual behavior
When starting the application, the agent logs the message
DEBUG c.m.a.a.i.c.ConfigurationBuilder - read environment variable: APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL=OFF
All log messages (log severity from 0 to at least 3) are transmitted to azure.
If the value is set to "FATAL", the agent transmits log messages with serverity level 3 (ERROR). 

### To Reproduce
-

#### Sample Application
-

### System information
 - SDK Version: 3.3.0, 3.4.1
 - OS type and version: Linux
 - Application Server type and version (if applicable): Tomcat
 - Using spring-boot? No
 - Additional relevant libraries (with version, if applicable): log4j-2-17.0

### Logs
-

### Screenshots
-

trask:
Resolves #2590

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
@@ -52,7 +52,7 @@ public class Configuration {
   }
 
   public void validate() {
-    instrumentation.logging.getSeverity();
+    instrumentation.logging.getSeverityThreshold();
     preview.validate();
   }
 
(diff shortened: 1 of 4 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -220,32 +220,33 @@ public static class DatabaseMaskingConfiguration {
...
     public String level = "INFO";
 
-    public Severity getSeverity() {
-      return getSeverity(level);
+    public int getSeverityThreshold() {
+      return getSeverityThreshold(level);
     }
 
-    public static Severity getSeverity(String level) {
+    public static int getSeverityThreshold(String level) {
```
jeanbisutti, line 227: _(suggestion)_ This method might be moved in `AgentLogExporter` and called around line 84. So all the log logic would be in the place where we need it. A `boolean exportLog(LogData log, String configuredLogThreshold)` method might then be added and contains all the log level logic for the export. The configuration object would then represent the log level as a string an not an int.

heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
trask: this looks a bit more challenging, as we are also using this method for fail-fast validation of the configuration:

https://github.com/microsoft/ApplicationInsights-Java/blob/560c7d58ff29aebcfffa5d88f970edfe0de0a91b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java#L54-L55

PR #2634, opened 2022-10-26
More dependabot and GitHub actions cleanup

trask:
A few things:

* Split out spotless into a separate CI job so that it's easier to see when that's the only failure
* Better organization/reuse of github actions (mirroring https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6975)
* Apply dependabot to `settings.gradle.kts` and `build.gradle.kts` in a rather roundabout way, see README.md for details (port of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6977)
* Add gradle wrapper validation to common CI checks


.github/project-root-duplicates/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
+plugins {
```
heyams, line 1: in the future, it will be better if we can automate the creation of the duplicates so that it doesn't require another 'out of sync' check. just a thought for future improvement.

heyams (approved):

PR #2647, opened 2022-10-28
Pre-sampled live metrics

heyams:


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/livemetrics/LiveMetricsSpanProcessor.java
```diff
@@ -30,7 +31,6 @@ public boolean isStartRequired() {
   @Override
   public void onEnd(ReadableSpan readableSpan) {
     if (quickPulse.isEnabled()) {
-      // TODO (trask) can we do anything better here in terms of double conversion?
```
heyams, line 33: not sure what this double conversion is referring to?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSampler.java
```diff
@@ -138,13 +146,23 @@ public String getDescription() {
   private static class RecordAndSampleWithItemCount implements SamplingResult {
 
     private final Attributes attributes;
+    @Nullable private final QuickPulse quickPulse;
 
-    RecordAndSampleWithItemCount(long itemCount) {
+    RecordAndSampleWithItemCount(long itemCount, @Nullable QuickPulse quickPulse) {
```
trask, line 151: instead of passing in `quickPulse` here, it may be better to use quickpulse to get the decision and pass the decision in here instead (that way this object is a nice simple immutable object)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -159,6 +159,8 @@ public void trackAsync(TelemetryItem telemetryItem) {
       throw new AssertionError("telemetry item is missing time");
     }
 
+    // TODO (heya) once we migrate performance counters and jvm metrics to OpenTelemetry, we can
+    // remove this
     if (quickPulse != null) {
       quickPulse.add(telemetryItem);
     }
```
trask, line 166: I don't think this code is needed, because we don't get quickpulse metrics from trackMetrics anyways (since those are only once a minute)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/livemetrics/LiveMetricsLogProcessor.java
```diff
@@ -0,0 +1,31 @@
...
+    this.mapper = mapper;
+    this.quickPulse = quickPulse;
+  }
+
+  @Override
+  public void onEmit(ReadWriteLogRecord logRecord) {
+    LogRecordData log = (LogRecordData) logRecord;
+    String stack = log.getAttributes().get(SemanticAttributes.EXCEPTION_STACKTRACE);
+    if (quickPulse.isEnabled()) {
+      quickPulse.add(mapper.map(logRecord.toLogRecordData(), stack, null));
```
trask, line 28: I think we want to pass itemCount here (I think can get from logRecord AiSemanticAttributes.ITEM_COUNT)

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/livemetrics/LiveMetricsSpanProcessor.java
trask: I couldn't figure out either 😂 , I agree with deleting this comment 👍

PR #2678, opened 2022-11-09
Update release action

trask:


.github/workflows/release.yml
```diff
@@ -37,14 +37,56 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          if [[ $VERSION != *-BETA* ]]
+          if [[ $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
           then
-            GA=" (GA)"
+            opt_ga=" (GA)"
+          else
+            opt_prerelease="--prerelease"
```
heyams, line 44: 👍

.github/scripts/update-version.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+
+sed -i "s/version=.*/version=$1/" gradle.properties
```
heyams, line 3: can we add a comment saying this line will replace `version=.*` with `version=$1` where $1 is the input parameter? 
nit: maybe print error when $1 is not provided? 

i am thinking we can run update-version.sh independently not just used by release.yml.

heyams (approved):

.github/scripts/update-version.sh
heyams: i ran it without $1, it will result to "version=".

heyams: in that case subsequent "s/version=.*" search will fail

PR #2690, opened 2022-11-12
Reformat Release Draft

heyams:


CHANGELOG.md
```diff
@@ -2,19 +2,19 @@
 
 ## Version 3.4.4 GA
 
-Enhancements:
+# Enhancements:
```
trask, line 5: probably these should be `###` since they are sub-items of `## Version 3.4.4 GA`

trask (approved):

CHANGELOG.md
heyams: make sense.. # in the draft release probably is equivalent to ###

PR #2717, opened 2022-11-18
Display only the import certificate message for the GET endpoint

jeanbisutti:
GET endpoint:  ../api/profiles/../appId


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/logging/NetworkFriendlyExceptions.java
```diff
@@ -172,17 +172,20 @@ private static String getCustomJavaKeystorePath() {
     }
 
     private static String getSslFriendlyExceptionAction(String url) {
+      if (!url.contains("profiles")) { // ../api/profiles/../appId
```
heyams, line 175: api/profiles/../appid is used by profiler only now. 
are you sure we want to ignore ssl friendly message for eveything else?

trask (approved):

PR #2756, opened 2022-12-01
Add a diagnostic to check Application Insights root certificate

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AppInsightsCertificate.java
```diff
@@ -0,0 +1,121 @@
...
+      writer.flush();
+      writer.close();
+
+      InputStream inputStream = process.getInputStream();
+      result = toString(inputStream);
+
+      process.destroy();
+    } catch (Exception e) {
+      throw new IllegalStateException(
+          "Error related to the execution of " + processBuilder.command() + ".", e);
```
trask, line 110: is this catch needed, since also catching above in `executeWithoutException`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AppInsightsCertificate.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private static String executeWithoutException(ProcessBuilder processBuilder) {
+    try {
+      return execute(processBuilder);
+    } catch (RuntimeException e) {
+      String header = "Error when executing command " + processBuilder.command() + ": ";
+      if (e.getSuppressed().length == 1) {
+        return header + e.getMessage() + " (Suppressed: " + e.getSuppressed()[0] + ")";
+      }
+      return header + e.getMessage();
```
trask, line 89: it doesn't look like this return value is displayed anywhere, what about?

```suggestion
      logger.error(e.getMessage(), e);
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AppInsightsCertificate.java
```diff
@@ -0,0 +1,114 @@
...
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.nio.charset.Charset;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+
+class AppInsightsCertificate {
+
+  private static final String APP_INSIGHT_ROOT_CERTIFICATE =
```
trask, line 20: comment where this comes from?

trask, line 20:
```suggestion
  // DigiCert Global Root G2
  private static final String APP_INSIGHT_ROOT_CERTIFICATE =
```

trask (approved):

PR #2765, opened 2022-12-02
Prevent runtime attach from application failure

jeanbisutti:


agent/runtime-attach/src/main/java/com/microsoft/applicationinsights/attach/ApplicationInsights.java
```diff
@@ -50,15 +51,19 @@ public static void attach() {
...
+      }
 
-    String appInsightResourceName = findAppInsightResourceName();
-    CoreRuntimeAttach runtimeAttach = new CoreRuntimeAttach(appInsightResourceName);
+      String appInsightResourceName = findAppInsightResourceName();
+      CoreRuntimeAttach runtimeAttach = new CoreRuntimeAttach(appInsightResourceName);
 
-    runtimeAttach.attachJavaagentToCurrentJVM();
+      runtimeAttach.attachJavaagentToCurrentJVM();
+    } catch (Exception e) {
```
trask, line 64: to be safe
```suggestion
    } catch (Throwable t) {
```

trask (approved):

Issue #2774, opened 2022-12-06
Capture of InProc Telemetry Type not working properly


zoomingrocket:
### Expected behavior
We were currently using App Insight 3.3.1 GA and started evaluating 3.4.5 GA. Immediately after switching over to the new release, we started noticing a huge spike of InProc invocations getting captured as "Request" type.

We are using "captureControllerSpans" as true in config JSON & customInstrumentation

```
"preview": {
    "captureControllerSpans": true,
}
,
    "customInstrumentation": [
      {
        "className": "com.boomi.execution.ExecutionTask",
        "methodName": "call"
      }
```

3.3.1 GA Release behavior
![image](https://user-images.githubusercontent.com/2420449/206006871-d6f23a8b-f307-4af4-87de-5e21e0310814.png)

### Actual behavior
3.4.5 GA Release Behavior
![image](https://user-images.githubusercontent.com/2420449/206006937-2332bfc0-0379-447a-b40b-b8f8295cccab.png)

### To Reproduce
We are using a COTs Middleware system and this is noticed for all requests captured where inProc is now classified as "Request" type instead

#### Sample Application
N/A

### System information
Please provide the following information:
 - SDK Version: 3.4.5 GA
 - OS type and version: Cent OS 7.9
 - Application Server type and version (if applicable): N/A
 - Using spring-boot? N/A
 - Additional relevant libraries (with version, if applicable):  N/A

### Logs
Let me know what specific log we should capture which will be helpful in debugging this behavior

### Screenshots
Posted above

zoomingrocket:
So after further analysis, the primary difference with the latest GA release is any custom instrumentation method invocations are by default tagged with telemetry type as "Request" instead of earlier behavior where they were marked inProc when part of another parent invocation.

We have tried two combinations
1. Disable inProc and leave the customInstrumentation but it resulted in the same behavior where all invocations for custom instrumentation class were captured with Telemetry Type REQUEST
2. Keep InProc ON and remove custom instrumentation - This resulted in no inProc capture at all, on top of it we missed actual custom instrumentation requests starting which had no traceparent as just standalone trace capture.

So looking for suggestions to have the best of both worlds :) We are ok to disable InProc capture, but we want the custom instrumentation requests to be captured as standalone requests only when there is no traceparent or invocation not part of a parent request span. This is currently possible in 3.3.1 GA with inProc capture disabled and only having custom instrumentation enabled

trask:
[...]

PR #2776, opened 2022-12-06
Fix custom instrumentation type

trask:
See #2774


jeanbisutti (approved):

agent/instrumentation/methods/src/main/java/io/opentelemetry/javaagent/instrumentation/methods/ai/MethodSingletons.java
```diff
@@ -33,15 +35,25 @@ public final class MethodSingletons {
...
   }
 
+  // START APPLICATION INSIGHTS MODIFICATIONS
+  private static class MethodSpanKindExtractor implements SpanKindExtractor<ClassAndMethod> {
+
+    @Override
+    public SpanKind extract(ClassAndMethod classAndMethod) {
+      // we emit SERVER spans instead of INTERNAL spans when there is no parent, so that it works
+      // well with Application Insights' customInstrumentations feature
+      return Span.current().getSpanContext().isValid() ? SpanKind.INTERNAL : SpanKind.SERVER;
```
trask, line 53:
```suggestion
      return hasParent() ? SpanKind.INTERNAL : SpanKind.SERVER;
```

PR #2809, opened 2022-12-22
Runtime connection string

trask:


classic-sdk/core/src/main/java/com/microsoft/applicationinsights/connectionstring/ConnectionString.java
```diff
@@ -0,0 +1,27 @@
...
+   * <pre>
+   *   {
+   *     "startWithoutConnectionString": true
+   *   }
+   * </pre>
+   *
+   * <p>If the connection string has already been set then this method logs a warning and does If
+   * the connection string has already been set then this method logs a warning and does nothing.
+   */
+  public static void init(String connectionString) {
```
jeanbisutti, line 22: Rename into `configure`?

smoke-tests/apps/RuntimeAttachWithDelayedConnectionString/src/main/resources/applicationinsights.json
```diff
@@ -0,0 +1,10 @@
+{
+  "startWithoutConnectionString": true,
```
jeanbisutti, line 2: When the user uses the runtime attachment API, I feel it would be more natural and convenient to enable the "start without connection string" mode in a programmatic way.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/DynamicConfiguration.java
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.applicationinsights.agent.internal.init;
+
+import com.microsoft.applicationinsights.agent.internal.configuration.Configuration;
+import java.util.ArrayList;
+import java.util.List;
+
+public class DynamicConfiguration {
```
jeanbisutti, line 10: I would prefer something more explicit and from a user point of view, `ProgrammaticConfiguration`?

trask: `RuntimeConfiguration`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -33,6 +33,8 @@
 public class Configuration {
 
   public String connectionString;
+  public boolean startWithoutConnectionString =
```
heyams, line 36: is this used by azure function only?

heyams: no, it is default to function's value.

smoke-tests/apps/RuntimeAttachWithDelayedConnectionString/src/main/resources/applicationinsights.json
trask: planning to wait on adding API for now, will see user feedback / amount of usage

heyams (approved):

PR #2820, opened 2023-01-05
Add final to Configuration fields

jeanbisutti:


trask (commented):
my hesitation is that these fields are dynamically bound via jackson, which is probably ["ignoring"](https://stackoverflow.com/questions/3301635/change-private-static-final-field-using-java-reflection) the final declaration, in which case it seems more honest not to mark them `final`

PR #2837, opened 2023-01-11
Allow service loader util to detect services provided by an extension

johnoliver:
Fix # .

- [ ] CHANGELOG.md updated


trask (approved):
probably we should invest in at least one profiler `smoke-test` to catch this sort of issue

PR #2893, opened 2023-01-31
Support metircIntervalSeconds backward compatibility 

heyams:
#2875


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -304,6 +304,7 @@ public static class PreviewConfiguration {
     @Deprecated public boolean openTelemetryApiSupport;
     public PreviewInstrumentation instrumentation = new PreviewInstrumentation();
     // this is just here to detect if using this old setting in order to give a helpful message
```
trask, line 306:
```suggestion
    // these are just here to detect if using this old setting in order to give a helpful message
```

trask (approved):

trask:
can you add backcompat for `APPLICATIONINSIGHTS_PREVIEW_METRIC_INTERVAL_SECONDS` also?

heyams:
> can you add backcompat for `APPLICATIONINSIGHTS_PREVIEW_METRIC_INTERVAL_SECONDS` also?

added.

trask (approved):

PR #2900, opened 2023-02-03
snippet injection add smoke test

siyuniu-ms:
use the snippet-injection version otl and add dependency at application insights


siyuniu-ms:
Things that could be improved:
1. Enable user have two connection string, one for server, and the other for browser. This could help when authentication is required under some cases.
2. Check the connection string to be set before setting the snippet, and provide alternative plan if sth went wrong.

siyuniu-ms:
the github build failed because it could not get my local version of otl java.

siyuniu-ms:
add smoke test

settings.gradle.kts
```diff
@@ -125,6 +125,8 @@ hideFromDependabot(":smoke-tests:apps:RoleNameOverrides")
 hideFromDependabot(":smoke-tests:apps:RuntimeAttach")
 hideFromDependabot(":smoke-tests:apps:RuntimeAttachWithDelayedConnectionString")
 hideFromDependabot(":smoke-tests:apps:Sampling")
+hideFromDependabot(":smoke-tests:apps:SnippetInjection")
+hideFromDependabot(":smoke-tests:apps:SnippetInjection:MockExtension")
```
trask, line 129:
```suggestion
```

dependencyManagement/build.gradle.kts
```diff
@@ -70,6 +70,7 @@ val CORE_DEPENDENCIES = listOf(
   "io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api:${otelInstrumentationAlphaVersion}",
   "io.opentelemetry.javaagent:opentelemetry-javaagent-bootstrap:${otelInstrumentationAlphaVersion}",
   "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${otelInstrumentationAlphaVersion}",
+  "io.opentelemetry.javaagent.instrumentation:opentelemetry-javaagent-servlet-common-bootstrap:${otelInstrumentationAlphaVersion}",
```
trask, line 73: is this needed in this PR?

trask (approved):

PR #2906, opened 2023-02-11
Better mapping of cosmos telemetry

trask:
cc @lmolkova


agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java
```diff
@@ -392,16 +396,31 @@ private static void applyDatabaseClientSpan(
       } else {
         type = "SQL";
       }
+    } else if (dbSystem.equals("Cosmos")) {
```
heyams, line 399: make "Cosmos" a constant string?

heyams (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/SpanDataMapper.java
```diff
@@ -392,16 +396,31 @@ private static void applyDatabaseClientSpan(
...
+      // this has special icon in portal (documentdb was the old name for cosmos)
+      type = "azure documentdb";
```
lmolkova, line 401: we had a long conversation between portal and cosmos team and decided that using RP namespace is the best choice (`Microsoft.DocumentDB` for client spans, `InProc | Microsoft.DocumentDB` for internal spans)

https://github.com/microsoft/ApplicationInsights-dotnet/blob/18be714998a846aba7e9b673d47eece67be2d3dc/WEB/Src/DependencyCollector/DependencyCollector/Implementation/AzureSdk/AzureSdkDiagnosticsEventHandler.cs#L20-L22

If it's not implemented on the portal yet, it should be soon (cc @OsvaldoRosado)

trask: thx! `Microsoft.DocumentDB` didn't have an image yet when I tested, but went ahead and changed it to avoid churn on the data 👍

PR #2953, opened 2023-03-08
When diagnostics are enabled, automatically use DIAGNOSTIC_PROFILE

johnoliver:
Currently if enableDiagnostics is set to true, the user must also use the DIAGNOSTIC_PROFILE ProfileType. At the moment enabling diagnostics with a different profile is _probably_ an incorrect configuration as it is unlikely to work. 

This PR:

1. If enableDiagnostics == true && a custom JFR config is provided by the user:
    - A warning is logged that this is unlikely to work, however continues to use the custom profile
2. If enableDiagnostics == true && an alternative inbuilt config was selected by the user:
    - The configuration is overridden to the DIAGNOSTIC_PROFILE  and a warning is logged that this has happened.
3. If enableDiagnostics == true && the user has not configured any profile setting:
    - The configuration is overridden to the DIAGNOSTIC_PROFILE.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/AlternativeJfrConfigurations.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/AlternativeJfrConfigurations.java
@@ -52,35 +52,68 @@ class AlternativeJfrConfigurations {
       justification =
           "The constructed file path cannot be controlled by an end user of the instrumented application")
   private static RecordingConfiguration getRecordingConfiguration(
-      @Nullable String triggeredSettings, AlertMetricType type) {
-    if (triggeredSettings != null) {
-      try {
-        // Look for file on the local file system
-        FileInputStream fis = new FileInputStream(triggeredSettings);
-        return new RecordingConfiguration.JfcFileConfiguration(fis);
-      } catch (FileNotFoundException e) {
-        // NOP, to be expected if a configuration and not a file is provided
-      }
-
-      // Look for file in the class path
-      InputStream fis = AlternativeJfrConfigurations.class.getResourceAsStream(triggeredSettings);
-      if (fis != null) {
-        return new RecordingConfiguration.JfcFileConfiguration(fis);
-      }
-
-      try {
-        // Try parsing the triggeredSettings as a pre-configured type
-        // Convert from kebab case to enum type
-        String enumType = triggeredSettings.toUpperCase(Locale.ROOT).replaceAll("-", "_");
-        ProfileTypes profile = ProfileTypes.valueOf(enumType);
-
-        return AlternativeJfrConfigurations.get(profile, type);
-      } catch (IllegalArgumentException e) {
-        logger.error("Failed to find JFC configuration " + triggeredSettings);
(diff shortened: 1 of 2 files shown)


jeanbisutti:
@johnoliver 

> 1. If enableDiagnostics == true && a custom JFR config is provided by the user:
>    
>    * A warning is logged that this is unlikely to work, however continues to use the custom profile
> 2. If enableDiagnostics == true && an alternative inbuilt config was selected by the user:
>    
>    * The configuration is overridden to the DIAGNOSTIC_PROFILE  and a warning is logged that this has happened.

For me, it is a bit confusing not to have the same behaviors for the first two cases. If enableDiagnostics == true && a custom JFR config is provided by the use, I would also have overriden the configuration to DIAGNOSTIC_PROFILE.

A second point is about the  "diagnostics" naming. Maybe, we could replace "Diagnostics is enabled" by "Auto-diagnostics from triggers" in the warning logs to have something more explicit (if "Auto-diagnostics from triggers" would represent what is enabled in AI).

johnoliver:
> @johnoliver
> 
> > 1. If enableDiagnostics == true && a custom JFR config is provided by the user:
> >    
> >    * A warning is logged that this is unlikely to work, however continues to use the custom profile
> > 2. If enableDiagnostics == true && an alternative inbuilt config was selected by the user:
> >    
> >    * The configuration is overridden to the DIAGNOSTIC_PROFILE  and a warning is logged that this has happened.
> 
> For me, it is a bit confusing not to have the same behaviors for the first two cases. If enableDiagnostics == true && a custom JFR config is provided by the use, I would also have overriden the configuration to DIAGNOSTIC_PROFILE.
> 
> A second point is about the "diagnostics" naming. Maybe, we could replace "Diagnostics is enabled" by "Auto-diagnostics from triggers" in the warning logs to have something more explicit (if "Auto-diagnostics from triggers" would represent what is enabled in AI).

There would be some danger in overriding a config explicitly set by a user, if a user does this to remove sensitive events from their profiles I would not want to override them.

jeanbisutti:
> > @johnoliver
> > > 1. If enableDiagnostics == true && a custom JFR config is provided by the user:
> > >    
> > >    * A warning is logged that this is unlikely to work, however continues to use the custom profile
> > > 2. If enableDiagnostics == true && an alternative inbuilt config was selected by the user:
> > >    
> > >    * The configuration is overridden to the DIAGNOSTIC_PROFILE  and a warning is logged that this has happened.
> > 
> > 
> > For me, it is a bit confusing not to have the same behaviors for the first two cases. If enableDiagnostics == true && a custom JFR config is provided by the use, I would also have overriden the configuration to DIAGNOSTIC_PROFILE.
> > A second point is about the "diagnostics" naming. Maybe, we could replace "Diagnostics is enabled" by "Auto-diagnostics from triggers" in the warning logs to have something more explicit (if "Auto-diagnostics from triggers" would represent what is enabled in AI).
> 
> There would be some danger in overriding a config explicitly set by a user, if a user does this to remove sensitive events from their profiles I would not want to override them.

@johnoliver Thanks! Could you please add a comment in the implementation or the test to have the "why" in the code?

jeanbisutti (approved):

Issue #2941, opened 2023-02-28
Using Application Insights java Agent is impossible to suppress some handled exceptions


apescione:
### Context

I'm using a Spring Boot Application (Java) and WebFlux as the web layer (Reactor), and Tomcat is used as an Embedded web server.
In my application, I've Server-Sent-Event API like this:
```
  @RequestMapping(value = "/messages",
        produces = {"text/event-stream", "application/stream+json"},
        method = RequestMethod.GET)
    public Mono<ResponseEntity<Flux<ServerSentEvent<String>>>> getSseMessages() {
```
This API sends events to clients using a HOT Stream.
I'm using the application insight java agent to instrument my application deployed into a docker container.
When a client disconnects himself from my application, the Spring application, in the next update through the client raises a 
ClientAbortException and inner cause java.io.IOException: Broken pipe.

### Expected behavior

Having a way to choose what exception I want to track. So, in my scenario, I don't want to consider this kind of exception "ClientAbortException and inner cause java.io.IOException: Broken pipe." on a particular API, because I know that sooner or later a client will disconnect itself (Close front end application, change page, etc...)

### Actual behavior

Reading Microsoft documentation, it seems there is a way to choose the "Sampling" for each metric, so that, if you set percentage to 0, you can suppress the metric and avoid that agent sents it to cloud.
Here's the documentation: https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-sampling-overrides

Despite it is a preview, it works fine for *AgentLogExporter*, in fact, if the "exception" is coming from a log, it can be managed, but if the metric is from *TelemetryItemExporter*, it seems there is no way to suppress it.
Here's the metric that I'd like to suppress:
```
2023-02-27 16:25:38.808Z DEBUG c.a.m.o.e.i.p.TelemetryItemExporter - sending telemetry to ingestion service:
....
[...]

PR #2970, opened 2023-03-20
Add a config to turn off capturing exception on request

heyams:
Address #2941


apescione:
When will this pull request be merged? will it be released with 3.4.12?

heyams:
> When will this pull request be merged? will it be released with 3.4.12?

Hi @apescione we want to integrate this as part of sampling override instead of creating a new config flag. I will update you when the new solution becomes available. Thank you for following up.

PR #3030, opened 2023-04-27
Make Micrometer Namespace Configurable

### LINKED ISSUE #3027: Micrometer metrics custom namespace (i.e. not "azure.applicationinsights")
Using Spring Boot with the `applicationinsights-runtime-attach` dependency. 

    <dependency>
        <groupId>com.microsoft.azure</groupId>
        <artifactId>applicationinsights-runtime-attach</artifactId>
        <version>3.4.11</version>
    </dependency>

In the business logic I am logging metrics with Micrometer counters:

    import io.micrometer.core.instrument.Counter;
    import io.micrometer.core.instrument.Metrics; 

    public class MapReduceController {

        private final Counter myCounter = Metrics.counter("my-counter");

        public String mostPopular(List<String> names) {
            myCounter.increment();
            // logic...
            return "Lisa";
        }  
   
according to MS documentation this all fine and dandy and I do get my metrics recorded in App Insights, BUT like the issue #2088 they end up in the Custome Namespace `azure.applicationinsights`.
There are no examples nor documentation stating how the metrics _should_ be recorded to get it into a specific namespace.

It should be possible to get Micrometer metrics into a custom namespace. If it is possible already please provide an example how Micrometer and/or appinsights should be configured to do so.

Dragging in `applicationinsights-core` and create metrics directly with the core MetricTelemetry class instead of Micrometer, like in #2088. This feels like a work around and not a solution to using [the supported Micrometer library](https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable?tabs=java#metrics).

**Additional context**

heyams:
Fix #3027

### DIFF
--- a/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
+++ b/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
@@ -200,7 +200,7 @@ public class AzureMonitorMeterRegistry extends StepMeterRegistry {
             meterId.getType(),
             meterId.getBaseUnit());
   }
-  
+
   private static String getNamespace() {
     return AzureMonitorRegistryConfig.INSTANCE.namespace();
   }
--- a/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
+++ b/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
@@ -200,8 +200,8 @@ public class AzureMonitorMeterRegistry extends StepMeterRegistry {
             meterId.getType(),
             meterId.getBaseUnit());
   }
-
-  private String getNamespace() {
+  
+  private static String getNamespace() {
     return AzureMonitorRegistryConfig.INSTANCE.namespace();
   }
 
--- a/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
+++ b/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
@@ -28,12 +28,9 @@ public class AzureMonitorMeterRegistry extends StepMeterRegistry {
   public static final AzureMonitorMeterRegistry INSTANCE =
       new AzureMonitorMeterRegistry(Clock.SYSTEM);
 
-  private final AzureMonitorRegistryConfig config;
-
   // visible for testing
   public AzureMonitorMeterRegistry(Clock clock) {
-    super(new AzureMonitorRegistryConfig(), clock);
-    config = new AzureMonitorRegistryConfig();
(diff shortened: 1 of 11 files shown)


agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorMeterRegistry.java
```diff
@@ -28,9 +28,12 @@ public class AzureMonitorMeterRegistry extends StepMeterRegistry {
   public static final AzureMonitorMeterRegistry INSTANCE =
       new AzureMonitorMeterRegistry(Clock.SYSTEM);
 
+  private final AzureMonitorRegistryConfig config;
+
   // visible for testing
   public AzureMonitorMeterRegistry(Clock clock) {
     super(new AzureMonitorRegistryConfig(), clock);
+    config = new AzureMonitorRegistryConfig();
```
heyams, line 36: can't init instance variable before the supertype constructor. not sure a better way to init it, it will return the same AzureMonitorRegistryConfig(). I new it twice at line 35 and 36. please let me know if you have a better way.

smoke-tests/apps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerNamespaceTest.java
```diff
@@ -0,0 +1,49 @@
...
+    List<Envelope> metricItems =
+        testing.mockedIngestion.waitForItems(
+            MicrometerTest::isMicrometerMetricWithValueOne, 1, 10, TimeUnit.SECONDS);
+
+    MetricData data = (MetricData) ((Data<?>) metricItems.get(0).getData()).getBaseData();
+    List<DataPoint> points = data.getMetrics();
+    assertThat(points).hasSize(1);
+
+    DataPoint point = points.get(0);
+    assertThat(point.getMetricNamespace()).isEqualTo("testNamespace");
```
trask, line 47: can you add an assertion to the existing `MicrometerTest` to ensure that the namespace still reports null when there's no namespace configured?

agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/AzureMonitorRegistryConfig.java
```diff
@@ -10,12 +10,18 @@
...
+  public static final AzureMonitorRegistryConfig INSTANCE = new AzureMonitorRegistryConfig();
+
+  private AzureMonitorRegistryConfig() {
     step =
         InstrumentationConfig.get()
             .getDuration(
                 "applicationinsights.internal.micrometer.step.millis", Duration.ofSeconds(60));
+    namespace =
+        InstrumentationConfig.get()
+            .getString("applicationinsights.internal.micrometer.namespace", "");
```
trask, line 24: I think I'd suggest using `@Nullable` on namespace everywhere to make it more clear that it is optional

```suggestion
    namespace =
        InstrumentationConfig.get()
            .getString("applicationinsights.internal.micrometer.namespace");
```

smoke-tests/apps/Micrometer/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerNamespaceTest.java
trask: I don't see this change (in the `MicrometerTest` smoke test)

trask (approved):
one last suggestion, thx 👍

PR #3036, opened 2023-05-01
Fix 2.x SDK interop

trask:
Found this while testing locally. Prior to 3.4.0 the 2.x integration didn't require iKey to be configured on the 2.x side (only on the 3.x side). This broke in 3.4.0, and currently the 2.x integration doesn't emit any telemetry if there's no iKey configured on the 2.x side.

### DIFF
--- a/smoke-tests/apps/CoreAndFilter2x/src/main/resources/ApplicationInsights.xml
+++ b/smoke-tests/apps/CoreAndFilter2x/src/main/resources/ApplicationInsights.xml
@@ -7,7 +7,8 @@
 
   <!-- cannot use ConnectionString here since testing CoreAndFilter against very old SDK
        prior to 2.5.1 when when ConnectionString support was introduced -->
-  <InstrumentationKey>BAD1BAD1-BAD1-BAD1-BAD1-BAD1BAD1BAD1</InstrumentationKey>
+  <!-- testing to ensure 3.x integration still works even with no InstrumentationKey configured -->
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/TelemetryClientClassFileTransformer.java
```diff
@@ -405,6 +431,29 @@ private void overwriteTrackMetricMethod(MethodVisitor mv) {
...
+      mv.visitVarInsn(ALOAD, 0);
+      mv.visitFieldInsn(
+          GETFIELD,
+          unshadedPrefix + "/TelemetryClient",
+          "configuration",
+          "L" + unshadedPrefix + "/TelemetryConfiguration;");
+      mv.visitMethodInsn(
+          INVOKEVIRTUAL,
+          unshadedPrefix + "/TelemetryConfiguration",
+          "isTrackingDisabled",
```
heyams, line 447: I didn't find any tests for isTackingDisabled or isDisabled. How does customers use it? no ikey is only when they disable tracking, correct?

trask: they don't use it directly, the 2.x TelemetryClient calls `isDisabled()` is called in several places

jeanbisutti (approved):

PR #3067, opened 2023-05-11
Use upstream micrometer instrumentation for 1.5+

### LINKED ISSUE #2806: valueMax reporting wrong number
### Expected behavior
I would expect `valueMax` to have a value less than `valueSum`.

### Actual behavior
`valueMax` is greater than `valueSum`.

### To Reproduce
This is my method that is called to gather metrics using `micrometer`:

```
@RequiredArgsConstructor
@Slf4j
public class TimeRecorder {
    private final static String METHOD = "method";
    private final static String CLASS = "class";
    private final MeterRegistry registry;

    public <T> T recorded(Supplier<T> supplier) {
        StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();

        if (stackTrace.length >= 3) {
            StackTraceElement stackTraceElement = stackTrace[2];

            return registry.timer("method.timing",
                            List.of(
                                    Tag.of(CLASS, stackTraceElement.getClassName()),
                                    Tag.of(METHOD, stackTraceElement.getMethodName())
                            ))
                    .record(supplier);
        } else {
            log.error("unable to determine callee, skipping metrics");
            return supplier.get();
        }
    }
}
```

### System information
 - SDK Version: applicationinsights-agent-3.4.6.jar
 - OS type and version: eclipse-temurin:17-jdk-alpine
 - Using spring-boot? Yes (version 2.7.1)
 - Additional relevant libraries (with version, if applicable): Micrometer 1.9.1

### Logs

### Screenshots

Metric

trask:
TODO need to manually test old actuator instrumentation, because I haven't been able to get old spring boot versions to work since [upgrading to Gradle 8.1.1](#3046).

Resolves #2806


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -177,14 +178,13 @@ private static void enableInstrumentations(Configuration config, Map<String, Str
     properties.put("otel.instrumentation.undertow.enabled", "true");
 
     if (config.instrumentation.micrometer.enabled) {
-      // TODO (heya) replace with below when updating to upstream micrometer
       properties.put("otel.instrumentation.ai-micrometer.enabled", "true");
-      properties.put("otel.instrumentation.ai-actuator-metrics.enabled", "true");
-      // properties.put("otel.instrumentation.micrometer.enabled", "true");
-      // properties.put("otel.instrumentation.spring-boot-actuator-autoconfigure.enabled", "true");
+      // properties.put("otel.instrumentation.ai-actuator-metrics.enabled", "true");
```
jeanbisutti, line 182: Why to leave code as a comment?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
```diff
@@ -139,7 +139,7 @@ void generateHttpClientMetrics() {
     MetricTelemetryBuilder builder = MetricTelemetryBuilder.create();
     MetricData metricData = metricDataCollection.iterator().next();
     MetricDataMapper.updateMetricPointBuilder(
-        builder, metricData, metricData.getData().getPoints().iterator().next(), true, true);
+        builder, metricData, metricData.getData().getPoints().iterator().next(), true, true, null);
```
jeanbisutti, line 142: metricData.getData() can never return null?

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
```diff
@@ -94,6 +97,15 @@ private List<TelemetryItem> convertOtelMetricToAzureMonitorMetric(
       MetricData metricData, boolean isPreAggregatedStandardMetric) {
     List<TelemetryItem> telemetryItems = new ArrayList<>();
 
+    String namespace = null;
+    if (micrometerMetricNamespace != null
+        && metricData
```
jeanbisutti, line 102:
```
metricData
            .getInstrumentationScopeInfo()
            .getName()
            .equals("io.opentelemetry.micrometer-1.5")
```

Something shorter to have something easier to read? For example, with an extract local variable (as boolean isMicrometer1_5) or an extract method.

smoke-tests/apps/Micrometer1_5/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
```diff
@@ -0,0 +1,31 @@
...
+  @Test
+  @TargetUri("/test")
+  void doMostBasicTest() throws Exception {
+    Telemetry telemetry = testing.getTelemetry(0);
+
+    assertThat(telemetry.rd.getName()).isEqualTo("GET /test");
+    assertThat(telemetry.rd.getSuccess()).isTrue();
+
+    // sleep a bit and make sure no micrometer metrics are reported
+    Thread.sleep(10000);
```
jeanbisutti, line 27: To make the test faster, would it be possible to check if assertions just after are successful every 1 second for example, and and this up to 10 seconds maximum?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
```diff
@@ -59,6 +59,13 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
       return CompletableResultCode.ofSuccess();
     }
     for (MetricData metricData : metrics) {
+      String name = metricData.getName();
+      if (metricData
+          .getInstrumentationScopeInfo()
+          .getName()
```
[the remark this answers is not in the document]
trask: yes, it can't return null

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
trask: (this is in test code)

smoke-tests/apps/Micrometer1_5/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
trask: this is ensuring that _no_ metrics are reported, so verifying that no metrics are reported after just 1 second is not enough

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MetricDataMapper.java
trask: done

smoke-tests/apps/Micrometer1_5/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
jeanbisutti: The test could fail if metrics are reported after 2 seconds for example (this could be checked every second for example)?

trask: that's true, I'm not too worried about test slowness when it comes to failures, but if you would like, we should do it across all of our smoke tests (it looks like this pattern is used in ~20 places)

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
jeanbisutti: Maybe an assertion checking that metricData.getData() is not null could be useful in test code to have an explicit error message

trask: getData never returns null

smoke-tests/apps/Micrometer1_5/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/MicrometerDisabledTest.java
jeanbisutti: Yes, it could be great to do it across all of our smoke tests , or start to do this for this test in this PR (and apply the same pattern for other smoke tests in another PR). In one previous project, I have applied this pattern and the reduction on build duration was significant.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
trask: brought this code back

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentMetricExporter.java
```diff
@@ -59,6 +59,13 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
...
+          .equals("io.opentelemetry.micrometer-1.5")) {
```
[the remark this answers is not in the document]
heyams: can we extract out isFromMicrometerInstrumentation as a static method in the utils class and then reuse it here?

trask:
Converting to draft because of the 63 character limit for OTel metric names which could introduce regressions for existing users by moving to the upstream instrumentation

PR #3111, opened 2023-06-02
Prefer forwarded url scheme

trask:
In current OTel Java Instrumentation release, url scheme will reflect scheme provided in `Forwarded` or `X-Forwarded-Proto`. This is preferred for us, because otherwise when SSL is terminated at the web server (e.g. in App Services) users would see url like `http://...` and think the request was unencrypted.

In the next OTel Java Instrumentation release, this behavior will change, and adding this setting is needed in order to preserve the existing behavior.

TODO add a smoke test for this, will need to add an option to send `X-Forwarded-Proto` header from SmokeTestExtension.


trask:
replaced by #3125

PR #3119, opened 2023-06-12
Fix Snippet Injection

siyuniu-ms:
Fix 1. remove static for snippetInjection 2. change the way of readfile 3. enable snippet injection with initial connectionString

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfiguration.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfiguration.java
@@ -4,12 +4,11 @@
 package com.microsoft.applicationinsights.agent.internal.configuration;
 
 import io.opentelemetry.javaagent.bootstrap.servlet.ExperimentalSnippetHolder;
+import java.io.BufferedReader;
 import java.io.IOException;
-import java.net.URISyntaxException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfiguration.java
```diff
@@ -20,19 +19,27 @@ public class SnippetConfiguration {
...
-      LOGGER.error("Failed to read javascript-snippet file", e);
-    }
-    return "";
-  }
-
-  private static Path getSnippetFilePath(String resourceName) throws URISyntaxException {
     ClassLoader classLoader = SnippetConfiguration.class.getClassLoader();
-    return Paths.get(classLoader.getResource(resourceName).toURI());
+    String resourceName = "javascript-snippet.txt";
+    InputStream inputStream = classLoader.getResourceAsStream(resourceName);
```
trask, line 24: early return will help reduce the amount of indentation below

```suggestion
    InputStream inputStream = classLoader.getResourceAsStream(resourceName);
    if (inputStream == null) {
      LOGGER.error("Resource not found: " + resourceName);
      return "";
    }
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfiguration.java
```diff
@@ -20,19 +19,27 @@ public class SnippetConfiguration {
...
+          new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+        String line;
+        while ((line = reader.readLine()) != null) {
+          // Process each line of the file
+          returnString += line + "\n";
+        }
+      } catch (IOException e) {
+        // Handle any IO exceptions that occur
+        LOGGER.error("Failed to read javascript-snippet file", e);
+      }
```
trask, line 37: let's create private static method similar that is used elsewhere:

```
  private static String toString(InputStream inputStream) throws IOException {
    try (BufferedReader bufferedReader =
        new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
      return bufferedReader.lines().collect(Collectors.joining(System.lineSeparator()));
    }
  }
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfiguration.java
```diff
@@ -20,19 +20,27 @@ public class SnippetConfiguration {
...
     return "";
   }
 
-  private static Path getSnippetFilePath(String resourceName) throws URISyntaxException {
-    ClassLoader classLoader = SnippetConfiguration.class.getClassLoader();
-    return Paths.get(classLoader.getResource(resourceName).toURI());
+  private static String toString(InputStream inputStream) throws IOException {
+    try (BufferedReader bufferedReader =
+        new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+      return bufferedReader.lines().collect(Collectors.joining(System.lineSeparator()));
```
trask, line 42: probably makes sense to always use `\n` instead of `\r\n` on windows server, since this is going to end up on client

```suggestion
      return bufferedReader.lines().collect(Collectors.joining("\n"));
```

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/configuration/SnippetConfigurationTest.java
```diff
@@ -5,17 +5,16 @@
...
 public class SnippetConfigurationTest {
   private static final String CONNECTION_STRING =
       "InstrumentationKey=00000000-0000-0000-0000-000000000000;IngestionEndpoint=https://fake-ingestion-endpoint";
 
   @Test
-  void testSnippetSetup() throws IOException {
+  void testSnippetSetup() {
     String snippet = SnippetConfiguration.readSnippet();
-    assertThat(snippet).contains("connectionString: \"YOUR_CONNECTION_STRING\"\n");
+    assertThat(snippet).contains("connectionString: \"YOUR_CONNECTION_STRING\"");
```
trask, line 17: then can bring back the `\n` here

trask (approved):

PR #3137, opened 2023-06-19
Fix log4j timestamp

trask:
this is needed temporarily until https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/8761


heyams (approved):

showpune:
@trask , can I confirm with you will it impact all the cases or it is just a corner case

trask:
I believe it affects all log4j2 users

PR #3163, opened 2023-06-28
Update azure sdk statsbeat tracking

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/StatsbeatSpanExporter.java
```diff
@@ -22,9 +24,11 @@ public StatsbeatSpanExporter(SpanExporter delegate, StatsbeatModule statsbeatMod
   @Override
   public CompletableResultCode export(Collection<SpanData> spans) {
     for (SpanData span : spans) {
-      statsbeatModule
-          .getInstrumentationStatsbeat()
-          .addInstrumentation(span.getInstrumentationScopeInfo().getName());
+      String instrumentationScopeName = span.getInstrumentationScopeInfo().getName();
```
jeanbisutti, line 27: Lines 27 to 30 could be brought together in one method, for example with an `instrumentationScopeName` method. The `export` method would be faster to read.

trask: I think it's ok either way, since `export` is already such a small method

trask (approved):

PR #3171, opened 2023-06-30
Change azure instrumentation scope name prefix for statsbeats

lmolkova:
@heyams @trask 

I'm sorry I gave you the wrong prefix for Azure SDK instrumentation scope name. We don't include group id and use artifact id as is. So, it should be not "com.azure:", but "azure-".

This should fix it.


trask (approved):
thx!

PR #3191, opened 2023-07-11
Restrict connectionString to app service only

### LINKED ISSUE #3187: In Java agent 3.4.14, facing issues related to env variables setting of app insights content
using applicationinsights-runtime-attach 3.4.14 for spring boot 3

1. I wanted to filter out all assets folder calls from app insights so saw the documentation and used the sampling overrides now facing issues with the env variable setting of the APPLICATIONINSIGHTS_CONFIGURATION_CONTENT. so in this would be the key in env variable APPLICATIONINSIGHTS_CONFIGURATION_CONTENT and value i have kept the entire file content including connectionString.

{"connectionString":"…","preview": {"sampling": {"overrides": [{"telemetryType": "request","attributes": [{"key": "http.url","value": "https?://[^/]+/assets/","matchType": "regexp"}],"percentage": 0}]}}}

when i run the application getting this error
 java.lang.IllegalStateException:com.microsoft.applicationinsights.agent.internal.configuration.ConfigurationBuilder$ConfigurationException: "connectionString" attribute is not supported inside of APPLICATIONINSIGHTS_CONFIGURATION_CONTENT, please use APPLICATIONINSIGHTS_CONNECTION_STRING to specify the connection string

heyams:
Fix #3187

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -841,7 +841,9 @@ public class ConfigurationBuilder {
 
     Configuration configuration = getConfiguration(json, JsonOrigin.ENV_VAR);
 
-    if (configuration.connectionString != null) {
+    // restrict connection string in APPLICATIONINSIGHTS_CONFIGURATION_CONTENT for App Service only
+    if (configuration.connectionString != null
+        && DiagnosticsHelper.useAppSvcRpIntegrationLogging()) {
       throw new ConfigurationException(
           "\"connectionString\" attribute is not supported inside of "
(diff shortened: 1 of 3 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -841,7 +841,9 @@ static Configuration getConfigurationFromEnvVar(String json) {
 
     Configuration configuration = getConfiguration(json, JsonOrigin.ENV_VAR);
 
-    if (configuration.connectionString != null) {
+    // restrict connection string in APPLICATIONINSIGHTS_CONFIGURATION_CONTENT for App Service only
+    if (configuration.connectionString != null
+        && DiagnosticsHelper.useAppSvcRpIntegrationLogging()) {
```
trask, line 846: I don't love that this method name looks like it's related to logging, i'll send a follow-up PR to rename

trask (approved):

PR #3195, opened 2023-07-12
Preserve db.cosmosdb.* span attributes

trask:
cc @lmolkova

### DIFF
--- a/agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MappingsBuilder.java
+++ b/agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MappingsBuilder.java
@@ -21,7 +21,16 @@ class MappingsBuilder {
       new HashSet<>(
           asList(
               "http.",
-              "db.",
+
+              // specifically listing out all standard db.* attributes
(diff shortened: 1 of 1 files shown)


lmolkova (approved):

agent/azure-monitor-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/MappingsBuilder.java
```diff
@@ -21,7 +21,16 @@ class MappingsBuilder {
...
+              // so that db.cosmosdb.* attributes will be preserved
+              SemanticAttributes.DB_SYSTEM.getKey(),
+              SemanticAttributes.DB_NAME.getKey(),
+              SemanticAttributes.DB_STATEMENT.getKey(),
+              SemanticAttributes.DB_OPERATION.getKey(),
+              SemanticAttributes.DB_CONNECTION_STRING.getKey(),
+              SemanticAttributes.DB_USER.getKey(),
+              SemanticAttributes.DB_SQL_TABLE.getKey(),
               "message.",
               "messaging.",
```
lmolkova, line 35: can I do a similar change for messaging so we don't lose `messaging.servicebus` and similar attributes ?

PR #3208, opened 2023-07-19
Add a config to disable all logging appenders

heyams:


smoke-tests/apps/JavaUtilLogging/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/JavaUtilLoggingDisabledTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.microsoft.applicationinsights.smoketest.schemav2.Data;
+import com.microsoft.applicationinsights.smoketest.schemav2.Envelope;
+import com.microsoft.applicationinsights.smoketest.schemav2.RequestData;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+@UseAgent("disabled_applicationinsights.json")
+abstract class JavaUtilLoggingDisabledTest {
```
trask, line 25: and then can remove all the subclasses (similar to the other disabled tests, no need to run these tests against envs)

```suggestion
@Environment(JAVA_8)
class JavaUtilLoggingDisabledTest {
```

smoke-tests/apps/JavaUtilLogging/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/JavaUtilLoggingTest.java
```diff
@@ -25,7 +25,7 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
-@UseAgent
+@UseAgent("applicationinsights.json")
```
trask, line 28: I think this is the default behavior

```suggestion
@UseAgent
```

smoke-tests/apps/JavaUtilLogging/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/JavaUtilLoggingDisabledTest.java
heyams: the purpose of doing subclass is to test all java versions.. this will cover one case? 
our repo has this setup everywhere though
i can make the change but i think it needs to be consistent

trask: check the other `*DisabledTest` smoke tests, they also only test one case

trask (approved):
👍

smoke-tests/apps/JavaUtilLogging/src/main/java/com/microsoft/applicationinsights/smoketestapp/JavaUtilLoggingDisabledServlet.java
```diff
@@ -0,0 +1,21 @@
...
+
+@WebServlet("/testDisabled")
+public class JavaUtilLoggingDisabledServlet extends HttpServlet {
+
+  private static final Logger logger = Logger.getLogger("smoketestapp");
+
+  protected void doGet(HttpServletRequest request, HttpServletResponse response) {
+    logger.log(Level.WARNING, "this message will get suppressed.");
+  }
+}
```
trask, line 21: I don't think we need a separate Servlet for these tests, can just hit the same servlet that the normal test does (this is how other `*DisabledTest` smoke tests work)

heyams: i followed the existing setup.. one test per servlet. i think we can clean up later for all

heyams: i noticed those are spring boot apps. not servlet

PR #3220, opened 2023-07-27
Update vm prefix using exporter beta.11

heyams:
This is depending on https://github.com/Azure/azure-sdk-for-java/pull/36121

MetadataInstanceResponse is package protected in beta.10.


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -94,7 +94,7 @@ public static TelemetryClient createForTest() {
     return builder()
         .setCustomDimensions(new HashMap<>())
         .setMetricFilters(new ArrayList<>())
-        .setStatsbeatModule(new StatsbeatModule())
+        .setStatsbeatModule(new StatsbeatModule(null))
```
trask, line 97:
```suggestion
        .setStatsbeatModule(new StatsbeatModule(response -> {}))
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java
```diff
@@ -73,6 +77,12 @@ public static boolean isFunctionsRpIntegration() {
     return functionsRpIntegration;
   }
 
+  public static void lazyUpdateVmRpIntegration(MetadataInstanceResponse response) {
+    rpIntegrationChar = 'v';
```
trask, line 81: I think this is a bit confusing here, as DiagnosticsHelper is only for integrated rp attach, and we don't have that for VMs, maybe it would help to move `rpIntegrationChar` somewhere else?

heyams: where do you recommend? i can move it

heyams: move it to a new class call SdkVersionPrefixHelper.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/SdkVersionPrefixHelper.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.microsoft.applicationinsights.agent.internal.configuration;
+
+import com.azure.monitor.opentelemetry.exporter.implementation.statsbeat.MetadataInstanceResponse;
+import com.azure.monitor.opentelemetry.exporter.implementation.statsbeat.RpAttachType;
+import com.azure.monitor.opentelemetry.exporter.implementation.utils.PropertyHelper;
+import com.azure.monitor.opentelemetry.exporter.implementation.utils.SystemInformation;
+import org.slf4j.LoggerFactory;
+
+public final class SdkVersionPrefixHelper {
```
trask, line 12:
```suggestion
public final class SdkVersionPrefixHolder {
```

trask (approved):

PR #3229, opened 2023-07-28
Telemetry processor - Allow multimatch on log body and span name

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/ProcessorUtil.java
```diff
@@ -24,7 +24,7 @@ public static String applyRule(
     StringBuilder sb = new StringBuilder();
     int lastEnd = 0;
     // As of now we are considering only first match.
```
trask, line 26:
```suggestion
```

trask (commented):
it may make sense to capture multimatches in an attribute of type "string array" (to capture all of the values, instead of only the last value)

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/processors/ExporterWithLogProcessorTest.java
```diff
@@ -184,7 +184,7 @@ void multiRuleToAttributesTest() {
     LogRecordExporter logExporter = new ExporterWithLogProcessor(config, mockExporter);
     TestLogRecordData mockLogA =
         TestLogRecordData.builder()
-            .setBody("yyyPassword=123 aba Pass=555 xyx Pass=777 zzz")
+            .setBody("yyyPassword=123 aba Pass=555 xyx")
```
trask, line 187: it seems nice to keep the multi-match example here so we can see/confirm the result?

trask (commented):
I'm not sure about implementing different behavior for log body and span name

I think my preference would be to keep them consistent (either break neither, or break both)

PR #3245, opened 2023-08-08
Use upstream micrometer instrumentation

### LINKED ISSUE #2806: valueMax reporting wrong number

trask:
TODO need to manually test old actuator instrumentation, because I haven't been able to get old spring boot versions to work since [upgrading to Gradle 8.1.1](#3046).

Resolves #2806 and https://portal.microsofticm.com/imp/v3/incidents/incident/409304026

The 63 character limit for OTel metric names which could introduce regressions for existing users by moving to the upstream instrumentation is resolved now in OTel Java v1.30.0.


heyams:
@trask is this ready for review?

agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/ActuatorInstrumentationModule.java
```diff
@@ -24,7 +25,10 @@ public ActuatorInstrumentationModule() {
 
   @Override
   public ElementMatcher.Junction<ClassLoader> classLoaderMatcher() {
-    return hasClassesNamed("io.micrometer.core.instrument.Metrics");
+    return hasClassesNamed("io.micrometer.core.instrument.Metrics")
+        .and(
+            // added in micrometer 1.5
+            not(hasClassesNamed("io.micrometer.core.instrument.config.validate.Validated")));
```
heyams, line 31: is this the only reason to keep using custom micrometer?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -189,11 +189,12 @@ private static void enableInstrumentations(
     properties.put("otel.instrumentation.undertow.enabled", "true");
 
     if (config.instrumentation.micrometer.enabled) {
-      // TODO (heya) replace with below when updating to upstream micrometer
+      // ai micrometer instrumentation is still used for micrometer versions prior to 1.5
```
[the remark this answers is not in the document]
heyams: can we also add that upstream doesn't support versions that are prior to 1.5?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/BackCompatMetricData.java
```diff
@@ -0,0 +1,78 @@
...
+  BackCompatMetricData(MetricData delegate) {
+    this.delegate = delegate;
+    this.name = backCompatName(delegate.getName());
+  }
+
+  private static String backCompatName(String name) {
+    return NAME_AND_TAG_KEY_PATTERN.matcher(toSnakeCase(name)).replaceAll("_");
+  }
+
+  private static String toSnakeCase(String value) {
```
heyams, line 32: what does it mean `toSnakeCase`?

trask: https://en.wikipedia.org/wiki/Snake_case

smoke-tests/apps/MicrometerOldVersion/build.gradle.kts
```diff
@@ -0,0 +1,8 @@
+plugins {
```
heyams, line 1: Can we rename this smoke test to be more specific, something like "MicrometerPrior1_5"

jeanbisutti:
[This comment](https://github.com/microsoft/ApplicationInsights-Java/blob/df1bb680abbc8b2864a81bba7d71d63d9840da40/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/MicrometerInstrumentationModule.java#L19) is only for Micrometer < 1.5?

heyams (approved):

trask:
> [This comment](https://github.com/microsoft/ApplicationInsights-Java/blob/df1bb680abbc8b2864a81bba7d71d63d9840da40/agent/instrumentation/micrometer-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/ai/MicrometerInstrumentationModule.java#L19) is only for Micrometer < 1.5?

I reworded it a bit, also check out https://github.com/microsoft/ApplicationInsights-Java/pull/3245/files#diff-1b736bfc6f05aef23179b4913d150fb5dad450dbf3297fe3a875dccfed3736b4R191-R199

trask:
Closing, can revisit in the future.

PR #3277, opened 2023-09-07
Fix AKS cloud_RoleName

heyams:


heyams:
Rmanuj helped me test it in AKS and it worked.

trask (approved):

PR #3288, opened 2023-09-13
Fix applicationinsights.json config is not loaded in manual rp attach scenario

### LINKED ISSUE #3283: The configuration file present in the same directory as the agent is not being picked-up in 3.4.16
### Expected behavior
Default configuration file `applicationinsights.json` present in the same directory as the agent should be automatically picked-up by the agent when the application starts.

### Actual behavior
The default application insights configuration is being applied and the configuration file is being ignored.

### To Reproduce
* Provide a configuration file that sits next to the agent where you change the selfDiagnostics level to DEBUG.
* Do NOT specify the path to the configuration file using the environment variable `APPLICATIONINSIGHTS_CONFIGURATION_FILE`

```
"selfDiagnostics": {
  "destination": "console",
  "level": "DEBUG"
}
```

### System information
 - SDK Version: 3.4.16
 - OS type and version: Debian GNU/Linux 11 (bullseye)
 - Application Server type and version (if applicable): OpenJDK Runtime Environment 18.9 (build 11.0.16+8)
 - Using spring-boot? Yes

### Logs
```
Picked up JAVA_TOOL_OPTIONS: -Xms512m -Xmx1024m -XX:MaxMetaspaceSize=300m -javaagent:/applicationinsights-agent-3.4.16.jar
2023-09-11 09:05:27.791Z INFO c.m.applicationinsights.agent - Application Insights Java Agent 3.4.16 started successfully (PID 1, JVM running for 3.386 s)
2023-09-11 09:05:27.796Z INFO c.m.applicationinsights.agent - Java version: 11.0.16, vendor: Oracle Corporation, home: /usr/local/openjdk-11
```

heyams:
Fix #3283

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -480,8 +480,7 @@ public class ConfigurationBuilder {
     }
 
     // any manual RP integration will support loading user's applcationinsights.json
-    if (SdkVersionPrefixHolder.isRpIntegration()
-        && RpAttachType.getRpAttachType() != RpAttachType.MANUAL) {
+    if (RpAttachType.getRpAttachType() == RpAttachType.INTEGRATED_AUTO) {
       // users do not have write access to agent directory in rp integrations
       // and rp integrations should not use applicationinsights.json because that makes it difficult
       // to merge rp intent and user intent
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -479,7 +479,7 @@ public class ConfigurationBuilder {
       return getConfiguration(runtimeAttachedConfigurationContent, JsonOrigin.RUNTIME_ATTACHED);
     }
 
-    // any manual RP integration will support loading user's applcationinsights.json
+    // only RP auto integrations do not support loading applicationinsights.json
     if (RpAttachType.getRpAttachType() == RpAttachType.INTEGRATED_AUTO) {
       // users do not have write access to agent directory in rp integrations
       // and rp integrations should not use applicationinsights.json because that makes it difficult
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -3,6 +3,7 @@
 
 package com.microsoft.applicationinsights.agent.internal.configuration;
 
+import com.azure.monitor.opentelemetry.exporter.implementation.statsbeat.RpAttachType;
 import com.azure.monitor.opentelemetry.exporter.implementation.utils.HostName;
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -478,7 +479,9 @@ private static Configuration loadConfigurationFile(Path agentJarPath) {
       return getConfiguration(runtimeAttachedConfigurationContent, JsonOrigin.RUNTIME_ATTACHED);
     }
 
-    if (SdkVersionPrefixHolder.isRpIntegration()) {
+    // any manual RP integration will support loading user's applcationinsights.json
+    if (SdkVersionPrefixHolder.isRpIntegration()
+        && RpAttachType.getRpAttachType() != RpAttachType.MANUAL) {
```
trask, line 484: would this work?
```suggestion
    if (RpAttachType.getRpAttachType() == RpAttachType.AUTO) {
```

heyams: latest exporter has 3 types 

MANUAL("Manual"), // Manually writing code to start collecting telemetry, e.g. via azure-monitor-opentelemetry-exporter
  STANDALONE_AUTO("StandaloneAuto"), // RP attach is enabled via a custom JAVA_OPTS
  INTEGRATED_AUTO("IntegratedAuto"); // RP attach is on by default

heyams: beta.12 will have this..  no config is allowed for non MANUAL

trask: how about then?
```suggestion
    if (RpAttachType.getRpAttachType() != RpAttachType.MANUAL) {
```

heyams: getRpAttachType will run into assertError in the non-attach scenario setRpAttachType is never called

heyams: getRpAttachType will return null in that case if we don't throw assertError

heyams: we need to check both

trask: > getRpAttachType will run into assertError in the non-attach scenario setRpAttachType is never called

can we set it in all cases?

heyams:
i've tested it in AKS env, it works.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -478,7 +479,8 @@ private static Configuration loadConfigurationFile(Path agentJarPath) {
       return getConfiguration(runtimeAttachedConfigurationContent, JsonOrigin.RUNTIME_ATTACHED);
     }
 
-    if (SdkVersionPrefixHolder.isRpIntegration()) {
+    // any manual RP integration will support loading user's applcationinsights.json
```
trask, line 482: (or do we need this, see comment in line below)
```suggestion
    // only RP integrations do not support loading applicationinsights.json
```

trask (approved):

PR #3291, opened 2023-09-14
(not for merging) Debug duplicate logging

trask:


jeanbisutti:
What do you think about adding this check on the self-daignostics at TRACE level or with a new flag?

PR #3310, opened 2023-09-27
Delay JFR profiling startup

milderhc:
Adding @trask suggestion to fix logging crash, moving profiler startup to AfterAgentListener.

- [ ] CHANGELOG.md updated

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AfterAgentListener.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AfterAgentListener.java
@@ -4,12 +4,19 @@
 package com.microsoft.applicationinsights.agent.internal.init;
 
 import com.google.auto.service.AutoService;
+import com.microsoft.applicationinsights.agent.internal.configuration.Configuration;
 import com.microsoft.applicationinsights.agent.internal.httpclient.LazyHttpClient;
+import com.microsoft.applicationinsights.agent.internal.profiler.ProfilingInitializer;
+import com.microsoft.applicationinsights.agent.internal.telemetry.TelemetryClient;
 import io.opentelemetry.javaagent.extension.AgentListener;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @AutoService(AgentListener.class)
 public class AfterAgentListener implements AgentListener {
+
+  private static final Logger logger = LoggerFactory.getLogger(AfterAgentListener.class);
 
   @Override
   public void afterAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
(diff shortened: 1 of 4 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AfterAgentListener.java
```diff
@@ -4,20 +4,46 @@
...
   @Override
   public void afterAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
     // only safe now to make HTTPS calls because Java SSL classes
     // trigger loading of java.util.logging (starting with Java 8u231)
     // and JBoss/Wildfly need to install their own JUL manager before JUL is initialized.
     LazyHttpClient.safeToInitLatch.countDown();
 
-    PerformanceCounterInitializer.initialize(FirstEntryPoint.getConfiguration());
+    Configuration configuration = FirstEntryPoint.getConfiguration();
+    TelemetryClient telemetryClient = TelemetryClient.getActive();
```
jeanbisutti, line 29: TelemetryClient is only used in the if at line 33. The code would be a little more readable for me if the TelemetryClient local variable was declared just before the if.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/RuntimeConfigurator.java
```diff
@@ -141,26 +134,6 @@ public void apply(RuntimeConfiguration runtimeConfig) {
       updateSampling(enabled, runtimeConfig.sampling, runtimeConfig.samplingPreview);
     }
 
-    // initialize Profiler
```
jeanbisutti, line 144: Not sure this code has to be removed. It seems to be used for Azure function and RP integration. @trask could you confirm that this code should be left?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -92,18 +91,22 @@ public class SecondEntryPoint implements AutoConfigurationCustomizerProvider {
...
   private static final String STATSBEAT_FOLDER_NAME = "statsbeat";
+  private static File tempDir;
 
   @Nullable public static AgentLogExporter agentLogExporter;
 
   @Nullable private static BatchLogRecordProcessor batchLogProcessor;
   @Nullable private static BatchSpanProcessor batchSpanProcessor;
   @Nullable private static MetricReader metricReader;
 
+  public static File getTempDir() {
```
jeanbisutti, line 103: It seems it could have the default visibility (package private).

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/RuntimeConfigurator.java
trask: yes, I believe we should keep this for Azure Functions consumption plan

jeanbisutti (approved):

PR #3356, opened 2023-10-23
Update azure-monitor-opentelemetry-exporter

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java
```diff
@@ -37,21 +37,21 @@ public static void initRpIntegration(Path agentPath) {
     // TODO (heya) how should we report functions windows users who are using app services
     //  windows attach by manually setting the env vars (which was the old documented way)
     if ("java".equals(System.getenv("FUNCTIONS_WORKER_RUNTIME"))) {
-      SdkVersionPrefixHolder.setRpIntegrationChar('f');
+      PropertyHelper.setRpIntegrationChar('f');
```
trask, line 40: we may be able to simplify now by using ResourceProvider.initResourceProvider() which is already doing the detection using the env vars (ok to follow-up on this after the release)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -385,12 +385,66 @@ private static Set<Feature> initStatsbeatFeatureSet(Configuration config) {
...
+      featureList.add(Feature.PREVIEW_GENERAL_EXPORT_QUEUE_CAPACITY);
+    }
+    if (config.preview.metricsExportQueueCapacity != 65536) {
+      featureList.add(Feature.PREVIEW_METRICS_EXPORT_QUEUE_CAPACITY);
+    }
+    if (config.preview.diskPersistenceMaxSizeMb != 50) {
+      featureList.add(Feature.PREVIEW_DISK_PERSISTENCE_MAX_SIZE_MB);
+    }
+    if (!config.preview.useNormalizedValueForNonNormalizedCpuPercentage) {
+      featureList.add(Feature.PREVIEW_USE_NORMALIZED_VALUE_FOR_NON_NORMALIZED_CPU_PERCENTAGE);
```
trask, line 440: maybe we should name this feature something like **DONT**_USE_NORMALIZED

heyams: sure, it was way too long

smoke-tests/apps/SpringScheduling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/SpringSchedulingTest.java
```diff
@@ -25,11 +25,7 @@
 @UseAgent
 abstract class SpringSchedulingTest {
 
-  @RegisterExtension
-  static final SmokeTestExtension testing =
-      SmokeTestExtension.builder()
-          .setSelfDiagnosticsLevel("debug")
-          .build(); // SmokeTestExtension.create();
+  @RegisterExtension static final SmokeTestExtension testing = SmokeTestExtension.create();
```
trask, line 28: 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
heyams: Updated in azure-sdk PR to PREVIEW_DONT_USE_NORMALIZED

smoke-tests/apps/RoleNameOverrides/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/AksRoleNameOverridesTest.java
```diff
@@ -35,7 +35,13 @@ abstract class AksRoleNameOverridesTest {
   @Test
   @TargetUri("/testAks")
   void test() throws Exception {
-    verifyRoleNameAndInstance("test-deployment-name", "test-pod-name");
+    String jbossHome = System.getenv("JBOSS_HOME");
+    if (jbossHome != null) {
+      // jboss/wildfly has a default service.name="jboss-module"
```
trask, line 40: 😮

heyams: upstream added it correct? it didn't have it before though

trask: oh, I understand. Upstream has some service name detectors, and now with latest azure-sdk changes, we're factoring in otel resource `service.name` into role name. this could be breaking change for Java agent users. I think we should override exactly which (if any) resource providers (detectors) we want to include in our Java agent distro

trask: see https://github.com/open-telemetry/opentelemetry-java/blob/e06d35add577fd702104b5b2836201e76f72d88b/sdk-extensions/autoconfigure/README.md?plain=1#L203

trask: we may need just one resource provider: `io.opentelemetry.sdk.autoconfigure.internal.EnvironmentResourceProvider`

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java
heyams: will do that after. require a newer version of exporter if we want to move logic to ResourceProvider and reuse it everywhere.

heyams: added a todo

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -385,12 +385,66 @@ private static Set<Feature> initStatsbeatFeatureSet(Configuration config) {
...
+      featureList.add(Feature.PREVIEW_GENERAL_EXPORT_QUEUE_CAPACITY);
+    }
+    if (config.preview.metricsExportQueueCapacity != 65536) {
+      featureList.add(Feature.PREVIEW_METRICS_EXPORT_QUEUE_CAPACITY);
+    }
+    if (config.preview.diskPersistenceMaxSizeMb != 50) {
+      featureList.add(Feature.PREVIEW_DISK_PERSISTENCE_MAX_SIZE_MB);
+    }
+    if (!config.preview.useNormalizedValueForNonNormalizedCpuPercentage) {
+      featureList.add(Feature.PREVIEW_DONT_USE_NORMALIZED);
```
trask, line 440: sorry, I think I was too brief on my prior comment, we should probably make this a bit clearer (longer 😅), but can do that later

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
heyams: it's just a way to identify the preview feature. do you prefer PREVIEW_DONT_USE_NORMALIZED_VALUE_FOR_NON_NORMALIZED_CPU_PERCENTAGE because of the negation? 
this is even longer

trask: ya, it's the negation that I was commenting on

maybe `PREVIEW_REPORT_NON_NORMALIZED_CPU_PERCENTAGE`

since essentially this setting opts in to (correctly) reporting the non-normalized cpu percentage (we're only not enabling it by default yet due to it being a breaking change)

PR #3379, opened 2023-11-09
Change statsbeat to system properties in smoke tests

heyams:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -478,19 +478,17 @@ public Builder setAadAuthentication(Configuration.AadAuthentication aadAuthentic
...
         this.connectionString = null;
         this.statsbeatConnectionString = null;
       } else {
         this.connectionString = ConnectionString.parse(connectionString);
         this.statsbeatConnectionString =
             StatsbeatConnectionString.create(
-                this.connectionString, statsbeatInstrumentationKey, statsbeatEndpoint);
+                this.connectionString,
+                System.getProperty("statsbeat.ikey"),
+                System.getProperty("statsbeat.endpoint"));
```
trask, line 491: we have a couple system properties already under `applicationinsights.testing.*`, how about
* `applicationinsights.testing.statsbeat.ikey`
* `applicationinsights.testing.statsbeat.endpoint`

trask: also, looks like we could simplify a bit and make StatsbeatConnectionString.create() method only take a single param, and handle the system property inside that method

heyams: yes, that will need to be changed in the standalone exporter.

heyams: maybe don't change it yet.. since it's only for smoke test?

trask (approved):

PR #3380, opened 2023-11-10
Set statsbeat disk storage to only 1mb

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -288,17 +287,17 @@ public void customize(AutoConfigurationCustomizer autoConfiguration) {
...
+    TelemetryPipeline telemetryPipeline = new TelemetryPipeline(httpPipeline, null);
 
     TelemetryPipelineListener telemetryPipelineListener;
     if (tempDir == null) {
       telemetryPipelineListener = new StatsbeatTelemetryPipelineListener(statsbeatModule::shutdown);
     } else {
       LocalStorageTelemetryPipelineListener localStorageTelemetryPipelineListener =
           new LocalStorageTelemetryPipelineListener(
-              diskPersistenceMaxSizeMb,
+              1, // only store at most 1mb of statsbeat telemetry
```
heyams, line 300: 1MB is sufficient for Statsbeat. 

(24 * 60 / 15 (network per 15 mins) + 3 (attach + feature + instrumentation per 24 hours)) * 362 bytes (avg custom metrics billable size)) = around 36 KB per day, purge allows up to 2 days.. 72 KB max.. unless there is constant failure every 1 min.. then that is the worst case. 1MB still good enough.

heyams (approved):

PR #3388, opened 2023-11-15
Support non-string type telemetry processor attributes

### LINKED ISSUE #3378: Telemetry processors bug
### Expected behavior
Include filter with non-string value should be correctly applied.

### Actual behavior
Currently only string filter fields are supported. 

### To Reproduce

Create such applicationinsights.json file:
```
{
    "preview": {
        "processors": [
            {
                "type": "attribute",
                "include": {
                    "matchType": "regexp",
                    "attributes": [
                        {
                            "key": "http.status_code",
                            "value": 400
                        }
                    ]
                },
                "actions": [
                    {
                        "action": "insert",
                        "key": "key",
                        "value": "value"
                    }
                ]
            }
        ]
    }
}
```

As "http.status_code" is not a string key (`AttributeKey.stringKey`) but rather long (`AttributeKey.longKey`), it will not be resolved, and this code will return null instead of actual value.

https://github.com/microsoft/ApplicationInsights-Java/blob/547ec61e8cfe8b7dbe77de73c8c53d51c03377f5/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java#L88

Same for regexp:

https://github.com/microsoft/ApplicationInsights-Java/blob/547ec61e8cfe8b7dbe77de73c8c53d51c03377f5/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java#L126

https://github.com/microsoft/ApplicationInsights-Java/blob/547ec61e8cfe8b7dbe77de73c8c53d51c03377f5/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java#L179

heyams:
Fix #3378


heyams:
need to fix test

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -935,7 +935,7 @@ public void validate(ProcessorType processorType, IncludeExclude includeExclude)
                   + " processors here: https://go.microsoft.com/fwlink/?linkid=2151557");
         }
         if (matchType == MatchType.REGEXP && attribute.value != null) {
-          validateRegex(attribute.value, processorType);
+          validateRegex(String.valueOf(attribute.value), processorType);
```
trask, line 938: we should probably use same format we use in the exporter:

https://github.com/Azure/azure-sdk-for-java/blob/0f379118d823884583bd43f1aa4f5e6ac9decab9/sdk/monitor/azure-monitor-opentelemetry-exporter/src/main/java/com/azure/monitor/opentelemetry/exporter/implementation/Mappings.java#L58

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1094,7 +1094,30 @@ private static void validateRegex(String value, ProcessorType processorType) {
 
   public static class ProcessorAttribute {
     public String key;
-    public String value;
+    public Object value;
+    public AttributeType type;
```
trask, line 1098: I think we need to default to STRING for backwards compatibility

smoke-tests/apps/TelemetryProcessors/src/main/java/com/microsoft/applicationinsights/smoketestapp/TestController.java
```diff
@@ -33,4 +34,12 @@ public String sensitiveData() {
   public String sensitiveDataInUrl() {
     return "Test sensitive data in URL";
   }
+
+  @GetMapping("/test-non-string-span-attributes")
+  public String testNonStringSpanAttributes() {
+    Span.current()
+        .setAttribute("myLongAttributeKey", 1234)
+        .setAttribute("myBooleanAttributeKey", true);
```
trask, line 42: can you add a list (array) attribute also?

smoke-tests/apps/TelemetryProcessors/src/smokeTest/resources/applicationinsights-non-string-span-attributes.json
```diff
@@ -0,0 +1,38 @@
...
+  "preview": {
+    "processors": [
+      {
+        "type": "attribute",
+        "include": {
+          "matchType": "strict",
+          "attributes": [
+            {
+              "key": "myLongAttributeKey",
+              "value": 1234
```
trask, line 18: don't we need this?
```suggestion
              "value": 1234,
              "type": "long"
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: default to String type.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: https://github.com/Azure/azure-sdk-for-java/pull/37712

heyams: probably no. that will be super complicated: regex on nonstring.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java
```diff
@@ -85,15 +90,21 @@ public boolean isMatch(Attributes attributes, String name) {
...
           // user specified key not found
           return false;
         }
-        if (attribute.value != null && !existingAttributeValue.equals(attribute.value)) {
+        String existingAttributeValue;
+        if (attribute.type == BOOLEAN_ARRAY
+            || attribute.type == LONG_ARRAY
+            || attribute.type == DOUBLE_ARRAY
+            || attribute.type == STRING_ARRAY) {
+          existingAttributeValue = valueObject.toString();
```
jeanbisutti, line 103: could you please add a comment explaining why `String.valueOf` would not work in this case

smoke-tests/apps/TelemetryProcessors/src/smokeTest/resources/applicationinsights-non-string-span-attributes.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "sampling": {
+    "percentage": 100
+  },
+  "preview": {
+    "processors": [
+      {
+        "type": "attribute",
+        "include": {
+          "matchType": "strict",
```
jeanbisutti, line 14: From my understanding, the Github issue is related to regex?  It my be worth having `"matchType": "regexp"` in the test?

heyams: but based on his json, it should be strict.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java
heyams: no diff. updated.

smoke-tests/apps/TelemetryProcessors/src/smokeTest/resources/applicationinsights-non-string-span-attributes.json
heyams: we don't support non string regex.

heyams:
https://github.com/microsoft/ApplicationInsights-Java/issues/3378#issuecomment-1832658810
customer confirmed both strict and regex work.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/processors/AgentProcessor.java
```diff
@@ -85,15 +85,13 @@ public boolean isMatch(Attributes attributes, String name) {
...
-        // String.valueOf(existingAttributeValue);
-        if (!(existingAttributeValue instanceof String)) {
+        Object valueObject = attributes.get(attribute.getAttributeKey());
+        if (valueObject == null) {
           // user specified key not found
           return false;
         }
-        if (attribute.value != null && !existingAttributeValue.equals(attribute.value)) {
+        String existingAttributeValue = String.valueOf(valueObject);
+        if (attribute.value != null && !existingAttributeValue.equals(attribute.getStringValue())) {
```
trask, line 94: for strict can we do equals on object values? (e.g. comparing double object / boolean object)
```suggestion
        if (attribute.value != null && !valueObject.equals(attribute.getValue())) {
```

heyams: this failed strict type

trask: what was the problem? I think it would be good to see what it would require to make it work

smoke-tests/apps/TelemetryProcessors/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/TelemetryProcessorsNonStringAttributesTest.java
```diff
@@ -0,0 +1,75 @@
...
+  @Test
+  @TargetUri("/test-non-string-strict-span-attributes")
+  void testNonStringStrictSpanAttributes() throws Exception {
+    Telemetry telemetry = testing.getTelemetry(0);
+    Map<String, String> properties = telemetry.rd.getProperties();
+
+    assertThat(properties.get("myLongAttributeKey")).isEqualTo("1234");
+    assertThat(properties.get("myBooleanAttributeKey")).isEqualTo("true");
+    assertThat(properties.get("myDoubleArrayAttributeKey"))
+        .containsAnyOf("1.0", "2.0", "3.0", "4.0");
```
trask, line 35: ?
```suggestion
        .containsExactly("1.0", "2.0", "3.0", "4.0");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1094,7 +1095,44 @@ private static void validateRegex(String value, ProcessorType processorType) {
...
+    public Object getAttributeValue() {
+      if (value instanceof Integer) {
+        if (type == AttributeType.LONG) {
+          return ((Integer) value).longValue();
+        } else if (type == AttributeType.DOUBLE) {
+          return ((Integer) value).doubleValue();
+        }
+      }
+      return value;
+    }
```
trask, line 1135: add `validate()` method to throw exception if `type` and `value` aren't aligned

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1094,7 +1095,80 @@ private static void validateRegex(String value, ProcessorType processorType) {
...
+
+    public void validate() {
+      if (type == AttributeType.STRING && !(value instanceof String)) {
+        throw new FriendlyException(
+            "Attribute type is String, but value is not a string.",
+            "Please provide a valid String attribute value.");
```
trask, line 1141: a bit more info would be helpful for users to pinpoint what the problem in their json file is, e.g.

```suggestion
            "Telemetry processor attribute <name> has type is String, but the value is not a string.",
            "Please provide a json value which matches the attribute type.");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -1094,7 +1095,80 @@ private static void validateRegex(String value, ProcessorType processorType) {
...
+      } else if (type == AttributeType.LONG && !(getAttributeValue() instanceof Long)) {
```
trask, line 1142: I think Json binding will be Integer?

heyams: getAttributeValue() will convert Integer to Long.

trask: oh, right 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -937,7 +937,10 @@ public void validate(ProcessorType processorType, IncludeExclude includeExclude)
         if (matchType == MatchType.REGEXP && attribute.value != null) {
           validateRegex(String.valueOf(attribute.value), processorType);
         }
-        attribute.validate();
+
+        if (matchType == MatchType.STRICT) {
```
trask, line 941:
```suggestion
        // regex matches always convert the value to string first
        if (matchType == MatchType.STRICT) {
```

trask (approved):

PR #3472, opened 2023-12-22
Ability to disable the Reactor and Reactor Netty instrumentations with env variables

### LINKED ISSUE #3470: Add ability to disable reactor instrumentation

Reactor instrumentation doesn't always do a great job. Here're some of the known problems:
- makes exception stacktraces unreadable (TracingSubscriber is added on every reactor operator)
- adds perf overhead
- sometimes propagates context in situations it's not really desireable.

All these could be good reasons to disable reactor instrumentation.
When using vanilla OTel agent, I can disable it with `OTEL_INSTRUMENTATION_REACTOR_ENABLED=false` env var or system property,
but it has no effect on ApplicationInsights agent.

We need an appInsights specific option to disable reactor instrumentation or allow to disable it with OTel one.

jeanbisutti:
Fix #3470

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
@@ -182,10 +182,13 @@ public class AiConfigCustomizer implements Function<ConfigProperties, Map<String
         "otel.instrumentation.opentelemetry-instrumentation-annotations.enabled", "true");
     properties.put("otel.instrumentation.opentelemetry-api.enabled", "true");
     properties.put("otel.instrumentation.opentelemetry-instrumentation-api.enabled", "true");
-    if (otelConfig.getBoolean("otel.instrumentation.reactor.enabled", true)) {
+    if (otelConfig.getBoolean("otel.instrumentation.reactor.enabled", true)
+        && !"false".equalsIgnoreCase(System.getenv("OTEL_INSTRUMENTATION_REACTOR_ENABLED"))) {
       properties.put("otel.instrumentation.reactor.enabled", "true");
     }
-    if (otelConfig.getBoolean("otel.instrumentation.reactor-netty.enabled", true)) {
+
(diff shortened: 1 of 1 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -182,10 +182,13 @@ private static void enableInstrumentations(
         "otel.instrumentation.opentelemetry-instrumentation-annotations.enabled", "true");
     properties.put("otel.instrumentation.opentelemetry-api.enabled", "true");
     properties.put("otel.instrumentation.opentelemetry-instrumentation-api.enabled", "true");
-    if (otelConfig.getBoolean("otel.instrumentation.reactor.enabled", true)) {
+    if (otelConfig.getBoolean("otel.instrumentation.reactor.enabled", true)
+        && !"false".equalsIgnoreCase(System.getenv("OTEL_INSTRUMENTATION_REACTOR_ENABLED"))) {
```
heyams, line 186: what is the equivalent of system property for OTEL_INSTRUMENTATION_REACTOR_ENABLED?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -182,10 +182,13 @@ private static void enableInstrumentations(
...
       properties.put("otel.instrumentation.reactor.enabled", "true");
     }
-    if (otelConfig.getBoolean("otel.instrumentation.reactor-netty.enabled", true)) {
+
+    if (otelConfig.getBoolean("otel.instrumentation.reactor-netty.enabled", true)
+        && !"false".equalsIgnoreCase(System.getenv("OTEL_INSTRUMENTATION_REACTOR_NETTY_ENABLED"))) {
```
heyams, line 191: same here. then you can use otelConfig.getBoolean("<system property>")

jeanbisutti:
Already implemented

PR #3482, opened 2024-01-03
Ability to set the connection string several times

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java
```diff
@@ -67,10 +69,22 @@ public void setConnectionString(String connectionString) {
               + " \"connectionStringConfiguredAtRuntime\" to true");
       return;
     }
-    if (TelemetryClient.getActive().getConnectionString() != null) {
-      logger.warn("Connection string is already set");
-      return;
+    if (connectionStringProgrammaticallySet.getAndSet(true)) {
+      String applicationInsightsConnectionStringEnv =
```
heyams, line 73: Since you make `APPLICATIONINSIGHTS_CONNECTION_STRING_ENV` public, you can import the constant directly. I don't think it's necessary to create a local variable.

heyams: should we just check `!TelemetryClient.getActive().getConnectionString().equals(System.getenv(APPLICATIONINSIGHTS_CONNECTION_STRING_ENV)`?

heyams: static import of constant, you simply use 'APPLICATIONINSIGHTS_CONNECTION_STRING_ENV', a string constant.

jeanbisutti: I don't mind

jeanbisutti: It's just preferences

trask: > should we just check `!TelemetryClient.getActive().getConnectionString().equals(System.getenv(APPLICATIONINSIGHTS_CONNECTION_STRING_ENV)`?

this is similar to my original thought, but

`!TelemetryClient.getActive().getConnectionString().equals(connectionString)`

jeanbisutti: The current verification of the connection string value is applied [only once](https://github.com/microsoft/ApplicationInsights-Java/blob/f15c27664daacb6a02d7ce47812693d01371c5b6/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java#L72) to avoid many logs in cases where the application often changes the connection string. The connection string from the environment variable could be equal to the programmatic connection string value when the connection is programmatically set the first time but not  the following times.  For these reasons, the current version verifies that a connection string from an environment variable exists.

trask:
```
if (setConnectionStringFirstTimeCalled && TelemetryClient.getActive().getConnectionString() != null) {
  // info log that programmatic call is overriding value from env var / json
}
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java
```diff
@@ -67,10 +69,15 @@ public void setConnectionString(String connectionString) {
               + " \"connectionStringConfiguredAtRuntime\" to true");
       return;
     }
-    if (TelemetryClient.getActive().getConnectionString() != null) {
-      logger.warn("Connection string is already set");
-      return;
+
+    if (connectionStringProgrammaticallySet.getAndSet(true)
```
trask, line 73: I forget, is the goal to log only the first time? if so, should this be:
```suggestion
    if (!connectionStringProgrammaticallySet.getAndSet(true)
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java
```diff
@@ -67,10 +69,15 @@ public void setConnectionString(String connectionString) {
               + " \"connectionStringConfiguredAtRuntime\" to true");
       return;
     }
-    if (TelemetryClient.getActive().getConnectionString() != null) {
-      logger.warn("Connection string is already set");
-      return;
+
+    if (!connectionStringProgrammaticallySet.getAndSet(true)
```
trask, line 73: maybe easier to read with something like `private static final AtomicBoolean firstCall = new AtomicBoolean(true);`
```suggestion
    if (firstCall.getAndSet(false)
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java
```diff
@@ -67,10 +69,15 @@ public void setConnectionString(String connectionString) {
...
+        && TelemetryClient.getActive().getConnectionString() != null) {
+      logger.info(
+          "The connection string is programmatically set. It will take precedence over the value defined from the applicationinsights.json file or the "
+              + APPLICATIONINSIGHTS_CONNECTION_STRING_ENV
+              + " environment variable.");
```
trask, line 78: can use placeholder `{}`

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/classicsdk/BytecodeUtilImpl.java
jeanbisutti: Not sure to follow because no value to change

PR #3533, opened 2024-02-01
Ability to disable auto instrumentations

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -135,6 +135,11 @@ public Map<String, String> apply(ConfigProperties otelConfig) {
 
   private static void enableInstrumentations(
       ConfigProperties otelConfig, Configuration config, Map<String, String> properties) {
+
+    if (!otelConfig.getBoolean("otel.instrumentation.common.default-enabled", true)) {
```
trask, line 139:
```suggestion
    if (otelConfig.getBoolean("applicationinsights.testing.instrumentation.disabled")) {
```

```suggestion
    if (otelConfig.getBoolean("applicationinsights.internal.instrumentation.disabled")) {
```

trask (approved):
we may want to add some of these

ivan-zaitsev:
@trask Is it documented feature or only for testing and internal usage?

PR #3560, opened 2024-02-28
Enable ETW in integrated_auto only

heyams:


trask (commented):
I think we should instead change the definition (and name?) of `isAppSvcRpIntegration` ( -> `isAppSvcRpAutoIntegration`)

It looks like there are a couple of other places where this method is used which also only apply to auto-integration

trask:
also maybe `isFunctionsRpIntegration`?

trask (approved):

PR #3564, opened 2024-03-01
Reduce logging

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -236,7 +236,7 @@ private BatchItemProcessor initBatchItemProcessor(
     if (tempDir == null) {
       telemetryPipelineListener =
           new DiagnosticTelemetryPipelineListener(
-              "Sending telemetry to the ingestion service", false, " (telemetry will be lost)");
+              "Sending telemetry to the ingestion service", true, " (telemetry will be lost)");
```
trask, line 239: this looks like a bug

jeanbisutti (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
heyams: tempDir is null.. written to disk is not possible. retries will be false. i don't get it why it becomes true.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -245,8 +245,8 @@ private BatchItemProcessor initBatchItemProcessor(
               // will log if that retry from disk fails
               new DiagnosticTelemetryPipelineListener(
                   "Sending telemetry to the ingestion service",
-                  true,
-                  " (telemetry will be stored to disk and retried)"),
+                  false,
```
heyams, line 248: not get it why it's false here.

trask: so that we only log the 2nd time a batch fails (when retrying from disk)

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
trask: 2nd arg controls logging, not writing to disk

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -236,17 +236,15 @@ private BatchItemProcessor initBatchItemProcessor(
     if (tempDir == null) {
       telemetryPipelineListener =
           new DiagnosticTelemetryPipelineListener(
-              "Sending telemetry to the ingestion service", false, " (telemetry will be lost)");
+              "Sending telemetry to the ingestion service", true, " (telemetry will be lost)");
     } else {
       telemetryPipelineListener =
           TelemetryPipelineListener.composite(
               // suppress warnings on retryable failures, in order to reduce sporadic/annoying
```
heyams, line 243: please update the comment.. this is not what it's actually doing

trask: I think the comment is correct, can you double check?

PR #3578, opened 2024-03-06
Fix ingestion sampling

trask:
Requires https://github.com/Azure/azure-sdk-for-java/pull/39103

TODO add smoke test


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -168,6 +169,12 @@ public static class Sampling {
     @Deprecated @Nullable public Double limitPerSecond;
 
     public List<SamplingOverride> overrides = new ArrayList<>();
+
+    private void validate() {
+      for (SamplingOverride samplingOverride : overrides) {
+        samplingOverride.validate();
+      }
+    }
```
trask, line 177: validation of the GA'd config was missing

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -257,6 +257,13 @@ private static void logConfigurationWarnings(Configuration config) {
       configurationLogger.warn(
           "\"clientsecret\" typed of AAD authentication has been deprecated since 3.5.0 GA. Please use \"user-assigned identity\" or \"system-assigned identity\" instead.");
     }
+    if (config.sampling.percentage != null && config.sampling.requestsPerSecond != null) {
+      configurationLogger.warn(
+          "Sampling \"requestsPerSecond\" and \"percentage\" should not be used at the same time."
+              + " Please remove one of them.");
+      config.sampling.percentage = null; // requestsPerSecond takes priority
+    }
```
trask, line 265: this was missing from 3.4.0+

don't want to add this validation to `Configuration`, since we can only throw exception there, and someone could have misconfigured in the meantime

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
heyams: good catch

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
heyams: not sure if that is a valid customer scenario though.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingPercentage.java
```diff
@@ -22,4 +24,8 @@ static SamplingPercentage fixed(double percentage) {
   static SamplingPercentage rateLimited(double targetPerSecondLimit) {
     return new RateLimitedSamplingPercentage(targetPerSecondLimit, 0.1);
   }
+
+  static SamplingPercentage ingestion() {
```
heyams, line 28: can we rename this to ingesitonSampling? `ingestion` is overuse everywhere.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingPercentage.java
```diff
@@ -22,4 +24,8 @@ static SamplingPercentage fixed(double percentage) {
...
+    return () -> USE_INGESTION_SAMPLING;
```
heyams, line 29:
```suggestion
    return () -> INGESTION_SAMPLING_PERCENTAGE;
```

heyams: it's not boolean return type.

heyams:
can we test https://github.com/Azure/azure-sdk-for-java/pull/39103 in this PR? i didn't find azure-sdk snapshot.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
trask: it's not, but don't want to fail the agent startup suddenly with new release if they had misconfigured it (but it was working) previously

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingPercentage.java
trask: not following?

.github/workflows/build-common.yml
```diff
@@ -81,6 +81,15 @@ jobs:
           distribution: temurin
           java-version: 17
 
+      - run: |
+          mvn -version
+          # double quotes are needed on Windows (https://stackoverflow.com/a/16349133/295416)
+          mvn install:install-file -Dfile="azure-monitor-opentelemetry-exporter-1.0.0-beta.20.jar" \
```
heyams, line 87: make the whole statement in one line works for me.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplingPercentage.java
heyams: it's ok since you rename the method name to match with the constant.

heyams (approved):

PR #3603, opened 2024-03-14
Fix OWASP $GITHUB_WORKFLOW permission issue

heyams:


.github/workflows/owasp-dependency-check-daily.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   analyze:
     runs-on: ubuntu-latest
-
+    permissions: write-all
```
trask, line 13: I don't think we need the permission on this part
```suggestion

```

trask (approved):

.github/workflows/owasp-dependency-check-daily.yml
heyams: let me run it again without it. the other one is green

heyams:
analyze will fail whenever there is a violation.

Issue #3414, opened 2023-11-30
CVE-2023-6378 


dreamglobe:
### Expected behavior
not to include shaded dependencies inside jar

### Actual behavior
shaded dependency has a vulnerability and I cannot fix it

### To Reproduce

Run trivy check and it will alert:

Java (jar)

Total: 3 (HIGH: 3, CRITICAL: 0)

┌─────────────────────────────────────────────────┬─────────────────────┬──────────┬────────┬───────────────────┬────────────────┬─────────────────────────────────────────────────────────────┐
│                     Library                     │    Vulnerability    │ Severity │ Status │ Installed Version │ Fixed Version  │                            Title                            │
├─────────────────────────────────────────────────┼─────────────────────┼──────────┼────────┼───────────────────┼────────────────┼─────────────────────────────────────────────────────────────┤
│ ch.qos.logback:logback-classic                  │ CVE-2023-6378       │ HIGH     │ fixed  │ 1.2.12            │ 1.3.12, 1.4.12 │ A serialization vulnerability in logback receiver component │
│ (applicationinsights-runtime-attach-3.4.11.jar) │                     │          │        │                   │                │ part of l ......                                            │
│                                                 │                     │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2023-6378                   │
├─────────────────────────────────────────────────┤                     │          │        │                   │                │                                                             │
│ ch.qos.logback:logback-core                     │                     │          │        │                   │                │                                                             │
│ (applicationinsights-runtime-attach-3.4.11.jar) │                     │          │        │                   │                │                                                             │
│                                                 │                     │          │        │                   │                │                                                             │
├─────────────────────────────────────────────────┼─────────────────────┤          │        ├───────────────────┼────────────────┼─────────────────────────────────────────────────────────────┤
│ io.netty:netty-codec-http2                      │ GHSA-xpw8-rcwv-8f8p │          │        │ 4.1.90.Final      │ 4.1.100.Final  │ io.netty:netty-codec-http2 vulnerable to HTTP/2 Rapid Reset │
│ (applicationinsights-runtime-attach-3.4.11.jar) │                     │          │        │                   │                │ Attack                                                      │
│                                                 │                     │          │        │                   │                │ https://github.com/advisories/GHSA-xpw8-rcwv-8f8p           │
[...]

PR #3605, opened 2024-03-14
Fix CVE-2023-52428

### LINKED ISSUE #3599: CVE-2023-52428
### Expected behavior
We expect no issues detected when .jar file is scanned with Sonatype Nexus. ApplicationInsights .jar file should include non-vulnerable library versions of Connect2id Nimbus JOSE+JWT (versions before 9.37.2 are vulnerable to CVE-2023-52428 according to https://nvd.nist.gov/vuln/detail/CVE-2023-52428). 

### Actual behavior
Our Sonatype Nexus detects CVE-2023-52428 in the ApplicationInsights .jar file (versions affected 3.2.0-BETA to latest 3.5.1) with root cause:
`applicationinsights-agent-3.5.1.jarinst/com/nimbusds/jose/crypto/PasswordBasedDecrypter.classdata[4.0-rc1, 9.37.2)`

**Description from CVE**
>   In Connect2id Nimbus JOSE+JWT before 9.37.2, an attacker can cause a denial of service (resource consumption) via a large JWE p2c header value (aka iteration count) for the PasswordBasedDecrypter (PBKDF2) component.

**Explanation**
> The nimbus-jose-jwt package is vulnerable to Denial of Service (DoS) attacks. The decrypt() method in the PasswordBasedDecrypter class fails to properly validate the length of the JWE p2c header. A remote attacker can exploit this vulnerability by supplying an oversized PBES2Count value, causing the application to consume all available resources and ultimately leading to a DoS condition.

### To Reproduce

Perform a Sonatype Nexus scan on the ApplicationInsights .jar file or a Docker image file that includes the ApplicationInsights .jar file.

### System information
Application Insights Java 3.5.1 (GA)

### Logs
None applicable

### Screenshots
image

heyams:
Fixed #3599, https://github.com/microsoft/ApplicationInsights-Java/issues/3414#issuecomment-1966066991


lmolkova (approved):
Thanks @heyams , using newest msal4j is indeed the best course of action today. New azure-identity version that brings it will be shipped in April.

It would be nice to add a comment/create a follow up issue to remove the workaround once new identity ships.

trask (approved):

dependencyManagement/build.gradle.kts
```diff
@@ -65,6 +65,9 @@ val CORE_DEPENDENCIES = listOf(
   "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${otelInstrumentationAlphaVersion}",
   // temporarily overriding transitive dependency from azure-core until next azure-bom release
   // which targets at least reactor-netty-http:1.1.1
+  // fix CVE-2023-52428 https://github.com/advisories/GHSA-gvpg-vgmx-xg6w
+  // TODO (heya) remove this once azure-identity is updated in April
+  "com.microsoft.azure:msal4j:1.14.3",
   "io.projectreactor.netty:reactor-netty-http:1.1.17",
```
trask, line 71: looks like the new lines split the above comment from it's target

trask, line 71: looks like the new lines split the above comment from it's target

heyams: it was from a merge conflict.

heyams: it's currently using io.projectreactor.netty:reactor-netty-http:**1.1.15**, should we remove the comment above? 

```
// temporarily overriding transitive dependency from azure-core until next azure-bom release
  // which targets at least reactor-netty-http:1.1.1
```

trask: I think it would be ok to remove both the comment and the pinned dependency version

ddduarte:
@heyams Hi! Do you have an idea when a new lib version  will be released?
Thanks.

PR #3649, opened 2024-04-15
Update OpenTelemetry depencencies

trask:


heyams (approved):
we might want to update standalone exporter.

PR #3663, opened 2024-04-23
Legacy SDK - Remove Java 7 check

jeanbisutti:


trask (approved):

trask:
I think this is ok since we still have Java 7 smoke tests running against this (legacy sdk) branch

PR #3683, opened 2024-05-04
feat: unique id is required for custom availability

opensourcevk:


classic-sdk/core/src/main/java/com/microsoft/applicationinsights/internal/schemav2/AvailabilityData.java
```diff
@@ -4,12 +4,13 @@
...
 
 import com.microsoft.applicationinsights.telemetry.Duration;
+import javax.annotation.Nonnull;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
 public class AvailabilityData {
 
-  private String id;
+  @Nonnull private String id;
```
opensourcevk, line 13: I am tempted to set id by default
```
Private String id = UUID.randomUUID().toString()
```

heyams: this class is not final and you can extend it and have your own id default to UUID. Does it work for you?

opensourcevk: Yes, it will work. However I was thinking if an agent can take care of creating a unique id, it will be easier for the end user.

heyams: Some end users might not like to use UUID though. Will close the PR. Thanks.

PR #3700, opened 2024-05-17
Display a message about sampling at start-up

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -178,6 +178,13 @@ public void onStartupSuccess() {
     String javaHome = System.getProperty("java.home");
     startupLogger.info("Java version: {}, vendor: {}, home: {}", javaVersion, javaVendor, javaHome);
 
+    boolean userHasChangedDefaultSampling =
+        configuration.sampling.percentage == null && configuration.sampling.requestsPerSecond == 5;
```
trask, line 182: if we log this in ConfigurationBuilder, we could avoid logging it if user explicitly sets `requestsPerSecond` to `5` (e.g. in case they want to remove the message)

jeanbisutti: I have moved the message display in `ConfigurationBuilder`.

The other advantage is that the sampling message is now the first one displayed by Application Insights:

```
2024-05-17 17:36:52.977+02:00 INFO  c.m.a.a.i.c.ConfigurationBuilder - Some telemetry data may miss. Indeed, Application Insights applies by default from the 3.4.0 a sampling to reduce the cost. You can change the default sampling configuration: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config
2024-05-17 17:36:54.690+02:00 INFO  c.m.applicationinsights.agent - Application Insights Java Agent 3.5.3-SNAPSHOT started successfully (PID 26512, JVM running for 2.805 s)
2024-05-17 17:36:54.691+02:00 INFO  c.m.applicationinsights.agent - Java version: 21, vendor: Oracle Corporation, 
```
The log level is info for the sampling log.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -301,6 +301,8 @@ private static void overlayConfiguration(
     // only fall back to default sampling configuration after all overlays have been performed
     if (config.sampling.requestsPerSecond == null && config.sampling.percentage == null) {
       config.sampling.requestsPerSecond = 5.0;
+      configurationLogger.info(
+          "Some telemetry may be sampled out because the default rate-limited sampling configuration has been applied starting from 3.4.0 in order to reduce default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config");
```
trask, line 305:
```suggestion
          "Some telemetry may be sampled out because the default rate-limited sampling configuration has been applied (from version 3.4.0) in order to reduce the default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config");
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -301,6 +301,8 @@ private static void overlayConfiguration(
     // only fall back to default sampling configuration after all overlays have been performed
     if (config.sampling.requestsPerSecond == null && config.sampling.percentage == null) {
       config.sampling.requestsPerSecond = 5.0;
+      configurationLogger.info(
+          "Some telemetry may be sampled out because the default rate-limited sampling configuration has been applied (from version 3.4.0) in order to reduce the default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config#sampling");
```
trask, line 305:
```suggestion
          "Some telemetry may be sampled out because the default rate-limited sampling configuration is being applied in order to reduce the default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config#sampling");
```

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -301,6 +301,8 @@ private static void overlayConfiguration(
     // only fall back to default sampling configuration after all overlays have been performed
     if (config.sampling.requestsPerSecond == null && config.sampling.percentage == null) {
       config.sampling.requestsPerSecond = 5.0;
+      configurationLogger.info(
+          "Some telemetry may be sampled out because a default sampling configuration was added in version 3.4.0 to reduce the default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config#sampling");
```
heyams, line 305: we might want to update this message based on https://github.com/microsoft/ApplicationInsights-Java/pull/3713. 
from 3.5.3, customer can disable ingestion sampling?

trask: this message is only emitted when customer has no sampling configuration (and so the default sampling configuration is used)

Issue #3707, opened 2024-05-20
APPLICATIONINSIGHTS_CONNECTION_STRING does not work in 3.5.x, works in 3.4.x


mercer:
### Expected behavior
APPLICATIONINSIGHTS_CONNECTION_STRING works in 3.5.x

### Actual behavior

APPLICATIONINSIGHTS_CONNECTION_STRING does not work in 3.5.x (tried 3.5.0, 3.5.1, 3.5.2)

### To Reproduce

1. add dependency com.microsoft.azure:applicationinsights-runtime-attach:3.4.19
2. add applicationinsights.json in src/resources, see bellow
3. provide connection string via APPLICATIONINSIGHTS_CONNECTION_STRING environment variable
4. attach to agent via ApplicationInsights.attach();
5. start app
6. agent connects to appinsights, traces are sent to apinsights; we had this setup for more than 1 year
7. upgrade to com.microsoft.azure:applicationinsights-runtime-attach:3.5.2
8. start app (no other changes, no changes to applicationinsights.json)
9. agent does not connect, see error bellow in app console

```
*************************
Application Insights Java Agent 3.5.2 startup failed (PID 46492)
*************************

Description:
Error loading connection string from a file ("unknown-from-application-insights-json").
Please use this format instead:
{ "connectionString": "${file:connection-string-file.txt}" }


Action:
Learn more about configuration options here: https://go.microsoft.com/fwlink/?linkid=2153358

```

### System information
Please provide the following information:
 - SDK Version: 3.5.x
 - OS type and version: Mac M2
 - Application Server type and version (if applicable):
 - Using spring-boot? yes
 - Additional relevant libraries (with version, if applicable):

### Logs
```
2023-05-30 17:02:01.940+03:00 ERROR c.m.applicationinsights.agent - Application Insights Java Agent 3.4.13 startup failed (PID 20200)
java.lang.IllegalArgumentException: Missing 'InstrumentationKey'
	at com.azure.monitor.opentelemetry.exporter.implementation.configuration.ConnectionStringBuilder.mapToConnectionConfiguration(ConnectionStringBuilder.java:85)
	at com.azure.monitor.opentelemetry.exporter.implementation.configuration.ConnectionStringBuilder.setConnectionString(ConnectionStringBuilder.java:39)
	at com.azure.monitor.opentelemetry.exporter.implementation.configuration.ConnectionString.lambda$parse$0(ConnectionString.java:37)
[...]

PR #3709, opened 2024-05-20
Fix misleading connection string message

jeanbisutti:
The following error messages were displayed for cases not related to a misconfiguration of a connection string from a file:

`Error loading connection string from a file`

```
Please use this format instead:
{ "connectionString": "${file:connection-string-file.txt}" }
```

For example `totoInstrumentationKey` for a connection string creates these error messages.

It can be misleading for the users.

Related to #3707

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
@@ -743,7 +743,7 @@ public class ConfigurationBuilder {
         && !replacedConnectionString.startsWith("InstrumentationKey=")
         && config.connectionString.equals(replacedConnectionString)) {
       throw new FriendlyException(
-          "You connection string seems to have a wrong format: \""
+          "Your connection string seems to have a wrong format: \""
               + config.connectionString
               + "\").\n"
               + "If you want to load the connection string from a file, please use this format:"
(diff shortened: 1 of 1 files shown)


trask (approved):
👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -743,10 +743,10 @@ static void overlayFromEnv(
         && !replacedConnectionString.startsWith("InstrumentationKey=")
         && config.connectionString.equals(replacedConnectionString)) {
       throw new FriendlyException(
-          "Error loading connection string from a file (\""
+          "You connection seems to have a wrong format: \""
               + config.connectionString
               + "\").\n"
-              + "Please use this format instead:"
+              + "If you want to read the connection string from a file, please use this format:"
```
heyams, line 749:
```suggestion
              + "If you want to load the connection string from a file, please use this format:"
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -743,10 +743,10 @@ static void overlayFromEnv(
         && !replacedConnectionString.startsWith("InstrumentationKey=")
         && config.connectionString.equals(replacedConnectionString)) {
       throw new FriendlyException(
-          "Error loading connection string from a file (\""
+          "You connection seems to have a wrong format: \""
```
heyams, line 746:
```suggestion
          "You connection string seems to have a wrong format: \""
```

PR #3710, opened 2024-05-20
Disable _otelresource_ custom metric in non-AKS envs

trask:
test https://github.com/Azure/azure-sdk-for-java/pull/40312

PR #3713, opened 2024-05-21
Introduce explicit preview option to disable ingestion sampling

trask:
relies on https://github.com/Azure/azure-sdk-for-java/pull/40338


heyams (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -269,12 +269,18 @@ private static void logConfigurationWarnings(Configuration config) {
...
+    if (config.sampling.requestsPerSecond != null && config.sampling.percentage != null) {
       configurationLogger.warn(
           "Sampling \"requestsPerSecond\" and \"percentage\" should not be used at the same time."
               + " Please remove one of them.");
       config.sampling.percentage = null; // requestsPerSecond takes priority
     }
+    if (config.sampling.requestsPerSecond != null && config.sampling.ingestion != null) {
+      configurationLogger.warn(
+          "Sampling \"requestsPerSecond\" and \"ingestion\" should not be used at the same time."
+              + " Please remove one of them.");
```
jeanbisutti, line 281:
```suggestion
              + " Only \"requestsPerSecond\" is taken into account. Please remove \"ingestion\" or \"requestsPerSecond\" from your configuration.");
```
Same type of comment for the previous `if`.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/Samplers.java
```diff
@@ -21,12 +21,15 @@ public static Sampler getSampler(
       sampler = new AiSampler(requestSamplingPercentage, parentlessDependencySamplingPercentage);
     } else if (sampling.percentage != null) {
       SamplingPercentage samplingPercentage;
-      if (sampling.percentage == 100) {
+      if (sampling.percentage == 100 && !Boolean.FALSE.equals(sampling.ingestion)) {
```
jeanbisutti, line 24: At line 173, `ingestion` may be `FALSE` (`Boolean ingestion = Boolean.FALSE)`? In this way, line 24 could become 
```suggestion
      if (sampling.percentage == 100 && Boolean.TRUE.equals(sampling.ingestion)) {
```
Which could be easier to read

smoke-tests/apps/Sampling/src/smokeTest/resources/applicationinsights-no-ingestion-sampling.json
```diff
@@ -0,0 +1,10 @@
+{
+  "role": {
+    "name": "testrolename",
+    "instance": "testroleinstance"
+  },
+  "sampling": {
+    "percentage": 100,
+    "ingestionSide": false
```
jeanbisutti, line 8: It may be more explict from a customer perspective:

```suggestion
    "portailIngestion": false
```

trask: i'm not sure about "portal" since portal is technical the U/X (I think)

what do you think of just `"ingestion"`?

here's the relevant docs:
https://learn.microsoft.com/en-us/azure/azure-monitor/app/sampling-classic-api#ingestion-sampling

jeanbisutti: I am OK with that. 

We could provide the doc link (https://learn.microsoft.com/en-us/azure/azure-monitor/app/sampling-classic-api#ingestion-sampling) in the documentation of the `ingestion` property.

heyams: can we stick with product's term: ingestion sampling. 
I would suggest "ingestionSamplingEnabled" default is on?

trask: btw, note that I've put this directly under the stable `"sampling"` config node

another option is to put it under the preview node

```
{
  "preview": {
    "ingestionSampling": false
  }
}
```

heyams: i think preview is better. sampling is applicable to java agent's sampling rule.

trask: > sampling is applicable to java agent's sampling rule.

I didn't understand this comment

there are three different types of sampling (fixed-rate, adaptive and ingestion) https://learn.microsoft.com/en-us/azure/azure-monitor/app/sampling-classic-api

so I think it makes sense to keep them together (unless we want to preview the config option first)

heyams: i meant java agent sampling rates vs. ingestion's sampling rate

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
trask: this code is gone now after rewrite

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/Samplers.java
trask: this code is gone now after rewrite

smoke-tests/apps/Sampling/src/smokeTest/resources/applicationinsights-no-ingestion-sampling.json
trask: resolving this conversation as everything has changed now, ptal 😅

trask:
ConnectionStringOverrides and RoleNameOverrides smoke tests should pass once #3710 is merged (since this PR is using latest azure sdk snapshot)

jeanbisutti (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSampler.java
```diff
@@ -112,9 +125,10 @@ private static SamplingResult useLocalParentDecisionIfPossible(Context parentCon
       return SamplingResult.drop();
     }
     if (parentSpan instanceof ReadableSpan) {
-      Long itemCount = ((ReadableSpan) parentSpan).getAttribute(AiSemanticAttributes.ITEM_COUNT);
-      if (itemCount != null) {
-        return new RecordAndSampleWithItemCount(itemCount);
+      Double sampleRate =
```
jeanbisutti, line 128:
```suggestion
      Double parentSampleRate =
```

At first, I have not noticed it was the sample rate from the parent span

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/Configuration.java
```diff
@@ -202,6 +202,8 @@ public static class SamplingPreview {
     // future goal: make parentBased sampling the default if item count is received via tracestate
     public boolean parentBased;
 
+    public boolean suppressIngestionSampling;
```
heyams, line 205: can we rename it to disableIngestionSampling? we use enable/disable in our json config. it's better to be consistent.

trask: renamed to `ingestionSamplingEnabled` (with default value true)

PR #3731, opened 2024-06-04
Add rate limited sampling smoke tests for ingestion sampling

trask:


smoke-tests/apps/RateLimitedSampling/src/smokeTest/resources/applicationinsights-ingestion-sampling-disabled.json
```diff
@@ -0,0 +1,14 @@
...
+  "role": {
+    "name": "testrolename",
+    "instance": "testroleinstance"
+  },
+  "sampling" : {
+    "requestsPerSecond": 100000
+  },
+  "preview": {
+    "sampling": {
+      "ingestionSamplingEnabled": false
```
heyams, line 11: might want to track this via Statsbeat feature as well.

smoke-tests/apps/RateLimitedSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/RateLimitedSamplingIngestionSamplingDisabledTest.java
```diff
@@ -0,0 +1,65 @@
...
+  void testSampling() throws Exception {
+    // give some time to collect
+    Thread.sleep(SECONDS.toMillis(5));
+
+    List<Envelope> rdEnvelopes = testing.mockedIngestion.getItemsEnvelopeDataType("RequestData");
+
+    assertThat(rdEnvelopes.size()).isEqualTo(1000);
+
+    for (Envelope rdEnvelope : rdEnvelopes) {
+      assertThat(rdEnvelope.getSampleRate()).isEqualTo(99.99f);
```
rajkumar-rangaraj, line 38: When sampling is disabled shouldn't this be 100%?

smoke-tests/apps/RateLimitedSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/RateLimitedSamplingIngestionSamplingTest.java
```diff
@@ -0,0 +1,65 @@
...
+  void testSampling() throws Exception {
+    // give some time to collect
+    Thread.sleep(SECONDS.toMillis(5));
+
+    List<Envelope> rdEnvelopes = testing.mockedIngestion.getItemsEnvelopeDataType("RequestData");
+
+    assertThat(rdEnvelopes.size()).isEqualTo(1000);
+
+    for (Envelope rdEnvelope : rdEnvelopes) {
+      assertThat(rdEnvelope.getSampleRate()).isNull();
```
rajkumar-rangaraj, line 38: Shouldn't this have a meaningful value when sampling is enabled? Am I missing something here?

trask: ah, this is about ingestion (server-side) sampling

smoke-tests/apps/RateLimitedSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/RateLimitedSamplingIngestionSamplingDisabledTest.java
trask: IIRC when it's 100 then ingestion sampling will take effect

but then again, I thought ingestion sampling only takes effect when we send no sample rate, so maybe it can be 100, I'll need to re-test this to confirm

harsimar: Looking at this part of code where sample rate is set to 99.99 when sampler returns 100: https://github.com/microsoft/ApplicationInsights-Java/blob/ab07cc3b437e9110bdbdf7a0ea391d6724aec8b0/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplerUtil.java#L30. I do see node.js only setting sample rate attr if the rate returned from sampler isn't 100. Probably for the same reason - service side sampling applies at 100 or empty.

harsimar: @trask you may want to put a comment about why you expect this to be 99.99 in the test so folks don't get confused

smoke-tests/apps/RateLimitedSampling/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/RateLimitedSamplingIngestionSamplingTest.java
harsimar: Can the name of this test be changed to make this more clear? To my understanding, this test asserts that when rate limited sampler is used, everything is sampled in at 100% & ingestion sampling enabled in json config, then we expect the rate to be null?

PR #3772, opened 2024-07-09
Fix broken com.azure.core.http instrumentation

heyams:
It seems that it is required to add InstrumentationPolicy to HTTP client instance. It's added by each library when pipeline is being built via 
`HttpPolicyProviders.addAfterRetryPolicies(policies)`;

ref: https://github.com/Azure/azure-sdk-for-java/blob/1cc65cda356ca8aece1213662ca1939e4a729364/sdk/core/azure-core/src/main/java/com/azure/core/http/policy/HttpPolicyProviders.java#L83


heyams:
test app is https://github.com/Azure-Samples/ApplicationInsights-Java-Samples/pull/458

PR #3774, opened 2024-07-11
Sharing OpenTel Metrics Support 

navsukumar:


.github/workflows/build-common.yml
```diff
@@ -8,7 +8,7 @@ on:
         required: false
         
 env:
-  EXPORTER_VERSION: 1.0.0-beta.25 # to be updated with the latest version
+  EXPORTER_VERSION: 1.0.0-beta.26 # to be updated with the latest version
```
heyams, line 11: you need to add exporter jar to this PR + the pom to the root folder.

heyams: image

like this.. you need 1.0.0-beta.26 instead.

navsukumar: just added the jar and pom file to the root folder.

PR #3776, opened 2024-07-15
Fail if proxy host contains http or https

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -874,6 +874,12 @@ private static Configuration.Proxy overlayProxyFromEnv(
                 + " APPLICATIONINSIGHTS_PROXY environment variable instead which supports passing"
                 + " the username and password, e.g."
                 + " APPLICATIONINSIGHTS_PROXY=https://myuser:mypassword@myproxy:8888");
+        if (proxy.host != null
+            && (proxy.host.startsWith("http") || proxy.host.startsWith("https"))) {
```
trask, line 878: just in case someone has a hostname of `http.example.org` 😄
```suggestion
            && (proxy.host.startsWith("http://") || proxy.host.startsWith("https://"))) {
```

trask (approved):

PR #3788, opened 2024-07-18
Help the user to configure the connection string

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -105,7 +105,8 @@ public void customize(AutoConfigurationCustomizer autoConfiguration) {
     if (Strings.isNullOrEmpty(configuration.connectionString)) {
       if (!configuration.connectionStringConfiguredAtRuntime) {
         throw new FriendlyException(
-            "No connection string provided", "Please provide connection string.");
+            "No connection string provided",
+            "Please provide connection string: https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config");
```
jeanbisutti, line 109: All the ways to configure the things can't be explained in a few words. So, a documentation link is provided.

trask: do we already have a go.microsoft.com link for this url?

trask: https://go.microsoft.com/fwlink/?linkid=2153358

trask (approved):

PR #3791, opened 2024-07-20
Make thread dump file configurable through env var

heyams:
default path:  C:\Users\<username>\AppData\Local\Temp\applicationinsights

now, customer can configure it to anywhere they want via APPLICATIONINSIGHTS_STARTUP_PROFILTER_FILE_PATH.

todo: update wiki when this is released https://github.com/microsoft/ApplicationInsights-Java/wiki/Thread-dump

### DIFF
--- a/agent/agent/src/main/java/com/microsoft/applicationinsights/agent/StartupProfiler.java
+++ b/agent/agent/src/main/java/com/microsoft/applicationinsights/agent/StartupProfiler.java
@@ -17,12 +17,21 @@ import java.nio.file.Files;
 
 final class StartupProfiler {
 
+  private static final String APPLICATIONINSIGHTS_STARTUP_PROFILTER_FILE_PATH =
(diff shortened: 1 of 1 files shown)


agent/agent/src/main/java/com/microsoft/applicationinsights/agent/StartupProfiler.java
```diff
@@ -17,12 +17,21 @@
 
 final class StartupProfiler {
 
+  private static final String APPLICATIONINSIGHTS_STARTUP_PROFILTER_FILE_PATH =
+      "APPLICATIONINSIGHTS_STARTUP_PROFILTER_FILE_PATH";
```
trask, line 21:
```suggestion
  private static final String APPLICATIONINSIGHTS_STARTUP_PROFILER_FILE_PATH=
      "APPLICATIONINSIGHTS_STARTUP_PROFILER_FILE_PATH";
```

trask (approved):

PR #3759, opened 2024-06-25
Don't validate the JSON connection string if APPLICATIONINSIGHT__CONNECTION_STRING

jeanbisutti:
Because the connection string from the APPLICATIONINSIGHT__CONNECTION_STRING environment variable takes precedence over the connection string coming from the JSON file.

Fix #3707

PR #3794, opened 2024-07-23
Check connection string env var before loading from a file

heyams:
Fix https://github.com/microsoft/ApplicationInsights-Java/issues/3707

replaced #3759

regression from https://github.com/microsoft/ApplicationInsights-Java/pull/3471


trask (commented):
I thought we were going to revert the change you made in #3471?

heyams:
> I thought we were going to revert the change you made in #3471?

you cannot simply revert because we support loading from the file feature.

trask (approved):

jeanbisutti (approved):

PR #3796, opened 2024-07-23
Add smoke test showing OOM due to recursive log capture on Wildfly

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/BatchItemProcessor.java
```diff
@@ -169,6 +171,23 @@ private void addItem(TelemetryItem item) {
...
     @Override
     public void run() {
+      // incrementing CallDepth for LoggerProvider causes the OpenTelemetry Java agent logging
+      // instrumentation to back off
+      //
+      // note: recursive log capture is only known to be an issue on Wildfly, because it redirects
+      // System.out back to a logging library which is itself instrumented by the Java agent
+      //
+      // see OutOfMemoryWithDebugLevelTest for repro that will fail without this code
+      CallDepth callDepth = CallDepth.forClass(LoggerProvider.class);
```
heyams, line 181: is it necessary to use CallDepth here? Log makes sense. BatchItemProcessor is used for span and metrics too. 
we passed a copy of BatchItemProcessor into AgentLogExporter.

trask: this method is run in a different thread

heyams: can we add it as comment?

heyams (approved):

PR #3878, opened 2024-09-24
Update Java versions for tests

trask:


trask:
@johnoliver @dsgrieve I disabled the G1 test on Java 21+ so that we can merge this and make sure at least our other tests are passing on 21 and 23

agent/agent-gc-monitor/gc-monitor-tests/src/test/java/com/microsoft/gcmonitortests/VariousCollectorsTest.java
```diff
@@ -6,6 +6,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.condition.JRE.JAVA_11;
 import static org.junit.jupiter.api.condition.JRE.JAVA_13;
+import static org.junit.jupiter.api.condition.JRE.JAVA_17;
```
heyams, line 9: does it need java 21 and java 23 test here since you added java 17?

trask: this import is for an exclusion, see below:

> @EnabledForJreRange(max = JAVA_17)

agent/agent-gc-monitor/gc-monitor-tests/src/test/java/com/microsoft/gcmonitortests/VariousCollectorsTest.java
```diff
@@ -33,6 +34,7 @@ void testParallel() throws Exception {
   }
 
   @Test
+  @EnabledForJreRange(max = JAVA_17)
```
jeanbisutti, line 37:
```suggestion
  @EnabledForJreRange(max = JAVA_17) // see https://github.com/microsoft/ApplicationInsights-Java/pull/3878#issuecomment-2431980712
```

jeanbisutti (approved):

PR #3899, opened 2024-10-09
Improve smoke tests around exceptions

trask:
Improving smoke tests and testing the fix in https://github.com/Azure/azure-sdk-for-java/pull/42241


smoke-tests/apps/SpringScheduling/src/main/java/com/microsoft/applicationinsights/smoketestapp/SpringSchedulingApp.java
```diff
@@ -22,4 +22,9 @@ protected SpringApplicationBuilder configure(SpringApplicationBuilder applicatio
   public void fixedRateScheduler() {
     System.out.println("Hello world.");
   }
+
+  @Scheduled(fixedRate = 100)
+  public void exceptional() {
+    throw new RuntimeException("exceptional");
+  }
```
trask, line 29: we were missing a non-Servlet test that captures an exception on a "request" telemetry

while Servlets do capture an exception in OTel, we suppress the exception if it is identical to the logged exception that is also captured

heyams:
need to wait for the new exporter release.

heyams (approved):

PR #3901, opened 2024-10-09
Remove gradle enterprise plugin

trask:
Trying to resolve one of the S360 issues


heyams:
this is done via #3911

PR #3908, opened 2024-10-14
Display byos message for profiler

jeanbisutti:


johnoliver (approved):

jeanbisutti:
> For the BYOS error, there was a link that was included and opening that link, shows the error message below:

In this case, the user can click on the link contained in the Application Insight message and find https://aka.ms/profiler/docs/configure-byos.

PR #3960, opened 2024-11-14
Add warning message if the Java agent is in the class path

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -118,9 +118,19 @@ public void init(EarlyInitAgentConfig earlyConfig) {
         startupLogger.debug("JAVA_TOOL_OPTIONS: " + System.getenv("JAVA_TOOL_OPTIONS"));
       }
 
+      String classPath = System.getProperty("java.class.path");
+      // class path null with a Spring Boot executable JAR
+      if (classPath != null
+          && classPath.contains(
+              "applicationinsights-agent")) { // JAR name in Maven central, the user could have
+        // renamed it
```
trask, line 126: looks like spotless mangled this a bit

trask (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
trask, line 126:
```suggestion
          // JAR name in Maven central, the user could have renamed it
          && classPath.contains("applicationinsights-agent")) {
```

PR #3976, opened 2024-12-04
Remove user sensitive data from self-diagnostics

jeanbisutti:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java
```diff
@@ -138,6 +140,20 @@ public void init(EarlyInitAgentConfig earlyConfig) {
...
+    return env.entrySet().stream()
+        .filter(entry -> !isSensitiveProperty(entry.getKey()))
+        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
+
+  private static boolean isSensitiveProperty(String key) {
+    return key.contains("user")
+        || key.contains("password")
+        || key.contains("pwd")
+        || key.contains("secret");
```
trask, line 154: maybe worth contains case insensitive?

also, I'd suggest still displaying the env var, but displaying the value as `****`

trask (approved):

PR #4000, opened 2025-01-03
Check Azure Spring Cloud before AKS in RP integration

jeanbisutti:
Related to https://github.com/Azure/azure-sdk-for-java/pull/43465


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java
```diff
@@ -47,15 +47,15 @@ public static void initRpIntegration(Path agentPath) {
       if (RpAttachType.INTEGRATED_AUTO.equals(RpAttachType.getRpAttachType())) {
         appSvcRpIntegratedAuto = true;
       }
+    } else if (!Strings.isNullOrEmpty(
+        System.getenv("APPLICATIONINSIGHTS_SPRINGCLOUD_SERVICE_ID"))) {
+      PropertyHelper.setRpIntegrationChar('s');
```
trask, line 52:
```suggestion
      // Spring Cloud needs to be checked before AKS since it runs on AKS
      PropertyHelper.setRpIntegrationChar('s');
```

trask (approved):
thanks for catching this!

PR #4010, opened 2025-01-08
Service version

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java
```diff
@@ -304,11 +303,6 @@ private static void overlayConfiguration(
       configurationLogger.info(
           "Some telemetry may be sampled out because a default sampling configuration was added in version 3.4.0 to reduce the default billing cost. You can set the sampling configuration explicitly: https://learn.microsoft.com/azure/azure-monitor/app/java-standalone-config#sampling");
     }
-    // only set role instance to host name as a last resort
-    if (config.role.instance == null) {
-      String hostname = HostName.get();
-      config.role.instance = hostname == null ? "unknown" : hostname;
-    }
```
trask, line 311: need to remove this, it's already a fallback provided by the azure monitor opentelemetry autoconfigure module, and by adding it too early here it prevents other (higher priority) fallbacks from getting used

PR #4012, opened 2025-01-08
[POC only] Use upstream detectors

trask:


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -36,7 +36,9 @@ public Map<String, String> apply(ConfigProperties otelConfig) {
     // enables all resource provider by default
     properties.put(
         "otel.java.enabled.resource.providers",
-        "io.opentelemetry.sdk.autoconfigure.internal.EnvironmentResourceProvider");
+        "io.opentelemetry.sdk.autoconfigure.internal.EnvironmentResourceProvider,"
+            + "io.opentelemetry.instrumentation.resources.ManifestResourceProvider,"
+            + "io.opentelemetry.instrumentation.spring.resources.SpringBootServiceVersionDetector");
```
trask, line 41: I'm not sure if we should add these two upstream detectors without a major version bump since these will start populating role name and app version in cases where they were previously empty

PR #4014, opened 2025-01-09
Better sampling override behavior

trask:
(icm 577578458)

Currently, this configuration:

```
      {
        "telemetryType": "request",
        "attributes": [
          {
            "key": "http.url",
            "value": ".*/health-check",
            "matchType": "regexp"
          }
        ],
        "percentage": 0,
        "id": "filter out health check"
      },
      {
        "telemetryType": "dependency",
        "attributes": [
          {
            "key": "db.statement",
            "value": "select count(*) from abc",
            "matchType": "strict"
          }
        ],
        "percentage": 50,
        "id": "filter out noisy jdbc"
      }
```

has the unexpected effect of capturing 50% of `select count(*) from abc` calls, even when they occur within the excluded health check.

this PR updates the behavior so that when a sampling override percentage is <100 and a parent dropped, then the span (with override percentage < 100) will also be dropped


smoke-tests/apps/SamplingOverrides/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/SamplingOverrides3Test.java
```diff
@@ -0,0 +1,109 @@
...
+    assertThat(logCount).isLessThanOrEqualTo(20);
+
+    testing
+        .mockedIngestion
+        .getItemsEnvelopeDataType("RequestData")
+        .forEach(
+            item -> {
+              assertThat(item.getSampleRate()).isEqualTo(10);
+            });
+    testing
```
jeanbisutti, line 58: Perhaps add a comment that in this case the sampling percentage should be that of the request.

smoke-tests/apps/SamplingOverrides/src/smokeTest/resources/applicationinsights3.json
```diff
@@ -0,0 +1,39 @@
...
+      {
+        "telemetryType": "dependency",
+        "attributes": [
+          {
+            "key": "db.statement",
+            "value": "select * from abc",
+            "matchType": "strict"
+          }
+        ],
+        "percentage": 50,
```
jeanbisutti, line 30: It may be worth to add a test for the user configuration given in the PR description (0% for the request). I have not seen it.

jeanbisutti (commented):
Could you please make the PR title more specific? Perhaps something like "Propagate the request sampling percentage"?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSamplerForOverride.java
```diff
@@ -0,0 +1,109 @@
...
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class AiSamplerForOverride implements Sampler {
```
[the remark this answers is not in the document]
trask: there's already an `AiOverrideSampler` 😭, I agree the names are confusing, but I haven't been able to come up with anything good

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplerUtil.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final Cache<Double, SamplingResult> recordAndSampleWithSampleRateMap =
+      Cache.bounded(100);
+
+  static SamplingResult shouldSample(String traceId, double sp) {
+    SamplingResult samplingResult;
+    if (sp == 0) {
+      return SamplingResult.drop();
+    }
+
+    if (sp != 100 && !shouldRecordAndSample(traceId, sp)) {
```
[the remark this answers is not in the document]
trask: this line isn't new (sorry, I had to move things around a bit)

https://github.com/microsoft/ApplicationInsights-Java/blob/6667d98d436fe1e9a646c84538c9545dfee3e0ee/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSampler.java#L93

smoke-tests/apps/SamplingOverrides/src/smokeTest/resources/applicationinsights3.json
trask: I think it's going to be the same code path as 10% for the request (the key is to test request percentage < dependency percentage < 100%)

jeanbisutti: My suggestion was to be fully sure.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSamplerForOverride.java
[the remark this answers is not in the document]
jeanbisutti: One is used to configure a specific sampler in the OTel SDK and the other one used at the exporter level?

Perhaps one could be named to something like `AiOverrideExportSampler`?

smoke-tests/apps/SamplingOverrides/src/smokeTest/resources/applicationinsights3.json
trask: added 👍

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentLogExporter.java
```diff
@@ -96,45 +99,62 @@ private CompletableResultCode internalExport(Collection<LogRecordData> logs) {
...
-          // capture it
-          continue;
-        }
+      boolean hasSamplingOverride = sampler != null;
 
-        logger.debug("exporting log: {}", log);
+      if (!hasSamplingOverride
+          && spanContext.isValid()
+          && !spanContext.getTraceFlags().isSampled()) {
+        // if there is no sampling override, and the log is part of an unsampled trace,
```
jeanbisutti, line 130: Because the use cases are not super simple, the code comments could try to refer to a test checking this scenario

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/AiSamplerForOverride.java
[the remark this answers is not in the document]
trask: > One is used to configure a specific sampler in the OTel SDK and the other one used at the exporter level?

not quite, they are both used (directly or indirectly) by the trace sampling, but one of them is also re-used by log exporter

i've done some renaming, let me know if it helps

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/sampling/SamplerUtil.java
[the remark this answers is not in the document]
trask: let me know if ok to resolve this discussion

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentLogExporter.java
jeanbisutti: My suggestion was to refer to tests in the comments, in addition to the current comment explanations.

trask: I'd prefer to just improve the comments if they are unclear. Or if you make specific suggestions linking these to tests I'm happy to hit accept.

jeanbisutti (approved):

PR #4038, opened 2025-01-22
Revert "Add warning message if the Java agent is in the class path"

trask:
Reverts microsoft/ApplicationInsights-Java#3960

@jeanbisutti it looks like this warning is always displayed on Java 8. you can see it if you look at the Java 8 smoke test runs in CI


trask:
closing in favor of #4039

PR #4121, opened 2025-03-04
Support for AAD Audience in connection string (for sovereign cloud)

harsimar:
Fix # .
Note: Don't merge until this [related change](https://github.com/Azure/azure-sdk-for-java/pull/44482) from the autoconfigure module is released. 

This PR incorporates connection string parsing changes from the autoconfigure module to use the AAD audience that is either provided via the connection string (or a default audience if that isn't provided). This is applicable for sovereign cloud scenarios where customers may be using AAD auth - the LazyHttpClient would use the provided audience when creating a new http pipeline, and all calls to quickpulse/breeze/profiler would be made with that audience. 

Tested in sovereign cloud environment via a local build.

- [x] CHANGELOG.md updated


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/telemetry/TelemetryClient.java
```diff
@@ -353,6 +359,13 @@ public ConnectionString getConnectionString() {
     return connectionString;
   }
 
+  public String getAadAudienceWithScope() {
+    if (connectionString == null) {
+      return APPLICATIONINSIGHTS_AUTHENTICATION_SCOPE;
```
harsimar, line 364: This line was added because ConsumptionPlanEnabledTest & RunTimeAttachWithDelayedConnectionStringTest were failing without this check. This raises a few questions ~

1) Are azure functions w/ consumption plan enabled supported in sovereign clouds? If so, would the connection  string always be set with some delay? Or is it still possible in azure function consumption plans to not set the connection string with a delay?
2) I noticed in these tests that the connection string is still null by the time the http pipelines are being created. Is there a need to refactor code such that http pipelines shouldn't be created unless the connection string is set to some value?

jeanbisutti: The runtime attachment is related to https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-spring-boot#enabling-programmatically.

The connection string can be configured at runtime (programmatically) with [-javaagent](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#connection-string-configured-at-runtime) or [the runtime attachment](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-spring-boot#programmatically-configure-the-connection-string).

With Azure function, the connection string retrieval is delayed:    https://github.com/microsoft/ApplicationInsights-Java/blob/d89b2563ad8e9786952f9cc319a23feed1e492a9/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AzureFunctionsInitializer.java#L90

About question 1, I don't know.

About 2, it may have several options. The place where the connection string is set at runtime (programmatic configuration or Azure function): https://github.com/microsoft/ApplicationInsights-Java/blob/d89b2563ad8e9786952f9cc319a23feed1e492a9/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/RuntimeConfigurator.java#L121

PR #4130, opened 2025-03-11
Fix request exception not captured when INSTRUMENTATION_LOGGING_LEVEL=OFF

### LINKED ISSUE #4095: INSTRUMENTATION_LOGGING_LEVEL suppresses exceptions to be recorded
### Expected behavior

Exceptions are recorded regardless of what level is specified by the `APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL` variable.

### Actual behavior

When the `APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL` variable is set to `OFF`, exceptions are filtered out.
Auto-collected exceptions or those recorded with <a href="https://javadoc.io/doc/io.opentelemetry/opentelemetry-api/latest/io/opentelemetry/api/trace/Span.html#recordException(java.lang.Throwable)"><code>recordException()</code></a> are exported since there is an `AgentSpanExporter` log, but they do not appear in the App Insights UI.
UPD: checked with `FATAL` level, also doesn't work, but it works with `ERROR`.

#### References

- On the "[Configuration options: Azure Monitor Application Insights for Java](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-config#autocollected-logging)" page, it says:
  > Log4j, Logback, JBoss Logging, and java.util.logging are autoinstrumented. Logging performed via these logging frameworks is autocollected.

  and the variable can be used to set the level.
- On the "[Exceptions | OpenTelemetry](https://opentelemetry.io/docs/specs/otel/trace/exceptions/)" page, it says:
  > An exception SHOULD be recorded as an **Event** on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.
- I can see a reason why disabled logging filters exceptions (because exceptions are actually events), but I don't think it's a clear behavior.
- On the "[Sampling overrides - Azure Monitor Application Insights for Java](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-sampling-overrides#how-it-works)" page, it says:
  > telemetryType (telemetryKind in Application Insights 3.4.0) must be one of request, dependency, trace (log), or exception.

  where the exception is of a different type from `trace` (`log`; events) and therefore must be recorded in a different way.

### To Reproduce

1. Exceptions appear when `slf4j` is not used - [`issue-4095/exceptions-without-slf4j/`](https://github.com/microsoft/ApplicationInsights-Java-Repros/pull/14/files#diff-7c3b01dea8f5a7e5f4e6ccaf04cef6a48fcd9f45d72aaba4403689719a8df0ed).
2. Exceptions disappear when `slf4j` is used - [`issue-4095/no-exceptions-with-slf4j/`](https://github.com/microsoft/ApplicationInsights-Java-Repros/pull/14/files#diff-e13d9d9b742119aebdcba207431ab0c9a77da38e1a9e2989dd6c1849ba9e2b8b).

### System information

 - SDK Version: 3.7.0
 - OS type and version: Linux
 - Using spring-boot? No
 - Additional relevant libraries (with version, if applicable): `org.slf4j:slf4j-api`, `io.opentelemetry:opentelemetry-api`, `io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations`.

### Logs

- [no-exceptions-with-slf4j.log](https://github.com/user-attachments/files/18892399/no-exceptions-with-slf4j.log)
- [exceptions-without-slf4j.log](https://github.com/user-attachments/files/18892428/exceptions-without-slf4j.log)

### Screenshots

`LOGGING_LEVEL` is `OFF` and `slf4j` is used:
Image

`LOGGING_LEVEL` is still `OFF`, but `slf4j` is removed:
Image

trask:
Resolves #4095


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/exporter/AgentSpanExporter.java
```diff
@@ -58,6 +58,9 @@ public CompletableResultCode export(Collection<SpanData> spans) {
     }
     for (SpanData span : spans) {
       logger.debug("exporting span: {}", span);
+
+      // TODO apply exception sampling overrides to span events!
```
trask, line 62: not ready for review yet, need to do this still

trask: nevermind, sampling overrides are already applied to span events, there was just an existing bug in it

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -425,7 +437,7 @@ private static Set<Feature> initStatsbeatFeatureSet(Configuration config) {
     if (config.preview.browserSdkLoader.enabled) {
       featureList.add(Feature.BROWSER_SDK_LOADER);
     }
-    if (!config.preview.sampling.overrides.isEmpty()) {
+    if (!config.sampling.overrides.isEmpty()) {
```
jeanbisutti, line 440: Many users have configured the sampling override.

Should the next release warns about deprecation, a following one removes `preview`?

It may also break the behaviors for App service users with the automatic agent upgrade.

trask: I don't think this change breaks anything, see 

https://github.com/microsoft/ApplicationInsights-Java/blob/84d364409bbd0264c69fb37d10847bbd06c3d672/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/configuration/ConfigurationBuilder.java#L208-L212

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -264,13 +262,27 @@ public void customize(AutoConfigurationCustomizer autoConfiguration) {
         .addPropertiesCustomizer(new AiConfigCustomizer())
         .addLogRecordProcessorCustomizer(
             (logRecordProcessor, configProperties) -> {
-              if (firstLogRecordProcessor.getAndSet(false)) {
-                // hack to run our log record processors first, before any other log processors
-                // (in particular before the batch log processor which performs the export)
-                // see https://github.com/open-telemetry/opentelemetry-java/issues/6599
+              if (logRecordProcessor instanceof BatchLogRecordProcessor) {
```
jeanbisutti, line 265: Could you please add a comment about why things have to be done at the `BatchLogRecordProcessor` level.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -264,13 +262,27 @@ public void customize(AutoConfigurationCustomizer autoConfiguration) {
...
+                        .collect(Collectors.toList());
+
+                logFilteringProcessor =
+                    new AzureMonitorLogFilteringProcessor(
+                        logSamplingOverrides,
+                        exceptionSamplingOverrides,
+                        (BatchLogRecordProcessor) logRecordProcessor,
+                        configuration.instrumentation.logging.getSeverityThreshold());
+
+                logRecordProcessors.add(logFilteringProcessor);
```
jeanbisutti, line 285: The code may be easier to read by extracting this code to a method.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AzureMonitorLogProcessor.java
```diff
@@ -3,103 +3,23 @@
...
-import io.opentelemetry.instrumentation.api.instrumenter.LocalRootSpan;
-import io.opentelemetry.sdk.internal.AttributesMap;
 import io.opentelemetry.sdk.logs.LogRecordProcessor;
 import io.opentelemetry.sdk.logs.ReadWriteLogRecord;
 import io.opentelemetry.sdk.trace.ReadableSpan;
-import io.opentelemetry.semconv.ExceptionAttributes;
-import java.lang.reflect.Field;
-import javax.annotation.Nullable;
 
 public class AzureMonitorLogProcessor implements LogRecordProcessor {
```
jeanbisutti, line 14:
```suggestion
// See also AzureMonitorLogFilteringProcessor
public class AzureMonitorLogProcessor implements LogRecordProcessor {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java
```diff
@@ -264,13 +262,18 @@ public void customize(AutoConfigurationCustomizer autoConfiguration) {
         .addPropertiesCustomizer(new AiConfigCustomizer())
         .addLogRecordProcessorCustomizer(
```
jeanbisutti, line 263: Suggestion: an extract method could be applied to this customizer, and  `addLogRecordProcessorCustomizer` could be used next to `.addLogRecordExporterCustomizer`. In this way it would be easier to notice the whole picture for logging.

jeanbisutti (approved):
Approved with comments.

jeanbisutti (approved):

PR #4141, opened 2025-03-17
Nicer smoke test assertions

trask:
Easier to read, also more aligned with upstream OpenTelemetry which uses similar assertj wrappers


smoke-tests/apps/gRPC/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/GrpcTest.java
```diff
@@ -32,174 +23,156 @@ abstract class GrpcTest {
...
+                .hasRequestSatisying(
+                    request ->
+                        request
+                            .hasName("example.Greeter/SayHello")
+                            .hasSuccess(true)
+                            .hasProperty("_MS.ProcessedByMetricExtractors", "True")
+                            .hasTag("ai.operation.name", "example.Greeter/SayHello")
+                            .hasParent(trace.getDependencyId(0)))
+                // auto-collected grpc events are suppressed by exporter because they are noisy
+                .hasMessageCount(0));
```
trask, line 56: this is what a test looks like with the new assertj wrappers

jeanbisutti (approved):

Issue #4140, opened 2025-03-17
App insights not exporting jmx metrics that are available later


numirov:
We have confluent kafka [connect](https://hub.docker.com/r/confluentinc/cp-kafka-connect) to which we are attaching app insights java agent. We want to expose connector metrics through jmx. We added them as described in the [app insights doc](https://learn.microsoft.com/en-us/azure/azure-monitor/app/java-jmx-metrics-configuration), confluent [doc](https://docs.confluent.io/platform/current/connect/monitoring.html#use-jmx-to-monitor-kconnect):
`{
                  "jmxMetrics": [
                    {
                      "name": "Connector JMX - type",
                      "objectName": "kafka.connect:type=connector-metrics,connector=connector-name",
                      "attribute": "connector-type"
                    },
                    {
                      "name": "Test Connector JMX - version",
                      "objectName": "kafka.connect:type=connector-metrics,connector=connector-name",
                      "attribute": "connector-version"
                    },
                    {
                      "name": "Test Connector JMX - class",
                      "objectName": "kafka.connect:type=connector-metrics,connector=connector-name",
                      "attribute": "connector-class"
                    },
                    {
                      "name": "Test Connector JMX - status(sandbox)",
                      "objectName": "kafka.connect:type=connector-metrics,connector=connector-name",
                      "attribute": "status"
                    },
                    {
                      "name": "Test java iometrics - count",
                      "objectName": "java.nio:name=mapped,type=BufferPool",
                      "attribute": "Count"
                    },
                    {
                      "name": "test connector - failed startup",
                      "objectName": "kafka.connect:type=connect-worker-metrics",
                      "attribute": "connector-startup-failure-total"
                    },
                    {
                      "name": "test connector - success startup",
                      "objectName": "kafka.connect:type=connect-worker-metrics",
                      "attribute": "connector-startup-success-total"
                    }
                  ],  
                  "selfDiagnostics": {
                    "level": "DEBUG"
                  }
}`
So from above only **kafka.connect:type** are not appearing in the portal. 
[...]

PR #4161, opened 2025-04-03
Log as warn invalid JMX metrics

jeanbisutti:
Related to #4140

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/AbstractJmxPerformanceCounter.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/AbstractJmxPerformanceCounter.java
@@ -7,7 +7,9 @@ import static com.microsoft.applicationinsights.agent.internal.diagnostics.MsgId
 
 import com.microsoft.applicationinsights.agent.internal.telemetry.TelemetryClient;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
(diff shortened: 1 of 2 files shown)


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/AbstractJmxPerformanceCounter.java
```diff
@@ -46,14 +51,22 @@ public synchronized void report(TelemetryClient telemetryClient) {
               value += Double.parseDouble(String.valueOf(obj));
             }
           } catch (RuntimeException e) {
+            if (!invalidJmxMetrics.contains(metricName)) {
+              invalidJmxMetrics.add(metricName);
```
trask, line 55:
```suggestion
            if (invalidJmxMetrics.add(metricName)) {
```

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/perfcounter/AbstractJmxPerformanceCounter.java
```diff
@@ -24,6 +26,8 @@ public abstract class AbstractJmxPerformanceCounter implements PerformanceCounte
   private final Collection<JmxAttributeData> attributes;
   private boolean alreadyLogged = false;
 
+  private final Set<String> invalidJmxMetrics = new HashSet<>();
```
trask, line 29: I forget if this class is only used by a single thread

```suggestion
  private final Set<String> invalidJmxMetrics = ConcurrentHashMap.newKeySet();
```

jeanbisutti: It does not seem but I take your suggestion just in case.

trask (approved):

PR #4350, opened 2025-07-18
Add timeout on profiler settings pull

johnoliver:
Add a time out on HTTP pull of the profiler settings. Defaulting to 30 seconds


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/profiler/service/ServiceProfilerClient.java
```diff
@@ -144,7 +149,18 @@ public Mono<ProfilerConfiguration> getSettings(Date oldTimeStamp) {
...
+        .send(request)
+        .timeout(
+            Duration.ofSeconds(httpTimeoutSeconds),
+            Mono.error(
+                new HttpResponseException(
+                    "Timed out after "
+                        + httpTimeoutSeconds
+                        + " seconds while waiting for response from "
+                        + requestUrl,
+                    null)))
```
trask, line 162: do we need timeout both here and via `TimeoutPolicy`?

johnoliver:
Converting to draft as more is probably needed for this

PR #4370, opened 2025-07-21
Use non-deprecated identity classes

trask:


trask:
Closing, I realized our VSCODE option is deprecated, so will just suppress deprecation warning here instead

PR #4374, opened 2025-07-21
Fix missing javadoc warnings

trask:


buildSrc/src/main/kotlin/ai.smoke-test.gradle.kts
```diff
@@ -87,6 +87,21 @@ configurations.all {
 }
 
 tasks {
+  // FIXME (trask) copy-pasted from ai.java-conventions.gradle
```
rajkumar-rangaraj, line 90: nit: Not sure what is the intent of this comment. Should this be removed or modified?

trask: there are other comments in this file like this too, ideally we wouldn't need the duplication

harsimar (approved):

PR #4429, opened 2025-08-19
Update OpenTelemetry versions

trask:


agent/agent-bootstrap/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -387,6 +393,49 @@ private void propagateOperationListenersToOnEnd() {
     propagateOperationListenersToOnEnd = true;
   }
 
+  private static <REQUEST, RESPONSE> void applyCustomizers(
```
rajkumar-rangaraj, line 396: Are these changes related to version bump?

trask: yes, see the top of the file, it's a copy of a file from upstream, which needs to be kept up-to-date with the upstream version

but also see https://github.com/microsoft/ApplicationInsights-Java/pull/4430 which deletes this file because there was a new feature in the latest upstream version that allows us to add the hook we need without having our own copy!!!

xiang17 (approved):

PR #4438, opened 2025-08-27
Update semconv artifact version

trask:


agent/agent-tooling/gradle.lockfile
```diff
@@ -71,7 +71,7 @@ io.opentelemetry:opentelemetry-sdk-logs:1.53.0=runtimeClasspath
 io.opentelemetry:opentelemetry-sdk-metrics:1.53.0=runtimeClasspath
 io.opentelemetry:opentelemetry-sdk-trace:1.53.0=runtimeClasspath
 io.opentelemetry:opentelemetry-sdk:1.53.0=runtimeClasspath
-io.opentelemetry:opentelemetry-semconv:1.30.1-alpha=runtimeClasspath
+io.opentelemetry:opentelemetry-semconv:0.14.1=runtimeClasspath
```
rajkumar-rangaraj, line 74: is this update correct?

trask: yeah, this file is autogenerated and this PR exposed that we are picking up this (unused) dependency from autoconfigure module, we could explicitly exclude it, but it's not causing any harm (other than 50kb of bloat in the Java agent size), and this line will go away completely once we bring in next autoconfigure release (due to https://github.com/Azure/azure-sdk-for-java/pull/46520)

rajkumar-rangaraj (approved):

PR #4446, opened 2025-09-02
Rollback #4373

jeanbisutti:
#4373 may lead to a network isolation issue in the release build

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -8,10 +8,6 @@ pluginManagement {
   }
 }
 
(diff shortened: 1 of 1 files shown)


settings.gradle.kts
```diff
@@ -8,10 +8,6 @@ pluginManagement {
   }
 }
 
-plugins {
```
trask, line 11: can you add a comment here about the problem with adding toolchain support so we don't accidentally add it back in?

PR #4458, opened 2025-09-12
Read environment variables for AKS metrics otlp support

harsimar:
This PR will read the following env vars so that the MetricDataMapper can append the necessary attribute to metric / send the metrics to the right places:

OTEL_METRICS_EXPORTER
OTEL_METRICS_EXPORTER_OTLP_ENDPOINT
APPLICATIONINSIGHTS_METRICS_TO_LOGANALYTICS_ENABLED

Note: still in draft mode because this depends on the [azure-sdk-for-java](https://github.com/Azure/azure-sdk-for-java/pull/46680) pr, and won't build until I build on top of [this pr](https://github.com/microsoft/ApplicationInsights-Java/pull/4445). I also need to test the e2e behavior once I have a built java agent.


smoke-tests/framework/src/main/java/com/microsoft/applicationinsights/smoketest/SmokeTestExtension.java
```diff
@@ -422,6 +425,12 @@ private void startTestApplicationContainer() throws Exception {
     Testcontainers.exposeHostPorts(6060);
     Testcontainers.exposeHostPorts(4318);
 
+    if (useOtlpViaEnvVars) {
+      envVars.put("OTEL_METRICS_EXPORTER", "otlp,azure_monitor");
+      envVars.put("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", FAKE_OTLP_INGESTION_ENDPOINT);
+      envVars.put("OTEL_EXPORTER_OTLP_PROTOCOL", "http/protobuf");
+    }
```
trask: this seems duplicative, I think we can simplify and just change `useOtlpEndpoint` to use the env vars (there's no specific need for us to test it _also_ with system properties)

smoke-tests/framework/src/main/java/com/microsoft/applicationinsights/smoketest/SmokeTestExtensionBuilder.java
```diff
@@ -29,6 +29,7 @@ public class SmokeTestExtensionBuilder {
   private final List<String> jvmArgs = new ArrayList<>();
   private boolean useDefaultHttpPort;
   private boolean useOtlpEndpoint;
+  private boolean useOtlpViaEnvVars;
```
trask: then we can revert changes to this file

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
```diff
@@ -136,7 +136,7 @@ void generateHttpClientMetrics() {
     MetricTelemetryBuilder builder = MetricTelemetryBuilder.create();
     MetricData metricData = metricDataCollection.iterator().next();
     MetricDataMapper.updateMetricPointBuilder(
-        builder, metricData, metricData.getData().getPoints().iterator().next(), true, true);
+        builder, metricData, metricData.getData().getPoints().iterator().next(), true, true, null);
```
trask, line 139: better to pass `false` explicitly (and ideally change the function from `Boolean` to `boolean`)
```suggestion
        builder, metricData, metricData.getData().getPoints().iterator().next(), true, true, false);
```

smoke-tests/apps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTestToAmw.java
```diff
@@ -0,0 +1,202 @@
...
+    // validate custom dimension
+    Map<String, String> properties3 = md3.getProperties();
+    assertThat(properties3).containsEntry("tag1", "abc");
+    assertThat(properties3).containsEntry("tag2", "def");
+    assertThat(properties3).containsEntry("name", "lemon");
+    assertThat(properties3).containsEntry("color", "yellow");
+    assertThat(properties3).containsEntry("_MS.SentToAMW", "True");
+  }
+
+  private void validateOtlpMetricsReceived(String name) throws Exception {
```
trask: it doesn't look like `name` is used in this function

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
xiang17: I set it to `Boolean` in case anyone would want to skip the `_MS.SentToAMW` flag. It would have different values.

smoke-tests/apps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTestToAmw.java
xiang17: Removed.

trask: can you confirm if it's correct to remove it, or if some validation on name should be added instead?

agent/agent-tooling/src/test/java/com/microsoft/applicationinsights/agent/internal/PreAggregatedMetricsTest.java
trask: I think it would be good to clarify in the spec if we can skip sending this flag. That would be better whenever `false` to save ingestion costs. And then we wouldn't need tri-state here.

smoke-tests/apps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTestToAmw.java
xiang17: I didn't find any reference to this variable.

smoke-tests/framework/src/main/java/com/microsoft/applicationinsights/smoketest/SmokeTestExtension.java
xiang17: I've replaced this variable with that existing variable.

smoke-tests/framework/src/main/java/com/microsoft/applicationinsights/smoketest/SmokeTestExtensionBuilder.java
xiang17: same as the other comment

smoke-tests/apps/HttpPreaggregatedMetrics/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HttpPreaggregatedMetricsTestOtlpEnabled.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // 2nd pre-aggregated metric
+    Envelope envelope2 = metrics.get(1);
+    MetricData md2 = (MetricData) ((Data<?>) envelope2.getData()).getBaseData();
+    validateMetricData(md2);
+
+    // 3rd pre-aggregated metric
+    Envelope envelope3 = metrics.get(2);
+    MetricData md3 = (MetricData) ((Data<?>) envelope3.getData()).getBaseData();
+    validateMetricData(md3);
```
trask, line 75: is sorting needed? looks like same verification is performed for all 3


```suggestion
    for (Envelope envelope : metrics) {
      MetricData md = (MetricData) ((Data<?>) envelope.getData()).getBaseData();
      validateMetricData(md);
    }
```

smoke-tests/apps/HttpPreaggregatedMetrics/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/HttpPreaggregatedMetricsTestOtlpEnabled.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  private static void verifyHttpServerPreAggregatedMetrics(List<Envelope> metrics) {
+    assertThat(metrics.size()).isEqualTo(1);
+    // 1st pre-aggregated metric
```
trask, line 80: (there's only 1)
```suggestion

```

smoke-tests/apps/OpenTelemetryMetric/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/OpenTelemetryMetricTestToAmw.java
```diff
@@ -0,0 +1,201 @@
...
+                      .mockedOtlpIngestion
+                      .getCollectorServer()
+                      .retrieveRecordedRequests(request());
+
+              assertThat(otlpRequests).isNotEmpty();
+
+              List<Metric> otlpMetrics =
+                  testing.mockedOtlpIngestion.extractMetricsFromRequests(otlpRequests);
+
+              assertThat(otlpMetrics).hasSizeGreaterThanOrEqualTo(3);
```
trask, line 162: it may be worth adding assertions about the name and attributes of these 3 metrics

trask (approved):

trask (approved):

PR #4517, opened 2025-10-23
Fix Release Tag

### LINKED ISSUE #4500: Missing release 3.7.5
According to https://central.sonatype.com/artifact/com.microsoft.azure/applicationinsights-agent I can see that 3.7.5 was released on the 26th September.

Also https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-enable?tabs=java states
> Point the Java virtual machine (JVM) to the jar file by adding -javaagent:"path/to/applicationinsights-agent-3.7.5.jar" to your application's JVM args.

However on that same page it links to the 3.7.4 release to download the agent.

https://github.com/microsoft/ApplicationInsights-Java/releases shows that the latest release is 3.7.4 and there is no 3.7.5 tag.

Should we therefore be using 3.7.5 at present - was it released and then removed, or never released in the first place?

xiang17:
Fix #4500.

Enable creating a release based on a previous commit. The release pipeline needs to be initiated from a previous commit.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,7 +46,7 @@ jobs:
           else
             opt_prerelease="--prerelease"
           fi
-          gh release create --target main \
+          gh release create --target HEAD \
                             --title "Application Insights Java $VERSION$opt_ga" \
                             --notes-file /tmp/release-notes.txt \
                             $opt_prerelease \


.github/workflows/release.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           else
             opt_prerelease="--prerelease"
           fi
-          gh release create --target main \
+          gh release create --target HEAD \
                             --title "Application Insights Java $VERSION$opt_ga" \
```
trask, line 50: I think `HEAD` is the default

```suggestion
          gh release create --title "Application Insights Java $VERSION$opt_ga" \
```

xiang17: Default is the `main` branch:

>  --target <branch>
    Target branch or full commit SHA (default [main branch])

https://cli.github.com/manual/gh_release_create

trask (approved):

Issue #4526, opened 2025-10-29
Filter logs using attributes does not work starting version 3.7.2


dvanackere-lpg:
### Expected behavior
Adding an attribute to a log and add a filter with this attribute should remove the log in Application Insights.
This was the behavior since at least version 3.4.x until version 3.7.1

### Actual behavior
The attribute is added, the log is modified accordingly, but the log is not filtered.

### To Reproduce

applicationinsights.json:
`
{
  "sampling": {
    "percentage": 100,
    "overrides": [{
        "telemetryType": "trace",
        "attributes": [{
            "key": "TODELETE",
            "value": ".*",
            "matchType": "regexp"
          }
        ],
        "percentage": 0
      }
    ]
  },
  "preview": {
    "processors": [{
        "type": "log",
        "body": {
          "toAttributes": {
            "rules": [
              "^Noisy log (?<TODELETE>.*)"
            ]
          }
        }
      }
    ]
  }
}
`

In the application, add a noisy log:
`
logger.info { "Noisy log that should not be sent to AppInsights" }
`

### System information
Please provide the following information:
 - SDK Version: 3.7.2 to 3.7.5 KO, 3.4.x to 3.7.1 OK
 - Using spring-boot: yes

I suspect that starting version 3.7.2, the 'sampling overrides' and the 'attribute processors' are not handled in the same order (processor then sampling before 3.7.2, sampling then processor starting 3.7.2).

dvanackere-lpg:
Hello @jeanbisutti , thanks for your response, unfortunately your suggestions would not work for us as the noisy logs come from third party libraries in the framework itself (from Azure, sorry) and disable them completely (or raising the level to ERROR) would remove some important monitoring (in Azure Functions, some log configurations must be at Information level to enable the monitoring).

PR #4527, opened 2025-10-29
Fix log sampling overrides by running preview processors before filtering

trask:
**Summary**
- Add a smoke test (`TelemetryProcessorLogFilterSamplingOverridesTest`) that reproduces issue #4526 by exercising sampling overrides based on attributes derived by preview processors.
- Adjust the log filtering pipeline so `AzureMonitorLogFilteringProcessor` applies preview attribute/log processors before executing severity filtering and sampling overrides, ensuring those overrides see the processed attributes.

**Root Cause**
- [PR #4130](https://github.com/microsoft/ApplicationInsights-Java/pull/4130) introduced `AzureMonitorLogFilteringProcessor`, moving severity filtering and sampling overrides from `AgentLogExporter` to a processor that now runs ahead of preview attribute and log processors.
- Because these preview processors still wrap only the exporter, overrides that relied on attributes they extract (e.g., regex-based fields) were evaluated against unprocessed log records, causing the regression described in issue #4526.

**Fix Details**
- Apply preview attribute and log processors inside `AzureMonitorLogFilteringProcessor` before sampling decisions so overrides can match on transformed attributes.
- Thread the processor configs from `SecondEntryPoint` into the filtering processor to maintain the prior processing order.

### DIFF
--- a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AzureMonitorLogFilteringProcessor.java
+++ b/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AzureMonitorLogFilteringProcessor.java
@@ -5,8 +5,13 @@ package com.microsoft.applicationinsights.agent.internal.init;
 
 import com.azure.monitor.opentelemetry.autoconfigure.implementation.AiSemanticAttributes;
 import com.microsoft.applicationinsights.agent.internal.configuration.Configuration;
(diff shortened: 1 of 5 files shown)


trask:
Closing, as sampling overrides should occur before telemetry processors are run. The previous behavior was unfortunately more of a bug. Attempting to work around this and preserve the prior behavior is complex and leading to potentially other inconsistencies and behavioral changes.

PR #4575, opened 2026-01-05
Add azure_monitor to metrics exporter for AKS

xiang17:
Fix # .

If `APPLICATIONINSIGHTS_METRICS_TO_LOGANALYTICS_ENABLED` is set to `true`, send OTel metrics to Breeze even if `OTEL_METRICS_EXPORTER` doesn't have `azure_monitor`.

- [ ] CHANGELOG.md updated


agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -115,9 +116,25 @@ public Map<String, String> apply(ConfigProperties otelConfig) {
     }
 
     String metricsExporter = otelConfig.getString("otel.metrics.exporter");
-    if (metricsExporter == null) {
-      // this overrides the default "otlp" so the exporter can be configured later
-      properties.put("otel.metrics.exporter", "none");
+    String azureMonitorName = AzureMonitorExporterProviderKeys.EXPORTER_NAME;
+    boolean metricsToLogAnalyticsEnabled =
+        otelConfig.getBoolean("applicationinsights.metrics.to.loganalytics.enabled", false);
```
jeanbisutti: The specifications seem to indicate that the `APPLICATIONINSIGHTS_METRICS_TO_LOGANALYTICS_ENABLED` should only be used in case of AKS and Auto-Attach. AKS detection can be done in this way: https://github.com/microsoft/ApplicationInsights-Java/blob/25f901b3677d6978a54c9bacc20c939d70ba749a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java#L55

Auto-Attach detection requires the Java agent JAR path . See [DiagnosticHelper.initRpIntegration.(Path agentPath)](https://github.com/microsoft/ApplicationInsights-Java/blob/25f901b3677d6978a54c9bacc20c939d70ba749a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java#L35) and [DiagnosticsHelper.setRpAttachType(Path agentPath, String markerFile)](https://github.com/microsoft/ApplicationInsights-Java/blob/25f901b3677d6978a54c9bacc20c939d70ba749a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/diagnostics/DiagnosticsHelper.java#L81C23-L81C73) .

`AiConfigCustomizer` is instanciated in the `SecondEntry` point: https://github.com/microsoft/ApplicationInsights-Java/blob/25f901b3677d6978a54c9bacc20c939d70ba749a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/SecondEntryPoint.java#L266

The RP integration of DiagnosticsHelper is initiliazed in the first entry point: https://github.com/microsoft/ApplicationInsights-Java/blob/25f901b3677d6978a54c9bacc20c939d70ba749a/agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/FirstEntryPoint.java#L92

I am not sure that the first entry point always finishes to execute before the second one. If the first entry point always finishes to execute before the second one, you could use the RP initialization done in the first entry point from the second one. Otherwise, you could reimplement the same kind of things in the second entry point.

xiang17: Updated to check AKS from the `AKS_ARM_NAMESPACE_ID` env var.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -115,9 +117,36 @@ public Map<String, String> apply(ConfigProperties otelConfig) {
...
+
+    boolean isAks = false;
+    try {
+      String aksNamespaceId = System.getenv("AKS_ARM_NAMESPACE_ID");
+      isAks = !Strings.isNullOrEmpty(aksNamespaceId);
+    } catch (SecurityException ignored) {
+      // if env is not accessible, assume not AKS
+    }
+
+    if (metricsToLogAnalyticsActivated && isAks) {
```
jeanbisutti: The specifications mention the case of "Auto-Attach scenarios on AKS". Perhaps the code should check that the Applicaction Insights Java agent is running with the Auto-Attach mode (https://github.com/microsoft/ApplicationInsights-Java/pull/4575#discussion_r2665056993)?

xiang17: We'll do for AKS regardless of mode.

lzchen (approved):

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -326,6 +330,31 @@ private static void enableInstrumentations(
     }
   }
 
+  private static boolean isAksIntegratedAttach(ConfigProperties otelConfig) {
+    String envVar = otelConfig.getString("applicationinsights.metrics.to.loganalytics.enabled");
+    boolean metricsToLogAnalyticsEnabled = envVar == null || Boolean.parseBoolean(envVar);
```
trask: do we want `envVar == null` condition? I was thinking we want to restrict this to AKS integrated attach only? but not sure

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -326,6 +330,31 @@ private static void enableInstrumentations(
...
+
+    boolean isAks = false;
+    try {
+      String aksNamespaceId = System.getenv("AKS_ARM_NAMESPACE_ID");
+      isAks = !Strings.isNullOrEmpty(aksNamespaceId);
+    } catch (SecurityException ignored) {
+      // if env is not accessible, assume not AKS
+    }
```
trask: I don't think we catch `SecurityException` elsewhere when we call System.getenv

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -326,6 +330,26 @@ private static void enableInstrumentations(
...
+
+    String aksNamespaceId = System.getenv("AKS_ARM_NAMESPACE_ID");
+    boolean isAks = !Strings.isNullOrEmpty(aksNamespaceId);
+
+    return metricsToLogAnalyticsEnabled && isAks;
+  }
+
+  private static String addAzureMonitorIfNotPresent(String metricsExporter) {
+    String azureMonitorName = AzureMonitorExporterProviderKeys.EXPORTER_NAME;
+    if (metricsExporter == null || metricsExporter.isEmpty() || metricsExporter.equals("none")) {
```
trask: based on this table https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/AutoAttach_Env_Vars.md#metrics-exporter

it seems like `none` + AMLE unset shouldn't add the exporter?

maybe worth generating a unit test to all the possibilities in that table

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
xiang17: Removed the catch clause.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
xiang17: Added unit tests to cover all cases.

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -343,4 +352,37 @@ private static <T> String join(List<T> values, char separator) {
...
+  // visible for tests
+  static boolean isAksAttach(String aksNamespaceId) {
+    return !Strings.isNullOrEmpty(aksNamespaceId);
+  }
+
+  static String updateMetricsExporter(String metricsExporter, String metricsToLogAnalyticsEnabled) {
+    String azureMonitorName = AzureMonitorExporterProviderKeys.EXPORTER_NAME;
+    // If AMLE is true, configure both otlp and azure monitor exporters
+    if (Boolean.parseBoolean(metricsToLogAnalyticsEnabled)) {
+      return azureMonitorName + ",otlp";
```
trask: should this be in addition to what the user set as `metricsExporter`?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
```diff
@@ -343,4 +352,37 @@ private static <T> String join(List<T> values, char separator) {
...
+    }
+
+    // If AMLE is unset:
+    if (metricsToLogAnalyticsEnabled == null || metricsToLogAnalyticsEnabled.isEmpty()) {
+      if (metricsExporter == null || metricsExporter.isEmpty()) {
+        // default is "azure_monitor,otlp"
+        return azureMonitorName + ",otlp";
+      } else if (metricsExporter.contains(azureMonitorName) && !metricsExporter.contains("otlp")) {
+        // if azure monitor is already present and otlp is not, add otlp
+        metricsExporter += ",otlp";
```
trask: i don't understand why adding `otlp` here?

agent/agent-tooling/src/main/java/com/microsoft/applicationinsights/agent/internal/init/AiConfigCustomizer.java
xiang17: Updated.

PR #4608, opened 2026-02-02
Fix connection string env var precedence over ${file:...} config lookup

### LINKED ISSUE #4214: Connection string env variable does not take precedence over file defined connectionString
### Expected behavior
I have env variable `APPLICATIONINSIGHITS_CONNECTION_STRING` and applicationinsights.json file. When I define connection string in configuration file via file path variable, the env variable does not take the precedence. 
When I use connectionString hardcoded in configuration file it works well and I would expect the same from the file path variable. 
### Actual behavior
The startup will fail, because the sdk attempts to read the file which does not exists instead of using the env variable. 

### To Reproduce

use runtime-attach dependency in pom.xml
`<dependency>
            <groupId>com.microsoft.azure</groupId>
            <artifactId>applicationinsights-runtime-attach</artifactId>
            <version>3.7.2</version>
</dependency>`

configure env variable `APPLICATIONINSIGHITS_CONNECTION_STRING` with connection string to Application Insights

Define `applicationinsights.json` configuration file in your application directory with connection string defined as filepath variable
```
{
  "connectionString": "${file:/app/secrets/applicationinsights_connection_string}"
}

```

### System information
 - SDK Version: 3.7.1
 - OS type and version: Win10, Java 17
 - Application Server type and version (if applicable): Springboots tomcat
 - Using spring-boot? Yes - 3.2.3
 - Additional relevant libraries (with version, if applicable): No

### Logs

[appins.log](https://github.com/user-attachments/files/20437245/appins.log)

jeanbisutti:
Fix #4214.

Using a base applicationinsights.json with `${file:...}` across environments (e.g. dev, prod) cannot fall back to the env var in environments where the secret file isn't mounted.

The fix defers the file lookup using a `Supplier<String>`, so the file is only read when no env var or system property override is present.


smoke-tests/apps/ConnectionStringFileNotFound/src/smokeTest/java/com/microsoft/applicationinsights/smoketest/ConnectionStringFileNotFoundTest.java
```diff
@@ -0,0 +1,80 @@
...
+import static org.assertj.core.data.MapEntry.entry;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+// Reproduces https://github.com/microsoft/ApplicationInsights-Java/issues/4214
+// When applicationinsights.json contains a ${file:...} reference to a nonexistent file,
+// the APPLICATIONINSIGHTS_CONNECTION_STRING env var should still take effect.
+@UseAgent
+abstract class ConnectionStringFileNotFoundTest {
```
jeanbisutti: It reproduces #4214

PR #4616, opened 2026-02-06
Gradle 9 prereq: remove smoke test dependencies on Spring Boot launcher

trask:
Keeping dependencies on Spring, only removing dependencies on Spring Boot launcher.

Reason: upgrading to Gradle 9.3.1 forces an upgrade of Spring Boot Gradle plugin from 2.5.12 to 3.4.2 (Spring Boot 2.7.x uses deprecated Gradle APIs removed in Gradle 9). Spring Boot 3.x compiles its `bootJar` launcher classes for Java 17+, making it impossible to run these JARs on Java 8 (which we want to continue testing against).

### DIFF
--- a/smoke-tests/apps/AzureFunctions/src/main/java/com/microsoft/applicationinsights/smoketestapp/SpringBootApp.java
+++ b/smoke-tests/apps/AzureFunctions/src/main/java/com/microsoft/applicationinsights/smoketestapp/SpringBootApp.java
@@ -6,17 +6,24 @@ package com.microsoft.applicationinsights.smoketestapp;
 import com.microsoft.azure.functions.worker.handler.FunctionEnvironmentReloadRequestHandler;
 import java.lang.reflect.Field;
 import java.util.Map;
-import org.springframework.boot.SpringApplication;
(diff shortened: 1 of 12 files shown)


smoke-tests/apps/AzureFunctions/src/main/java/com/microsoft/applicationinsights/smoketestapp/SpringBootApp.java
```diff
@@ -27,7 +34,23 @@ public static void main(String[] args) throws Exception {
...
+    appContext.setServletContext(context.getServletContext());
+    appContext.register(SpringBootApp.class);
+    appContext.refresh();
+
+    DispatcherServlet dispatcher = new DispatcherServlet(appContext);
+    Tomcat.addServlet(context, "dispatcher", dispatcher).setLoadOnStartup(1);
+    context.addServletMappingDecoded("/*", "dispatcher");
+
+    tomcat.start();
+    tomcat.getServer().await();
```
trask, line 53: there's a good bit of duplication of this block of code among these tests

I think that's ok since these are "standalone applications under test"

PR #4727, opened 2026-05-26
Improve Dependabot grouping for OpenTelemetry version variables

xiang17:
## Why

After #4721, Dependabot didn't group all OpenTelemetry packages in the same PR.

Issue 1: Changes in #4724 should be grouped into #4719 as part of the opentelemetry group.
Issue 2: `otelInstrumentationAlphaVersion` is detected by Dependabot to create PRs like #4719. But `otelSdkVersion` and `otelContribVersion` are not detected. The difference seems to be there is Kotlin string interpolation.

## What Changed
- Introduce explicit alpha-version variables for OTel SDK and contrib versions instead of using `${...}-alpha` interpolation in dependency declarations
- Update OTel BOM and contrib dependency entries to use the new explicit alpha vars
- Expand Dependabot `opentelemetry` group patterns to include the new alpha vars and concrete contrib artifacts

This makes OTel dependency versioning clearer for humans and easier for Dependabot to detect and group consistently.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -46,15 +46,19 @@ updates:
       opentelemetry:
         patterns:
           - "otelSdkVersion"
+          - "otelSdkAlphaVersion"
           - "otelInstrumentationVersion"
           - "otelInstrumentationAlphaVersion"
           - "otelContribVersion"
+          - "otelContribAlphaVersion"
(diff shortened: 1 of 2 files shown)


rajkumar-rangaraj (approved):

xiang17:
This worked. After merging the PR, Dependabot closed #4724 and opened a new PR #4728.

PR #4761, opened 2026-06-19
Normalize line endings for gradlew.bat files

xiang17:
Follow up on #4711 as its step 3.

The files are updated as:

> git add --renormalize gradlew.bat perf-tests/gradlew.bat


mattsains-msft (approved):
consider adding a .gitattributes file

xiang17:
The repo already has a `.gitattributes` file. The existing file/index not normalized. So this issue has been occurring in dependabot's environment.

